PR #870, opened 2013-09-14
make `word-at-point' the default input value for magit-show-commit

mgalgs:


fommil:
ha! We have evidence of @tarsius saying "At the very least" :-P

tarsius:
:open_mouth:

Well, it's okay, since it doesn't proceed an action which a person should take. And it's also from before [the ban](https://github.com/magit/magit/blob/master/CONTRIBUTING.md#reporting-issues-and-suggesting-features) was imposed.  :smirk:

PR #1691, opened 2014-12-30
Docstrings.

mwfogleman:


tarsius:
> the conflicting licenses ... Do you think that's a problem in the docstrings?

I don't know and so the answer becomes *yes it is*. On one hand the licenses are not compatible and Linus intentionally picked v2 not "v2 or whatever comes next". On the other hand there's fair use. In any case copying doc-strings make it very easy for a troll to harass us.

So what we should do for the doc-strings is always first think from the Emacs side. Strictly follow the Emacs conventions, thereby making it our own. *Then* head over to the Git documentation to check whether anything crucial was forgotten, or whether the terminology is off.

tarsius:
I am afraid I have to say that many of the suggested changes are not improvements. Some of the changes are plain wrong, others are just marginal improvements. I think the best course of action is for me to resume working on the docs, and then for others to review and improve these changes, instead of the other way around.

The above is not all bad, but there are many commits changing rather little. If you don't want the effort you put into this goto waste, you should reduce this to a handful commits which I can then individually cherry-pick or reject (so not to small but also not to big commits, and *no* commits which fixup earlier commits in the same series).

Sorry about this, it's always hard to start contributing to a project, while the maintainer is taking a break, I am afraid.

mwfogleman:
Hi Tarsius, 

I just tried to clean up my commits. Before I did so, I made a local pre-clean branch, so if these are not cleaned to your liking, I can clean them in some other way.

No need to apologize. If my commits are helpful, feel free to take them. If they aren't, feel free to drop them.

I've learned a lot from poking around the magit code. Even if these commits aren't particularly helpful, I now have a cursory familiarity with the codebase, and that may come in handy in future contribution efforts.

tarsius:
Are you familiar with the `company` package (or would not mind learning more about it)? If so there might be something you could do which would be more exiting as it involves coding. I kinda promised a "all the commands that make sense here" menu. In a sense that would be documentation still, so it would be okay to implement this now, despite the feature freeze. (But if we run into complications this might end up not making it into `2.1.0` anyway).

What we need is a command that invoked on e.g. a hunk would display something like:

```
@@ -5,5 +5,5 best of all numbers
stage    s
unstage  u
discard  k
revert   v
...
```

That could however be harder to do than it looks on the surface. It's not enough to just create a menu from the respective keymaps. Some commands have internal logic which controls whether it can act on the particular thing at point, e.g. you cannot stage a committed change but `s` is never-the-less bound to `magit-stage` on such hunks.

Well, for an initial implementation it's probably better to just ignore that issue. In any case we should avoid complications that leak into other parts of the code. It's probably easier to figure out how to do that if we start with an implementation that ignores the issue altogether.

I would like to use `company` for this, not `popup` which is used by the competing `auto-complete`, because the former has a very good reputation and seems to be generally preferred by newer packages. I don't know however if `company` exposes the low-level menu-creating code in a way we could use.

tarsius:
I have just pushed the first round of my own massive doc-string updates. Covers most of the things you documented, except for some internal functions, which I consider to be named descriptively enough to not require a doc-string. Also I wrote these doc-strings without looking at the Git manpages.

So am afraid, this wasn't able to make it in.

mwfogleman:
:+1: Great! Fine by me.

Sorry, I forgot to reply to your message about company. In any case, I don't think I'm the one to help with that.

tarsius:
Just reopening to test some Github functionality.

tarsius:
That's what I expected.

Issue #1645, opened 2014-11-27
Magit 2.1.0 will be released on the 1th of July


tarsius:
## Magit 2.1.0 will be released on the 1th of July

----

**[Related issues](https://github.com/magit/magit/milestones/2.1.0)** | **[Update instructions](http://magit.vc/manual/magit/Installation.html)** | **[MAGIT 2.1.0 RELEASED](https://github.com/magit/magit/issues/1941)**

----

## Release steps

**Bold** items are important check-points.
_Italic_ items require that someone lends me a hand.

* [x] **announce release date**
* [x] inform maintainers of some
  * [x] extensions #1921
  * [x] themes #1922 
  * [x] starter-kits #1923
* [x] prepare pull requests
  * [x] melpa milkypostman/melpa#2892
  * [x] solarized bbatsov/solarized-emacs#189
  * [x] zenburn bbatsov/zenburn-emacs#200
  * [x] magit-gh-pulls sigma/magit-gh-pulls#31
* [x] test elpa packaging using fake release and local server
  * [x] melpa
  * [x] marmalade
  * [x] update without following instructions (went better than expected)
* [x] copy fixes from manual to respective blurbs elsewhere
* [x] update information and links on webpage and wiki
* [x] strip temporary information from README.md
* [x] regenerate Authors.md
* [x] finish release notes
* [x] **merge `next` branch into `master`**
  * [x] git-modes
  * [x] magit
* [x] rename `next` branch to `master`
  * [x] magit-stgit
  * [x] magit-svn
  * [x] magit-topgit
* [x] tag
  * [x] git-modes 1.2.0
  * [x] magit 2.1.0
  * [x] magit-stgit
  * [x] magit-svn
  * [x] magit-topgit
* [x] **push branch and tag**
* [x] _get melpa recipes merged_
* [x] _get themes merged_
  * [x] _solarized_ bbatsov/solarized-emacs#189
  * [x] _zenburn_ bbatsov/zenburn-emacs#200
* [ ] _get `next` of extensions merged_
  * [x] magit-annex https://github.com/magit/magit-annex/commit/cf1f2cfe9ed20564c7b5ecdeb8b4cc2d0f89936b
  * [ ] ...
* [x] _get milkypostman/melpa#2641 closed_
* [ ] upload to marmalade
  * [ ] magit
    * [x] with-editor
    * [ ] _get git-commit uploaded_ nicferrier/elmarmalade#110
    * [x] magit-popup
    * [x] magit
[...]

Issue #1428, opened 2014-07-13
Display latest commits


xmaillard:
Hello,

It would be cool if we could display a custom number of latest commits right in the status  buffer.

Regards

PR #1698, opened 2015-01-10
magit-status: add a section with recent commits

vincentbernat:
This section existed in the past but has been removed in favor of
`magit-insert-unpulled-or-recent-commits`. I am disturbed but the fact
that recent commits are only displayed when there is no unpulled nor
unpushed commits. I prefer to have those commits always displayed.

### DIFF
--- a/magit.el
+++ b/magit.el
@@ -4461,15 +4461,18 @@ can be used to override this."
             (apply-partially #'magit-wash-raw-diffs t)
           "diff-index" "--cached" base)))))
 
+(defun magit-insert-recent-commits ()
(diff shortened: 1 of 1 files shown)


tarsius:
Are you sure that ever existed on `master`? Please show me the commit when it was removed. On the `next` branch this function already exists, and in the spirit of getting a release out instead of debugging things that "can't possibly cause any issues", I am no longer backporting to `master`. Please give `next` a try instead. See #1645 for more information.

vincentbernat:
It's what I had understood from #1428. I'll just add the appropriate function in my `.emacs`.

Issue #1586, opened 2014-11-01
Create a fancier magit-copy-as-kill (and give it a better name)


jsyjr:
(Please ignore if the development version already addresses this -- I use the version installed by el-get.)

The description of C-w (magit-copy-item-as-kill) says "Copy sha1 of commit at point into kill ring".  This is alternately unclear or inaccurate.  In reality it copies the displayed portion of the SHA1.  If an abbreviated SHA1 is displayed then that is all you get.  My reading of that doc string lead me to expect a copy of the canonical 40-character SHA1.  My preference would be to have C-w uniformly copy the full SHA1.

Given the pending change to git there has been an active discussion the emacs-devel mailing list about how to refer to commits and the overhead of using copy-paste.  That prompts my second suggestion.  I would like to see 'W' (capital W) generalize C-w to allow copying various fragments associated with the commit at point.  These could include all of --pretty's built-in formats, a selection of the %placeholders and perhaps a means of supplying an arbitrary format string.

tarsius:
Also support killing a commit message as suggested in #1715.

tarsius:
Work on this has began in #2225 and will likely make it into v2.3.

I have renamed the commands. The way I intend to implement "fancier" is to save the value to a revision stack, provided it is a rev of course, and add a command to pop the latest rev and insert a representation at point or at point and eob. The user can customize how exactly the rev is formatted using an option.

PR #1715, opened 2015-01-31
magit-copy-item-as-kill: copy commit messages

abo-abo:
* magit.el (magit-copy-item-as-kill): With point on the commit message,
  copy it to kill ring, with leading indentation removed.

### DIFF
--- a/magit.el
+++ b/magit.el
@@ -7465,7 +7465,18 @@ Non-interactively DIRECTORY is always (re-)initialized."
   (magit-section-action copy (info)
     ((branch commit file diff)
      (kill-new info)
-     (message "%s" info))))
+     (message "%s" info))
+    ((message)
+     (kill-new
+      (mapconcat
+       (lambda (x) (if (> (length x) 4) (substring x 4) x))
+       (split-string
+        (buffer-substring-no-properties
+         (magit-section-beginning it)
+         (magit-section-end it))
+        "\n")
+       "\n"))
+     (message "COMMIT_MSG"))))
 
 (defun magit-ignore-item (edit &optional local)
   "Ignore the item at point.


tarsius:
Generally I don't want to add any new features to `master` anymore. I even make a fuss for bugfixes. See #1645. On the other hand you're a competent elisper I don't want to discourage you from contributing.

This seems like a good idea. But I am not quite sure that's the right way to do it. It depends on what you want to do with the string saved to the kill-ring. Yank it in a commit message buffer, or yank it anywhere? In the former case there is a better way now, see `git-commit-save-message` and `magit-maybe-save-head-message`. Not sure if something equivalent exists on `master` yet. On `next` this would have to save to the `log-edit-comment-ring` and only maybe also to the `kill-ring`.

There is also another open issue about `magit-copy-as-kill`:  #1586. I think it would be better to implement all this in one go and only on `next`, what do you think? By the way I think you should switch to `next`, except for the manual it's mostly done, and you should be able to figure it out.

abo-abo:
> This seems like a good idea. But I am not quite sure that's the
> right way to do it. It depends on what you want to do with the
> string saved to the kill-ring. Yank it in a commit message buffer,
> or yank it anywhere? In the former case there is a better way now,
> see git-commit-save-message and magit-maybe-save-head-message. Not
> sure if something equivalent exists on master too. On next this
> would have to save to the log-edit-comment-ring and only maybe also
> to the kill-ring.

Alright, if you don't want to touch master it's fine. I can keep this
in my config until the new version is released.  It's just that
currently this function does nothing useful when the point is located
on the commit message. It's nice UX-wise to do the expected thing
anywhere, not only on hunks.

My use-case was to copy the commit message before the soft reset with
<kbd>x</kbd>, then add some stuff and yank the commit message.  It
would also be useful to copy the commit message into Changelog, with
some minor modifications.

> By the way I think you should switch to next anyway -- except for
> the manual it's mostly done, and you should be able to figure it
> out.

I tried it just now, and it seems to be quite broken: <kbd>ll</kbd>
doesn't work.

tarsius:
> My use-case was to copy the commit message before the soft reset with <kbd>x</kbd>, then add some stuff and yank the commit message.

Aha! That's what I though -- and there's a much better way to do that now. Just to the reset without bothering saving the HEAD commit message, do your thing and initiate a new commit, and then press `p`. And the message is back -- it was saved for you when you reset to HEAD~. This only works when "discarding" a single commit, so being able to explicitly store the message is also useful.

This is done using the `magit-maybe-save-head-message` I mentioned before, when doing it explicitly this function can server as an example. *Also* saving it in the `kill-ring` is a good idea too. If you do stick to using `next`, then please implement this along these lines.

> It would also be useful to copy the commit message into Changelog, with some minor modifications.

I think that should be a separate command.

> I tried it just now, and it seems to be quite broken: <kbd>ll</kbd> doesn't work.

You need to be a lot more specific than that. There are quite a few people using `next` by now and so far nobody has reported that this quite essential command "doesn't work".

tarsius:
> and there's a much better way to do that now

And it's not even the best... Why do you reset at all? Sometimes it is needed but since you didn't mention the alternative approach also, I thought I would do so myself: see https://github.com/magit/magit/wiki/History-Manipulation for a description of both approaches.

PR #1721, opened 2015-02-03
magit-commit-compose-mail: new command

ueno:
Using `git send-email' from the command-line requires a separate e-mail
sending configuration, which would be cumbersome if one is already using
a mail client on Emacs.  Add new command `magit-commit-compose-mail' for
convenience.

### DIFF
--- a/magit.el
+++ b/magit.el
@@ -82,6 +82,7 @@ Use the function by the same name instead of this variable.")
 (require 'format-spec)
 (require 'grep)
 (require 'help-mode)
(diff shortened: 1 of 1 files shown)


npostavs:
magit is in feature freeze, see #1645.

ueno:
I see, thanks for the information.
it seems the ```next``` branch already has compose-mail support, though it is only for ```git request-pull```.
I withdraw this PR for now.

PR #1735, opened 2015-02-19
Add apply/format patches to help menu

Silex:
Hello,

Not sure the doc string/emplacement is adequate, please comment.


tarsius:
The bindings are correct. I will worry about the placement later. That involves the use of `:max-action-columns` which should likely be replaced with just `:action-columns' (which hasn't been implemented yet).

PR #1750, opened 2015-03-05
use completing-read-ido from the ido-ubiquitous package

tarsius:
`ido-completing-read` is not suitable as a replacement for
`completing-read`, because it lacks essential features and
has bugs.  Instead use the wrapper `completing-read-ido`
from the `ido-ubiquitous` package.

### DIFF
--- a/magit-utils.el
+++ b/magit-utils.el
@@ -47,7 +47,8 @@
 (require 'dash)
 
 (eval-when-compile (require 'ido))
-(declare-function ido-completing-read 'ido)
+(declare-function completing-read-ido 'ido-ubiquitous)
+(defvar ido-ubiquitous-next-override)
 
 (defvar magit-backup-mode)
 (defvar magit-backup-untracked)
@@ -276,14 +277,41 @@ results in additional differences."
 
 (defun magit-ido-completing-read
   (prompt choices &optional predicate require-match initial-input hist def)
-  "Ido-based completing-read almost-replacement."
+  "Ido-based `completing-read' almost-replacement.
+
+`ido-completing-read' is not suitable as a replacement for
+`completing-read', because it lacks essential features and
+has bugs.  Instead use the wrapper `completing-read-ido'
+from the `ido-ubiquitous' package."
   (require 'ido)
-  (let ((reply (ido-completing-read
-                prompt
-                (if (consp (car choices))
-                    (mapcar #'car choices)
-                  choices)
-                predicate require-match initial-input hist def)))
+  (require 'ido-ubiquitous)
+  (let* (;; Pretend these modes are on because otherwise we would
+         ;; end up using regular `completing-read'.  Also ensure all
+         ;; `ido-ubiquitous' fixes to be used, even those that are not
+         ;; enabled or were disabled by the user.
+         (ido-mode t)
+         (ido-ubiquitous-mode t)
+         (ido-ubiquitous-allow-on-functional-collection t)
+         (ido-ubiquitous-fallback-completing-read-function
+          #'completing-read-default)
+         ;; `ido-ubiquitous' calls this "old style" and a "quirk" but
+         ;; it actually is an essential feature which we depend on.
+         (ido-ubiquitous-old-style-default t)
+         ;; Magit properly supports ido, but to do so it does require
+         ;; `ido-ubiquitous', which in turn is configured to prevent
+         ;; that.  Override that, but only when magit's own wrapper
+         ;; (this function) is called, not when `completing-read-ido'
+         ;; is called due to `ido-ubiquitous-mode' being on.
+         (ido-ubiquitous-next-override
+          (remove '(disable prefix "magit-") ido-ubiquitous-next-override))
+         (reply (completing-read-ido
+                 prompt
+                 ;; Unlike `completing-read', `ido-completing-read'
+                 ;; and `completing-read-ido' cannot handle alists.
+                 (if (consp (car choices))
+                     (mapcar #'car choices)
+                   choices)
+                 predicate require-match initial-input hist def)))
     (or (and (consp (car choices))
              (cdr (assoc reply choices)))
         reply)))


tarsius:
@DarwinAwardWinner I have already merged this, but could you please have a look anyway and tell me whether you think this is sane.

Also I would very much appreciate it if you could get at least the bugfixes, which are currently implemented as advices, merged into `ido` itself. Later maybe even the mode.

DarwinAwardWinner:
Looks like you've basically done everything you'd need to do to make this work. Obviously that function wasn't really designed to be called as a library function by other code, as you can see from all the hoops you had to jump through. If more people are going to want to do this kind of thing, I should probably restructure the code to facilitate that better: https://github.com/DarwinAwardWinner/ido-ubiquitous/issues/64. I'll let you know before I make any breaking changes in that department. 

Meanwhile, I have a few tweaks I would make to your current solution, but I don't have time to detail them here. I'll reply again later after work.

DarwinAwardWinner:
Ok, a few corrections:

* The variable `ido-ubiquitous-next-override` just needs to be set to one of the symbols `disable`, `enable`, or `enable-old`. You probably just want to let-bind `ido-ubiquitous-next-override` to  `enable-old` and omit the let-binding of `ido-ubiquitous-old-style-default`.
* It's probably unnecessary to let-bind `ido-ubiquitous-fallback-completing-read-function`. Unless the user has customized `completing-read-function` before activating ido-ubiquitous mode, the fallback will already be set to `completing-read-default`, and if it isn't that should probably be respected anyway.

So the above would reduce it to:

```lisp
(defun magit-ido-completing-read
    (prompt choices &optional predicate require-match initial-input hist def)
  "Ido-based `completing-read' almost-replacement.
`ido-completing-read' is not suitable as a replacement for
`completing-read', because it lacks essential features and
has bugs. Instead use the wrapper `completing-read-ido'
from the `ido-ubiquitous' package."
  (require 'ido)
  (require 'ido-ubiquitous)
  (let* (;; Pretend these modes are on because otherwise we would
         ;; end up using regular `completing-read'. Also ensure all
         ;; `ido-ubiquitous' fixes to be used, even those that are not
         ;; enabled or were disabled by the user.
         (ido-mode t)
         (ido-ubiquitous-mode t)
         (ido-ubiquitous-allow-on-functional-collection t)
         ;; `ido-ubiquitous' calls this "old style" and a "quirk" but
         ;; it actually is an essential feature which we depend on.
         (ido-ubiquitous-next-override 'enable-old)
         (reply (completing-read-ido
                 prompt
                 ;; Unlike `completing-read', `ido-completing-read'
                 ;; and `completing-read-ido' cannot handle alists.
                 (if (consp (car choices))
                     (mapcar #'car choices)
                   choices)
                 predicate require-match initial-input hist def)))
    (or (and (consp (car choices))
             (cdr (assoc reply choices)))
        reply)))
```

* Are you sure you need old-style default mode? This is where you don't pass a DEF argument and `completing-read` returns an empty string to indicate that the default should be chosen. If you know what the default argument is anyway, can you just rewrite all your calls to `completing-read` to pass it? (If you do this, change the override from `enable-old` to `enable`.
* I notice you've written "`completing-read-ido` cannot handle alists." That's something I could probably fix, or at least trigger the fall back for.

tarsius:
> Are you sure you need old-style default mode?

It's required for e.g. `magit-branch-set-upstream` from the `next` branch, which prompts with "Change upstream to branch (empty to unset)". I could just use "" as the default, but then that would show up in the list of completions I believe. But regardless since `completing-read` returns the empty string on empty input, `magit-completing-read` should to, and so all possible `magit-completing-read-functions` have to so to (or be patched to do). Otherwise contributors and authors of extensions to magit, who use regular completion would end up implementing queries which don't work properly with `ido` (and maybe `helm` also, still have to check that).

> I notice you've written "completing-read-ido cannot handle alists." That's something I could probably fix

I was thinking the same, but thought I would start by suggesting the changes, I feel are more essential (this one is easy to work around).

> or at least trigger the fall back for.

I would strongly prefer a fix, similar to what you do for collections that are a function.

tarsius:
I agree with the other suggestions and have merged #1751 (with some slight comment and whitespace changes. Note to others: that patch uses `ido-ubiquitous-with-override`, so it isn't exactly the same as the above code.

npostavs:
> [`ido-ubiquitous-next-override` = `enable-old` is] required for e.g. magit-branch-set-upstream from the next branch, which prompts with "Change upstream to branch (empty to unset)".

I strongly prefer not to have `enable-old`. With the `enable-old` setting, if I press just `RET` at the "Push <foo-branch> to: ", it looks like I chose `origin/<foo-branch>` but then I get told "Nothing selected".

Even with the "new" style, the user can still pass empty by hitting `C-j` instead of `C-m` (aka `RET`). This is the common behaviour across all ido prompts.

DarwinAwardWinner:
If you need different overrides for different function, I think you could just have magit add specific ido-ubiquitous overrides for individual functions using `ido-ubiquitous-apply-function-override`, and then take out the line in `magit-ido-completing-read` that sets `ido-ubiquitous-next-override` (or change it to `(ido-ubiquitous-next-override (or ido-ubiquitous-next-override 'enable-old))`.

tarsius:
Yes, I noticed that too and it is quite annoying. So far I had it on my implicit TODO-list known as "it's so annoying, I certainly won't forget about it". But could you please add a new issue to track this more formally :-)

DarwinAwardWinner:
I once had dreams of writing some code that would look at where `completing-read` was being called and try to determine whether the result was being compared to an empty string and auto-switch to compatibility mode, but then I realized that was insane and it would be faster to just manually list all the functions that required old-style completion. :)

npostavs:
> But could you please add a new issue to track this more formally :-)

Okay, see #1781.

DarwinAwardWinner:
By the way, as far as I can tell, `completing-read-ido` actually does support alists, unlike stock ido completion.

PR #1751, opened 2015-03-06
Fixup ido-ubiquitous usage

DarwinAwardWinner:
As discussed in https://github.com/magit/magit/pull/1750#issuecomment-77513698

### DIFF
--- a/magit-utils.el
+++ b/magit-utils.el
@@ -292,26 +292,18 @@ from the `ido-ubiquitous' package."
          (ido-mode t)
          (ido-ubiquitous-mode t)
          (ido-ubiquitous-allow-on-functional-collection t)
-         (ido-ubiquitous-fallback-completing-read-function
-          #'completing-read-default)
-         ;; `ido-ubiquitous' calls this "old style" and a "quirk" but
-         ;; it actually is an essential feature which we depend on.
-         (ido-ubiquitous-old-style-default t)
-         ;; Magit properly supports ido, but to do so it does require
-         ;; `ido-ubiquitous', which in turn is configured to prevent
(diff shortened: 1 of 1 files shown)


tarsius:
Merged with some slight changes to comments and indentation: 9066e177ca9fd70c5f3c247ef8389aca65c7c4f5. Thanks a lot.

tarsius:
I had to put back

```lisp
(ido-ubiquitous-next-override
  (remove '(disable prefix "magit-") ido-ubiquitous-next-override))
```

Without this `completing-read-ido` did always use the fallback.

See 8fca5d08c76a7f19c6d7acf8cb82ab8c93786a8f.

DarwinAwardWinner:
The value of `ido-ubiquitous-next-override` should be one of `disable`, `enable-old`, or `enable`, or nil. The `ido-ubiquitous-with-override` macro simply let-binds this variable, but I see you were having issues with that since you don't want a hard dependency on ido-ubiquitous, so let-binding it yourself is fine. Anyway, values other than the above will will do "something", but I'm not sure exactly what. I'm pretty sure that just let-binding it to `enable-old` should do what you want (including forcing it to work on functional collections).

tarsius:
That's what I ended up doing in c6ae423baa82eb6fefa55fcc0bf5dad37b9d35cb.

DarwinAwardWinner:
You should also be able to get rid of the let binding for `ido-ubiquitous-allow-on-functional-collection`. The override should take care of that too: https://github.com/DarwinAwardWinner/ido-ubiquitous/blob/bleeding-edge/ido-ubiquitous.el#L590-L594

tarsius:
Okay, removed that too. Thanks.

PR #1754, opened 2015-03-08
Add --subject-prefix option to magit-patch-popup

andschwa:
Helps existing feature.

If this violates feature freeze, feel free to reject.

### DIFF
--- a/magit-remote.el
+++ b/magit-remote.el
@@ -253,6 +253,7 @@ branch as default."
               (?r "In reply to"      "--in-reply-to=")
               (?v "Reroll count"     "--reroll-count=")
               (?s "Thread style"     "--thread=")
+              (?S "Subject Prefix"   "--subject-prefix=")
               (?U "Context lines"    "-U")
               (?M "Detect renames"   "-M")
               (?C "Detect copies"    "-C")


tarsius:
It's okay to do things like this now. But in this particular case I have chosen to initially only add the most essential arguments. There are many more arguments all of which are used by someone but listing them all would be a bit overwhelming I'm afraid.

(On the other hand I've recently made some changes to the popup code, which would allow to better group arguments, so maybe I will reconsider. I always intended to go over this gain eventually anyway, but that is something I had scheduled for after the release).

For that reason  commit (63b55c1d5de677ef44e0375fb89af7a0f4471716) which added this popup contains:

> If I think that some important argument is missing, then please add it locally using `magit-define-popup-{switch,option}`.

In this case

```lisp
(magit-define-popup-switch 'magit-patch-popup %S "Subject Prefix" "--subject-prefix=" nil ?s)
```

andschwa:
Oh cool! I didn't know I could do that, thanks :smile: 

I would still recommend including this on the menu when this popup-patch is gone back over, because the subject prefix is pretty frequently used in mailing list communities. I don't use format-patch or send-email without it!

andschwa:
Yay Lisp.
``` elisp
(magit-define-popup-option 'magit-patch-popup ?S "Subject Prefix" "--subject-prefix=")
```
Gives me just what I needed.

tarsius:
Ups, yes option, I have corrected my suggestion. You might want to keep the last two arguments.

PR #1755, opened 2015-03-09
magit-git-command: create process buffer if needed

npostavs:
Was getting `Wrong type argument: bufferp, nil` when calling `magit-git-command` before any other git-running commands.

eg:
`emacs -Q -L . -L ../dash/ -l magit -f magit-status --eval '(magit-git-command "status" default-directory)'`


tarsius:
Thanks

Issue #1747, opened 2015-03-04
"git send-email" from magit


unhammer:
We already have "git send-email", but this requires adding various settings to ~/.gitconfig and installing perl scripts; it'd be much nicer if I could just press something like "N m" and it makes a format-patch out of the unpushed commits and puts it into a new email in message-mode.

tarsius:
I don't remember - do we already have that on `master`? If so, and that's what you are using, please update to `next` and try whether that satisfies your needs.

Silex:
On `next`, try the `W` key (`magit-patch-popup`)

unhammer:
"W p" runs format-patch, but without putting it into an email, while "W r" does request pull (which assumes I have some remote to push to first). I was talking about putting the format-patch into a compose-mail, so I can just fill in the To-field and hit send.

andschwa:
:+1:

andschwa:
If one were to try implementing `send-email` on next, should send-email be a sub-command, like "W s" or a new command, `magit-email-popup`? Git `send-email` has at least one required option, `--to`, and the range.

Can a particular Magit popup action make a certain option required (and prompt for it) when it's not otherwise, or should the to field just be asked for in interactive?

tarsius:
(Good questions. Generally speaking users either hate the popups (luckily just a few) or think they are totally awesome - but this goes straight to the limitations of the current implementation :-)

Ah, yes that's another reason why I am hesitating to add additional arguments to the patch popup: if we add send-mail to that popup, would the additional arguments still make sense. In the case of what you suggested in #1754, I am *guessing* that would be useful here too.

More likely though this needs a new popup. Look at blaming for an example of such a popup-with-just-one-action.

(For `idefix` the successor to `magit-popup` I plan to make it possible to define such things with just one command, by making it more like prefix arguments (except that they are *infix* arguments) and then optionally merge multiple commands that have such arguments into a shared popup (but it would still be possible to invoke the individual commands directly with just their own arguments shown in their own popup).

> Can a particular Magit popup action make a certain option required (and prompt for it) when it's not otherwise, or should the to field just be asked for in interactive?

Generally speaking this should be done inside `interactive`. If we go with a separate popup, then using visible arguments might also make sense. That should work with `--to` without any problems, but for positional arguments like "the range" there is no explicit support. It should be possible to fake it though. You would then have to post-process the arguments inside `interactive` and maybe even add a "helper mode", like for the "sequence popups".

[...]

PR #1756, opened 2015-03-09
Add send-email to magit-patch-popup

andschwa:
Per feature request #1747.

This isn't done, but I figured a PR would be a better place to continue discussing it than the issue.

* [x] Accept files and directories in addition to range
* [ ] Complete for files
* [ ] Add subject option

This adds `magit-read-repeated-options` for repeating an option (like `--to=`) using comma-separated input. It would be much better to use this as the reader for the popup option.

Currently `magit-send-email` interactively prompts for To and Cc fields, and should probably prompt for Subject. The man page says subject is prompted for, but it actually seems to default to the `subject-prefix` that `format-patch` uses.

There are thoughts about making `send-email` a separate popup from `format-patch`, but they do share some options.

### DIFF
--- a/magit-git.el
+++ b/magit-git.el
@@ -980,6 +980,17 @@ Return a list of two integers: (A>B B>A)."
                                  (magit-remote-at-point)
                                  (magit-get-remote))))))
 
+(defun magit-read-repeated-options (prompt &optional option)
+  "Read multiple comma-separated strings and repeat option for each.
+
+PROMPT is prepended to each token in the split input string in
+addition to being used as the prompt.
+
+If the empty string is given, no options are returned."
+  (mapcar (lambda (x) (and (not (string= "" x))
+                           (concat (or option prompt) x)))
+          (split-string (read-from-minibuffer prompt) ",")))
+
 ;;; Variables
 
 (defun magit-get (&rest keys)
--- a/magit-git.el
+++ b/magit-git.el
@@ -989,9 +989,8 @@ OPTION is prepended to each token of the split string, resulting
 in repeating options. If not given, prompt is used instead.
 
 If the empty string is given, no options are returned."
-  (mapcar (lambda (x) (and (not (string= "" x))
-                           (concat (or option prompt) x)))
-          (split-string (read-from-minibuffer (concat prompt ": ")) ",")))
+  (mapcar (lambda (x) (concat (or option prompt) x))
+          (completing-read-multiple (concat prompt ": ") nil nil nil)))
 
 ;;; Variables
 
--- a/magit-git.el
+++ b/magit-git.el
@@ -989,8 +989,8 @@ OPTION is prepended to each token of the split string, resulting
 in repeating options. If not given, prompt is used instead.
 
 If the empty string is given, no options are returned."
-  (mapcar (lambda (x) (concat (or option prompt) x))
-          (completing-read-multiple (concat prompt ": ") nil nil nil)))
+  (--map (concat (or option prompt) it)
+         (completing-read-multiple (concat prompt ": ") nil nil nil)))
 
 ;;; Variables
 
--- a/magit-git.el
+++ b/magit-git.el
@@ -980,17 +980,23 @@ Return a list of two integers: (A>B B>A)."
                                  (magit-remote-at-point)
                                  (magit-get-remote))))))
 
-(defun magit-read-repeated-options (prompt &optional option)
-  "Read multiple comma-separated strings and repeat option for each.
+(defun magit-repeat-options (opt &optional required)
+  "Convert --opt=a,b,c into --opt=a --opt=b --opt=c.
 
-PROMPT is concatenated with a colon and passed to read-from-minibuffer.
+When called within a popup, this retrieves the current arguments
+matching OPT and repeats the option for each comma-separated
+value given to the argument.  The argument may be specified
+multiple times.
 
-OPTION is prepended to each token of the split string, resulting
-in repeating options. If not given, prompt is used instead.
-
-If the empty string is given, no options are returned."
-  (--map (concat (or option prompt) it)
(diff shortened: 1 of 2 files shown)


tarsius:
For reading a comma separated list of values use `completing-read-multiple`, see `magit-log-read-files` and its use in the log popup for an example.

andschwa:
Oddly, I'd looked at `magit-log-read-files` and started this whole thing with a similar `mapconcat`, then decided against it, but never brought back `completing-read-multiple`. A fresh look is quite enlightening!

With the limit to files option, the strings are all appended comma-separated after `-- `; whereas with `send-email` the option itself needs to be repeated for each entry, so `magit-read-repeated-options` is still useful. It is much better with `completing-read-multiple` to get the list of strings now, and it gets rid of any empty strings!

tarsius:
> With the limit to files option, the strings are all appended comma-separated after `-- `; whereas with `send-email` the option itself needs to be repeated for each entry, so `magit-read-repeated-options` is still useful.

Kinda. We could also allow something like `--to=a"@a.a,b@b,b"` in the popup and then only inside `interactive` convert `"@a.a,b@b,b"` to `("--to=a@a.a" "--to=b@b,b")`. Given the current popup implementation this is the best we can do if we want to display this in the popup, which I think we should.

Also not that I recently made some changes to `magit-popup.el`  which allow faking subsections: e.g.

```lisp
:options '((?a "a" "--a=")
           (?b "b" "--b=")
           nil ; insert an empty line
           (?c "c" "--c=")
           (?d "d" "--d="))
```

or even

```lisp
:options '("Actions" ; the first item has to be a string, usually just use the default heading
           (?a "a" "--a=")
           (?b "b" "--b=")
           "Actions for foobar only"
           (?c "c" "--c=")
           (?d "d" "--d="))
```

andschwa:
Okay I can use `(magit-current-popup-args "--to=")` to get the option string `--to=foo,bar`, and I think now I need to `split-string` that on `,` and `--map` "--to=" onto each element of its `cdr` (since the `car` already has the option and I don't want to have to remove it).

andschwa:
@tarsius was this sort of what you were thinking?

P.S. I hope you don't mind, but I rebased the branch (since it's not merged yet).

tarsius:
Yes, that's the idea. And no I don't mind the rebasing. And since you rewrite history anyway, you might as well squash the last commit.

andschwa:
I'll just squash the whole branch when it's finished. Just a few questions:

Do you have any idea how I could simultaneously complete for files/directories *and* revisions in `magit-read-range-or-commit`?

Would you prefer a different name or location for the `magit-repeat-options` function? I also feel like it could be improved. I didn't like how I had to filter and then subsequently `setq magit-current-popup-args`.

I like the idea of the subsection, except for having to filter the subsection's args out of actions where they're not supported. The only options I could see adding are `--subject` and `--dryrun`, the former isn't really used over `--subject-prefix`, at least not in mailing list circles, and the latter seems like something to leave to the user in their `.emacs`. All SMTP settings are usually saved in a user's `.gitconfig`, and the remaining options are either obscure or meant to be used interactively (e.g. `--annotate` and `--compose`). In summary: I don't see a need for a `send-email` options subsection.

tarsius:
> Do you have any idea how I could simultaneously complete for files/directories *and* revisions in `magit-read-range-or-commit`?

Well that would be a different function then. If an existing function is modified then it should probably be `magit-read-commit` instead. To do this we need some kind of context aware completion. `pcomplete` seems to provide something like that, but it seems to be intended for in-buffer completion, not minibuffer completion. But maybe it can be used there too. Would be great if you could investigate this.

> Would you prefer a different name or location for the `magit-repeat-options` function?

It should probably be placed in `magit-popup.el` and be named something like `magit-popup-split-option`

> I also feel like it could be improved. I didn't like how I had to filter and then subsequently `setq magit-current-popup-args`.

It should not manipulate that variable, instead it should return a new list, which can then be let-bound inside `interactive`.

> I like the idea of the subsection, except for having to filter the subsection's args out of actions where they're not supported.

There is no way around that right now. That's why I initially suggested you use a separate popup.

> The only options I could see adding are `--subject` and `--dryrun`, the former isn't really used over `--subject-prefix`, at least not in mailing list circles, and the latter seems like something to leave to the user in their `.emacs`. All SMTP settings are usually saved in a user's `.gitconfig`, and the remaining options are either obscure or meant to be used interactively (e.g. `--annotate` and `--compose`).

No comment here. I didn't add this command myself because I didn't want to investigate these things yet.

> In summary: I don't see a need for a `send-email` options subsection.

Ah, then we don't need a separate popup after all.

tarsius:
I have played around with `git send-email` some more and have - once more - come to the conclusion, that it is not worth implementing some minimal viable wrapper, like you are suggesting here. The real challenge to users is getting the configuration right, not typing `git send-email from..to` in a shell. So I think a throughout tutorial on how to configure this would be more useful, even to Magit users, than a small wrapper.

I have opened a new issue  #1800 about implementing extended support for email and pull request collaboration eventually.

andschwa:
Sounds good. Thanks for helping me play around with Lisp :D

I'm not sure I agree with helping setting up SMTP, the only trouble with it is that it's incredibly provider dependent. The Git documentation for SMTP I found more than sufficient; it's just finding your server's settings that seems to give people trouble. But we'll cross that road much later.

tarsius:
I think that instead of letting the user specify the recipients in the popup it would be better to force her to configure this per repository (in .git/config). But even if we do that, it would still be necessary to catch the various questions Git might ask due to incomplete configuration. ETC. :-)

tarsius:
I think there are *many* corner cases neither of us has even run into yet. But sine you seem to agree that this should be delayed, I will stop arguing why that is so now :-)

PR #1760, opened 2015-03-11
Do not quote magit-blame-mode-lighter

alezost:
Currently a lighter for blame-mode does not appear in the mode-line because the lighter variable is quoted.


tarsius:
Thanks

PR #1790, opened 2015-04-02
magit-blame-mode: shadow view-mode bindings

### LINKED ISSUE #1780: [next][patch(es)] With view-read-only = t, magit-blame-mode fights for view-mode over keybindings
With `view-read-only` = `t`, the `view-mode` keybindings can shadow `magit-blame-mode`'s, for example, `q` is bound to `View-quit` (which leaves the buffer still in `magit-blame-mode`, and buries it) instead of `magit-blame-quit`.

I thought of 2 possible solutions:

- Set `view-read-only` buffer locally to `nil` (could be a problem if the user has setup their own buffer-local settings (or am I over thinking this?)) https://github.com/magit/magit/compare/magit:next...npostavs:blame-no-view?expand=1

- Make sure `magit-blame-mode` precedes `view-mode` in `minor-mode-map-alist` so `magit-blame-mode` shadows `view-mode` rather than the other way around. Seems less problematic, though having to manipulate `minor-mode-map-alist` feels a bit silly. https://github.com/magit/magit/compare/magit:next...npostavs:blame-over-view?expand=1

npostavs:
```
If view-read-only is set to `t', magit-blame-mode will turn it on when
it calls read-only-mode. To stop view-mode from shadowing
magit-blame-mode's keybindings, ensure that magit-blame-mode's entry in
minor-mode-map-alist precedes view-mode's.
```

fix #1780.

### DIFF
--- a/magit-blame.el
+++ b/magit-blame.el
@@ -115,6 +115,16 @@ and then turned on again when turning on the latter."
     map)
   "Keymap for `magit-blame-mode'.")
 
+(defun magit-blame-put-keymap-before-view-mode ()
+  "Put `magit-blame-mode' ahead of `view-mode' in `minor-mode-map-alist'."
+  (--when-let (assq 'magit-blame-mode
+                    (cl-member 'view-mode minor-mode-map-alist :key #'car))
+    (setq minor-mode-map-alist
+          (cons it (delq it minor-mode-map-alist))))
+  (remove-hook 'view-mode-hook #'magit-blame-put-keymap-before-view-mode))
+
+(add-hook 'view-mode-hook #'magit-blame-put-keymap-before-view-mode)
+
 (defvar-local magit-blame-buffer-read-only nil)
 (defvar-local magit-blame-cache nil)
 (defvar-local magit-blame-disabled-modes nil)


tarsius:
Don't you agree that this is more readable?

```lisp
(defun magit-blame-put-keymap-before-view-mode ()
 "Put `magit-blame-mode' ahead of `view-mode' in `minor-mode-map-alist'."
 (--when-let (assq 'magit-blame-mode
                   (cl-member 'view-mode minor-mode-map-alist :key #'car))
   (setq minor-mode-map-alist
         (cons it (delq it minor-mode-map-alist))))
 (remove-hook 'view-mode-hook #'magit-blame-put-keymap-before-view-mode))
```

npostavs:
Yeah, I suppose I'm not in the habit of using the anaphoric macros. 

I think the explicit names in the `let*` version help understand what's going on, but maybe not enough to outweigh the verbosity in this case.

tarsius:
> Yeah, I suppose I'm not in the habit of using the anaphoric macros.

It's not so much about `--when-let` as about not using variables that are only used once. The most explicit variant would have been `(-when-let (misplaced-blame-entry ...) ...)` (or `(let ((misplaced-blame-entry)) (when misplaced-blame-entry ...)`.

> I think the explicit names in the `let*` version help understand what's going on, but maybe not enough to outweigh the verbosity in this case.

I have an every increasing dislike for that approach. It's not only more verbose, I find it to be misleading, both when reading the code and when writing it. Take this as example.

```lisp
(let ((very-descriptive-name (equally-explicit-function-name input-value)))
  ...
  (use-only-once very-descriptive-name))
```

vs.

```lisp
...
(use-only-once (equally-explicit-function-name input-value))
```

Using a variable gains exactly nothing but the reader has to read all of `...` to know whether `very-descriptive-name` is used more than once. (Well at least for code that I have written, you can be fairly certain about that. I would only ever call the same function with the same input within the same code block twice, if doing so was extremely cheap, and avoiding to do it (e.g. by using a variable), would make the code much much uglier (which very rarely is the case)). 

When writing code using intermediate names can lead to doing more than necessary or doing something inefficiently, like you have done here (not that efficiency matters here, but only doing what is necessary does). If you had written the "retrieval code" in place then you would probably have used something like `cl-member` too - instead of walking the list twice, using `assq` and then `memq`.

When experimenting I sometimes start out with assigning names to intermediate steps too, but when I am done determining how to retrieve a value, then I usually convert to a more functional flow. Provided the user is used to read lisp, i.e. knows when to start reading "at the end", then that should not be a problem.

tarsius:
In this example it really doesn't matter much. But Magit used to contain *many* instances of things like this which have lead to (to me) unreadable code. And in many cases that code was also inefficient, and only after stopping to name the intermediate names, was I even able to see why it was inefficient. So I try to avoid that now :-)

npostavs:
> `(very-descriptive-name (equally-explicit-function-name`

In this case I was thinking the function name (`memq` or `cl-member`) didn't make explicit that we are interested in the tail specifically. But maybe that should be considered "obvious" to a lisper.

> Using a variable gains exactly nothing but the reader has to read all of ... to know whether very-descriptive-name is used more than once.

I hadn't considered that. How important is it for the reader to know how many times it's used though?

> If you had written the "retrieval code" in place then you would probably have used something like cl-member too - instead of walking the list twice, using assq and then memq.

Oh, shucks, I was aware of the double list walking when writing the original, but I somehow forogot about `cl-member` with `:key`. I've probably spent too long avoiding `cl` functions.

> In this example it really doesn't matter much. But Magit used to contain many instances of things like this [...]

Right, and consistency does matter, so I've amended the commit to use your suggested style. I'll try to avoid one-use names in the future and see how it feels. Thanks for the critique.

tarsius:
Thanks. And your contributions are always welcome.

Just though that, since I implicitly said in the `1.4.0` that former maintainers were to lax when accepting contributions, I should act accordingly :-)

PR #1794, opened 2015-04-03
.travis.yml: set EMACSBIN in calls to make

npostavs:
should let travis test pass


tarsius:
Could you please do that on the `master` branch please. I will then take care of the merge.

PR #1798, opened 2015-04-05
magit-reverse-files: pass file names to prompt

npostavs:
The (mapcar #'magit-section-value) call was dropped in
d0e04a140d4042976a1d654a45670eea24b54669.


tarsius:
Thanks.

Issue #1803, opened 2015-04-11
Upgrade warning message is too intrusive; magit-last-seen-setup-instructions


benma:
Hi

I just upgraded magit on melpa and am greeted with this message on emacs startup:

```You have just updated to version 1.4.0 of Magit, and have to
make a choice.```

To make that disappear, it instructs me to add this to my config:

```(setq magit-last-seen-setup-instructions "1.4.0")```

This behavior is too intrusive and annoying. The package and this message are not so important to force everyone to deal with this. Making people aware of `(setq magit-auto-revert-mode nil)` is not important enough to annoy people with this manual intervention, especially since it has been the default behavior before.

I strongly dislike this move. Imagine if every package started the habit of forcing the user to modify their config files to get rid of intrusive messages. If it is a bad idea for other packages, it is a bad idea for this package.

I suggest just dropping this message and config variable.

At the very least, please display the message only once and never again, without forcing the user to suppress the message manually.

tarsius:
Okay now we have this situation:

1. I, the maintainer of the package, think that it is important that this feature is enabled, because otherwise many things would not work as intended me and also because by now it is expected by most users.

2. One user comes along and argues that it is a very very dangerous feature, and that that enabling it by default is very bad. See #1783.

   I think this is only partially true, because (a) the action effect is reversible and (b) its effect is only undesirable when using a workflow which, in my opinion, doesn't make much sense when using Magit. 

   Never-the-less I do agree that this can take a user by surprise, especially because the effect of the feature can go unnoticed until it is to late to undo it. Also I am in the final steps of creating the first release in a very long time. I chose to inform users about the issue. This way most users will continue to use Magit as intended while those who would be harmed by it, are given the opportunity to turn it off, before anything bad happens.

   While I think this is an important feature, it also isn't absolutely essential, and if it conflicts with the way one uses Emacs, then it is perfectly fine to disable it. And I don't want to responsible for dataloss. (You got annoyed because I am forcing you to set a variable - imagine how annoyed someone would get because I am responsible for them losing valuable work).

3. Another user comes along claiming it is wrong of me to force users to make a choice.

   The way I see it, I have no other choice. Turning the feature of by default is a no-go because doing so would break many workflows. Also many users don't read release notes and faqs. So turning it of would cause many bug reports which I then would have to spend my time on.

[...]

PR #1806, opened 2015-04-13
don't spam "revert inhibited" in blame mode

tprk77:
Hi, so I know Magit is in feature freeze, but I have a small bug fix for the `next` branch. I would
consider it a bug fix anyway:

If the user has turned on `global-auto-revert-mode`, and then does a `magit-blame`, the user will
start getting the message "Reverting (buffer) inhibited due to magit-blame-mode" every few seconds
(depending on the value of `auto-revert-interval`). This seems pretty spammy to me. It's especially
annoying when you are trying to type in the minibuffer, and this keeps popping up.

So could we remove this message? My opinion is that it does more harm than good. The original author meant well I'm sure, but I'm guessing they didn't use `global-auto-revert-mode`, or perhaps their `auto-revert-interval` is much higher then mine.

### DIFF
--- a/magit-blame.el
+++ b/magit-blame.el
@@ -169,8 +169,7 @@ and then turned on again when turning on the latter."
 (defadvice auto-revert-handler (around magit-blame activate)
   "If Magit-Blame mode is on, then do nothing.
 See #1731."
-  (if magit-blame-mode
-      (message "Reverting %s inhibited due to magit-blame-mode" buffer-file-name)
+  (unless magit-blame-mode
     ad-do-it))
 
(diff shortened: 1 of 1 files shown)


tprk77:
Maybe I did this a little backwards, should I have made an issue before the pull request? Also, please let me know if you would rather fix this a different way. Any suggestions would be welcome.

npostavs:
> Maybe I did this a little backwards,

You didn't do anything wrong, maintainer is just offline/busy for a few days.
https://github.com/magit/magit/issues/1803#issuecomment-91959071

tarsius:
Thanks.

tprk77:
Thank *you!* I love this extension, and the `next` branch is even better.

PR #1823, opened 2015-04-24
Copyedit documentation

seh:
Function `magit-maybe-show-setup-instructions` presents a warning and
some configuration advice when the `magit-last-seen-setup-instructions`
variable's value suggests that the user hasn't seen it yet. Adjust the
language for an American English-appropriate tone.

### DIFF
--- a/magit.el
+++ b/magit.el
@@ -7782,28 +7782,27 @@ You have just updated to version 1.4.0 of Magit, and have to
 make a choice.
 
 Before running Git, Magit by default reverts all unmodified
-buffers which visit files tracked in the current repository.
-This can potentially lead to dataloss so you might want to
-disable this by adding the following line to your init file:
+buffers that visit files tracked in the current repository.  This
+can potentially lead to data loss, so you might want to disable
+this by adding the following line to your init file:
 
   (setq magit-auto-revert-mode nil)
 
-The risk is not as high as it might seem.  If snapshots on Melpa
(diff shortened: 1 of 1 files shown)


seh:
I realized after the fact that I started this change atop the `magit/next` branch, as opposed to `magit/master` as suggested in aa72a592b27edcda326255c4310cc4b60f5bbd4e. Please let me know if you want to try again with a different request.

npostavs:
I think you just need to rebase to `master` then `push -f`.

tarsius:
I fixed up the commit slightly to preserve the original (and quite intentional) filling.

In the process I learned that `--no-commit` does not what I would want it to do. With regards to this `git cherry-pick` is not consistent with other commands which allow "applying" a commit but also edit it, *without* having to manually preserve the message. I will look into a workaround.

I also realized that `magit-commit-extend` probably should not preserve the `GIT_COMMIT_DATE` when the `AUTHOR` is not "me".

seh:
I too find _git cherry-pick_ confusing in that regard. I guess it offers the `CHERRY_PICK_HEAD` ref that you can use to grab the commit message via

> _git commit (-C |--reuse_message=)CHERRY_PICK_HEAD_

, or if you want to edit the message, you can use the similar

> _git commit (-c |--reedit_message=)CHERRY_PICK_HEAD_

command.

PR #1830, opened 2015-04-29
Avoid unnecessary work to improve performance

tarsius:
Avoid re-highlighting when moving within a section, un-/highlighting of sections whose bodies are invisible, and many more such improvements.

Unfortunately it was a prerequisite for these changes to drop support for Emacsen before `24.4`.


tarsius:
If your distribution does not have Emacs 24.4, then please see https://github.com/magit/magit/wiki/FAQ#my-distribution-doesnt-have-emacs-244. Feel free to comment here.

PR #1837, opened 2015-05-04
DOC : Correct load path in install directions

tacaswell:
There is a missing `.` from the path to add to the `.emacs` as compared to the directions in the previous code blocks.


tarsius:
Thanks

PR #1845, opened 2015-05-09
Makefile: follow links in finding libraries

ilohmar:
If the ELPA_DIR is a link to another directory, ``dash`` etc were not found.  With the ```-L``` option find succeeds again.  I cannot think of any drawbacks this has, since emacs finds the library one way or the other.


tarsius:
Thanks-

Issue #1859, opened 2015-05-19
`make lisp` produces invalid autoloads file


Malabarba:
To install the next branch, I ran `make lisp` in the repo directory as instructed in the readme.
This generated an invalid autoloads file.

All of the autoloads pointed to the files like `"../../../../.emacs.d/packages/magit/magit-extras"` which is one level too deep. Changing all of them to things like `"magit-extras"` was enough to fix it.
This happened even though I ran the make command inside the `magit` directory

npostavs:
Are there any symlinks involved?

Malabarba:
Yes, the `.emacs.d` is a symlink to `~/Otherdir/emacs.d`.

npostavs:
Maybe it would work if you ran it from the non symlink location (but make sure you delete the bad one first)?

Malabarba:
Yes, that does generate a valid file. Thanks for the suggestion.

tarsius:
That should fix it.

tarsius:
It might be worth fixing this in Emacs itself too. It wouldn't surprise me if `package.el` had problems to when `~/.emacs.d` is a symlink.

Malabarba:
My `.emacs.d` (and my `elpa/` by extension) has always been inside a symlink and it never caused me any problems with `package.el` and looking at this part of the code I don't see any mentions of truename. All it does is call `update-directory-autoloads` with the package's directory.

But then, I see your setup is a bit more complicated because of `make`, so maybe that's not enough for your case.

PR #1866, opened 2015-05-22
Makefile: make paths work on cygwin with NTEmacs

ilohmar:
cygwin make passed cygwin paths to the emacs binary.  As a result, the library paths were not effective and autoloads could not be built.

This change makes magit's Makefile translate the paths in question, such that it builds properly on cygwin with NTEmacs.  It does not affect any non-cygwin system at all.  I did not test with cygwin's Emacs, but cygwin-built binaries generally can handle the mixed dos-drive-letter path notation as command-line arguments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,15 @@ ifeq "$(DASH_DIR)" ""
   DASH_DIR = ../dash
 endif
 
-LOAD_PATH ?= -L . -L $(CL_LIB_DIR) -L $(DASH_DIR)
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks

tarsius:
Please see #1859 and the fix in https://github.com/magit/magit/commit/8f5d1d2fc4e541fa26b4719b10bf21039c15b009. That should make setting CURDIR_SANE unnecessary. Could you please verify.

PR #1870, opened 2015-05-24
magit-show-commit: Fix commit check

kyleam:
Check passed rev rather than master.

----

Just out of curiousity, in what circumstances did the previous check (fc3355a25af4cf88aa4b9825ee1901b0d901427c) fail?


tarsius:
`magit-git-failure` returns t when the exit code is 1, else nil. `git cat-file commit COMMIT`s exit code is either 0 or 128.

kyleam:
I see.  Thanks for the explanation.

Issue #1848, opened 2015-05-12
Refined background does not disappear when selecting region


uu1101:
Refined background highlight (that of the word SHOW in the screenshot) remains even when outside the selected region, where before it was removed as well. Compare following screenshots of 3eb8982 (where it disapeared) and late `next`:

You have mentioned in https://github.com/magit/magit/issues/1581#issuecomment-99816426 that it might be related to a recent removal of `'priority 2`.

Before:

![before](https://cloud.githubusercontent.com/assets/552356/7592377/469fd4d8-f8d3-11e4-97f6-8ebedc84e188.png)

After:

![after](https://cloud.githubusercontent.com/assets/552356/7592384/4de6051e-f8d3-11e4-9602-706b6c97d9c5.png)

Thank you for your effort.  Best regards,
uu1101.

tarsius:
> You have mentioned in https://github.com/magit/magit/issues/1581#issuecomment-99816426 that it might be related to a recent removal of `'priority 2`.

Actually setting the priority does work, so I have done that now.

However it does not work if the theme sets `inverse-video`. Each face property is considered individually when constructing the final face while taking `priority` into account. When `inverse-video` is involved then that leads to strange result and because I only tested with the face I use which happens to do that, I though setting the priority does not work.

Both zenburn and solarized are guilty of that. I will propose fixes upstream, but first have to make sure that it doesn't break `diff-mode`. Also might end up having to define magit variants for the `diff-refine-*` faces.

In any case you can fix it for your theme by only setting the `:background`.

uu1101:
Thank you!  This works nicely here (I'm using the default theme).  Thank you for your kindness!

tarsius:
Since it works for the default theme now I am closing this. For the solarized themes I have fixed this (but not upstreamed yet) and for zenburn I will do the same. I don't expect there to be many more, if any, themes that do what the mentioned themes do.

Issue #1743, opened 2015-02-27
ediff: request for "simple" ediff instead of 3 way merge for everything except merges


Silex:
Hello,

When doing `M-e` on staged or unstaged changes, magit currently shows a 3-way merge ediff.

A classic ediff like what `ediff-buffers` does is more sensible IMHO.

jsyjr:
+1

tarsius:
Does it really? What I see is that the user is presented with a default choice `A..B` (which she may change to `A...B` to explicitly compare with the merge base) and that `magit-ediff-compare--read-revisions` then returns a list of that range and the two revisions individually (which seems like an unnecessary complication, but that's not relevant here). Then two buffers are created showing the blobs from these two revisions, resulting in a regular "compare two buffers" ediff.

What am I missing? How do you get the 3-way ediff? I need exact steps - maybe under certain circumstances a bug in `magit-ediff-dwim` kicks in and `magit-ediff-resolve` or `magit-ediff-merge` is used instead of `magit-ediff-compare`. Also try whether the same happens if you use `magit-ediff-compare` (<kbd>M-d</kbd>) directly.

jsyjr:
I have a bunch of changes that I want to review so as to understand what I have done:
* In the status dashboard I move my cursor to a file line
* Possibly I TAB to expand, possibly I cursor down to an ill-understood change
* I decide that I want to see my change(s) in context so I hit M-e (magit-ediff-dwim)

Currently magit opens ediff with three panes (A: ~HEAD~, B: workspace, C: ~[index]~) positioned at the top of file.  This is sub-optimal:
* Given magit's laudable effort to use stage / unstage terminology the ~[index]~ qualifier is confusing.  Pane labeling more consistent with the dashboard and magit conventions would be [HEAD], [unstaged] and [staged].  (Similarly since popup S is described as 'Stage all' popup U should be 'Unstage all', not 'Reset index'.)
* Positioning the unstaged changes pane in the center also conflicts with the dashboard.  I suggest left to right ordering: [unstaged], [staged], [HEAD].  (I could also live with right to left ordering.)
* Except when both both staged and unstaged changes are present only two panes should appear.  This way if magit actually opts to display three panes then one knows immediately that something atypically is afoot.
* If M-e happens with the cursor not on a file line but on an specific difference then ediff should open displaying that difference (not the top of file).

tarsius:
@Silex Ah sorry, I somehow understood you were comparing commits and got three buffers.

[...]

Issue #1724, opened 2015-02-04
magit-section-forward much slower than magit-goto-next-section


abo-abo:
With current `master`, I can simply hold <kbd>n</kbd> and the scrolling is almost fluid.

For the same repository, on `next`, the speed of <kbd>n</kbd> is *much* lower, to the
point where holding it is not an option.

Still, when typing <kbd>n</kbd> in rapid succession, the speed is close to acceptable.
Maybe there was some canceling mechanism in `master` that prevented hangups, and this
mechanism didn't make it into `next`.

tarsius:
> For the same repository, on `next`, the speed of <kbd>n</kbd> is *much* lower, to the
point where holding it is not an option.

I think that's mostly because it does more, most importantly it recreates the diff buffer for every commit on the way. What I do when moving to a far away commit is to use just regular `next-line` which bypasses updating the diff buffer and the other functions on `magit-section-movement-hook`.

abo-abo:
> What I do when moving to a far away commit is to use just regular
> next-line which bypasses updating the diff buffer and the other
> functions on magit-section-movement-hook.

But I like updating the diff buffer - it's a great feature.  Maybe you
could add some caching or lazy font-locking or something.

Although it's already supposed to be fast:

    time git log HEAD^..HEAD > temp && git diff HEAD^..HEAD >>temp
    
    real    0m0.004s
    user    0m0.004s
    sys     0m0.000s

tarsius:
> Maybe you could add some caching or lazy font-locking or something.

Doing it asynchronously using font-lock or like font-lock is a long time goal. For now we are stuck with what we have. Caching would introduce a lot of unnecessary complexity I am afraid - the approach I want to take is to make it fast, not work around it being slow.

> Maybe there was some canceling mechanism in `master` that prevented hangups, and this
mechanism didn't make it into `next`.

I believe there never was such a mechanism. This might be more feasible than caching.

> But I like updating the diff buffer - it's a great feature.

I like and use it all the time too and have so far considered the slow down you are reporting to be a documentation issue; having to use <kbd>C-n</kbd> instead of <kbd>n</kbd> when moving across many commits doesn't appear to be a big deal to me. Press <kbd>SPC</kbd> when you have arrived at the last commit.

> Although it's already supposed to be fast:

That's Git not Magit - the parsing is what takes long when displaying a diff.

> Still, when typing <kbd>n</kbd> in rapid succession, the speed is close to acceptable.

[...]

PR #1879, opened 2015-05-28
Switch status display to old-school style

abo-abo:


tarsius:
I am afraid I don't agree with any of these reverts. All these changes were intentional.

abo-abo:
I think it would be advantageous to be able to instantly select the old-school magit theme, instead of spending hours modifying the defaults.

A tonne of people, including myself, have been using `master` for years, and they're used to how things look.  The switch to `next` would be much easier if `magit` didn't change so much visually.

Silex:
Well, a lot of people are also using `next` for years now :-)

Personally I like the idea of a smaller modeline by moving the directory information into the `magit-status` buffer, but that's about it (maybe reintroducing ahead/behind is ok too).

The wording `upstream` is better than using `remote`. Ideally I'd like:

```
Repository:  ~/Dropbox/source/site-lisp/git/magit/
Head:        6748187 master Bring back old-school indendation for unstaged changes
Upstream:    6748187 onto origin/master Bring back old-school indendation for unstaged changes
Last Tag:    1.4.1 (2222 ahead)
```

tarsius:
> Personally I like the idea of a smaller modeline by moving the directory information into the `magit-status` buffer, but that's about it (maybe reintroducing ahead/behind is ok too).

When I introduced the buffer-name-format variables I intended to default to a shorter name eventually. I just haven't gotten around to it. But just chopping of everything but the last part of the path isn't the solution - what if there are two repositories located in directories with the same name?

In a sense this is related to the choices offered when calling `magit-status` with a single prefix argument. Here too we want something that is as short as possible - but not shorter. In other words choices have to be unique. In this case I have made multiple appeals for someone else to implement it but nobody volunteered.

What we need is something like `uniquify.el` but adapted to our needs.

tarsius:
It's not technically possible to let users pick what information should be displayed on the very first line. This has to be hard coded because the first line also serves as the heading of the other lines, making it possible to hide those other lines (to remove "noise"). The forms that insert the subsequent lines have to appear *inside* the form that inserts the first line/section (because they are actually part of that section).

And I really don't feel like making that hard coded first line the suggested "Repository" line. If you guys insist I can add a function which inserts that line and can be added to `magit-status-headers-hook`. But it would not be possible for it to appear before the Head and Upstream lines....

... oh! we could use the header-line. (Or the mode-line, but for some reason that seems to be out of question. Honestly I don't understand this. There is enough room on the mode-line for fairly long paths without pushing any other information of the edge. And file-visiting buffers, the most common buffers that is, do it like that too).

abo-abo:
> It's not technically possible to let users pick what information should be displayed on the very first line.

I don't see why. Just pass a lambda or an alist into `magit-insert-status-headers`.

tarsius:
> I don't see why. 

You'll just have to believe me or study `magit-insert-section` in detail, in particular the parts that have to do with the `magit-section` text property and the `children` struct slot.

> Just pass a lambda or an alist into `magit-insert-status-headers`.

We already "pass a list of sections to be inserted into `magit-insert-status-headers`" - the value of the variable `magit-status-headers-hook`.

As I said, in other words, there has to be an outer call to `magit-insert-section` which contains all other calls to `magit-insert-section`. This can only be done by hard coding it or by transforming *code* at runtime.

It boils down to:

```
(defvar inner-sections
  '((lambda () (magit-insert-section (x) (insert "foo\n")))
    (lambda () (magit-insert-section (x) (insert "bar\n")))
    (lambda () (magit-insert-section (x) (insert "baz\n")))))

(magit-insert-section (x)
  (insert "outer\n")
  (dolist (fn inner-sections)
    (funcall fn)))
```

Now, it is not inherently impossible to do it anyway. But doing it could result in a very nasty special case.  ... I just had an idea for a kludge which might turn out to be only half nasty:

```
(defvar inner-p nil)

;; This is require to allow re-using insert-{foo,bar,baz}
;; in other buffers.
(defvar inserting-all-sections-p nil)

(defvar all-sections '(insert-foo insert-bar insert-baz))

(defun insert-cdr-of-all-sections ()
  (when (and inserting-all-sections-p (not inner-p))
    (setq inner-p t)
    (dolist (fn (cdr all-sections))
      (funcall fn))))

(defun insert-foo ()
  (magit-insert-section (x)
    (insert "foo\n")
    (insert-cdr-of-all-sections)))

(defun insert-bar ()
  (magit-insert-section (x)
    (insert "bar\n")
    (insert-cdr-of-all-sections)))

(defun insert-baz ()
  (magit-insert-section (x)
    (insert "baz\n")
    (insert-cdr-of-all-sections)))

(let ((inner-p nil)
      (inserting-all-sections-p t))
  (funcall (car all-sections)))
```

So we got around manipulating code at runtime, and the resulting code is not *that* nasty. But the  fact remains that this is not understandable unless one is familiar with the implementation details. Some where in the documentation this would have to appear: "The body of section inserters which are intended to be used as a member of `magit-status-sections-hook` have to consist of a single form `magit-insert-sections`. The last form inside that form in turn has to be a call to `magit-status-sections-insert-rest`." That doesn't explain anything though and sounds like a very arbitrary restriction.

tarsius:
Ah no, that won't work without an additional kludge. In the "outer section" `magit-insert-heading` has to be used not `insert`, so we would need `insert-except-when-this-is-the-outer-section-then-magit-insert-heading`.

Silex:
> If you guys insist I can add a function which inserts that line and can be added to magit-status-headers-hook. But it would not be possible for it to appear before the Head and Upstream lines....

Well just to be clear I'm perfectly fine with how `next` is currently is. I just liked the idea... using something like `uniquify.el` for the modeline is good too.

johnmastro:
> In a sense this is related to the choices offered when calling magit-status with a single prefix argument. Here too we want something that is as short as possible - but not shorter. In other words choices have to be unique. In this case I have made multiple appeals for someone else to implement it but nobody volunteered.

> What we need is something like uniquify.el but adapted to our needs.

Do you already have a sense of what Magit's particular needs are? A naive solution is obviously pretty simple but you may already have a good idea of where naive won't be good enough.

tarsius:
> Do you already have a sense of what Magit's particular needs are? A naive solution is obviously pretty simple but you may already have a good idea of where naive won't be good enough.

I don't like how the current implementation loops over the collected data over and over again. It's not as bad anymore as it once was as I have removed some unnecessary looping. But maybe it can be trimmed down more. But that's mostly cosmetic.

If I remember correctly what happens if the last part of the paths of two repositories conflict is that *one of them* is prefixes with the second to last part. But this should happen to both representations.

In other words it should behave like `uniquify.el`. I am pretty sure we can either just call functions from the library or at least copy-paste to get started. I have not looked at that library though.

Also I haven't looked into this enough to be aware of potential pitfalls.

tarsius:
Please customize `magit-insert-status-headers-hook` to try out recent additions. I might be willing to include alternative implementations of "header inserteres", which are more like those used in former releases.

See dc11ad23533cd5ee331dd91fc671cc1a34eb6fe9 for the complications that arise when one wants to have the words ahead and behind appear when showing the number of commits that are in a tag but not the current branch or vice-versa. As you can see I then decided that it was not worth the trouble.

As for the other reverts suggest in this pr - I have decided against them. The changes were intentional and what looks prettier is essentially a personal preference (which likely is guided in great part by what one is used to).

abo-abo:
> As for the other reverts suggest in this pr - I have decided against them. The changes were intentional and what looks prettier is essentially a personal preference

`org-mode` (which `magit-status` tries to emulate, I guess) allows to customize all this stuff.  I don't see why `magit` shouldn't do it.  I'm sure there will be many complaints about the visual changes once `next` is in MELPA.

tarsius:
> `org-mode` (which `magit-status` tries to emulate, I guess) allows to customize all this stuff.  I don't see why `magit` shouldn't do it.

`org-mode`  has more than one active developer.

> I'm sure there will be many complaints about the visual changes once `next` is in MELPA.

And that will probably cause me to flip out.

abo-abo:
> And that will probably cause me to flip out.

No need to flip out. Just let add a bunch of defcustoms (set to your way by default).
Rebasing my changes on top of the ever-changing upstream will make me flip out.

Also `magit-insert-tags-header` doesn't seem to work with the latest changes:

```elisp
(setq magit-status-headers-hook
      '(magit-insert-repo-header
        magit-insert-remote-header
        magit-insert-head-header
        magit-insert-tags-header))
```

johnmastro:
> If I remember correctly what happens if the last part of the paths of two repositories conflict is that one of them is prefixes with the second to last part. But this should happen to both representations.

> In other words it should behave like uniquify.el. I am hopeful we can either just call functions from that library or at least copy-paste to get started. I have not looked at that library in a while though.

I'll take a look at this (but can't promise a particular timeline).

Do you want to offer a configuration style similar to `uniquify-buffer-name-style`? My first inclination is not to (to always behave like uniquify's `forward` style) but, clearly, many people like to be able to configure everything.

tarsius:
You are lucky to maintain relatively young packages you yourself started. Maintaining a package that has been around for years is quite a different story. Every time you change something - considering it a small or even major improvement - someone comes a long and is very outspoken about it being a step backward or even a regression. Several issues are opened every day. I do have to draw the line somewhere - sometimes it is quite arbitrary where exactly I draw the line. That's not because I am lazy or drunk with power, but because I spend a lot of time doing all this for free and cannot just put ever more time into maintaining Magit and providing support.

Often explaining why I won't do something takes just as much time or more than just doing it. It would be very helpful, if at least those users who ought to have an idea what it entails to maintain Magit because they themselves maintain other popular packages, did not insist on having it their way or alternatively to receive an exhaustive analysis of the problem which convinces them that things actually have to be the way they are and/or that their minor grip isn't that important after all.

What I sometimes do in such situations is to provide a proof-of-concept implementation which can be added to the local configuration, as I have done here: https://github.com/magit/magit/issues/1884#issuecomment-106803577. I do that in the hope that users will realize that the reason that I don't add the suggested feature to Magit isn't that I am to lazy to implement it and/or don't care about their needs, but that I have my, possibly unarticulated, reasons not to do it. Since a seldom receive a thank you or even just a comment when doing so, I sometimes wonder whether this is received in exactly the opposite way, i.e. "if it's so easy to implement this, then why doesn't this lazy/inconsiderate maintainer just do it".

I am open to suggestions (but have to learn to say no *more* often). I am aware that I sometimes make changes that aren't good (but sometimes those changes are made because making some insignificant thing slightly less nice enables major improvements elsewhere). It might not seem that way if you are on the receiving end of one of these tirades, but those who stick around here a lot will be able to tell you that I very often (probably to often) go out of my way to accommodate the wishes of users (and that I occasionally have a fit I suppose). 

However I am of the opinion that often I do know better what I am talking about than some user who happens to have a strong opinion on some subject, but doesn't know the back story of why things are they way they are. I won't hide this out of political correctness. For example if some users says that B is clearly superior to A, then he often is not aware that just a few days ago another user has showed up with the exact opposite opinion and managed to convince me. While I am being criticized for not providing some essential feature, I am also being criticized for providing to many features "nobody is ever going to use", i.e. feature creep. In short I am dealing with very opposing and strongly held view points on an almost daily basis. I try to take all those opinions into consideration, but it clearly is impossible to satisfy everyone all the time.

Its especially frustrating to be asked to "at least provide an option to change the behavior" because to me that's often the *worst* possible outcome, not a compromise. It means that I now have to maintain two alternative approaches. (And once you add an option to a package with as many users as Magit, you can be sure that someone will come along suggesting a third approach).

And guess what happens if I do add an option so that everyone can have it their way. Exactly, someone will consider that a failure on my part to decide on one approach and sticking with it.

And when all of this is only cause some user wants to have his shutters painted red, not pink, because traditionally shutters were either red or green, never pink, then ... well if you are having a bad day anyway, then you start wondering whether a lack of empathy and/or imagination is at play.

Please don't take this personally. I don't expect you will as I am sure you have the empathy to see where this is coming from and that you just got unlucky to push the button. Also note that I am currently particularly sensitive to criticism/suggestions about minor details because I *really* want to get 2.1.0 out but am being bombarded with new issues by users who have just switched to the next branch, haven't used it enough to have discovered the wealth of new and improved features but immediately start pointing out significant regressions but also nitpick about details which are just insignificant in the greater scheme of things. (Is it more important that you have tabs in front of your file names, or that thousands of users can finally start using the new version?) I am not saying that these real or perceived issues should not be reported, whether major or minor, or that they should be reported in a different way - I am only saying that it is stressful.

tarsius:
> Do you want to offer a configuration style similar to `uniquify-buffer-name-style`?

If it is based on uniquify, then we would probably get the customizabiliy for free mostly. In which case I would find this desirable. Otherwise it might not be worth it. I don't have a strong opinion here...

> My first inclination is not to (to always behave like uniquify's `forward` style) but, clearly, many people like to be able to configure everything.

... but tend to prefer for it to be customizable, provided that doesn't lead to a lot of additional complexity.

tarsius:
> Also `magit-insert-tags-header` doesn't seem to work with the latest changes:

Aha. For me it works. What are the symptoms of it not working for you? Are you sure you have pulled the relevant change? (I think I hadn't pushed them yet when I posted the above).

abo-abo:
> Is it more important that you have tabs in front of your file names, or that thousands of users can finally start using the new version?

Having the tabs, obviously. Thousands of users use magit not just for the sake of it, but because it's a superior interface.  If it's ridden with annoyances, it stops being superior. I've been using `master` for years and I think I could use it for years more.  There are practically zero features missing, zero annoyances, and it's faster. The only reason I've considered switching to `next` was that I wanted to see if I could customize `magit-blame-mode` to display commits on the right instead of spliced. But if you're too busy to accept (not even implement, just accept) changes that make it possible to make `next` look like `1.4.1`, I'll just stick with `1.4.1`.

npostavs:
> I've been using master ... and it's faster.

Huh. That contradicts recent reports from other users. I wonder if you're using a different `next` and `master` than everyone else.

> But if you're too busy to accept (not even implement, just accept) changes that make it possible to make next look like 1.4.1

Having the option of looking like both means every related **future** change might take up to twice as long because it has to take both options into account.

abo-abo:
> Having the option of looking like both means every related future change might take up to twice as long because it has to take both options into account.

Bullshit. This is Emacs. With Emacs Lisp. Cosmetic changes come at zero cost.

Silex:
@tarsius: please, keep your focus on releasing `next`. What `abo-abo` is doing is extremely demotivating because it makes you think there's still a lot of work to be done, but there is not! All this whining is about cosmetics and feeling entitled to "keep old interface back".

Tag the old version, merge the new and let the whiny people use the old version until we can figure a way to "please everyone", which is probably not possible without a lot of work or not even possible *at all*.

@abo-abo: most `next` users are very happy with it, I doubt that "there will be many complaints about the visual changes once next is in MELPA". Some will whine because "things changed", but for these people they should just use the old version. You can't just come after two years of work and 2000+ commits and tell @tarsius that `next` is "ridden with annoyances". It seems that you didn't even *really* looked at what `next` offers and how much superior it is compared to `master`.

Silex:
@abo-abo: I think patches going in the direction of `defcustom` and with a user-maintained wiki page showing example configurations of  "how to make it look like the old interface" would probably be accepted, but maybe not *right now*.

abo-abo:
> All this whining is about cosmetics and feeling entitled to "keep old interface back".

@Silex I don't appreciate the tone, especially coming from someone whose whole bulk of contributions are less than just one of the 7 commits that I propose.

> @abo-abo: I think patches going in the direction of defcustom and with a user-maintained wiki page showing example configurations of "how to make it look like the old interface" would probably be accepted, but maybe not right now.

A useless statement. You're not the maintainer, @tarsius is.

> most next users are very happy with it

A very bold and subjective statement. It's hard to believe that even 10% of users have the Github version and not the MELPA version.

npostavs:
I hope we can avoid having this thread degenerate into namecalling.

> This is Emacs. With Emacs Lisp. Cosmetic changes come at zero cost.

I can't see how this can be true. Mabye you have a very different interpretation of "zero cost" than me, but as far as I can tell, it takes a non zero amount of work to maintain Emacs themes. For example, solarized-emacs has an [issue tracker with open issues][1] that require work. https://github.com/magit/magit/issues/1848#issuecomment-101655880 is a recent case in magit where cosmetic choices of themes resulted in extra work.

[1]: https://github.com/bbatsov/solarized-emacs/issues

tarsius:
> A useless statement. You're not the maintainer, @tarsius is.

Following that logic you are not entitled to an opinion either, now are you? @Silex has participated in past discussions often enough to have confidence in his assumptions about what I might or might not do.

> especially coming from someone [who hasn't contributed anything significant]

@Silex and @npostavs might not have contributed that much in terms of code, but they have helped tremendously by supporting users, providing moral support, investigating issues, and giving technical advice - over a long period of time. Thank you guys, your help is much appreciated.

johnmastro:
> [M]ost next users are very happy with it, I doubt that "there will be many complaints about the visual changes once next is in MELPA". Some will whine because "things changed", but for these people they should just use the old version.

It's certainly true that you can't please everybody. I was quite happy with `master` and see `next` as a clear improvement. People (myself included, of course) always have some trouble adjusting to change, but my admittedly subjective feeling is that `next` will be very well received.

Regarding visual changes, I wonder if part of the disconnect is because themes still need to be updated for `next` - tarsius has already done most of the work for `solarized` and `zenburn`, but the changes haven't yet been merged upstream.

abo-abo:
> Regarding visual changes, I wonder if part of the disconnect is because themes still need to be updated for next

It's not that, I already updated my [eclipse-theme](https://github.com/abo-abo/eclipse-theme) for `next`.

My issues are the following:

1. I want the sections (`Untracked`, `Staged` etc.) to be less visible, since they're there only for structure.  So it's not great that the entire section is highlighted when the cursor is positioned there.

2. On the other hand, the highlighting of the actionable items is important, because acting on these items results in significant repository changes. Which is why I think they should be separated with an indent in addition to being highlighted, just like it was before.

3. Displaying the full repository path in the mode line and the buffer list is annoying. Not displaying it at the top is annoying as well. Emacs has some traditional interfaces like `dired` and `org-mode`, which have worked well for a lot of people over the years. It makes sense to mimic them.

4. Displaying the repository path for the upstream instead of the last commit message on the upstream is more significant to me, since I often want to check there <kbd>PP</kbd> will push.

5. Not being able to view the current change with ediff in a two buffer vertical split is very inconvenient: that is 50% of my workflow with magit.

If these issues were resolved, for which I offered help, I could stomach the slower log scrolling compared to `master`, and perhaps help with it in the future. Otherwise, I'll just have to use `master` for a year longer, until these issues are naturally raised by other users and resolved. I'm sure that will be the case.

tarsius:
1. I have already offered some code which addresses this. It is not true that these sections are only for structure, many commands which previously only acted on individual files now also do the appropriate thing when point is on the header of a section containing file sections. Which is why I do not intend to add that function I have written for you to magit, it gives the impression that these sections are not actionable but they are.

2. I consider the highlighting to be a sufficient indicator of the current section. And even if it wasn't the case I don't see how indenting the current *and* the non-current file sections would help empathizing the current section. I maintain my position that this is just an aesthetic preference. I will not offer an option to change this and you will just have to file it under "the maintainer is being stubborn". But please realize that for all other issues you have raised I have actually made at least a step into your direction.

3. I have already implemented and merged support for this. So far you have ignored that, except for saying "it doesn't work". It would help if you could respond to my question "in what way does it not work?".

4. See 3, you can have that now. Personally I don't need to see the url to know where I am pushing because if I know what repository I am in then I can always tell so from just the remote name.

5. In other words you use just a tiny fraction of magit. That makes it likely that you are unaware of many new and improved features and I would therefor appreciate it if you did not claim that next offers almost nothing over master. You simply don't know whether that is the case.

> If these issues were resolved, for which I offered help,

What would help is if you followed up on the issue you have raised, e.g. why doesn't the change I made to address 3 and 4 work for you?

> I could stomach the slower log scrolling compared to master

Did you run into other performance issues? Because above you sure made it sound that way.

> and perhaps help with it in the future

I tried to follow the suggestion you made in #1724 but it did not work. Maybe I did something wrong. I certainly would appreciate help with that. I have pushed my failed attempt to `delayed-log-following`.

tarsius:
I have added two new ediff commands now, see #1743.

abo-abo:
> appreciate it if you did not claim that next offers almost nothing over master. You simply don't know whether that is the case.

Sorry, didn't mean to offend. Of course, I meant "doesn't offer *for me*". I'm using only the basic stuff, nothing wrong with that.

> "it doesn't work"

There was a bad format spec. I've sent a PR.

tarsius:
Thanks.

tarsius:
Do you thing there should me *more* information be displayed these new header lines? Like the the branch name? Of course if one used the new *and* the old lines, then that would be redundant information. But it's worth at least considering it.

abo-abo:
> Do you thing there should me more information be displayed these new header lines?

It looks fine the way it is now, perhaps because I'm used to it. I think the perfect way to do it would be in a manner of `format-time-string`:

- Each header line is a string.
- In each string, the escape sequences can be used and combined with no restriction.

For example, if `%L` represents the local path and `%R` represents the remote, it could be even something strange like:

    '("Repository: %L from {%R}")

Here's what the current thing would look like:

    '("Repo:   %-10L"
      "Remote: %-10R"
      "Head:   %-10h %b %c"
      "Tag:    %-10t (%N ahead)")

Silex:
I won't answer the ad hominem attacks, partly because it's useless, partly because others said what needed to be said already.

I'm glad the discussion is turning into... a *discussion* about possible alternatives and implementations instead of strong assertions about how things are or should be.

PR #1889, opened 2015-06-02
therefor -> therefore

fxn:
"therefor" means "for or in exchange for that or this; for it", while "therefore" indicates logical consequence.

Thanks for Magit!


tarsius:
Just learned something new - I just assumed these were alternative spellings. Thanks!

PR #1910, opened 2015-06-23
Fix calls to display warning; don't use 24.3 in travis

npostavs:


tarsius:
Merged after rebasing. Thanks.

Issue #1810, opened 2015-04-16
RFH A "How to push on MS Windows" tutorial


tarsius:
Every other week a MS Windows user posts somewhere that she cannot push using Magit. I am pretty sure that almost never Magit is to blame, but never-the-less this isn't good for its reputation among Windows users.

We have two FAQ entries about this, but there are two problems with those. (a) Many users don't read the FAQ. Unfortunately we cannot do anything about that. (b) When we do point users to these explanations, then that might be enough for them, but it also *might not*.

I would like to get to the point where we can just point users to some resource, which is sufficient to "fix" the issue for 95% of them. A simple FAQ entry - or two - won't do. It is more complex than that.

So this is a request for someone to contribute such an in depth tutorial. It should cover "does it happen on the command line too?", **all** the options available on Windows (including pros and cons), and for each variant it would explain in depth what potential issues are, how to diagnose them, and how to remove failed attempts to work around them.

tarsius:
Ps: I am considering creating yet another issue on the [Emacs Stackexchange site](http://emacs.stackexchange.com) for this. That question would differ from other such issues, in that it would clearly state that it isn't about fixing the issue just for *one particular person*, and that only people who fully understand the issue should answer. Which I am afraid doesn't include me (I haven't used Windows in years). Maybe I should make that a "community wiki question" but then those contributing answers would not get brownie points.

tarsius:
Once more this request led nowhere so I am closing. I might still create the stackexchange question should there be another burst of users having issues configuring git.

npostavs:
At the time this was opened, I was really busy, and looking at it kind of fell off my TODO list. I've now written https://github.com/magit/magit/wiki/Pushing-with-Magit-from-Windows that should mostly cover this, probably the troubleshooting sections will need to be expanded as we find more people having trouble.

Oh, and I've assumed native Emacs with Git for Windows, I suppose there could be additional wrinkles using any combination of native/Cygwin/Msys2 Emacs/git.

tarsius:
Thanks a lot @npostavs! Looks good!

> Under Windows, Magit is unable to intercept the password prompt from git,

Have you verified that this is still the case. The way I remember it is that this only ever happened for some users and that we never figured out why that is so. In case that was a bug in Git that might have been fixed since.

npostavs:
I rechecked now with latest release: Emacs 24.5, and Git for Windows 1.9.5; it's still the case.

[...]

PR #1911, opened 2015-06-23
ssh-w32: new functions for ssh-agent on w32

npostavs:
Context: https://github.com/magit/magit/issues/1810#issuecomment-108123470

I'm still a bit unsure about

- naming
- whether this should depend on magit for locating the executables (that's the current code), or is it better to avoid dependency and copy the logic instead?
- entry points. Currently using `add-advice` on `magit-push` (might need `magit-fetch`, and `magit-pull` as well).

### DIFF
--- a/lisp/ssh-w32.el
+++ b/lisp/ssh-w32.el
@@ -1,0 +1,109 @@
+;;; ssh-w32.el --- use ssh-agent on windows-nt -*- lexical-binding: t -*-
+
+;; Copyright (C) 2015  The Magit Project Contributors
+;;
+;; For a full list of contributors, see the AUTHORS.md file
+;; at the top-level directory of this distribution and at
+;; https://raw.github.com/magit/magit/master/AUTHORS.md
+
+;; Author: Jonas Bernoulli <jonas@bernoul.li>
+;; Maintainer: Jonas Bernoulli <jonas@bernoul.li>
+
+;; Package-Requires: ((emacs "24.4") (cl-lib "0.5") (dash "2.10.0") (magit "2.0.50"))
+;; Homepage: https://github.com/magit/magit
+
+;; This file is not part of GNU Emacs.
+
+;; This file is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 3, or (at your option)
+;; any later version.
+
+;; This file is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this file.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; This file provides functions to startup ssh-agent, prompt for
+;; passphrases from the Windows console, and set the needed
+;; environment variables in Emacs, so that pushes and pulls from magit
+;; will not require entering any passphrase.
+
+;;; Code:
+
+(require 'dash)
+(require 'magit-git) ; for git.exe location
+
+;;; Options
+
+(defgroup ssh-w32 ()
+  "Using ssh-agent on `windows-nt'.")
+
+(defcustom ssh-w32-add-executable
+  (concat (file-name-directory magit-git-executable) "ssh-add.exe")
+  "Location of ssh-add.exe"
+  :group 'ssh-w32
+  :type 'string)
+
+(defcustom ssh-w32-agent-executable
+  (concat (file-name-directory magit-git-executable) "ssh-agent.exe")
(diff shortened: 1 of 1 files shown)


tarsius:
> naming

In Emacs there are some windows specific libraries and I believe their names *begin* with `w32-`. Maybe do the same here. Or maybe something that implies "it just works" or "servants doing the dirty work for you", oh oh! what about `daphne.el`? (I've been watching a lot of Frasier lately).

> entry points

I think it should be kept in a separate repository. In any case it should be a separate elpa package since the free world doesn't need it and because it could be used by other packages that don't depend on magit themselves. Unlike `with-editor` which also can be used outside of magit, it additionally is only a optional dependency of magit - that's why I would treat these two libraries differently. 

> is it better to avoid dependency and copy the logic instead?

....so: yes.

> might need magit-fetch, and magit-pull as well

There's little doubt, now is there?

> add-advice

I might add hooks, but for now that seems fine.

The library headers has some problems, e.g. I am not the author.

npostavs:
> In Emacs there are some windows specific libraries and I believe their names begin with w32-. Maybe do the same here.

Hmm, but then it's sort of clashing with their prefix, though perhaps using `w32-ssh` would be okay (i.e. unlikely to be used by Emacs in the future).

> Or maybe something that implies "it just works" or "servants doing the dirty work for you", oh oh! what about daphne.el? (I've been watching a lot of Frasier lately).

That seems too obscure (I've never watched Frasier). Maybe `ssh-agency`? 'Cause it handles the ssh-agent for you.

> the free world doesn't need it

`ssh-{un?}free-agent`, `ssh-{un?}leash-the-agent`

> I think it should be kept in a separate repository.

Okay.

> The library header has some problems, e.g. I am not the author.

Oh yeah, I just copied the header from with-editor, copyright stuff is sooo boring. :p

tarsius:
> Maybe `ssh-agency`? 'Cause it handles the ssh-agent for you.

That sounds good. `w32-ssh` and `ssh-w32` also work, but this is cooler!

>> the free world doesn't need it

> `ssh-{un?}free-agent`, `ssh-{un?}leash-the-agent`

I am afraid the implied side blow would be lost on (?) many people, so `ssh-agency-{start,ensure}` would probably be better.

Also consider whether `ssh-w32-add-keys` should be treated as a strictly internal function, as in `ssh-agency-start-1`. That would make sense given the current feature set (all or nothing). On the other hand it should probably be possible add only a specific key instead of all that were found.

tarsius:
Since it's gonna be a separate repository, would you like to keep it under your personal account or inside the magit "organization"? If it's the latter then I can create `magit/ssh-agency`. If you change your mind about the name, then we can of course still rename it.

npostavs:
> On the other hand it should probably be possible add only a specific key instead of all that were found.

Hah, the feature requests are coming in already! :) 

> would you like to keep it under your personal account or inside the magit "organization"? If it's the latter then I can create magit/ssh-agency.

I guess it makes sense to have it under magit "organization". 

> If you change your mind about the name, then we can of course still rename it.

I'm feeling pretty good about this name. Web search finds one other near use of it for a similar purpose: [a perl script](http://www.perlmonks.org/?node_id=706626).

> I tried to think of a name that represented something that keeps track of my ssh-agent(s) for me, a framework that is more persistant, so "ssh-agency" seemed a logical choice. But I found typing "ssh-agency" was too similar to "ssh-agent" so I just call it "agency".

Typing shouldn't be problem for our use case :)

npostavs:
Okay, code is now at https://github.com/magit/ssh-agency.

>> Currently using `add-advice` on `magit-push` (might need `magit-fetch`, and `magit-pull` as well).
>
> I might add hooks, but for now that seems fine.

Turned out to need all of 
```
'(magit-push magit-push-matching magit-push-tag magit-push-tags
  magit-pull magit-pull-current
  magit-fetch magit-fetch-current magit-fetch-all)
```
a bit awkward since it depends on the particulars of how the functions are implemented. Maybe it's better to advise `magit-start-process` and check the args for `"push"`, `"pull"`, or `"fetch"`?

tarsius:
Let's keep this open for a while.

tarsius:
> a bit awkward since it depends on the particulars of how the functions are implemented.

I'll probably add some hooks. But that needs some more thought before I can implement it, e.g. would users start using these hooks even in cases when they would be better of using the respective hooks provided by Git itself. Won't happen before the release.

> Maybe it's better to advise `magit-start-process` and check the args for `"push"`, `"pull"`, or `"fetch"`?

I don't think that's a good idea.

tarsius:
Do you think this is ready to be added to Melpa? I do.

Could you please adjust https://github.com/magit/magit/wiki/Pushing-with-Magit-from-Windows.

npostavs:
[Done][1] and done.

First contact with user didn't go quite as smooth as planned though: https://github.com/magit/magit/issues/1924#issuecomment-116238856 :( Inevitable, I suppose.

[1]: https://github.com/milkypostman/melpa/pull/2887

Issue #1912, opened 2015-06-23
prevent whitespace errors when running org-texinfo+export-to-texinfo from Makefile


tarsius:
For some reason running `org-texinfo+export-to-texinfo` in batch mode results in whitespace errors while running the same function interactively does not.

More precisely when this is run from a Makefile, then indented lines in code blocks lack one space. (And files don't end with a newline, but that we already work around).

tarsius:
Doesn't belong here.

PR #1928, opened 2015-06-28
docs: use ndash

dunn:
Using the ndash instead of a hyphen is more formally correct and (in my opinion) looks nicer.

I didn't need to run `make texi` for this change, which is lucky since it's not working for me:
```
🐙  make texi
Generating texi files
WARNING: This creates whitespace errors in code blocks.
WARNING: Discard these whitespace errors before committing.
WARNING: Also see https://github.com/magit/magit/issues/1912.
Generating magit.texi
Cannot open load file: no such file or directory, ox-texinfo+.el
make[1]: *** [texi] Error 255
make: *** [texi] Error 2
```

Probably something silly, but what am I doing wrong?

### DIFF
--- a/Documentation/magit-popup.org
+++ b/Documentation/magit-popup.org
@@ -150,7 +150,7 @@ a change to notice them.
 
   This option controls whether the section which lists the commands
   that are common to all popups is initially show.  We recommend you
-  set this to ~nil~ - after you have memorized that it can be shown on
+  set this to ~nil~ – after you have memorized that it can be shown on
   demand using ~C-t~.
 
 - Key: C-t, magit-popup-toggle-show-common-commands
--- a/Documentation/magit-popup.texi
+++ b/Documentation/magit-popup.texi
@@ -195,7 +195,7 @@ a change to notice them.
 
 This option controls whether the section which lists the commands
 that are common to all popups is initially show.  We recommend you
-set this to @code{nil} - after you have memorized that it can be shown on
+set this to @code{nil} – after you have memorized that it can be shown on
 demand using @code{C-t}.
 @end defopt
 
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -75,7 +75,7 @@ To stage or unstage a change one places the cursor on the change and
 then types ~s~ or ~u~.  The change can be a file or a hunk, or when the
 region is active (i.e. when there is a selection) several files or
 hunks, or even just part of a hunk.  The change or changes that these
-commands - and many others - would act on are highlighted.
+commands – and many others – would act on are highlighted.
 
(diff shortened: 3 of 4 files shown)


tarsius:
> Using the ndash instead of a hyphen is more formally correct and (in my opinion) looks nicer.

Thanks. I'll probably merge tomorrow, but first have to make sure this doesn't cause any problems when exporting to pdf and html. (See the webpage repository).

> Probably something silly, but what am I doing wrong?

You need org-mode v8.3beta and my [`ox-texinfo+`](https://github.com/tarsius/ox-texinfo-plus). I haven't had the time to document this yet and actually I hope I can get most of it merged into `ox-texinfo` which is part of org-mode. I usually don't use `make texi` and instead use `C-c C-e x i` to export directly from the buffer visiting an org file. If you do that then the warning above doesn't apply, see #1912.

dunn:
> There are various kinds of dashes; is this "the best one"? Why?

From what I can tell, [it's acceptable](https://en.wikipedia.org/wiki/Dash) to use either the en dash (–) or the em dash (—) in this context, as long as en dashes have a space on either side and em dashes don't.  The Chicago Manual of Style recommends the em dash, but I think the en dash with spaces is easier to read.  But I don't have a strong preference either way.

dunn:
With your `ox-texinfo+`, I get this message when trying to export buffers with `C-c C-e x i`:
```elisp
cond: Wrong type argument: numberp, (:export-options nil :input-file "/Users/cat/Dropbox/projects/clones/magit/Documentation/magit.org" :input-buffer "magit.org" :texinfo-filename "" :texinfo-class "info" ...)
```
(I know this is taking time away from actually releasing magit, so you don't have to hold my hand through this if you don't feel like it.)

tarsius:
@kyleam @npostavs what are your opinions on this?

I don't like the idea of space-less em dashes.

npostavs:
I'm more used to space-less em dashes; that's what I see in the newspaper. In my own writing, I pretty much avoid dashes (I would use parentheses, commas, or colons for all the occurrences in this commit) but more because I can't be bothered to type in fancy dash characters than out of any special editorial preference.

By the way, I can hardly tell the difference between ndash and hyphen.

kyleam:
I don't have a strong opinion about space-less em dashes versus en dashes with spaces.  Like @npostavs, I tend to just avoid them (except for using en dashes for ranges in academic writing).

tarsius:
I am probably going to delay my decision on this question until after the release. I don't really have a choice - there are still so many other things to do.

dunn:
No problem! There are instances of ` -- ` that need also need converting anyway.

PR #1930, opened 2015-06-29
Revisions to manual

kyleam:
I've split these into two commits.  The second one has more minor changes, where I don't think the original version hurts clarity.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -63,7 +63,7 @@ is so cumbersome that many users only use it in exceptional cases.
 In Magit staging a hunk or even just part of a hunk is as trivial as
 staging all changes made to a file.
 
-The most visible part of Magit's interface is the status buffer which
+The most visible part of Magit's interface is the status buffer, which
 displays information about the current repository.  Its content is
 created by running several Git commands and making their output
 actionable.  Among other things, it displays information about the
@@ -78,7 +78,7 @@ hunks, or even just part of a hunk.  The change or changes that these
 commands - and many others - would act on are highlighted.
 
 Magit also implements several other "apply variants" in addition to
-staging and unstaging.  One can also discard or reverse a change, or
+staging and unstaging.  One can discard or reverse a change, or
 apply it to the working tree.  Git's own porcelain only supports this
 for staging and unstaging and you would have to do something like ~git
 diff ... | ??? | git apply ...~ to discard, revert, or apply a single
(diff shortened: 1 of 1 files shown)


kyleam:
In addition to these changes:
- The section Customizing/Essential Settings/Performance seems to be incomplete, ending with "Now for the things that you can do to improve performance:"
- The plumbing section is inconsistent with how it capitalizes "Git". The "asynchronous" section uses "Git", while the other sections tend to use "git".

tarsius:
Thanks so much! I have just gone through the changes, and picked out the ones I am unsure about and/or would like to get a second opinion on. These changes touch on some rule sets I just don't know enough about. So this seems like a perfect learning opportunity. Except that I already have a lot of other things to do of course :-/

Anyway, later I will open a new, commented pr with the changes I have not applied yet.

> The section Customizing/Essential Settings/Performance seems to be incomplete, ending with "Now for the things that you can do to improve performance:"

That's one of these other things I still have to take care of ...

> The plumbing section is inconsistent with how it capitalizes "Git". The "asynchronous" section uses "Git", while the other sections tend to use "git".

I am aware that there is some inconsistency to be found here. Generally I use "git" when talking about the executable, "Git" otherwise. (Of topic: I know you don't like "..., <ALT> otherwise" much). But in many cases it is not so clear whether we are talking about the executable or "the version control system named Git".

kyleam:
> Thanks so much!

My pleasure.  Thanks for taking the time to write up such nice
documentation (in addition to all your other work).

> I have just gone through the changes, and picked out the ones I am
> unsure about and/or would like to get a second opinion on. These
> changes touch on some rule sets I just don't know enough about.

Great.  I hope that separating them out makes it a little easier to go
through and select the changes you want to keep.  The ones in the
first commit are things that are either clear typos or I think affect
readability/clarity.  The changes in the second commit are much more
picky (and in some cases, purely stylistic), and I think it will read
fine if you keep it as is.

> I am aware that there is some inconsistency to be found
> here. Generally I use "git" when talking about the executable, "Git"
> otherwise.  [...]  But in many cases it is not so clear whether we
> are talking about the executable or "the version control system
> named Git".

Yeah, I thought that might have been the distinction.  I pointed out
the plumbing section specifically because there are chunks where you
are clearly talking about the same git, but the case is different.

For example,

    - Function: magit-run-git &rest args

      Calls git synchronously with ARGS and then refreshes.

and

    - Function: magit-run-git-async &rest args

      Start Git, prepare for refresh, and return the process object.
      ARGS is flattened and then used as arguments to Git.

> (Of topic: I know you don't like "..., otherwise" much, but I kinda
> do :-).

Hopefully I put those changes in the "minor" commit.

tarsius:
I have merged most of the changes from the first commit with 8a01600096888ba48ebb7d707cac2fd788daa36d. The leftovers are in bdaf3c5204e15b47f90810e2a34f46385f5afb80.

tarsius:
I have merged the second commit with some minor changes. In hindsight, I should just have merged them as is - the differences are minimal. (The only place where you did get it wrong was the "file, or range/commit" thing).

tarsius:
Thanks again.

PR #1932, opened 2015-06-29
Remove needless condition

syohex:
Because Magit requires Emacs 24.4 and 'read-only-mode' was introduced at Emacs 24.3.


tarsius:
Thanks

PR #1935, opened 2015-06-30
Documentation corrections

dunn:
I tried to split up the changes into different commits based on what sort of changes they are – feel free to reject nit-picky or unhelpful ones.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -4315,6 +4315,18 @@ to do that yourself, then consider informing me about the resources
 that helped you figure what to do for your specific setup, so that I
 can post a link here.  That way those coming after you have it easier.
 
+*** Installing Emacs on Mac OS X
+
+The easiest way to get an up-to-date installation of Emacs on OS X is
+probably with the [[http://brew.sh][Homebrew]] package manager:
+
+#+BEGIN_SRC shell-script
+  brew update && brew install emacs
+#+END_SRC
+
+Other Mac package managers, including [[http://www.macports.org][MacPorts]] and [[http://www.finkproject.org][Fink]], also make
+installing Emacs relatively straightforward.
+
 ** I am using a Git release older than 1.9.4
 
 At least version 1.9.4 is required.  There is no way around it.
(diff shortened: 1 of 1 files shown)


Documentation/magit.org
```diff
@@ -1123,7 +1124,7 @@ that it should be bound globally.  We recommend using ~C-x g~:
   #+END_SRC
 
   Starting with Emacs 25.1 that keymap can also be customized in a
-  sane fashion:
+  more concise way:
```
dunn: This might seem like a pointless change but I agree with those who say that using terms like "sane" (and "crazy", etc.) as synonymous with "good" (or "bad") perpetuates negative attitudes towards people will mental illness and so should be avoided.

Documentation/magit.org
```diff
@@ -3676,18 +3677,18 @@ These functions are used to run git to produce some effect.  Most
 Magit commands that actually run git do so by using such a function.
 
 Because we do not need to consume git's output when using these
-functions.  Their output is instead logged into a per-repository
+functions, their output is instead logged into a per-repository
 buffer, which can be shown using ~$~ from a Magit buffer or ~M-x
```
kyleam: oops, sorry, I think my revisions introduced this by mistake.

dunn: 🙊
You caught a lot more errors than me, so I think you get a pass on that one.

kyleam:
These changes look good to me.

Documentation/magit.org
tarsius: But political correctness has its dangers too. I mean no offense and this is clearly not targeted at people with a mental illness. It's not even targeted at the author of the code that forced us to do it the more complicated way. (When I got my patch, addressing this issue, merged, he said that he had just copied that code from `icomplete.el`, without giving it much thought.) What I feel is "crazy" is that this clearly was a defect/misfeature (which many users must have run into) but nearly two decades passed until someone bothered to fix it.

tarsius:
I have merged most of this, except for the last commit and very few select hunks.

Thanks!

Documentation/magit.org
dunn: I know you meant no offense! But some people may find it hurtful anyway.  What about changing it to "that keymap can also be customized in a far more sensible way"?

tarsius: I would not call a person with a mental illness "crazy". I would only use that word in "the other sense". If I cannot use "crazy" here then I cannot use it anywhere. Not using a word (with the intended meaning A) because some people might be hurt because they might jump to the conclusion that I meant meaning B is, well, crazy. The reason why some people might jump to that conclusion and then feel hurt, is *because* we have been censuring our language so that nobody needs to feel hurt.

We start here and then we end up beeping when some says "fuck". It might be a cultural difference but the equivalent German "verrueckt" sounds perfectly fine and non-discriminating in my ears.

Note that I usually use "she" when talking about "a user". So it's not that I am not sensitive to language that might marginalize certain groups people, but I have to draw the line somewhere. And not using a perfectly fine word because someone *could* take offense - that's past that line.

PR #1936, opened 2015-06-30
magit-mode-bury-buffer: Kill aux buffers depending on a var

Malabarba:
magit-kill-aux-buffers: New var. Controls whether
magit-mode-bury-buffer should kill auxiliary buffers instead of
burying.

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -179,6 +179,13 @@ has to confirm each save."
                  (const :tag "Ask" t)
                  (const :tag "Save without asking" dontask)))
 
+(defcustom magit-kill-aux-buffers t
+  "If non-nil, `q' kills auxiliary buffers instead of burying.
+This does not affect the main *magit:...* buffer."
+  :type 'boolean
+  :group 'magit
+  :package-version '(magit . "2.1.0"))
+
 ;;; Magit Mode
 
 (defvar magit-mode-map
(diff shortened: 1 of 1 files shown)


Malabarba:
I wrote this when I realized I had about 15 magit buffers lying around from 3 different repos.
Do you think it's worth merging?

tarsius:
Maybe. I need to think about it.

Just burying a buffer has advantages. Most importantly point will in most cases still be were it used to be.

So it's to late for this change before the release. But let's consider it sometime after the release.

Malabarba:
Oh. Actually, I just realised I left the default value at `t`. My intent was to make it `nil` by default.

Malabarba:
But I agree 1 day before release is never a good idea. :)

tarsius:
@kyleam This could be an alternative to your #2146. (Cannot reopen this pr because "The next branch has been deleted".)

kyleam:
Hmm ... although this looks useful in general, I think there are some
disadvantages of using it specifically for avoiding sticky log/diffs arguments.

* I find it  useful to have the log and diff buffers stay around
  but not have sticky arguments, especially so that you can keep the
  history alive.

* Even if the user wants to kill log and diff buffers to get rid of
  sticky arguments, they may want to keep around other buffers,
  especially the process buffer, and this will kill any buffer aside
  from status buffer.

* It requires that the user explicitly hit q versus some other way of
  switching away from the buffer.

tarsius:
Yeah... Just something that came to mind while looking through old issues.

PR #1937, opened 2015-07-01
release notes: consistency and corrections

dunn:
Sometimes LaTeX-style
```
`single quotes'
```
are used, other times backticks are placed on both sides.  I switched everything to the latter.


tarsius:
Uhh, unfortunately the intention was to use the former :(

dunn:
No problem, I'll just switch them all to the former.

dunn:
Or should commands be surrounded by backticks and everything else quoted?

tarsius:
```
`everything' except "terms"
```

:-)

dunn:
Done, and another commit pushed with more general tweaks.

PR #1939, opened 2015-07-01
RelNotes/2.1.0: fix typos, grammar, etc.

npostavs:
rebased on top of @dunn's changes, includes some refilling of those. I guess there will some overlap with @kyleam's changes.

>  is it okay if I squash your corrections into a single commit authored by me

no problem

### DIFF
--- a/Documentation/RelNotes/2.1.0.txt
+++ b/Documentation/RelNotes/2.1.0.txt
@@ -30,8 +30,8 @@ user interface and the internal abstractions.
 
 Magit always provided the most intuitive staging interface, which is
 vastly superior to `git add --interactive'.  This release extends this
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks! Cherry-picked on top of other corrections.

npostavs:
You left out the deletion of a pretty much duplicate paragraph:

```
 * Most commands that previously could only act on the section at
   point can now also act on all sibling sections selected using the
   region.

 * Many commands that previously could only act on the section at
   point can now act on all sections selected by a region, which can
   span sibling sections of the same type.
```

tarsius:
Ah yes, I though that was dropped by mistake.

PR #1940, opened 2015-07-01
RelNotes/2.1.0: minor fixes

madscientist:
Two trivial changes which I don't think others caught yet.


tarsius:
Thanks! Cherry-picked on top of other corrections.

PR #1955, opened 2015-07-03
Add magit-insert-recent-commits to magit-status-sections-hook

mgalgs:
`(magit-insert-recent-commits)` is amazing. Enable it by default.

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -110,7 +110,8 @@ at all."
     magit-insert-staged-changes
     magit-insert-stashes
     magit-insert-unpulled-commits
-    magit-insert-unpushed-commits)
+    magit-insert-unpushed-commits
+    magit-insert-recent-commits)
   "Hook run to insert sections into a status buffer."
   :package-version '(magit . "2.1.0")
   :group 'magit-status


tarsius:
Personally I use `magit-insert-unpulled-or-commits`, and think we should considering replacing `magit-insert-unpulled-commits` with that. But I am not so found of `magit-insert-recent-commits` because using that often means that the same commits are displayed under "Recent commits" and "Unpulled commits".

What do others think?

npostavs:
> using that often means that the same commits are displayed under "Recent commits" and "Unpulled commits".

That seems too annoying to have as a default.

> Personally I use magit-insert-unpulled-or-commits, and think we should considering replacing magit-insert-unpulled-commits with that.

I guess you could still have the problem of having the same commits displayed under "Recent commits" and "Un*push*ed commits"?

tarsius:
> I guess you could still have the problem of having the same commits displayed under "Recent commits" and "Unpushed commits"?

Yes, and I think that's the reason I have not made it the default.

mgalgs:
Makes sense, just thought I'd initiate a discussion.  Thanks!

tarsius:
> Makes sense, just thought I'd initiate a discussion.

And you did. But if that is the intention, then create an issue, not a pull request. Or create a pull request but explicitly state that you would like to discuss the included variant *and others*. 

This does not sound like an invitation to a discussion:

> `(magit-insert-recent-commits)` is amazing. Enable it by default.

PR #1956, opened 2015-07-03
Add new line and fix indent

aaa707:
I guess a new line is missing.
Thanks,

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -94,7 +94,8 @@ at all."
              magit-insert-remote-header
              magit-insert-head-header
              magit-insert-upstream-header
-             magit-insert-tags-header))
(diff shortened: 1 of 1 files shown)


aaa707:
Also, I've added a current user info into header.
I'm using some accounts, so it's convenient to show the current user.

npostavs:
> I'm using some accounts, so it's convenient to show the current user.

I don't think it makes sense as a default though.

tarsius:
Thanks. I have cherry-picked.

> I don't think it makes sense as a default though.

I agree and left that out.

aaa707:
Thanks for merging.

> I don't think it makes sense as a default though.

Yes, so I don't set as a default.

PR #1965, opened 2015-07-03
magit-mode: Don't bind M-w

### LINKED ISSUE #1958: Binding M-w in magit-mode breaks normal Emacs behavior
I find it quite common to mark&copy text in my Magit to use elsewhere in Emacs.  Recently, it was rebound to some-long-function-I-can't-remember, which seems to only copy the current commit id.  I wouldn't mind this behavior being on a custom key, but rebinding M-w seems odd for a buffer that mostly contains text.  I know it's easy to change the binding back, which I've done, but I thought you might want to correct this bit of inconsistency.

mgalgs:
It's quite common for users to want to kill things from magit
buffers (commit messages, diffs, etc).  Currently this isn't possible
since M-w is bound to `magit-copy-buffer-thing-as-kill`.  Fix this by
not binding M-w.  Users who want M-w bound to
`magit-copy-buffer-thing-as-kill` can bind it themselves.

Fixes #1958

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -253,7 +253,6 @@ has to confirm each save."
     (define-key map "\C-xa"  'magit-add-change-log-entry)
     (define-key map "\C-x4a" 'magit-add-change-log-entry-other-window)
     (define-key map "\C-w"   'magit-copy-as-kill)
-    (define-key map "\M-w"   'magit-copy-buffer-thing-as-kill)
     (define-key map [remap evil-previous-line] 'evil-previous-visual-line)
     (define-key map [remap evil-next-line] 'evil-next-visual-line)
     map)


madscientist:
Minor terminology nit: as I know you already know, M-w doesn't kill things, it _saves_ things to the kill ring without killing them.  I do agree though that rebinding M-w in magit buffers is problematic.

tarsius:
> Fixes #1958

Well, half of it. To fix this properly the key binding has to be removed *and* a new one has to be found. At least! `magit-copy-buffer-thing-as-kill` also has to be further improved...

> M-w doesn't kill things, it saves things to the kill ring

... that for example. (Note that I was already aware of this, see #1586.)

There are various options and I am open to suggestions. But lets discuss this on #1958.

mgalgs:
Yes, I'm all about a proper fix but this really needs to be fixed *right now* since it leaves such a bad impression on the droves of people now jumping on the 2.1.0 release.  I was envisioning reverting the keybinding immediately so that we could work on a proper fix with more calm.

tarsius:
Lets discuss this on #1958.

Issue #1485, opened 2014-08-30
Add support for a publish branch


tarsius:
Once upon a time when I started contributing to Magit more frequently but did not have commit access yet, I got a bit annoyed with the fact that a branch can only have one "upstream" branch. And so the separate package `magit-push-remote` was born. Once I did have commit access for Magit that package started to bit-rot but I have since revived it.

I intend to implement support for a "publish" branch in addition to the "upstream" branch in Magit itself, and deprecate `magit-push-remote`. I am going to do this in a way compatible with 
Felipe Contreras' fork of Git. This basically implements the same feature but is less of a hack. Unfortunately it doesn't look like this is ever going to be merged into Git because Felipe and the Git maintainers do not get along. But that doesn't have to bother us, we can implement it in Magit. This implementation will be compatible with Felipe's without actually depending on his fork. That basically means that we are going to use the same git variables as his fork does. That will however be incompatible with `magit-push-remote` which never was much more than a kludge.

Also read Felipe's description of the "tringular workflow" which is made much easier if there is a publish branch in addition to the upstream branch: https://felipec.wordpress.com/2014/05/11/git-triangular-workflows. Also worth reading: https://felipec.wordpress.com/2013/09/01/advanced-git-concepts-the-upstream-tracking-branch.

---

Also make other related changes and take concerns raised in #2009 into account.

The `v2.4` release will focus on this, it's motto is:

> Deal with the fact that everyone has a different opinion on how branching and pushing should work.

fommil:
now that the **awesome** 2.1 release is out the door, will you be looking at this? :grin:

tarsius:
I have not yet had the change to set priorities, there's just to much going on right now.

fommil:
what with all those cigars to smoke and whiskies to sip :smile:

mgalgs:
This still sounds a little too restrictive...  I often push to 3 or more remotes.  Also, it sounds like it will require extra configuration (setting up a "publish" branch) which is excessive for pushing little one-off branches here and there.  How about something like what I proposed in #1967 (push the current branch to *any* remote)?

Of course, this "publish branch" approach could coexist with something like #1967, so it's not really an either/or decision...

tarsius:
I am not going to implement a "publish remote" compatible with Felipe's Git fork. Instead I will add support for Git's "push remote", which is configured using the variables `branch.<name>.pushRemote` and `remote.pushDefault`.

[...]

PR #1967, opened 2015-07-03
magit-remote: add action to push current branch to non-upstream

mgalgs:
In the fork+pull-request workflow, it's common to track an upstream
branch while pushing to a non-upstream remote (e.g. your forked repo).
None of the current pushing actions allow you to push your current
branch (and nothing more) to a non-upstream remote.  Add one.

### DIFF
--- a/lisp/magit-remote.el
+++ b/lisp/magit-remote.el
@@ -166,6 +166,7 @@ then read the remote."
               (?d "Dry run"       "--dry-run")
               (?u "Set upstream"  "--set-upstream"))
   :actions  '((?P "Current"   magit-push-current)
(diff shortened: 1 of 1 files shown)


tarsius:
#1485

vincentbernat:
After upgrading to 2.1, I am also faced with this problem. I am using an "upstream" remote and a "github" remote for most projects. Once I create a branch, I usually do <kbd>P</kbd> <kbd>-u</kbd> <kbd>P</kbd> the first time and I expect to be able to just tell a remote. Now, I have to tell a remote plus the name of the branch (for example `github/mynewbranch`).

The publish branch proposed in #1485 is not really a solution. I just want to be able to input a remote, like this was the case in 1.4.0.

PR #1974, opened 2015-07-04
.travis.yml: pass DASH_DIR to make

npostavs:
This should let travis tests pass again.

EDIT: passes, despite warnings about having git 1.8.5


tarsius:
Thanks. I have cherry-picked.

PR #1976, opened 2015-07-04
magit-mode: Use `C-c C-w' for killing magit buffer things

### LINKED ISSUE #1958: Binding M-w in magit-mode breaks normal Emacs behavior

mgalgs:
It's quite common for users to want to kill things from magit
buffers (commit messages, diffs, etc).  Currently this isn't possible
since M-w is bound to `magit-copy-buffer-thing-as-kill'.  Fix this by
moving this binding to C-c C-w.

Fixes #1958

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -253,7 +253,7 @@ has to confirm each save."
     (define-key map "\C-xa"  'magit-add-change-log-entry)
     (define-key map "\C-x4a" 'magit-add-change-log-entry-other-window)
     (define-key map "\C-w"   'magit-copy-as-kill)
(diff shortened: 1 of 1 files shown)


tarsius:
I have chosen a different approach, see https://github.com/magit/magit/issues/1958#issuecomment-118610448.

PR #1977, opened 2015-07-04
magit-log: add section hook for graph'd recent commits

mgalgs:
It can be useful to see recent commits with decorations and graph
annotations.  Add a status section hook function for this (that can be
added to `magit-status-sections-hook`).

Screenshot:

image


mgalgs:
Also added a commit to speed things up a bit by leveraging git revision while using `--graph`.  Without this additional patch the status buffer of the Linux kernel repository takes around 15 seconds to refresh.  With the patch it's under a second.

mgalgs:
Abandoning in favor of `magit-log-section-args` and #1990.

Issue #1963, opened 2015-07-03
magit-ediff-stage: Please don't set ediff-quit-hook


michael-heerdegen:
Hello,

today I got the upgrade to the new Magit version via Melpa, including the new "magit-ediff-stage" command.  I could hardly await it (thanks for working on this!).

However, there is a little problem with this.  I use "ediff-quit-hook" to close the selected frame (I have configured "ediff-window-setup-function" to pop up a new frame).  With the new Magit version, closing the selected frame doesn't work anymore because "magit-ediff-stage" resets the hook.

Apart from my personal problem - such a hook should never be set directly, better use add-hook with the LOCAL parameter.

Didn't look at master, sorry if this issue should already be fixed there.

Thanks,

Michael.

tarsius:
> such a hook should never be set directly, better use add-hook with the LOCAL parameter.

It's not that easy. This is a very recent change (353bc8ef0cc5a2f7ea8ff0ce20e04468836ed096) which fixes #1934.

Taking all that into account, what's your suggestion?

madscientist:
I only read the comment for the commit, not all the code, but it seems to me that if ediff-cleanup-mess is a no-op, why not leave it in ediff-quit-hook?  IOW, just put back the add-hook instead of seq-local, and have your local hook call magit-ediff-cleanup-auxiliary-buffers as it does now to do the cleanup.

Did I miss something?  Yes, it's a bit bizarre that ediff-cleanup-mess is still invoked and basically doesn't do anything, but it doesn't hurt...?

tarsius:
> if ediff-cleanup-mess is a no-op

I am not 100% sure that is the case. It might have undesirable side-effects.

> Did I miss something?

`magit-ediff-resolve` uses `smerge-ediff`, which doesn't use `add-hook`. We cannot change that from within Magit.

michael-heerdegen:
Jonas Bernoulli <notifications@github.com> writes:

> magit-ediff-resolve uses smerge-ediff, which doesn't use add-hook. We
> cannot change that from within Magit.

I consider that a bug, too.

Anyway, this is not too important.  I will not be angry if you just
close this.  I'll find a way to program around this.


Thanks,

Michael.

tarsius:
> I will not be angry if you just close this.

I do intend to do something about this, just cannot promise I will get to it soon. There's just to much going on right now!

tarsius:
Please try #1980.

michael-heerdegen:
[...]

PR #1980, opened 2015-07-05
Add new hook magit-ediff-quit-hook

tarsius:
See #1963 and c18264374d1a42edfc7e781a8c7fe39d78b4ea1b.

Please test and comment.

### DIFF
--- a/lisp/magit-ediff.el
+++ b/lisp/magit-ediff.el
@@ -32,6 +32,29 @@
 (require 'ediff)
 (require 'smerge-mode)
 
+(defvar smerge-ediff-buf)
+(defvar smerge-ediff-windows)
+
+(defgroup magit-ediff nil
+  "Ediff support for Magit."
+  :group 'magit-extensions)
+
+(defcustom magit-ediff-quit-hook
+  '(magit-ediff-cleanup-auxiliary-buffers
+    magit-ediff-restore-previous-winconf)
+  "Hooks to run after finishing Ediff, when that was invoked using Magit.
+The hooks are run in the Ediff control buffer.  This is similar
+to `ediff-quit-hook' but takes the needs of Magit into account.
+The `ediff-quit-hook' is ignored by Ediff sessions which were
+invoked using Magit."
+  :package-version '(magit . "2.1.1")
+  :group 'magit-ediff
+  :type 'hook
+  :options '(magit-ediff-cleanup-auxiliary-buffers
+             magit-ediff-restore-previous-winconf))
+
+(defvar magit-ediff-previous-winconf nil)
+
 ;;;###autoload (autoload 'magit-ediff-popup "magit-ediff" nil t)
 (magit-define-popup magit-ediff-popup
   "Popup console for ediff commands."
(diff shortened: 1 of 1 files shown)


michael-heerdegen:
Thanks!  That looks appropriate and solves by problem.

Two comments:

1.

> +(defvar smerge-ediff-buf)
> +(defvar smerge-ediff-windows)

I guess it's not good to reuse prefixes from different packages.  I would call it "magit-smerge-..."

2.
```
(defcustom magit-ediff-quit-hook
  '(magit-ediff-cleanup-auxiliary-buffers
    magit-ediff-restore-previous-winconf)
  ...)
```

Caveat: Filling hooks via variable definition can be problematic: If the user does "add-hook" to the variable before loading the definition (i.e. before requiring the package), the specified hook members will not become part of the hook (because the variable's value will not be touched).

AFAIK there is no perfect solution to this problem, but using a default value of nil and adding to the hook afterwards might be better.  A user wanting to setq the hook can still use "with-eval-after-load".

P.S.  I really hope you can switch to lexical binding soon and get rid of all these backquoted lambdas.

tarsius:
> I guess it's not good to reuse prefixes from different packages. I would call it "magit-smerge-..."

Magit only used the values of these variables, it is Smerge which sets it. Smerge should properly define these variables but unfortunately it does not, so we have to declare them to avoid compile warnings.

> Filling hooks via variable definition can be problematic:

This is being done a lot in Emacs too, so I am going to stick with this.

> I really hope you can switch to lexical binding soon and get rid of all these backquoted lambdas.

I intend to do it eventually, but am not in a hurry :-)

> solves by problem.

Could you please post your configuration. I am considering to use a new frame too.

michael-heerdegen:

> Magit only used the values of these variables, it is Smerge which sets
> it. Smerge should properly define these variables but unfortunately it
> does not, so we have to declare them to avoid compile warnings.

smerge-mode.el devfar's them in Emacs master.  But I guess you
nonetheless need to quite the byte compiler.

> Could you please post your configuration. I am considering to use a
> new frame too.

I wouldn't call it a configuration but a hack.  Anyway, don't you think
this should better be done in Ediff itself?

michael-heerdegen:
Anyway, here is my setup for using pop-up frames:

```
(defun my-ediff-setup-windows (&rest args)
  (let ((oframe (selected-frame))
	(frame (make-frame)))
    (select-frame-set-input-focus frame)
    (set-frame-parameter frame 'ediff-frame oframe)
    (apply #'ediff-setup-windows-plain args)
    (set-frame-parameter nil 'fullscreen 'maximized)
    (run-with-idle-timer
     0 nil
     (lambda ()
       (when (and (window-live-p ediff-control-window)
                  (not ediff-use-long-help-message))
         (minimize-window ediff-control-window))))))

(setq-default
 ediff-window-setup-function 'my-ediff-setup-windows
 ediff-split-window-function 'split-window-horizontally)

(defun my-ediff-quit-dwiw ()
  "Do what I want after quitting ediff."
  (let* ((current-frame (selected-frame))
	 (oframe (frame-parameter current-frame 'ediff-frame)))
    (when oframe
      (other-frame 1)
      (sit-for 0)
      (delete-frame current-frame)
      (sit-for 0)
      (when (frame-live-p oframe)  (select-frame-set-input-focus oframe)))))

(add-hook 'ediff-quit-hook #'my-ediff-quit-dwiw 'append)
(add-hook 'magit-ediff-quit-hook #'my-ediff-quit-dwiw 'append)
```

Unfortunately one can't get rid of timers and sit-for's because you have to wait until the frames are created or deleted "physically".

tee3:
This seems to work for me, `ediff-cleanup-mess` still runs and cleans up properly.

Issue #1945, opened 2015-07-02
Speedup log --graph by not using nested repetition in regex


steabert:
I recently swithed from magit 1.4.2 to magit 2.1.0.

When i show a log with "l l" (i.g. magit-log-current), outputting the graph takes forever, I have to kill it with Ctrl-G.
I read about the problems with showing logs with the "--graph" option, but since in version 1.4.2 everything worked fine, could this actually be some kind of regression?

Profiling indicates that code spends all of the time in "magit-log-wash-line", when calling "looking-at". If I disable that part, the log shows immediately with a graph, but obviously not fully washed (colors are there but sha1 hash and co are located after the graph)

Silex:
Do you have a public repo on which we can check the issue?

steabert:
Unfortunately not (and I don't control it), but looking through my repos I found that the git repo itself has the same problem, so you could test it on the git repository.

A (initial) log with --graph option using magit 1.4.2 takes about 1-2 s for the git repo, with magit 2.1.0 I'm waiting forever.

Silex:
You mean doing it on git@github.com:magit/magit.git? It's very snappy by me so obviously there must be some problem with your installation. I suggest you try to reinstall magit (and all the dependencies), and possibly test with emacs -Q to ensure the problem is not with your other configs.

tarsius:
I think he meant git's git repository.

My guess would have been bad commit metadata that causes the washing to cycle forever. But I expect there are no commits with bad metadata in git's own history.

I have increased the value of `magit-log-cutoff-length` you might want to lower that. But that probably won't help. The higher number might cause a delay, but I it would cause the washing to take forever.

Silex:
> I think he meant git's git repository.

Oh, ok I just tried and I can confirm it's very slow. The funny thing is that `l a` is fast because there's less graphs on the active branches.

steabert:
Sorry for any confusion, I meant indeed the https://github.com/git/git repository.
If I lower the `magit-log-cutoff-length` back to 100, magit 2.1.0 doesn't get stuck.

tarsius:
> magit 2.1.0 doesn't get stuck

I wonder what's at play here. What happens if you `git checkout HEAD~412` and then `l l`?

tee3:
Magit from MELPA 20150701.1326
GNU Emacs 24.5.1
OS X 10.10.4

[...]

PR #1984, opened 2015-07-06
Speedup log by improving regexp and not using --color

tarsius:
Includes @steabert's improved regexp (see #1945).

Not using `--color` by default is my idea. Here too the problem likely is an inefficient regexp, but it's out of our control, and probably not as easily fixable. I have chosen to not color graphs by default, because speed is important and many users who miss the colors should be able to spot that `--color` is now available from the log popup. Also gives them a change to learn how to set default arguments from popups.

Any comments or suggestions?

Looks like there's soon going to be a `2.2.0` release.

Off-topic: @steabert you might want to associate the email address used when creating that commit with your github account.

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -585,7 +585,7 @@ For internal use; don't add to a hook."
 
 (defconst magit-log-oneline-re
   (concat "^"
-          "\\(?4:\\(?: *[-_/|\\*o.] *\\)+ *\\)?"   ; graph
+          "\\(?4:[-_/|\\*o. ]*\\)"                 ; graph
           "\\(?1:[0-9a-fA-F]+\\) "                 ; sha1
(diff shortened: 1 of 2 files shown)


steabert:
Not using '--color' by default is indeed better.
I also used to modify 1.4.2 so that if i used next/previous keys in the log, only a diffstat of the commit would be displayed instead of the entire commit with diffs (makes more sense because n/p should be very fast, and n/p doesnt' move point to the commit buffer, so you want a short overview anyway). When pressing enter, the actual full commit with diffs would be displayed. Since this is about speeding up the log, I don't know if this belongs here or if I should open a feature request.

Thanks for spotting the email problem, I fixed it.

tarsius:
> Thanks for spotting the email problem

Well, it wasn't hard to spot. I had to click around quite a bit to get to your username :-)

PR #1990, opened 2015-07-06
magit-log: limit commit count as efficiently as possible

mgalgs:
Changes since v1 (ea4fffb98ea2332cc5c5423b3628da09e41d4892):

  - Addressed comments from @tarsius
  - Removed dependency on 97d25c4ada8ec268890750fb9536a42b26a33427, which is being dropped.

---

`--graph` is slow on large repositories [1], even when using `--max-count`
since `--graph` computes the entire history before doing its decoration.
We can speed things up a bit ourselves by limiting the number of commits
with a git revision [2].

This is especially useful for users who would like to use `--graph` in
their `magit-log-section-args`.

[1] http://www.spinics.net/lists/git/msg232230.html
[2] http://www.spinics.net/lists/git/msg232226.html

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -1059,13 +1059,28 @@ commits."
         (magit-insert-unpulled-commits)
       (magit-insert-recent-commits t))))
 
+(defun magit-commit-count-limiting-args (num-args)
+  "Returns a list of arguments for `git log' that limit the
+number of commits as efficiently as possible using a combination
+of git revisions and -NUM-ARGS."
+  (let ((args (list (format "-%d" num-args))))
+    (-when-let (nrevs (magit-git-string "rev-list"
+                                        (format "--max-count=%d" (1+ num-args))
+                                        "--count"
+                                        "HEAD"))
+      (setq nrevs (string-to-number nrevs))
+      (when (> nrevs num-args)
+        (setq args (cons (format "HEAD~%d.." (1- nrevs))
+                         args))))
+    args))
+
 (defun magit-insert-recent-commits (&optional collapse)
   "Insert section showing recent commits.
 Show the last `magit-log-section-commit-count' commits."
   (magit-insert-section (recent nil collapse)
     (magit-insert-heading "Recent commits:")
-    (magit-insert-log nil (cons (format "-%d" magit-log-section-commit-count)
-                                magit-log-section-arguments))))
+    (magit-insert-log nil (nconc (magit-commit-count-limiting-args magit-log-section-commit-count)
+                                 magit-log-section-arguments))))
 
 (defun magit-insert-unpulled-cherries ()
   "Insert section showing unpulled commits.


tarsius:
This doesn't seem to work:

```
$ git log --oneline --graph        
* ac62e00 m3
* 01434d3 m2
*   88149fc Merge branch 'other'
|\  
| * f12896c o3
| * d9ca182 o2
| * b83664a o1
|/  
* 9232bb5 m1

$ git log --oneline --graph --max-count=3
* ac62e00 m3
* 01434d3 m2
*   88149fc Merge branch 'other'
|\  

$ git log --oneline --graph HEAD~2..     
* ac62e00 m3
* 01434d3 m2

$ git log --oneline --graph HEAD~3..
* ac62e00 m3
* 01434d3 m2
* 88149fc Merge branch 'other'
* f12896c o3
* d9ca182 o2
* b83664a o1

$ git log --oneline --graph --max-count=99
* ac62e00 m3
* 01434d3 m2
*   88149fc Merge branch 'other'
|\  
| * f12896c o3
| * d9ca182 o2
| * b83664a o1
|/  
* 9232bb5 m1

$ git log --oneline --graph HEAD~99..     
fatal: ambiguous argument 'HEAD~99..': unknown revision or path not in the working tree.

```

mgalgs:
Yes, because of the way it's filtering out commits it might not draw the same lines.  The `HEAD~99..` case is because your repo must not have that many commits reachable from `HEAD`, and is the reason for `when`-checking [this](https://github.com/mgalgs/magit/blob/24ed82a7a310bd6b24d64230abdfc99d66884086/lisp/magit-log.el#L1072-1074).

tarsius:
> is the reason for when-checking this.

Okay, good.

But

```
$ git log --oneline --graph HEAD~3..
* ac62e00 m3
* 01434d3 m2
* 88149fc Merge branch 'other'
* f12896c o3
* d9ca182 o2
* b83664a o1
```

is no good:

* It shows more commits than the maximum we specified.
* The graph it displays is useless.

And I don't think this is a rare special case. It's the normal case.

If there is a merge in the last MAX commits (and in other cases we don't care about the graph), we only get a meaningful graph if there are at most MAX-3 commits on the branch being merged. Whenever there are more commits on the branch being merged, *all* of these commits will be displayed.

mgalgs:
> It shows more commits than the maximum we specified.

That's because you forgot the `-3`.  Note that `magit-commit-count-limiting-args` returns a combination of git revisions and -NUM-ARGS (see [here](https://github.com/mgalgs/magit/blob/log-optimization/lisp/magit-log.el#L1063-L1066)).

> The graph it displays is useless.

The graph above indeed looks useless.  But for the repositories I work on (like the linux kernel) and for a reasonably high value of `magit-log-section-commit-count` (15, for example) the context provided by these small graphs in my status buffer is invaluable.

tarsius:
Well, okay. But I would still strongly prefer if you or someone else could just role up their sleeves and fix this in Git itself, instead of asking me to add a kludge to Magit.

mgalgs:
I'm not sure this can even be "fixed" in git itself...  You really have to compute the whole history for `--graph` since anyone could branch off of the genesis commit at any time, make some commits there, and merge it into the tip.  I'm not sure you could figure out how to draw all the lines unless you traverse the entire history, even if you're not printing the commits at which those lines are terminating.

We have to do something though.  If anyone has any better ideas I'm all ears but magit is currently unusable for linux kernel development without this change.

tarsius:
Junio said:

> Surely. I am on a bus with terrible WiFi that does not let me use the usual terminal,
> but you would find a code in revision.c that sets revs->topo_order = 1 when it parses
> "--graph" option. If you disable it, that would stop "--graph" from wanting to compute
> the whole history before starting to emit stuff (and then stop at nth one with --max-count).

> I do not know what other side effects such a change would have, though.

He didn't say that he is sure that it won't work. Unless we try it, we won't know. And since it should clear by now that you will have a hard time getting this past me (at least until you exhaust that other option first), you might just as well spend that time trying to do fix this issue at its root. That might turn out to be impossible to do, but we won't know until we try. I am just not comfortable adding kludges unless necessary.

tarsius:
That was to harsh. I am sorry. I will consider this again. It's probably true that this usually gives good results, and that the simple test cases I constructed were not as typical as I though. I'll have to experiment some more. I think that's why I reacted so negatively: there's a lot going on right now, and so I wanted to avoid this work.

mgalgs:
Ok, take your time.  I've carried this in my fork for over two years now so no rush :smiley: 

Maybe we could start by making this optimization opt-in?  In other words, default behavior remains the same but provide an option for the adventurous (especially those graphing commits in their status buffer) to turn on this optimization?

tarsius:
Here's something I've been toying with:

```lisp
(defun magit-insert-recent-history ()
  (let* ((current (magit-get-current-branch))
         (tracked (and current (magit-get-tracked-branch)))
         (base    (and tracked (magit-git-string "merge-base" current tracked))))
    (magit-insert-section (recent)
      (magit-insert-heading "Recent commits:")
      (magit-insert-log
       (and base (list current tracked (concat "^" base "~")))
       (nconc (if base
                  (list "--graph"          ; doesn't make sense without this
                        (format "-%d" 50)) ; don't use v, it's to restrictive
                (list (format "-%d" magit-log-section-commit-count)))
              magit-log-section-arguments)))))
```

tarsius:
@mgalgs please checkout #2243.

Issue #1991, opened 2015-07-06
Respect .mailmap for log commands


glucas:
When a .mailmap is present it would be great to use it for the author name in magit log buffers. 

I first tried adding `--use-mailmap` to my `magit-log-arguments`, but that didn't work. After digging into the code a bit I found that changing the format string in `magit-insert-log` to use this:

    [%%aN]

instead of 

    [%%an]

works. I believe this change is generally safe: when there is no .mailmap file `%aN` behaves the same as `%an`.

For more on mapping authors with mailmap, see http://git-scm.com/docs/git-shortlog.

glucas:
The format change works for displaying the mapped author name in magit log buffers, but it looks like further changes would be required to also support mapped names for the `--author` filter. From the command line I can use for example `git log --use-mailmap --author="Joe"` to find commits for the user I've locally mapped to `Joe`. 

That isn't working in magit with the changes described above.

glucas:
If I add `"--use-mailmap"` to the log arguments in `magit-insert-log` then the author filter works as well.

I'll submit a pull request with these changes for consideration.

tarsius:
Okay, let's do it. Finish the pr and I will merge.

ORESoftware:
sorry but what does "respecting .mailmap" mean exactly?

glucas:
It means: if a .mailmap file is present, apply it to git log commands as described in the linked git documentation. 

The corresponding PR was merged years ago — I’ve been using this behavior for a long time now!

PR #1992, opened 2015-07-06
Respect .mailmap for git log commands

glucas:
Fix for  #1991.

This fix assumes that when a .mailmap file is present it should always be applied. It might be preferable to make this behavior configurable instead.

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -550,7 +550,7 @@ Type \\[magit-reset-head] to reset HEAD to the commit at point.
   "Insert a oneline log section.
 For internal use; don't add to a hook."
   (magit-git-wash (apply-partially 'magit-log-wash-log 'oneline)
-    "log" (magit-log-format-max-count) "--use-mailmap"
+    "log" (magit-log-format-max-count)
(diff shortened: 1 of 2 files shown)


tarsius:
Let's just use `%aN` for the time being (everywhere and including `%cN`), but not `--use-mailmap`.

I am quite undecided what to do about that argument. We could just use it unconditionally, or add it to the popup. Or add it to the popup and then also consult it's value to decide whether to use `%an` or `%aN`. I would kinda like that last variant, but when the user consults the respective git documentation then that would contradict what actually happens. On the other hand there is not other way for the user to control what %-sequence is used. So maybe the format string should be configurable. But that would make it very easy for users to shut themselves in the foot. And I plan changes to the popup functionality. And so on.

By the way you can easily add this argument to the popup yourself. The manual documents how.

tarsius:
On the other hand, lets just try and see. Keep the `--use-mailmap` (but move it further down in the list of arguments) and also replace occurrences of `%cn` too. I cannot guarantee I will keep this if it starts causing issues, then again I don't really expect there will be any.

tarsius:
A have cherry-picked and then reshuffled the changes a bit.

Thanks!

PR #2001, opened 2015-07-07
Expand FAQ on using magit with Tramp

jschwab:
Since magit requires git > 1.9.4, newer than the version shipped by
older distributions (e.g., Debian 7), some users may choose to install a
newer git version.  Tramp may not detect this git if it is installed in
a non-standard location.  This points the user to the documentation on
tramp-remote-path.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -4433,6 +4433,14 @@ only undo that setting in Magit by customizing
 Magit requires at least version ~1.9.4~.  When connecting to a remote
 machine using Tramp it requires that same version on that remote.
 
+By default Tramp searches for programs in the directories given by the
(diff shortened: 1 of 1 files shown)


jschwab:
It was not clear to me whether documentation patches should be against only the .org files or also the .texi files.

I tried to generate the .texi files, but when I did so, there were non-whitespace differences between the .texi files present in the repo and the new files.  So this is a patch against magit.org only.

tarsius:
> So this is a patch against magit.org only.

That's okay for now, the tool chain is still a bit wacky. I will just generate myself and amend.

tarsius:
Thanks.

Applied with the regenerated `magit.texi`.

Issue #2006, opened 2015-07-07
Ask user when commit selection leaves it undecided whether to use a...b, a..b, b...a, or b..a


tarsius:
When the user selects sibling commits (and branches) using the region and neither end-point is an ancestor of the other, with the intention to use diff or ediff, then it is ambiguous which comparison is actually desired.

Currently we always use a..b, a being the end-point which appears first in the buffer.

A tiny part of the solution can be found in `magit-ediff-compare--read-revisions`.

kyleam:
Another option would be to leave the behavior as is and allow both
these things (the order and range type) to be changed with a key in
magit-diff-refresh-popup or magit-diff-mode-map.

kyleam:
Ahh, but I guess that wouldn't handle ediff (which could do the swapping fine, but not the range change).

kyleam:
As of bd5a2d91ae95e7bf90ee830ddb6f66df0620d81e, it's possible to use a "..." range.  @tarsius I'll close this and let you reopen if you want something other than that solution.

PR #2010, opened 2015-07-08
[RFC] Diff actions to switch range arguments and type

kyleam:
These commits add two actions to the magit-diff-refresh-popup: one that switches the range type (back and forth between ".." and "...") and one that swaps the revision arguments in a range.  What do you think?

(somewhat related to #2006)


tarsius:
I'll decide on this later.

tarsius:
Thanks. That's really useful.

PR #2022, opened 2015-07-08
Replace magit-remote-update with magit-fetch-all

### LINKED ISSUE #2021: symbol's function definition is void: magit-remote-update
when I try to do remote update from the Magit menu in X11.

kyleam:
magit-remote-update was renamed to magit-fetch-all in 9f873f6f1c694330.

Fixes #2021.


tarsius:
Applied. Thanks.

PR #2027, opened 2015-07-09
magit-diff-visit-file: Support magit-diff-mode

kyleam:
If magit-refresh-args for magit-diff-mode looks like
("revA...?revB" _ _) or ("...?revB" _ _), use magit-find-file-noselect
with revB unless revB is HEAD.

Use find-file-noselect for all other cases ("rev...?[HEAD]", unstaged,
staged, working tree relative to a revision).


tarsius:
Thanks.

PR #2028, opened 2015-07-09
magit-log-read-revs: Fix splicing

kyleam:
Fix spliced list introduced in 5a974ff6153f15311865b34bf1eaaeea5277e48f.


tarsius:
Yikes

tarsius:
Added to the fruit basked.

PR #2036, opened 2015-07-10
WIP always show revision header but make it selectable

tarsius:
Display the same information in the `header-line` and the first line
of the buffer text, but make sure only one of them is every actually
displayed.

When the `window-start` is visible, then don't show the `header-line`.

There are still some minor issues, but I generally like it.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -1334,7 +1334,8 @@ Type \\[magit-reverse] to reverse the change at point in the worktree.
         (and (> window-start-pos 48)
              (concat
               (propertize " " 'display
-                          `((space :width (,(+ (car (window-fringes)) 2)))))
+                          `((space :width (,(+ (car (window-fringes))
+                                               2))))) ; FIXME get from face :box
               (propertize (concat "Commit "
                                   (magit-rev-parse (car magit-refresh-args)))
                           'face 'magit-header-line)))))
@@ -1355,8 +1356,9 @@ Type \\[magit-reverse] to reverse the change at point in the worktree.
       (magit-delete-line)
       (magit-insert-section (headers)
         (magit-insert-heading
-          (propertize (concat "Commit " rev)
-                      'face '(magit-header-line header-line)))
+          (propertize (concat "Commit " rev
+                              "\n") ; FIXME underline display line
+                      'face 'magit-header-line))
         (when refs
           (magit-insert (format "References: %s\n"
                                 (magit-format-ref-labels refs))))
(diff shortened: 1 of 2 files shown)


kyleam:
Once the fringe is accounted for, I think the transition will seem
pretty seamless.  One place where this breaks down is is calling
beginning-of-buffer at the end of a scrolled buffer, which results in
two hash lines until the next command is given.  (I'm not sure why
post-command-hook doesn't catch this case.)

npostavs:
> One place where this breaks down is is calling
> beginning-of-buffer at the end of a scrolled buffer, which results in
> two hash lines until the next command is given. (I'm not sure why
> post-command-hook doesn't catch this case.)

I'm also seeing it fail if I call `next-line` when point is on the bottom line of the window. Maybe the post-command-hook sometimes runs before the display updates??

But when it works, it's hardly noticeable, not annoying at all.

tarsius:
Pushed an update with the flickering fixed. I had to add two pixels to the fringe-width, I hope that's a constant value.

I see two more issues:

* The fringe is shown. But that could be considered a feature as it makes it more likely that one notices that the line can be entered.

* If the face has the `:underline` property set, then that is not used for the complete line, even if the trailing newline is also propertized.

We should probably take these issues to the stackexchange site.

> Conjecture: it works with scrolling commands and fails with point movement commands because point movement triggers scrolling only later in the command loop.

That brings back bad memories...

Is there another hook we could use. Some post motion hook. Or are they triggered at the wrong time too?

kyleam:
Hmm, also seems to break for magit-diff-show-or-scroll-up.

tarsius:
Unlikely this will  be done in time for `2.2.0` then.

npostavs:
Using `window-scroll-functions` seems to work:

```diff
diff --git i/lisp/magit-diff.el w/lisp/magit-diff.el
index 6367f3c..96b5956 100644
--- i/lisp/magit-diff.el
+++ w/lisp/magit-diff.el
@@ -1320,12 +1320,12 @@ (define-derived-mode magit-revision-mode magit-diff-mode "Magit"
 Type \\[magit-reverse] to reverse the change at point in the worktree.
 \n\\{magit-revision-mode-map}"
   :group 'magit-revision
-  (add-hook 'post-command-hook 'magit-revision-hideshow-header-line nil t)
+  (setq-local window-scroll-functions #'magit-revision-hideshow-header-line)
   (hack-dir-local-variables-non-file-buffer))
 
-(defun magit-revision-hideshow-header-line ()
+(defun magit-revision-hideshow-header-line (window window-start-pos)
   (setq header-line-format
-        (and (not (pos-visible-in-window-p 1))
+        (and (> window-start-pos 48)
              (concat
               (propertize " " 'display
                           `((space :width (,(+ (car (window-fringes)) 2)))))
```

Note: I thought we're supposed to use `add-function` for `*-functions` type variables, but when I do `(add-function :after (local 'window-scroll-functions) #'magit-revision-hideshow-header-line)` Emacs tries to `(apply nil ...)` when running the `window-scroll-functions` so I just used `setq-local` instead.

kyleam:
Using `window-scroll-functions` seems to be working nicely.

PR #2039, opened 2015-07-10
Enable --force flag in magit-branch-popup

jowl:
This enables doing `git branch --force some-branch origin/some-branch` without first checking out `some-branch`. While not the most common use-case of `git-branch`, it is something that I want to do from time to tome.

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -1006,7 +1006,9 @@ Non-interactively DIRECTORY is (re-)initialized unconditionally."
   "Popup console for branch commands."
   'magit-commands
   :man-page "git-branch"
-  :switches '((?t "Set upstream configuration" "--track"))
+  :switches '((?t "Set upstream configuration" "--track")
+              (?f "Force"                      "--force"))
+
   :actions  '((?b "Checkout"          magit-checkout)
               (?u "Set upstream"      magit-branch-set-upstream)
               (?k "Delete"            magit-branch-delete)


tarsius:
I'll consider that, but note that you can add your own arguments. http://magit.vc/manual/magit-popup/Customizing-existing-popups.html

jowl:
Nice, I didn't know that. That certainly reduces the need for my proposed change.

tarsius:
Alternatively you can use `C-u x` (`git reset --hard`). If point is on a branch or commit that will be offered as default. This is quite similar, but would not update the upstream branch, which I believe your approach would do.

tarsius:
It would probably be better to add a "Reset" variant to the branch popup. What do others think? Unnecessary? To confusing?

tarsius:
Better yet, do what `magit-branch-rename` does: ask for confirmation when there's a conflict and then use the appropriate argument if the user confirms.

jowl:
I really like your last alternative.

npostavs:
> Alternatively you can use C-u x (git reset --hard). If point is on a branch or commit that will be offered as default. This is quite similar, but would not update the upstream branch, which I believe your approach would do.

I think the point is moving a branch which is not currently checked out; `reset` acts on the checked out branch.

> It would probably be better to add a "Reset" variant to the branch popup, instead of the argument --force. What do others think? Unnecessary? To confusing?
> Better yet, do what magit-branch-rename does: ask for confirmation when there's a conflict and then use the appropriate argument if the user confirms.

I think adding a "Reset" action would be best. Otherwise, the user has to use the "Create" action to move a branch which is kind of weird (it would probably make sense to git experts, but nobody else). Possibly it should be called "Move" instead. Resetting is like a special case of move, in that it moves the current branch. But I don't think making "Move" into a special case of "Create" makes much sense.

jowl:
@npostavs is correct, the point was to move a branch that wasn't checked out. I guess it'd make more sense to call the action "Reset", but to clarify I'm happy with being able to adding my own arguments through the `magit-define-popup-switch` and wont pursue this further. Close at will.

npostavs:
I'm now using the following in my init.el to add a "Move" action:
```elisp
(defun magit-branch-move (branch to &optional args)
  (interactive
   (let ((branch (magit-read-branch "Move branch" (magit-get-current-branch))))
     (list branch
           (magit-read-other-branch-or-commit "to" branch)
           (magit-branch-arguments))))
  (magit-branch branch to (cl-adjoin "--force" args :test #'equal)))
(magit-define-popup-action 'magit-branch-popup
  ?m "Move" 'magit-branch-move)
```

tarsius:
"Rebase" is better here because "move" already means something else in the context of `git branch`: it's a synonym for "rename".

npostavs:
> "move" already means something else in the context of git branch: it's a synonym for "rename".

Oops, forgot about that.

tarsius:
I have added `magit-branch-reset`.

PR #2041, opened 2015-07-10
magit-{change,remove}-popup-key: fix doc :keywords

npostavs:
The docstrings and manual should refer to :actions, :switches, and
:options (not :action, :switch, and :option).

---

I didn't try to update the .texi as I understand some fancy hacks are needed.


tarsius:
Hm. I would say the documentation is right and the code is wrong. These functions change just one event not multiple. I think I fixed that on a feature branch which later was abandoned, because the refactoring and cleanup was just going to far, to be feasible before the release.

Could you please change the code instead? I think it would be best if both the singular and, for backward compatibility reasons, the plural were accepted.

npostavs:
ok

PR #2043, opened 2015-07-10
magit-{define,change,remove}-popup-key: singular type

npostavs:
Let these functions handle :action, :option, and :switch as
documented. The plural versions :actions, :options, and :switches are
still accepted.

---
context: https://github.com/magit/magit/pull/2041#issuecomment-120403699


tarsius:
Thanks

PR #2050, opened 2015-07-11
Docstring fixes

kyleam:


tarsius:
Thanks. Picked.

tarsius:
I have given you commit access. For more complex things please still open a pr, but trivial stuff like this you can apply directly - saves us both work. And thanks for all these fixes!

tarsius:
@npostavs I have given you commit access too.

@npostavs @kyleam As you might have noticed, I have been cherry-picking a lot lately. Please do the same for the time being and avoid merged (especially merged "in the wrong direction", but I am sure pointing that out was unnecessary).

@sigma @dudebout @vanicat Unless you intend to return soon and with fireworks, I would like to remove you from the Owners and Developers teams, and instead add @npostavs and @kyleam to Owners too (in addition to Developers).

kyleam:
> I have given you commit access. For more complex things please still
> open a pr, but trivial stuff like this you can apply directly

Great, thank you

dudebout:
@tarsius I did not know about this concept of teams on Github. Your proposal seemed fair, so I went ahead and did the three changes for myself.

Issue #2011, opened 2015-07-08
magit-rebase-autosquash can choose a bad commit


phil-s:
`magit-rebase-autosquash` uses `(magit-get-tracked-branch)` for its interactive `COMMIT` argument, which works for typical use cases, but breaks in some situations:

`magit-get-tracked-branch` returns the upstream HEAD reference. Typically this will be the commit prior to your unpushed changes, so this results in a rebase target which spans (only) all of your commits, which is more or less what you want.

I was just experimenting in a state where I had unpulled commits as well as unpushed, and I was trying to rebase/autosquash some of my unpushed commits. `r f` fails here, as the upstream HEAD is the wrong commit to rebase against.

A manual `(magit-rebase-autosquash "COMMIT^")` for the appropriate `COMMIT` (as I'd choose for `magit-rebase-interactive`) works as desired, as does `r -a e` with an interactive selection. I wonder if `magit-rebase-autosquash` and friends should similarly be using the commit at point, rather than trying to figure it out by itself?

I guess automatically picking the parent of the oldest unpulled commit would be correct in my scenario; but I don't know offhand whether there are other scenarios to consider as well?

tarsius:
Haven't looked at the code but I would think that we should start from the merge-base of the current branch and its upstream.

phil-s:
Thanks tarsius! That works nicely in my tests.

I believe the same thing needs to be done to `magit-rebase-unpushed` as well (it's essentially the same scenario), and perhaps also for the default suggestions made by `magit-rebase` and `magit-rebase-from`, which are also currently using `magit-get-tracked-branch`.

That would cover everything in `magit-sequence.el` calling that function.

phil-s:
On second look, I think `magit-rebase` and `magit-rebase-from` are correct as they are (I concluded that using the merge base makes no sense for a non-interactive rebase); so it's just `magit-rebase-unpushed` which needs a copy of that fix?

PR #2053, opened 2015-07-11
magit-rebase-unpushed: start from merge-base

phil-s:
Replica of the magit-rebase-autosquash fix from commit
f1afebf6b6e6128c701f87f6bfa3d9ffca938de3. See #2011.


tarsius:
Thanks! I forgot about that (when fixing the other instance, but also your reminder).

PR #2056, opened 2015-07-11
Use uniquify.el for uniquifiying buffer names

npostavs:
There might be a problem if `topdir` is `nil`. I couldn't remember under what scenario that happens.


tarsius:
@npostavs @kyleam Please push your feature branches directly to this repository. Then we can all add fixup commits, like @npostavs has done for my #2036.

I am closing this pr and creating my own, so that I can start doing so right away and demonstrate a little feature I have added.

Issue #2054, opened 2015-07-11
(magit-status "/other/foo") from a *magit-status: foo* buffer doesn't create new buffer


npostavs:
when using a `%b` in `magit-status-buffer-name-format`. It's the same for the interactive case: `C-u magit-status RET /other/foo RET`.

### Recipe
`emacs -Q -L ..\dash -L lisp -l magit -l testing.el` where `testing.el` contains
```elisp
(require 'magit)
(setq magit-status-buffer-name-format "*magit-status: %b*")

(magit-status "~/src/dash")
;; (switch-to-buffer "*scratch*")
(magit-status "~/.emacs.d/el-get/dash")
(setq Buffer-menu-name-width 40)
(call-interactively 'list-buffers)
```

Then `*Buffer List*` will contain only
```
.%  *magit-status: dash*                        164 Magit            
    *scratch*                                      0 Lisp Interaction 
 %* *Messages*                                   155 Messages         
```
Uncommenting the `(switch-to-buffer "*scratch*")` line in `testing.el` results in 2 `magit-status` buffers as expected:
```
.%  *magit-status: dash*<2>                      164 Magit            
    *scratch*                                      0 Lisp Interaction 
 %  *magit-status: dash*                         152 Magit            
 %* *Messages*                                   155 Messages               
```

### Analysis

The problem is `magit-status-internal` let-binds `default-directory` to `/other/foo` but this affects the buffer local binding, so then `magit-mode-get-buffer` thinks `*magit-status: foo*` is already at `/other/foo` and reuses it.

This is kind of a weird corner case: I only found it because I was looking at using `uniquify.el` with magit. Though I guess it might actually be fairly common to want to see the status of your other/foo while looking your first foo.

This could be fixed by wrapping the body of `magit-status-internal` in a `with-temp-buffer`, although that feels kind of silly. Thoughts?

tarsius:
That seems like a good solution, but not exactly obvious, so write it like `(with-temp-buffer ; #2054`.

PR #2057, opened 2015-07-11
Use uniquify.el for uniquifiying buffer names

tarsius:
(Re-roll of  #2056) @npostavs said:

There might be a problem if `topdir` is `nil`. I couldn't remember under what scenario that happens.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -74,9 +74,7 @@
 (defcustom magit-diff-buffer-name-format "*magit-diff: %a*"
   "Name format for buffers used to display a diff.
 
-The following `format'-like specs are supported:
-%a the absolute filename of the repository toplevel.
-%b the basename of the repository toplevel."
+See `magit-mode-get-buffer' for an explanation of formats."
   :package-version '(magit . "2.1.0")
   :group 'magit-diff
   :type 'string)
@@ -243,9 +241,7 @@ many spaces.  Otherwise, highlight neither."
 (defcustom magit-revision-buffer-name-format "*magit-rev: %a*"
   "Name format for buffers used to display a commit.
 
-The following `format'-like specs are supported:
-%a the absolute filename of the repository toplevel.
-%b the basename of the repository toplevel."
+See `magit-mode-get-buffer' for an explanation of formats."
   :package-version '(magit . "2.1.0")
   :group 'magit-revision
   :type 'string)
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -59,9 +59,7 @@
 (defcustom magit-log-buffer-name-format "*magit-log: %a*"
   "Name format for buffers used to display log entries.
 
-The following `format'-like specs are supported:
-%a the absolute filename of the repository toplevel.
-%b the basename of the repository toplevel."
+See `magit-mode-get-buffer' for an explanation of formats."
   :package-version '(magit . "2.1.0")
   :group 'magit-log
   :type 'string)
@@ -204,9 +202,7 @@ be nil, in which case no usage information is shown."
 (defcustom magit-cherry-buffer-name-format "*magit-cherry: %a*"
   "Name format for buffers used to display commits not merged upstream.
 
-The following `format'-like specs are supported:
-%a the absolute filename of the repository toplevel.
-%b the basename of the repository toplevel."
+See `magit-mode-get-buffer' for an explanation of formats."
   :group 'magit-log
   :type 'string)
 
@@ -223,9 +219,7 @@ The following `format'-like specs are supported:
 (defcustom magit-reflog-buffer-name-format "*magit-reflog: %a*"
   "Name format for buffers used to display reflog entries.
 
-The following `format'-like specs are supported:
-%a the absolute filename of the repository toplevel.
-%b the basename of the repository toplevel."
+See `magit-mode-get-buffer' for an explanation of formats."
   :package-version '(magit . "2.1.0")
   :group 'magit-log
   :type 'string)
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -459,6 +459,8 @@ the function `magit-toplevel'."
                    (equal default-directory topdir)))
             (buffer-list)))
 
+(add-to-list 'uniquify-list-buffers-directory-modes 'magit-status-mode)
+
 (defun magit-mode-get-buffer (format mode &optional topdir create)
   (if (not (string-match-p "%[ab]" format))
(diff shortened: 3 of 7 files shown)


tarsius:
My squash commit was created using the new `magit-commit-augment` instead of `magit-commit-squash` which does not allow editing the message. I used that to comment on the issue you raised above.

What's not so nice about this is that you cannot see that comment here, but first have to click on the link above. But in the commit message I have mentioned you - did that cause a Github notification?

tarsius:
Okay, now back to your patch...

This doesn't seem to do anything for me, do I have to do anything to enable this feature?

Also this should be optional, probably depending on what %-sequence is used in the buffer-name format. We could repurpose %b, or replace it with %r.

npostavs:
> But in the commit message I have mentioned you - did that cause a Github notification?

Yup. It showed up in by inbox first.

> This doesn't seem to do anything for me, do I have to do anything to enable this feature?

No it should be turned on from the start. You can use the recipe from #2054 to test it. Um, just realized that only tests magit-status buffers, and I forgot to enable it on other modes.

> Also this should be optional, probably depending on what %-sequence is used in the buffer-name format. We could repurpose %b, or replace it with %r.

Oh, it's already dependent on using `%b` in the format, otherwise there will be no buffer name conflicts, and uniquify.el doesn't do anything.

tarsius:
I have tried this again, and I like it.

The name is uniquified by adding a string *after* the `*`. If it's not feasible to put it before that, then I would be okay with dropping both stars. But first try if you can keep the trailing star and keep it at the end.

I think we could also make it the default like this:

* Drop support for plain "basename" variant. As we have seen it has issues.
* Use `%u` to mean "uniquify".
* Treat `%b` as a synonym for `%u`, for the sake of people who have already customized the formats.
* Default to uniquify by replacing `%a` with `%u` in all formats.
* Add a variable to turn of uniquify globally, so that users don't have to customize each format individually.
* Adjust the documentation. This should also make it into the manual.

npostavs:
> The name is uniquified by adding a string after the `*`. If it's not feasible to put it before that, then I would be okay with dropping both stars. But first try if you can keep the trailing star and keep it at the end.

Okay, I'll investigate this.

> Drop support for plain "basename" variant. As we have seen it has issues.

If you refer to #2054, I believe the issue is the same with uniquify added because uniquify only comes into effect after creating the new buffer.


> Use %u to mean "uniquify".

I think having uniquify acting on %b formats (what's currently pushed here) makes sense. In current magit without uniquify, %b formats get uniquified in the old style `*magit-status: foo*<2>`. Since 24.4, Emacs enables uniquify by default so it makes sense that magit will use `uniquify-buffer-name-style` for this instead.

> Add a variable to turn of uniquify globally, so that users don't have to customize each format individually.

They could use `uniquify-ignore-buffers-re`.

npostavs:
> The name is uniquified by adding a string after the *. If it's not feasible to put it before that, then I would be okay with dropping both stars. But first try if you can keep the trailing star and keep it at the end.

The uniquify code really wants to treat the whole name as a single unit, it doesn't want to put stuff in the middle. It can be worked around by renaming after unquify has done it's job, a bit hacky but it seems to work (b6cf00f23bd7cfe1579f62b82c69c54f648b6508).

> But in the commit message I have mentioned you - did that cause a Github notification?

Fun fact: when I rebased it, I sent myself the notification again.

tarsius:
Alright. Could you please also mention uniquify in the relevant doc-strings. I just noticed that only `magit-status-buffer-name-format` is mentioned in the manual. The other variables should also be documented in the appropriate sections, and, more importantly, the concept should be described somewhere in "Modes and Buffers". If you could do that, that would be great.

Do you, and @kyleam, think we should default to %b?

Also did you investigate the use of uniquify.el when listing known repositories?

npostavs:
> Do you, and @kyleam, think we should default to %b?

I think `%b` with uniquifying is more readable, so yes.

> Also did you investigate the use of uniquify.el when listing known repositories?

No, is this for `magit-list-repos-uniquify` called from `magit-read-repository`? I *suspect* uniquify won't be helpful there because there are no buffers involved, but I'll see if something can be hacked in.

kyleam:
> think we should default to %b?

yes, I think so

npostavs:
As I was going to update the manual, I noticed the `magit-rename-buffer` function which made me realize there is a flaw in this unquifiying scheme: it doesn't distinguish between buffers renamed by uniquify and buffers renamed by the user. So if you call `(magit-status "sub2/repo")` twice (assuming there is already a `"sub1/repo"` open), on the second call you get a new status buffer called `*magit-status: repo<sub2>*<2>`. The problem is present even without uniquify, where you would get a new `*magit-status: repo*<3>` instead of the existing `*magit-status: repo*<2>`.

tarsius:
> As I was going to update the manual, I noticed the `magit-rename-buffer` function which made me realize there is a flaw in this unquifiying scheme:

Just thought about that too yesterday and was going to mention it here :-)

> on the second call you get a new status buffer called `*magit-status: repo<sub2>*<2>`.

First, it doesn't make sense to use `magit-rename-buffer` with the status buffer, it is intended for diff, log, and similar buffers, where you might actually want multiple buffers (per-repo) open at the same time.

Secondly, I occasionally use `magit-rename-buffer` (without the uniquify changes) and never noticed any problems. So you might want to try again with a diff buffer, and without using uniquify. Also I don't quite understand your description of what is going wrong.

npostavs:
> Secondly, I occasionally use magit-rename-buffer (without the uniquify changes) and never noticed any problems. So you might want to try again with a diff buffer, and without using uniquify. Also I don't quite understand your description of what is going wrong.

I omitted some details, let me try again.

The problem occurs when using a `%b` format and having at least 2 buffers in repositories with the same base name open. So if you have `magit-diff-buffer-name-format` = `*magit-diff: %b*`, and you open a diff buffer in `/foo/repo` and then a 2nd diff buffer in `/other-foo/repo`, then the 2nd diff buffer will be named `*magit-diff: repo*<2>`, i.e., Emacs already renames without you calling `magit-rename-buffer`.

 Opening a diff buffer in the 2nd repo again will open yet another `*magit-diff: repo*<3>`.

So this isn't directly related to the uniquify stuff, except that probably nobody would want to use `%b` without uniquify.

tarsius:
I would like to postpone this after v2.2.0, which I would like to release on the 1th August (or if necessary a few days after that).

tarsius:
This will conflict heavily with #2192. I am closing this, but intend to port these changes onto the things I am working on in `jb/get-buffer-by-mode-and-args`, so don't delete this branch just yet.

Issue #1953, opened 2015-07-02
magit-fullscreen is gone


ihodes:
Is there a replacement or suggested library/approach to making magit full screen? Previously I did the following:

```elisp
(defadvice magit-status (around magit-fullscreen activate)
    (windowt-configuration-to-register :magit-fullscreen)
    ad-do-it
    (delete-other-windows))

(defun magit-quit-session ()
    "Restores the previous window configuration and kills the magit buffer"
    (interactive)
    (kill-buffer)
    (jump-to-register :magit-fullscreen))

(bind-keys :map magit-status-mode-map
           ("q" . magit-quit-session))
```

tarsius:
Yes.

```lisp
(setq magit-restore-window-configuration t) ; that's the default actually
(setq magit-status-buffer-switch-function
      (lambda (buffer) ; there might already be an Emacs function which does this
        (pop-to-buffer buffer)
        (delete-other-windows)))
```

kyleam:
This also works:

```
(add-hook 'magit-status-mode-hook 'delete-other-windows)
```

ihodes:
@tarsius superb, thank you.

tarsius:
@kyleam So the buffer gets displayed before `(funcall (or switch-function ....))` in `magit-mode-display-buffer`? I don't see where that would happen.

kyleam:
I've been using that since I switched to next.  Seems to work.

garyo:
Kyle's solution works for me too; very nice!

On Thu, Jul 2, 2015 at 4:37 PM, Kyle Meyer <notifications@github.com> wrote:

> I've been using that since I switched to next. Seems to work.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/magit/magit/issues/1953#issuecomment-118160600>.
>



-- 
Gary

codeitagile:
Excellent.

Silex:
@kyleam 's solution suddenly stopped working :-(

jguenther:
I think the `fullframe` [package](https://github.com/tomterl/fullframe) might be what you're looking for:

```
(require 'fullframe)
(fullframe magit-status magit-mode-quit-window)
```

tarsius:
There were some big changes in that area. Have a look at the "Switching Buffers" node in the magit manual. I would be interested in a solution that goes full screen just using display actions.

bmag:
This is what I am currently trying, and it seems to work:
```elisp
(setq magit-display-buffer-function
[...]

PR #2060, opened 2015-07-12
magit-status-internal: switch buffer permanently

npostavs:
This should fix the tests.

Context: #2054 and 3753d07c8b1655c1a165a104d25ff9e7f1fdc6e5,

@tarsius: does this make sense? I'm a bit confused as to how `magit-status` and `magit-status-internal` are *supposed* to affect current buffer. In particular, how come the tests were broken, while the interactive commands continued to work fine??

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -383,14 +383,19 @@ then offer to initialize it as a new repository."
 (put 'magit-status 'interactive-only 'magit-status-internal)
 
 (defun magit-status-internal (directory &optional switch-function)
-  (with-temp-buffer ; don't let-bind status buffer's `default-directory', #2054
-    (let ((default-directory (file-name-as-directory
-                              (expand-file-name directory))))
-      (magit-mode-setup magit-status-buffer-name-format
-                        (or switch-function
-                            magit-status-buffer-switch-function)
-                        #'magit-status-mode
-                        #'magit-status-refresh-buffer))))
+  (let ((temp-buffer (generate-new-buffer " *temp*")))
+    (unwind-protect
+        (progn ; let-bind `default-directory' in temp buffer, see #2054
+          (set-buffer temp-buffer)
+          (let ((default-directory (file-name-as-directory
+                                    (expand-file-name directory))))
+            (magit-mode-setup magit-status-buffer-name-format
+                              (or switch-function
+                                  magit-status-buffer-switch-function)
+                              #'magit-status-mode
+                              #'magit-status-refresh-buffer)))
+      (and (buffer-name temp-buffer)
+           (kill-buffer temp-buffer)))))
 
 (defun ido-enter-magit-status ()
   "Drop into `magit-status' from file switching.


tarsius:
>I'm a bit confused as to how `magit-status` and `magit-status-internal` are *supposed* to affect current buffer.

When they are called the current buffer only matters because that's where we get the value of `default-directories` from.

> In particular, how come the tests were broken, while the interactive commands continued to work fine??

That's where I start being confused too. Maybe we hit the "FIXME: kill-buffer can change current-buffer in some odd cases." found inside `with-temp-buffer`? How did you come up with your solution? I mean all you do is use `set-buffer` twice instead of `with-current-buffer`. I wouldn't expect that to make a difference. The again I have run into odd behavior with the latter, so maybe I would have tried that too.

I think `with-current-buffer` should be defined like this:

```lisp
(defmacro with-current-buffer (buffer-or-name &rest body)
  (let ((temp-buffer (make-symbol "temp-buffer")))
    `(let ((,temp-buffer ,buffer-or-name))
       (save-current-buffer
         (set-buffer ,temp-buffer)
         ,@body))))
```

instead of

```lisp
(defmacro with-current-buffer (buffer-or-name &rest body)
  `(save-current-buffer
     (set-buffer ,buffer-or-name)
     ,@body))
```

Because if `buffer-or-name` is evaluated inside `save-current-buffer` and the concrete `buffer-or-name` also changes buffers (which `magit-mode-setup` does by using `magit-mode-display-buffer`), then odd things start happening. Maybe that doesn't matter here because we use `with-temp-buffer` which does generate the buffer outside of `with-current-buffer`, but maybe the use of `with-current-buffer` inside `magit-mode-display-buffer` still messes it up somehow.

I think that the underlying issue could be that `save-current-buffer` doesn't nest well and ends up restoring the wrong buffer. But I didn't dive into this further.

Anyway, it's quite possible that our logic is sound and that we just managed to hit some odd courner case. Especially because the tests actually succeed when run interactively. I suspect (and could probably look that up somewhere) that making a buffer temporarily current also affects what buffers are considered to be currently visible in the current frame (but apparently it makes a difference if Emacs is run interactively or in batch mode).

And that matters to `magit-mode-get-buffer`.

----

**Aha!!!** So while the above is probably rather confusing. And I myself am far from understanding what the hell is going on, I am on the right track:

```diff
diff --git a/lisp/magit-mode.el b/lisp/magit-mode.el
index e8b8123..5d2b3ce 100644
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -423,13 +423,15 @@ (defun magit-mode-display-buffer (buffer mode &optional switch-function)
          (setq buffer (magit-mode-get-buffer-create buffer mode)))
         ((not (bufferp buffer))
          (signal 'wrong-type-argument (list 'bufferp nil))))
-  (let ((section (magit-current-section)))
-    (with-current-buffer (get-buffer-create buffer)
+  (let* ((section (magit-current-section))
+         (buffer (get-buffer-create buffer))
+         (window (get-buffer-window buffer (selected-frame))))
+    (with-current-buffer buffer
       (setq magit-previous-section section)
       (if magit-inhibit-save-previous-winconf
           (when (eq magit-inhibit-save-previous-winconf 'unset)
             (setq magit-previous-window-configuration nil))
-        (unless (get-buffer-window buffer (selected-frame))
+        (unless window
           (setq magit-previous-window-configuration
                 (current-window-configuration))))))
   (funcall (or switch-function
diff --git a/lisp/magit.el b/lisp/magit.el
index 2819c54..667588d 100644
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -383,14 +383,13 @@ (defun magit-status (&optional directory)
 (put 'magit-status 'interactive-only 'magit-status-internal)
 
 (defun magit-status-internal (directory &optional switch-function)
-  (with-temp-buffer ; don't let-bind status buffer's `default-directory', #2054
-    (let ((default-directory (file-name-as-directory
-                              (expand-file-name directory))))
-      (magit-mode-setup magit-status-buffer-name-format
-                        (or switch-function
-                            magit-status-buffer-switch-function)
-                        #'magit-status-mode
-                        #'magit-status-refresh-buffer))))
+  (let ((default-directory (file-name-as-directory
+                            (expand-file-name directory))))
+    (magit-mode-setup magit-status-buffer-name-format
+                      (or switch-function
+                          magit-status-buffer-switch-function)
+                      #'magit-status-mode
+                      #'magit-status-refresh-buffer)))
 
 (defun ido-enter-magit-status ()
   "Drop into `magit-status' from file switching.
```

Now the tests pass in batch mode too. So what did we learn? `(with-current-buffer DANGER-ZONE &rest BODY)` :-/

Some more notes (I am gonna take a break before finishing this):

* Limited testing indicates that moving `(get-buffer-window buffer (selected-frame))` outside of `with-current-buffer` is not actually necessary. But I no longer trust `with-current-buffer` and want to play it save.

* `magit-mode-display-buffer` returns the value of its BUFFER argument, a buffer *or* buffer name. It probably should always return the buffer returned by `get-buffer-create`. But I don't think it really matters, at least not to the current caller.

* This might break https://github.com/magit/magit/issues/1953#issuecomment-118155182, which would be a good thing, because I could not figure out how that could possibly work. (We should then however provide a substitute for https://github.com/magit/magit/issues/1953#issuecomment-118152788.

tarsius:
Pushed this to https://github.com/magit/magit/tree/jb/semi-current-buffer.

npostavs:
> When they are called the current buffer only matters because that's where we get the value of default-directory from.

But what is supposed to have happened after they are called?

> How did you come up with your solution? I mean all you do is use set-buffer twice instead of with-current-buffer.

I was guessing that `magit-status-internal` was supposed to switch to the (possibly new) status buffer, and wrapping `with-current-buffer` was undoing it. But I saw that `M-x magit-status` still switched buffers despite `with-current-buffer`, hence my confusion.


> Especially because the tests actually succeed when run interactively.

Just to clarify, when run via `ert-run-tests-interactively` I saw the same failures. By "interactive commands continued to work fine" I just meant calling `M-x magit-status RET` interactively.

---

This little corner case turned out to be pretty "interesting"!

tarsius:
> But what is supposed to have happened [to the previously current buffer] after [magit-status{,-internal}] are called?

Well, since those functions are supposed to make the status buffer current, the previously buffer is supposed to not be current anymore. Unless of course that's the same buffer. The previously current buffer might, or might not, be visible anymore, in another window or frame. That depends on what the user has picked for `magit-status-buffer-switch-function`. Are you over-thinking this or am I missing something?

npostavs:
> Are you over-thinking this or am I missing something?

Probably I'm over-thinking. But this whole topic has rather confused me.

npostavs:
> Now the tests pass in batch mode too.

However it rebreaks #2054, in other words only the last hunk seems significant; I can't tell how this is different from just reverting 3753d07c8b1655c1a165a104d25ff9e7f1fdc6e5.

npostavs:
> In particular, how come the tests were broken, while the interactive commands continued to work fine??

I think I figured this out: there is a somewhat surprising interaction between `save-current-buffer` and `switch-to-buffer` (or `pop-to-buffer` in our case). From `emacs -Q` execute the following in `*scratch*`:

```elisp
(progn
  (save-current-buffer (switch-to-buffer "*Messages*"))
  (current-buffer))
```
It returns `*scratch*`, but after executing it's actually the `*Messages*` buffer that is selected.

tarsius:
> I think I figured this out [...]

Interesting. And makes sense.

Anyway 01cf031f0ba54a3860571794f12195ddaf3f7323 should do the trick in our case.

PR #2061, opened 2015-07-12
Added magit reference card

mrkkrp:
Here is “Magit Reference Card”! It's open to corrections. I decided to highlight important parts of command names to make them stand out. PDF version is also included. I've not changed anything in makefile, it's up to you (if needed at all).

**TODO** List:

- [X] write/format first page;
- [x] write/format second page;
- [x] achieve consensus on content and look;
- [x] do a rebase to clear history.

### DIFF
--- a/Documentation/magit-refcard.tex
+++ b/Documentation/magit-refcard.tex
@@ -1,0 +1,188 @@
+% Reference Card for Magit
+% Copyright (C) 2008-2015  The Magit Project Contributors
+
+\documentclass[a4paper,10pt]{extarticle}
+
+\usepackage{textcomp}
+\usepackage{fullpage}
+\usepackage{boxedminipage}
+\usepackage{multicol}
+\pagestyle{empty}
+\usepackage[T1]{fontenc}
+\renewcommand{\baselinestretch}{1.3}
+
+\newcommand{\docheader}[1]{\Huge\centering{#1}\normalsize\bigskip}
+\newcommand{\boxedgroup}[1]{\par\noindent\centering{\textbf{\large#1}\medskip}}
+\newcommand{\helpkey}[2]{\par~~\noindent\textbf{#1}\hfill{#2}~~}
+\newcommand{\group}[1]{\bigskip\par\noindent\textbf{\large#1}\bigskip}
+\newcommand{\key}[2]{\par\noindent\textbf{#1}\hfill{#2}}
+\newcommand{\meta}[1]{\textlangle{#1}\textrangle}
+
+\begin{document}
+
+\docheader{Magit Reference Card}
+
+\begin{boxedminipage}[t]{11cm}
+
+  \medskip
+
+  \boxedgroup{Getting help in Emacs}
+
+  \helpkey{C-h k \meta{key}}{describe function bound to \meta{key}}
+  \helpkey{C-h b}{list the current key-bindings for the focus buffer}
+  \helpkey{C-h m}{describe mode}
+  \helpkey{C-h l}{show the keys you have pressed}
+  \helpkey{\meta{prefix} C-h}{show all key-bindings beginning with \meta{prefix}}
+
+  \medskip
+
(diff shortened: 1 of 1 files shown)


tarsius:
It's a start, but it's also a fairly random selection of commands. I also think bindings should be split into more groups, and that we should *describe* what keys do, not what commands they are bound to.

mrkkrp:
> we should describe what the keys do, not what commands they are bound to

@tarsius, this was my initial thought too, ref cards for Cider and Alchemist fooled me into thinking that names of functions suffice.

> but it's also a fairly random selection of commands. I also think bindings should be split into more groups

This is easy to improve that. I can use my own judgement to form the groups, but it would help if you could name them for me. Also, if we want to keep it on one page, we can't have *many* groups (4-5 maximum including popups), since popups take half of page anyway.

tarsius:
I think we need two pages - at least.

mrkkrp:
@tarsius, OK, let it be two pages. This can be printed on two sides of a sheet of paper (even manually if printer doesn't support that), but more than 2 pages is not a ref card, it's rather mini-manual :-)

tarsius:
> it would help if you could name them for me.

I think there should be "modify the diff/log shown in the current buffer"-section (but with a better buffer, and hinting at the fact that does commands not only work in diff/log buffer but also the status buffer). Also very important is the "apply variants" section. Generally speaking take inspiration from how things are structured in the manual, but depart from it where necessary.

mrkkrp:
Here is the first page, this has more sections and it's more structured. On the second page we can have descriptions of diff/log specific things (we can mention inlined diffs in status buffer too) and other operations. I don't think we should describe actions that are available via popups, because popups present them anyway.

----
You have to fix this section of manual: [5.3.1 Refreshing Diffs](http://magit.vc/manual/magit.html#Refreshing-diffs), `+` increases context for diff hunks, not decreases it.

----
[PDF link](https://github.com/magit/magit/blob/783f66b60b8285c2ee7f462abba4b593773de29b/Documentation/magit-refcard.pdf).

mrkkrp:
**TODO** List:

- [X] write/format first page;
- [ ] write/format second page;
- [ ] achieve consensus on content and look;
- [ ] do a rebase to clear history.

mrkkrp:
@tarsius, It seems like I've put entire manual into the ref card. Since popups remind user of most commands, the ref card doesn't get that big at all. Second page is a bit empty. What do you think about it?

mrkkrp:
@tarsius, So if it looks all right to you, I'm finishing with it.

mrkkrp:
@tarsius, Any decision/criticism on this?

tarsius:
Sorry, been a bit busy with other things.

The reference card is okay. I would have structured it differently (using the inspecting/manipulating/transfering distinction and grouping all apply variants). The descriptions of `n p N P` are wrong though, please fix that.

This document *is* helpful and I want to make it available to users soon. But I don't want to track the pdf to the source repository. Please add it to the webpage repository instead and update all the relevant Makefiles and also add links in the appropriate html files.

mrkkrp:
@tarsius, OK, I've amended that last commit so that:
- [X] description of <kbd>n</kbd>, <kbd>p</kbd>, <kbd>N</kbd>, and <kbd>P</kbd> key bindings is fixed (I hope);
- [X] PDF file is removed.

Now, let's make it clear, this should be all for this PR, now you want me to open a pull request for this repo: https://github.com/magit/magit.github.io -- this is webpage repo as I understand. There I should add PDF file, add links so users can see the refcard and also edit `Makefile`. Right?

tarsius:
Yes, exactly.

Add the file inside the `manual` directory and add an appropriate target to the Makefile in that directory. Then add links to the generated pdf from `index.html` and `manual/index.html`. Thanks.

tarsius:
Thanks @mrkkrp.

I have merged your pr against the webpage repo.

PR #2068, opened 2015-07-13
Adds recommended magit-status keybinding

slipset:
If `C-x C-g` is not already bound, we bind `magit-status` to it


npostavs:
I don't think the magit package should be taking over global keys, even if *at the time it's loaded* the key appears to be unbound.

tarsius:
I agree with @npostavs. It's not nice to take over global bindings. Some user would not like it if did so. And it's also the convention to not do it.

PR #2080, opened 2015-07-15
Push like Git would

tarsius:
The first commit replaces the old push variants with new ones which do what Git would do. This is not finished. It hasn't been tested and some commands don't do anything at all yet. But it gives an idea of what variants will be available once I have completed this.

The second commit adds stubs for the push variants which push to the planned publish remote. This requires work elsewhere too. I am adding this here so that we now have a list of all push variants that are currently in planning.


tarsius:
I have pushed a rebased version to `jb/git-push`. I am not going to merge this as-is, but there might be some useful bits in there.

Issue #2085, opened 2015-07-16
the new diff/rev view


pank:
Hi,

Thanks for putting all the work into Magit 2.1!  I really like the way rebase and patches are integrated. 

However, in Magit 2.1 the new diff view is pretty different from previous versions and from the "normal" diff view.  Here are the things that I still feel uncomfortable about since I installed Magit 2.1.

1. I don't understand the need for different from standard region behaviour.  ATM when I activate the region to kill something, I get weird red bars denoting my region (though only to the precision of the line above and below, not to the actual point).  It confuses me every time.  I looked for a `magit-revision-*` variable to disable this, but couldn't find one.
2. Why does Magit use different colors from normal diff?  The contrast on my system is bad (e.g. red on red and green on green, using the AFC theme).  I guess I could customize magit faces, but this sounds like a tedious exercise.
3. I think in the "old" Magit `RET` was bound to what now corresponds to `magit-diff-visit-file-worktree` (`C-RET`).  I think this is a more sensible default for `RET`, even if it cannot always get the line exactly correct.  It unfortunate when I browse the log file, click `RET` and get some old version of the file, as I might not notice and edit the wrong file.  If I explicitly want an old version of a file I'll happily go through the trouble of clicking a less intuitive key combination.  I can rebind this, but to me the default is wrong (you may of course disagree).

There are probably benefits of the above that are not obvious to me.  I'd be happy with just a variable enabling a more "conservative" take on the magit-rev.

Thanks,
Rasmus

npostavs:
> I don't understand the need for different from standard region behaviour. ATM when I activate the region to kill something, I get weird red bars denoting my region (though only to the precision of the line above and below, not to the actual point)

Because you can stage/unstage/discard the lines included in the region, but not parts of line.

pank:
On July 16, 2015 9:03:42 PM GMT+02:00, Noam Postavsky <notifications@github.com> wrote:
>> I don't understand the need for different from standard region
>behaviour. ATM when I activate the region to kill something, I get
>weird red bars denoting my region (though only to the precision of the
>line above and below, not to the actual point)
>
>Because you can stage/unstage/discard the lines included in the region,
>but not parts of line.

Why is this relevant for, say an unpushed commit?

I enjoy the (un)staging keys and using (plain) region for staging things before commit. But that seems irrelevant in this context.

npostavs:
You can apply/reverse lines of a commit.

tarsius:
(3)

[...]

Issue #2063, opened 2015-07-12
Staging / unstaging by selected region is confusing


kzar:
Hi again,

Apart from #2062 my only issue with "the big upgrade" to 2.1 has been that I now find it very confusing to partially stage / unstage a hunk by selecting a region.

I _think_ it works basically as before except now I can't see the region that I've selected. If so perhaps the region the user has selected should be highlighted in some visible way?

I might be wrong there, not sure but as this is something quite common it should be as intuitive as possible. Someone who doesn't understand the new model for how these hunks and cursors and regions work (me!) should still be able to highlight and stage a region.

Cheers, Dave.

tarsius:
> If so perhaps the region the user has selected should be highlighted in some visible way?

You are likely using a theme which does only set the foreground color of hunk lines. But the background should also be set.

tarsius:
The region isn't shown because what is being selected, using mark and point, is actually larger than the region. Say mark is in the middle of one line and ends at the very beginning of another line. In that case all changes on both of these lines, and of course all in between, are part of the "selected area", and `s` would stage all of that.

So that area, which is larger than the region, is being visualized instead of the region. This is done by removing the added/removed/contact background color of lines outside that area (which of course doesn't change anything if these lines don't have such a background color in the first place). Additionally there are thin lines shown right above and right below the selected area (but that doesn't work in a terminal).

kzar:
Oh I see, hitting `C-SPACE` moves one horizontal bar and then I can move the cursor down to move the other horizontal bar before hitting `s` to stage. It's actually pretty easy in fairness, just quite different from what I was expected and hence my confusion!

Perhaps _also_ highlighting the region between the two horizontal bars would help to illustrate what is happening and avoid the confusion?

tarsius:
> Perhaps also highlighting the region between the two horizontal bars would help to illustrate what is happening and avoid the confusion?

That would hide hunk refinements, so no.

PR #2091, opened 2015-07-17
magit-keep-region-overlay: new option

tarsius:
As explained in the doc-string of that option, I *strongly* urge user
to *not* change the value of this option.  Re #2085, #2063, tweets, and
https://www.reddit.com/r/emacs/comments/3bt2nh/magit_v210_released/cstul4y.

### DIFF
--- a/Documentation/RelNotes/2.3.0.txt
+++ b/Documentation/RelNotes/2.3.0.txt
@@ -86,5 +86,10 @@ THIS IS NOT COMPLETE.
   the same index as the previous hunk, but that often was very far
   away from where point was before the refresh.  #2227
 
+* New option `magit-keep-region-overlay' allows displaying the region
+  overlay together with the visualization of a valid Magit selection.
+  See the doc-string for why you don't want to do that, but if you
+  disagree then you can now have it your way.  #2091
+
 Authors
 -------
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -168,6 +168,53 @@ has to confirm each save."
                  (const :tag "Ask" t)
                  (const :tag "Save without asking" dontask)))
 
+(defcustom magit-keep-region-overlay nil
+  "Whether to keep the region overlay when there is a valid selection.
+
+By default Magit removes the regular region overlay if, and only
+if, that region constitutes a valid selection as understood by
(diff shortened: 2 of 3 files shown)


tarsius:
@kyleam so you think I should add this? What about you @npostavs?

kyleam:
No, I don't, but you have a better feel for the amount of pressure for
it.  Perhaps we could let things settle a bit more and revisit it if it
keeps coming up.

npostavs:
I vote yes for this change.

> But please note that when the region
does *not* constitute a valid selection, then the region is
*always* visualized as usual, and that it is usually under such
circumstances that you want to use a non-magit command to act on
the region.

I find I pretty often want to copy just some of the text from a single hunk, e.g. a function name. It's pretty rare (probably never) that I want to copy a region that crosses hunk boundaries.

> setting this option to t also
causes all face properties, except for `:foreground', to be
ignored for the faces used to highlight headings of selected
sections.

Does this affect anything when using the default faces? I'm trying this out now, and I can't see any obvious difference.

kyleam:
> I find I pretty often want to copy just some of the text from a single
> hunk, e.g. a function name. It's pretty rare (probably never) that I
> want to copy a region that crosses hunk boundaries.

That makes sense.  I often copy text from hunks too, but I guess the
lack of region highlighting doesn't bother me (perhaps I got used to C-w
and M-w operating on an unhighlighted region because for a while I had
transient-mark-mode disabled, preferring instead to C-SPC C-SPC).

> Does this affect anything when using the default faces? I'm trying
> this out now, and I can't see any obvious difference.

Hmm, I can see the shade of gray change compared to the diff context and
see gray as opposed for green/red for insertions/deletions.

npostavs:
> Hmm, I can see the shade of gray change compared to the diff context and see gray as opposed for green/red for insertions/deletions.

I guess this is just too subtle for me. I would probably need pictures with circles around and/or arrows pointing to the differences and then *maybe* I'd get it.

Bronsa:
I would love to see this merged, any update?

tarsius:
Well, users have stopped screaming for this, but since it is cheap to offer this option I am now doing it anyway. But I have also added a note saying that it might be removed again if it complicates other things.

PR #2092, opened 2015-07-17
docs: fix broken link to bug report about emacs hanging due to tramp

vyp:
---

I'm getting a broken link [here](https://debbugs.gnu.org/cgi/bugreport.cgi?bug~20015):

>  An error occurred. Error was: No bug number 

However, I'm not 100% I found the correct replacement link because it [mentions emacs 25.0.50](https://lists.gnu.org/archive/html/bug-gnu-emacs/2015-03/msg00203.html), whereas the faq mentions emacs 24.5. But it *seems* to be the correct bug report?

(Original link has a robots.txt apparently, so it seems as though it was not archived anywhere.)

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -4491,9 +4491,9 @@ also affects the diffs displayed inside Magit.
 ** Emacs 24.5 hangs when loading Magit
 
 This is actually triggered by loading Tramp.  See
-https://debbugs.gnu.org/cgi/bugreport.cgi?bug~20015 for details. You
-can work around the problem by setting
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks.  Fixed in d5d23470a4b6277bcfdc54e6c0dc7c398fb26946.

vyp:
Thanks!

npostavs:
> However, I'm not 100% I found the correct replacement link because it mentions emacs 25.0.50

FYI: it has since been fixed in 25.0.50, so it only applies to 24.5

PR #2098, opened 2015-07-19
Hunk file posititioning

kyleam:
re: the second commit

When point is on the file or hunk header, I think it is more useful to
go to the first changed line rather than the first context line.
Thoughts?


npostavs:
> When point is on the file or hunk header, I think it is more useful to
> go to the first changed line rather than the first context line.

Makes sense to me.

tarsius:
Thanks

PR #2100, opened 2015-07-20
magit-git-command displaying wrong process buffer

kyleam:
I'm not very certain of these changes, so please consider this a bug
report rather than a PR.

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -233,7 +233,7 @@ variable `magit-process-buffer-name-format'."
 Process output goes into a new section in a buffer specified by
 variable `magit-process-buffer-name-format'."
   (cl-destructuring-bind (process-buf . section)
-      (magit-process-setup program args)
+      (magit-process-setup program args default-directory)
     (magit-process-finish
      (let ((inhibit-read-only t))
        (apply #'process-file program nil process-buf nil args))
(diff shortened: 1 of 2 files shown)


tarsius:
The idea was to use one process buffer per repository, so we should not use `default-directory`, but always `(magit-toplevel default-directory)` (and I don't think it should be the callers who have to take care of this. Except of course when we are not inside a repository, if I remember correctly, we then use the global (usually startup) value of `default-directory`.

kyleam:
> The idea was to use one process buffer per repository, so we should
> not use `default-directory`, but always `(magit-toplevel
> default-directory)`

OK.  I thought magit-git-command might be an exception because, if the
user is calling some git command from a subdirectory, the output may
depend on that directory, and it avoids running a command with a
different default-directory than the process buffer's default-directory.

I've pushed a second branch (km/command-subdir-2) that instead changes
magit-mode-get-buffer to compare its topdir argument with magit-toplevel
of the searched buffers.  This seems to work, while keeping everything
in the same process buffer, but perhaps it breaks things elsewhere.

tarsius:
Maybe a compromise would be in order: use the top-level process buffer, but prefix the command in the header with the relative path `  0 lisp/$ git ...`.

kyleam:
I like that idea.  Added to the km/command-subdir-2 branch.  Due to the face differences, I think it looks fine even without the "$".

tarsius:
I have pushed minor terminology fixes to the v2 branch.

tarsius:
I have installed v2 with some cosmetic changes. And one big change: the second commit was replaced with 4de98951c680b17d3806add121de2b0cfb8cf4c1.

tarsius:
I had to revert 4de98951c680b17d3806add121de2b0cfb8cf4c1, but will try again later.

PR #2109, opened 2015-07-23
Protect against empty git-blame chunks

kanru:
In some codebase the Author or other field might be missing. Which leads to
error in process filter: magit-blame-format-heading: Wrong type argument: stringp, nil


tarsius:
Thanks. I have cherry-picked this.

PR #2110, opened 2015-07-23
magit-define-popup-key: defer if var not yet bound

npostavs:
This allows adding entries to the ediff popup even if the popup variable
hasn't been defined yet.

See https://github.com/magit/magit/issues/1743#issuecomment-120789326 and https://github.com/magit/magit/issues/1743#issuecomment-123402314


tarsius:
Thanks. I have cherry-picked this.

PR #2113, opened 2015-07-24
Make magit-clone expand relative file names.

another-david:
Invoking magit-clone from /home/user/subdir with a request to clone a
repo into ~/subdir/repo results in a repo in
/home/user/subdir/~/subdir/repo.

Expand the directory name to an absolute path before passing to git-clone.

### DIFF
--- a/lisp/magit-remote.el
+++ b/lisp/magit-remote.el
@@ -38,10 +38,11 @@
   "Clone the REPOSITORY to DIRECTORY."
   (interactive
    (let  ((url (magit-read-string "Clone repository")))
-     (list url (read-directory-name
-                "Clone to: " nil nil nil
-                (and (string-match "\\([^./]+\\)\\(\\.git\\)?$" url)
-                     (match-string 1 url))))))
+     (list url (expand-file-name
+                (read-directory-name
+                 "Clone to: " nil nil nil
+                 (and (string-match "\\([^./]+\\)\\(\\.git\\)?$" url)
+                      (match-string 1 url)))))))
   (make-directory directory t)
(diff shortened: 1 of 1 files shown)


kyleam:
Thanks.

tarsius:
@another-david @invalid-email-address David Edmonds <david@localhost.localdomain>:

Your email is indeed invalid. Could you please let me know your real address so that I can fix this in our `.mailmap`. We need your proper email address in case we ever attempt to become a gnu project (which we might or might not do).

tarsius:
@kyleam @npostavs Since the release of v2.1 this has been the third contributor who has accidentally or intentionally not set `user.name` and/of `user.email`. I want to keep `AUTHORS.md` neat, so for first time contributors we now have to check whether the author information is correct.

`git-commit` already tries to make users write good commit messages - maybe we should extend that by warning users if they have not set these essential variables.

kyleam:
> so for first time contributors we now have to check whether the author information is correct

Sorry about that.   I'll make sure to do this in the future.

tarsius:
No problem, I think it was me who did the last two merges.

It's a bit ironic that I have a single unpushed commit sitting in my local `master` which is titled "update AUTHORS.md and .mailmap" :-)

tarsius:
I have just removed this commit from the public `master`. The whole point of the @invalid-email-address account is to not provide a proper email address - so there is reason to believe that we won't be given the correct address. Public history shouldn't be modified but when it has to be modified, then that should happen as quickly as possible.

If David Edmonds decides his email address then we can push it again (or a commit with fixed metadata). Otherwise I can fix the bug myself without looking at the provided patch.

PR #2116, opened 2015-07-25
magit-mode-map: stubs for section-only commands

### LINKED ISSUE #2106: Some suggestions to make key bindings more discoverable
Original title was: **support reseting individual hunks within unstage changes section**, it was changed because the focus of this issue has quickly shifted elsewhere. -- @tarsius

----

Recently I was looking at a diff in a magit-status buffer showing an unstaged change that was nothing more than corrupted indentation.  It would have be very convenient to hit 'x' on that hunk and have it vanish.

npostavs:
in order to give nicer error messages and make such bindings more
discoverable.

resolves #2106

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -181,6 +181,24 @@ has to confirm each save."
 
 ;;; Magit Mode
 
+(defmacro magit-define-stub-at-point-function (op)
+  `(defun ,(intern (format "magit-%s-thing-at-point" op)) ()
+     "This is a placeholder command.
+
+Section-specific keymaps will bind a different command instead,
+where applicable."
+     (interactive)
+     (user-error "There is nothing to %s at point." ,(symbol-name op))))
+
+(magit-define-stub-at-point-function apply)
+(magit-define-stub-at-point-function add-log-of)
+(magit-define-stub-at-point-function delete)
+(magit-define-stub-at-point-function untrack)
+(magit-define-stub-at-point-function rename)
+(magit-define-stub-at-point-function visit)
+(magit-define-stub-at-point-function visit-worktree-of)
+(magit-define-stub-at-point-function reverse)
+
 (defvar magit-mode-map
   (let ((map (make-keymap)))
     (suppress-keymap map t)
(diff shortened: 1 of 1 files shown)


tarsius:
I think we shouldn't do this

Some time ago many commands were defined using "action commands". Instead of `magit-section-case` the used a similar macro `magit-section-action`, which additionally defined a hook. I don't think any user ever used those hooks (Magit itself abused them in a few places).

These hooks emulated something like text property keymaps, but were much less flexible, non-standard, mostly undocumented, implemented using fragile code, in short quite awful and unnecessary.

I removed them and instead started using a mixture of `magit-section-case` and text property keymaps. I think that's a huge win. The only downside is that those commands don't automatically appear in the output of `C-h m`. But we fixed that by listing them explicitly.

We should improve those explicit listings. E.g. make sure all keys are listed and maybe change the format.

special "define magit command" macro and consisted mostly of `magit-section-action` which was very similar to `magit-section-case` but the 

. They defined commands which mostly consisted of something like `magit-section-case` but also defined a hook that allowed users to add specialized for certain sections. I think nobody ever used that.

tarsius:
> I think we shouldn't do this

Or maybe we should. I am pretty sure though that I should go to bed now...

npostavs:
What if we bound the main command to `magit-VERB-thing`, and then the section keymaps would just have `<remap> <magit-VERB-thing>` to `magit-SPECIFIC-VERB`?

tarsius:
I have just pushed 79cd46938ddf2030bde9b313c4ae1c8c21493ddf which is inspired by this to some extend.

It doesn't use a definer macro. Such macros, which do almost nothing but generate some strings, are usually a bad idea, mostly because the strings they generate are not as good as what one would get if each string is worded individually.

`a` and `v` are bound to existing commands. `k`, `R`, `RET`, and `C-return` to new stub commands. `C` and `K` are not bound to any new stub. Of course we could bind those too, but at some point users have to start reading documentation instead of [retracted]. Also `magit-add-log-of-thing-at-point` stinks anyway. (We should instead allow adding changelog/commit-message stubs for all modified files/definitions at once, using something like `git-commit-insert-message-stub` (which could be added to a hook if desired.)

So I think that all we should do, beyond what I have already merged, is to make sure all mode doc-strings mention all relevant section-specific bindings (#2120).

---

I know, I suggested this - and now that you have implemented it, I am rejecting it. Sorry about that.

npostavs:
> I know, I suggested this - and now that you have implemented it, I am rejecting it. Sorry about that.

No worries, I'm not especially attached to this code.

PR #2117, opened 2015-07-26
Fix one sentence wording as well as two incorrect key bindings

emacs18:
Changed "off" to of".


tarsius:
Please amend that additional correction and also use your real name as `user.name`. Thanks.

emacs18:
I added my name in github "settings" page.
I hope that does the trick.

tarsius:
Not really no, at least not if you do it after creating the commit. From your changes I have created one new commit attributed to you.

PR #2120, opened 2015-07-27
Improve dispatch popup and major-mode doc-strings

tarsius:
Just a suggestion. What do you think?

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -355,16 +355,26 @@ deep."
 This mode is documented in info node `(magit)Status'.
 
 \\<magit-status-mode-map>\
-Type \\[magit-refresh] to refresh the current buffer.
-Type \\[magit-dispatch-popup] to see available action popups.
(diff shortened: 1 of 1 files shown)


npostavs:
I like the reorganization of the dispatch popup, but I'm not really sold on turning `describe-mode` into `describe-keybindings`.

tarsius:
> I like the reorganization of the dispatch popup, but I'm not really sold on turning `describe-mode` into `describe-keybindings`.

Well, we already do that. This just adds the missing commands from section keymaps and presents the bindings that are explicitly listed in a different format.

tarsius:
I just pushed to commits which modify `magit-dispatch-popup` and add `magit-file-buffer-{mode,popup}`. Rebasing this pr...

tarsius:
a80751189a93b891362810eab8bac707745ec7f8 has to do it.

Issue #1800, opened 2015-04-05
Support sending and receiving patches and working with pull requests


tarsius:
Instead of implementing basic support for `git send-email` and the fetching of pull requests, we should implement complete workflows for interacting with other developers per mail, github, bitbucket, gitlab, and probable some others.

Getting this right will be a lot of work and will likely end up as the major new feature of some later release. I do not intend to accept patches that implement just parts of this. `2.1.0` will provide basic support for creating patches using `git format-patch` and sending pull requests per mail using `git request-pull`. These features already exist on the `next` branch, and are here to stay until they are replaced with something better. However I do not intend to add support for `git send-email` anytime soon, especially not one that doesn't offer much over just typing that command into a shell.

andschwa:
:+1:

tarsius:
Tabula rasa

Closing, to avoid pressuring myself into implementing this before the prerequisites are fully ready, just because there's an open issue for this but not the prerequisites.

Also see the [`3.1.0`](https://github.com/magit/magit/milestones) milestone (click on "more").

zabbal:
Note: while integration with mu4e is certainly nice to have it's not necessary for the first step. For example right now git send-email works for me via msmtp so just a basic wrapper on top of git-send-email which takes care of setting all relevant options and getting cover-letter from emacs would do.

tarsius:
> while integration with mu4e is certainly nice to have it's not necessary for the first step.

The problem with intermediate implementations is that users get used to them. Later when I remove something which is no longer necessary and/or gets in the way of the "ultimate version", some users would urge us to restore the dropped feature/workflow because "that was the best part of it".

zabbal:
In general you're absolutely right. But in this particular case I doubt we can convince every single magit user to switch to use mu4e for emails (although it would be cool :) So I think even if you won't add plain git-send-email support and go straight for mu4e, users would keep demanding it because it's "least common denominator" regardless of personal choice for MUA.

tarsius:
>  plain git-send-email support

I probably won't object to merging a minimal viable version if someone else implemented it.

> mu4e ... "least common denominator"

Did I mention just mu4e? I intend to support Gnus too.

zabbal:
mu4e, gnus, whatever - it doesn't matter how many clients you support directly, there'll always be another guy who only uses mutt or or thunderbird or, gods forbid, outlook. Hence inevitably there will be requests to support sending patches indirectly via git-send-email instead of interacting with some email client directly.

[...]

PR #2125, opened 2015-07-30
magit-format-patch: support <since> commit

kyleam:
Thoughts?

### DIFF
--- a/lisp/magit-remote.el
+++ b/lisp/magit-remote.el
@@ -365,14 +365,23 @@ branch as default."
 (defun magit-format-patch (range args)
   "Create patches for the commits in RANGE."
   (interactive
-   (list (-if-let (revs (magit-region-values 'commit))
-             (concat (car (last revs)) "^.." (car revs))
-           (let ((range (magit-read-range-or-commit "Format range or commit")))
-             (if (string-match-p "\\.\\." range)
-                 range
-               (format "%s~..%s" range range))))
+   (list (magit-read-patch-range "Format range or commit")
          (magit-patch-arguments)))
   (magit-run-git-no-revert "format-patch" range args))
+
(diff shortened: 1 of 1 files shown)


tarsius:
Lets delay this until we start working on `2.3.0` (branching and pushing). https://github.com/magit/magit/issues/1800 is the umbrella issue for this sort of thing.

I had a short look and think that this change, by itself, does not provide much. We don't have to support <since>, just because git does.

>Do not convert a single commit argument read by magit-format-patch to '<commit>^..<commit>' because this prevents the user from giving a single commit argument that has the same meaning as it does for 'git format-patch', where it is interpreted as '<commit>..HEAD'.

On the other hand what you propose makes it harder to create a patch for a single commit. We can get `<commit>..HEAD` quite easily: set mark on the first commit in "Unpushed commits", move down, press `W p`.

kyleam:
Thanks for your comments.

> Lets delay this until we start working on `2.3.0` (branching and
> pushing).

Sure.

> We don't have to support <since>, just because git does.

Yes, but I'm in favor of interpreting user-inputted ranges in the same
way as the underlying git command would, rather than modifying them
before passing them down.

> On the other hand what you propose makes it harder to create a patch
> for a single commit.

With point at the commit, commit^..commit would be the default prompt,
so it is only more difficult if the user wants to manually input a
commit to get commit^..commit (which I suspect for many users is not the
expectation, given the behavior of git format-patch).

tarsius:
Well, I am not completely against this. (Really I just haven't had a close look yet.) So you probably want to keep the branch around. Unless you have something better cooking of course.

kyleam:
Yep, I'll keep it around (plus it's in pull/origin), but, since this PR is mentioned in #1800 and can be revisited with that, I figured there is no advantage to leaving this open.

PR #2146, opened 2015-08-06
magit-{log,diff}-use-sticky-arguments: new options

kyleam:


tarsius:
I like it but think the doc-string have to be a lot more explicit. "Whether to re-use the state from the previous diff popup call." is fine as the summary but it's also not really true. It's the same call in the *same repository* and the call might be to the diff-refresh popup instead of the diff popup.

It's probably necessary to explain that the arguments are being saved in a buffer-local value.

Maybe there should be a third value `use-the-current-arguments-but-only-when-invoked-from-the-diff-buffer-itself`.

tarsius:
Somewhat related: `magit-{log,diff}-arguments` rely on the value of `magit-current-popup` and the `major-mode` to detect how it should get the relevant arguments currently in effect. That's not good enough. When e.g. `magit-log` is invoked directly then it should still the arguments from the log buffer.

Do you have any suggestions how to do that. I suppose we have to `(memq this-command the-log-commands)`, but how do we get the `the-log-commands`? A list which has to be updated everytime a new command is added? Or something like `(put 'magit-log 'magit-real-log t)`?

kyleam:
Thanks, I've pushed an update on top of your recent changes.

> call might be to the diff-refresh popup instead of the diff popup

Won't the diff-refresh popup work the same regardless of the value for
this option?

tarsius:
> Won't the diff-refresh popup work the same regardless of the value for this option?

Yes, but that's not what I meant. 

If you
1. `magit-log-current`
2. `magit-log-refresh`
3. `magit-log-other`

then the option you are proposing allows the user to choose whether (3) uses the arguments set in (2) or the default value. You are claiming the choice is between the arguments used in (1) and the default value.

tarsius:
Are you proposing this because you personally don't want "sticky" arguments or because you are anticipating that we would sooner or later have to add this option because some users are not going to like the new behavior?

If it's the latter, then I would rather delay this until some users actually asks for such an option, or if that doesn't happen even until we replace `magit-popup` with something better.

kyleam:
> Yes, but that's not what I meant. [...]

Yep, sorry.  Looking again, it is clear what you meant.

> Are you proposing this because you personally don't want "sticky"
> arguments or because you are anticipating that we would sooner or
> later have to add this option because some users are not going to like
> the new behavior?

I don't prefer sticky arguments (at least I haven't come around to them
yet).  I think the main reason is that I rarely kill Magit buffers, so
when I call a log/diff command after some time, I forget that I had
changed the default arguments in the last command.  (I don't really look
at the popups if I know what command I want, so I won't catch the visual
indication of the changed arguments.)  Depending on the argument, it may
be more or less clear from the resulting buffer that something other
than what I intended was used.

It is simpler for me to just know that a new command call will always
behave the same, and I'm happy to explicitly use a refresh popup to
change the previous call.

> If it's the latter, then I would rather delay this until some users
> actually asks for such an option, or if that doesn't happen even until
> we replace `magit-popup` with something better.

Even though it is not the latter, I think waiting to see how other users
respond makes sense.

uu1101:
Another option (I think something along the lines has been discussed in other issues) is to have some kind of popup-history.  In this way one could begin having the default popup, and pressing <kbd>M-p</kbd> would restore the last-used options.

tarsius:
Yes. I haven't forgotten about that.

alezost:
>Are you proposing this because you personally don't want "sticky" arguments or because you are anticipating that we would sooner or later have to add this option because some users are not going to like the new behavior?

>If it's the latter, then I would rather delay this until some users actually asks for such an option, or if that doesn't happen, even until we replace magit-popup with something better.

I am also the user who doesn't like these sticky arguments (sorry for bothering).

hooptie45:
> Another option (I think something along the lines has been discussed in other issues) is to have some kind of popup-history. In this way one could begin having the default popup, and pressing M-p would restore the last-used popup with whatever options had been specified.

:+1:  That would be a great feature.

tarsius:
Yes, lets do that and skip this intermediate step.

tarsius:
> Yes, lets do that and skip this intermediate step.

Actually let's not skip it, because it's not really an intermediate step. Even with argument history some value still has to be the default, and these options would control what that would be. @kyleam could you please revive this ;-)

kyleam:
Sure.  I'll open a new PR.

PR #2152, opened 2015-08-06
Always use RFC2822 format for GIT_COMMITTER_DATE

thomasf:
Fixes issue mentioned here: https://github.com/magit/magit/issues/1645#issuecomment-117636222


tarsius:
Applied. Thanks.

PR #2155, opened 2015-08-07
Double-check that directory is a directory

quodlibetor:
For some reason, magit blame was failing with exactly one file that I
have, saying that I was not visiting a tracked file.

After some debugging, it looked like magit was setting
`default-directory' to `/path/to/file.py/' instead of `/path/to/'.

Something might be wrong with the way that emacs keeps track of whether
or not a node is a file or a directory on OSX.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -294,7 +294,8 @@ call function WASHER with no argument."
                (file-name-as-directory (if file
                                            (expand-file-name file)
                                          default-directory)))))
-       (while (not (file-accessible-directory-p default-directory))
+       (while (not (and (file-accessible-directory-p default-directory)
+                        (file-directory-p default-directory)))
          (when (string-equal default-directory "/")
            (throw 'unsafe-default-dir nil))
          (setq default-directory


quodlibetor:
I'm not sure if this is the best way to handle this problem, and I can't reproduce it. Let me know if there's anything I can do to help debug.

tarsius:
```lisp
(file-accessible-directory-p "/path/to/existing/file.txt")
(file-accessible-directory-p "/path/to/existing/file.txt/")
```

Both of these forms return nil, so your addition is redundant.

After some debugging, it looked like magit was setting `default-directory` to `/path/to/file.py/` instead of `/path/to/`.

Can you be more specific? Was the buffer local value of `default-directory` of the buffer visiting `/path/to/file.py` set to `/path/to/file.py/`? If so, then that's a bug in some other code, because Magit never sets the `default-directory` in file visiting buffers. Or did you come to the conclusion that Magit somewhere let-bound `default-directory` to the bogus value?

quodlibetor:
The code that I modified (in `magit--with-safe-default-directory`) sets `default-directory` multiple times, using `setq`.

Here's my debug messages from while I'm trying to debug this, `new default directory:` happens at line 296 in `magit-git` and `default-directory` happens at line 299.

```
file is: nil
new default directory: /Users/bwm/projects/config/remote_checks/discovery/src/create_discovery_service_checks.py/
default-directory /Users/bwm/projects/config/remote_checks/discovery/src/create_discovery_service_checks.py/
new default directory: /Users/bwm/projects/config/remote_checks/discovery/src/create_discovery_service_checks.py/
file is: /Users/bwm/projects/config/remote_checks/discovery/src/create_discovery_service_checks.py, toplevel is nil
new default directory: /Users/bwm/projects/config/remote_checks/discovery/src/create_discovery_service_checks.py/
default-directory /Users/bwm/projects/config/remote_checks/discovery/src/create_discovery_service_checks.py/
new default directory: /Users/bwm/projects/config/remote_checks/discovery/src/create_discovery_service_checks.py/
user-error: Buffer isn't visiting a tracked file
```

I agree that in general the two functions (`file-accessible-directory-p` and `file-directory-p`) should (and normally do) return `nil` for files-that-are-not-directories.

I should point out that I use `magit-blame` all the time, and this is the first time that this particular problem has ever come up.

I'm working off magit-master 46e2af5 and emacs compiled version `GNU Emacs 25.0.50.1 (x86_64-apple-darwin13.4.0, NS appkit-1265.21 Version 10.9.5 (Build 13F34)) of 2015-03-24 on muon.local`.

tarsius:
I don't really know what I could do here. I don't want to merge this because I still believe it doesn't make a difference, and even if it does make a difference it would still be a kludge for an issue  whose cause is unknown.

All I can do is suggest that if you run into this again, then test the state outside of Magit first, e.g. do the values of `default-directory` and `buffer-file-name` make sense? If you can establish that these values are wrong even before using Magit, then investigate what else is messing it up.

It things only start going wrong after Magit touched the file-visiting buffer, then you have to figure out where it goes wrong. The debug output you provided above did not make much sense to me, there appear more instances of `message` to be involved than the ones you mentioned. If you can reproduce the issue then please post the debug output as well as the diff which shows what debug statements you added. `magit--with-safe-default-directory` is probably not the right place for the debug statements, it's more interesting what `default-directory` is before we try using that invalid value, not what happens while we try using the invalid value.

PR #2156, opened 2015-08-07
depend on async to guarantee untainted builds

tarsius:
Usually building a new version of Magit does not require a clean environment (one in which no older version of Magit has been loaded yet). But when updating from `v1.*` to `v2.*` then it is required. Unfortunately many users who get Magit from Melpa never saw the update instructions, which explain that they have to do the uninstall-restart-reinstall dance.

That's not their fault because when they updated to `>=20150701` there was nothing that indicated that this would be different from previous updates. Likewise it's not the fault of the Melpa maintainers, because they are not responsible for `package.el` itself. It's nobodies fault, but we still have to fix it - I've had it with those "impossible bugs" where the user and the code don't agree about what can possibly happen.

So I have added `magit` to `async-bytecomp-allowed-packages` in https://github.com/jwiegley/emacs-async/pull/50. Once that has been merged and hit Melpa, I will update the `async` snapshot version `magit` depends on, and then merge this.

Before the Magit `2.2.0` release we have to ask @jwiegley to create a `async` release because that is the only way for this to also take effect for users of Melpa-Stable. Then after the release we have to set back the `async` version we depend on to a snapshot version, because not all Melpa (unstable) users will have updated between now and then. Sight.

Also see http://emacs.stackexchange.com/questions/13574 where Stefan acknowledges that the issue is real and cannot always be blamed on bad development practices.

### DIFF
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -11,7 +11,7 @@
 ;;	Marius Vollmer <marius.vollmer@gmail.com>
 ;; Maintainer: Jonas Bernoulli <jonas@bernoul.li>
 
-;; Package-Requires: ((emacs "24.4") (dash "2.10.0") (with-editor "2.1.0"))
+;; Package-Requires: ((emacs "24.4") (dash "2.11.0") (with-editor "20150808"))
 ;; Keywords: git tools vc
 ;; Homepage: https://github.com/magit/magit
 
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -47,12 +47,6 @@
 (defvar magit-blame-mode)
 (defvar git-rebase-line)
 
-;; Work around the dreaded Elpa issue.
-(cl-eval-when (compile)
-  (cl-defstruct magit-section
-    type value start content end hidden washer refined
-    source diff-header process parent children))
-
 (require 'diff-mode)
 (require 'smerge-mode)
 
--- a/lisp/magit-popup.el
+++ b/lisp/magit-popup.el
@@ -12,7 +12,7 @@
 ;; Author: Jonas Bernoulli <jonas@bernoul.li>
 ;; Maintainer: Jonas Bernoulli <jonas@bernoul.li>
 
-;; Package-Requires: ((emacs "24.4") (dash "2.10.0"))
+;; Package-Requires: ((emacs "24.4") (async "20150807") (dash "2.11.0"))
 ;; Keywords: bindings
 ;; Homepage: https://github.com/magit/magit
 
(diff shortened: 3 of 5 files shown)


jwiegley:
@tarsius I'm happy to make a release for you.  Is everything in that you want to be there?

tarsius:
@jwiegley As far as the changes to `async` are concerned, yes I think that's it. I still have to update the dependencies of `magit` a bit, but you can tag a release now.

@kyleam @npostavs Please don't push anything to `master` today. That way I can bump all "versions" the various packages depend on to "20150808" and be sure that this triggers a rebuild off everything.

kyleam:
> Please don't push anything to master today.

ok

tarsius:
@jwiegley Could you please create that tag now, so that I can release Magit v2.2.0. Also mentioning @thierryvolpiatto in case you have permission to create tags and John isn't around today.

Thanks.

thierryvolpiatto:

> @jwiegley Could you please create that tag now, so that I can release
> Magit v2.2.0. Also mentioning @thierryvolpiatto in case you have
> permission to create tags and John isn't around today.

Done, we are now at 1.4.

-- 
Thierry
Get my Gnupg key:

tarsius:
Thanks!

syl20bnr:
Interesting thread, I understand now why in Spacemacs we  empirically ended up with an update process that first delete packages and asks to restart. This was the only way I found to get stable updates.

syl20bnr:
@tarsius FYI I will set `async-bytecomp-allowed-packages` to nil in Spacemacs so be aware of this if a user of Spacemacs reports a bug.

In the releas note it is metioned that I should set `async-bytecomp-allowed-packages` to nil but the question is where exactly ? I would like to completely remove this package from being operational in Sapcemacs, we don't need this hack ;-(

Thank you.

thierryvolpiatto:
Sylvain Benner <notifications@github.com> writes:

> @tarsius FYI I will set async-bytecomp-allowed-packages to nil in
> Spacemacs so be aware of this if a user of Spacemacs reports a bug.
>
> In the releas note it is metioned that I should set
> async-bytecomp-allowed-packages to nil but the question is where
> exactly ? I would like to completely remove this package from being
> operational in Sapcemacs, we don't need this hack ;-(

If you want to disable async compilation, use 
(async-bytecomp-package-mode -1) in your config after loading package
using it (AFAIK for the moment helm and magit).
Or at end of your config, even better.
So when the mode is disabled you don't have to worry about
async-bytecomp-allowed-packages value.

-- 
Thierry
Get my Gnupg key:

tarsius:
> but the question is where exactly ?

Before `with-editor` and `magit-popup` are loaded. Loading any one of these two libraries does this:

```lisp
(and (require 'async-bytecomp nil t)
     (memq 'magit (bound-and-true-p async-bytecomp-allowed-packages))
     (fboundp 'async-bytecomp-package-mode)
     (async-bytecomp-package-mode 1))
```

Or just turn the mode of again like @thierryvolpiatto suggested.

PR #2157, opened 2015-08-09
Allow switching to toplevel directory from a subdirectory of repo.

thierryvolpiatto:
Hi, 

with 1.4.2 I was able to open magit-status from helm, now it is possible only from toplevel directory.
Now I am prompted for creation of a new repo:

.../github/magit/ is a repository.  Create another in .../github/magit/lisp/? (y or n)

I added a new var to not be prompted when toplevel dir exists, otherwise if it doesn't exists
you are anyway prompted.

Perhaps you will want to change the var name which is not the best, I hope also my changes don't modify the original behavior too much.

Thanks.

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -142,7 +142,9 @@ The following `format'-like specs are supported:
   :type 'string)
 
 (defcustom magit-status-prompt-for-init nil
-  "Prompt for initializing a git repo when current directory is not a git repo."
+  "Prompt for initializing a git repo when current directory is not a git repo.
+When top level directory is not found prompt for creation even if
+`magit-status-prompt-for-init' is nil."
   :package-version '(magit . "2.1.0")
   :group 'magit-status
   :type 'boolean)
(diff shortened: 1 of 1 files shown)


tarsius:
I don't know how helm calls `magit-status` but unless it does so using `call-interactively` then it should be using `magit-status-internal` instead (which is a misnomer, it should be called `magit-status-non-interactive` or something like that).

Have you tried using that function?

thierryvolpiatto:

> I don't know how helm calls magit-status but unless it does so using
> call-interactively then it should be using magit-status-internal
> instead (which is a misnomer, it should be called
> magit-status-non-interactive or something like that).
>
> Have you tried using that function?

No, will try soon, thanks.

-- 
Thierry
Get my Gnupg key:

thierryvolpiatto:
Yes, `magit-status-internal` is working fine.
It is not autoloaded though.
So perhaps you can close this PR.

Thanks.

tarsius:
I've added the autoload.

thierryvolpiatto:

> I've added the autoload.

Great, thanks.

-- 
Thierry
Get my Gnupg key:

PR #2158, opened 2015-08-09
always show revision header but make it selectable

tarsius:
Rerun of #2036.


tarsius:
I am giving up on this. Maybe it's possible to make the pseudo header-line look like the header-line. But it appears that this requires many tricks, which cannot be generalized. So users would have to customize this, and I would have to help them do it. That's not worth it.

PR #2168, opened 2015-08-13
Fix bug where magit-status wouldn't work because of files in working directory

niligulmohar:
In some working directories `magit-status` would only open an empty buffer and display the message

    magit-process-insert-section: Wrong type argument: arrayp, nil

The problem was in `magit-rev-format`. Enabling `magit-git-debug` revealed that its git invocation failed with this message:

    128 git … show -s --format=%h %s FOO
    fatal: ambiguous argument 'FOO': both revision and filename

This fix adds the disambiguating "--", which makes `magit-status` work even if there are unfortunately named files in the working directory.


tarsius:
Applied. Thanks!

PR #2177, opened 2015-08-16
Add debug declare to most defmacro

vanicat:


tarsius:
Thanks. I have taken the liberty to but `debug` on the same line as `indent`.

PR #2178, opened 2015-08-17
Clarify magit-popup-use-prefix-argument options

phil-s:
The value options in the Customize interface for `magit-popup-use-prefix-argument` can be confusing at first glance, because the menu of options does not show the context for the "if-then-else" values.

It's not ambiguous as such, but a small wording tweak means no one will have to double-check the name of the variable to be certain of what each value means.

### DIFF
--- a/lisp/magit-popup.el
+++ b/lisp/magit-popup.el
@@ -117,8 +117,8 @@ that without users being aware of it could lead to tears.
 `nil'      Ignore prefix arguments."
   :group 'magit-popup
   :type '(choice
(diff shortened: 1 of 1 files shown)


phil-s:
Actually, this can be clearer still. I'll re-roll it.

tarsius:
You can just force push to the same remote. No need to create a new pr.

phil-s:
Ah, thanks. I've always found the whole github "pull request" system incredibly confusing, but that goes a long way towards making it seem saner.

Issue #2086, opened 2015-07-16
Add a template for new issues to CONTRIBUTING.md ?


phil-s:
I imagine this would be helpful in obtaining key information from users (such as Magit, Emacs, and Git version details) when they are submitting a new issue. This seems especially pertinent now that Magit 2 is released.

As I understand it, CONTRIBUTING.md is the closest thing in Github to providing any kind of issue template, so I'm just envisioning some content that users could copy & paste into the issue textarea, and then edit in the appropriate values.

(I did notice https://github.com/kentcdodds/issue-template but I don't know how that's supposed to work. It says it was written for the AngularJS project, but https://github.com/angular/angular.js/issues/new doesn't present any kind of template, so that's not a ringing endorsement.)

tarsius:
Unfortunately, I am under the impression that this won't make a big difference, if any. I mean, I have just recently added a "frequently asked, frequently asked questions" thingy to the readme and wiki front page. With two hands pointing to it... Many users just don't read instructions.

I think the only technical solution to this issue would be if Github allowed developers to style the note which asks users to read the contribution guidelines before they open a new issue. Currently it pretty much looks like this:

> Please review the [**guidelines for contributing**](https://github.com/magit/magit/blob/master/CONTRIBUTING.md) to this repository.

Sure the background is a light yellow and the link is bold, but that simply ain't enough to get the attention of many users. If I could, then I would add geocity-style animated rainbow lines, pointing fingers, and dancing babies around that box.

----

You also seem to imply that users should be instructed to report what versions they are using. That request is indeed missing from the instructions and *maybe* it should be added. But I have a feeling that those users who would read the instructions do mention the used version when appropriate, even when not told to do so. While the behavior of users who don't read instructions are not affected by instructions they haven't read.

Also I think that I am usually quite good at spotting when a user is talking about some old version. A bigger problem is when I explicitly ask about it and then get an incorrect answer. We had a similar problem after starting to involve the Emacsclient in the committing process. I ended up (sadly much to late) writing a debugging function (`with-editor-debug`) which answered my questions, so that I did not have to rely on answers which were not actually verified by the user providing them.

[...]

PR #2179, opened 2015-08-17
Clarify magit-popup-use-prefix-argument options

phil-s:
The value options in the Customize interface for `magit-popup-use-prefix-argument` can be confusing at first glance, because the menu of options does not show the context for the "if-then-else" values.

It's not ambiguous as such, but a small wording tweak means no one will have to double-check the name of the variable to be certain of what each value means.

n.b. I can see that the code base is pretty good at sticking within 80 column lines, and this change makes these lines longer, but I wasn't sure what your preferred format would be if that was to be avoided. Happy to re-roll as required.

### DIFF
--- a/lisp/magit-popup.el
+++ b/lisp/magit-popup.el
@@ -117,8 +117,8 @@ that without users being aware of it could lead to tears.
 `nil'      Ignore prefix arguments."
   :group 'magit-popup
   :type '(choice
-          (const :tag "Use default action, else show popup" default)
-          (const :tag "Show popup, else use default action" popup)
+          (const :tag "With prefix use default action (without prefix show popup)" default)
+          (const :tag "With prefix show popup (without prefix use default action)" popup)
           (const :tag "Ignore prefix argument" nil)
           (const :tag "Abort and show usage information" disabled)))
 


tarsius:
> Author:     phil-s <phil-s@users.noreply.github.com>

You still haven't configured Git properly. You have to set the Git variables `user.name` and `user.email`.

tarsius:
@kyleam @npostavs what do you think? I think the current descriptive strings are good enough, but don't really care.

kyleam:
It reads ok to me because customize shows "Control how prefix arguments
affect infix argument popups" as the description, making me interpret
the first part as what the prefix does.

@phil-s Would the rephrasing below help or do you think it needs to
include "with prefix"?

    (const :tag "Call default action instead of showing popup" default)
    (const :tag "Show popup instead of calling default action" popup)

npostavs:
I like @kyleam's proposal, it flows best. The "With prefix X (without prefix Y)" is a little clunky.

The original "Do X, else Y" is slightly ambiguous, in that the "else Y" could be read as another possible effect of the prefix argument rather than the unprefixed one.

phil-s:
FWIW I used "With prefix X (without prefix Y)" because setting this option to either of those values affects *both* cases, and to me it seemed best to make that unambiguous. "X instead of Y" isn't as "clunky", but is also not explicit about the fact that, *without* the prefix, the inverse will occur.

> You still haven't configured Git properly. You have to set the Git variables user.name and user.email.

I queried that in https://github.com/magit/magit/issues/2086#issuecomment-122249027 but I think you missed it. <blockquote>As for the name and email address, I'm completely happy with it being listed as you've shown. (I'm not sure if you raised it because you thought I'd want it to be different, or because you want it to be different?)</blockquote>

tarsius:
> I did

Then something went wrong. What I have quoted above is what is actually stored in the commit.

phil-s:
Crossed wires, I think. I'm just saying that, for my part, I'm perfectly happy with being listed as `phil-s <phil-s@users.noreply.github.com>`. I could change it if you feel it's a problem, but I don't otherwise have any inclination to do so.

tarsius:
> I'm just saying that, for my part, I'm perfectly happy with being listed as `phil-s <phils@users.noreply.github.com>`.

Do you have ideological reasons for this, i.e. do you want to take a stand against real-name-policies?

Otherwise please just configure git properly, because while you might not care about proper metadata in commits, most maintainers of the projects you contribute to probably do, and they shouldn't all have to add an entry to `.mailmap` just because you don't care.

> I could change it if you feel it's a problem, but I don't otherwise have any inclination to do so.

Yes please do that. Also for the benefit of projects which might otherwise have to add `.mailmap` just because of you. Even if they already have such a file you would do them a favor if you did not force them to add an entry and ask you for the information they need to be able to add that entry.

----

I have installed @kyleam's change. They have the advantage that they are not cut of in a somewhat narrow window, and they are good enough.

phil-s:
> Do you have ideological reasons for this, i.e. do you want to take a stand against real-name-policies?

I guess? That's a stronger statement than I'd use, though. I'm not "taking a stand" (and if I was I'd be failing dismally in lots of places), but my usual position has always been that if I don't need to hand out my name or email address, then I won't.

To be honest I still don't understand why it's important here, but I'm not looking to make things difficult or to waste your time, so I'll aim to make any future commits with these details configured.

PR #2192, opened 2015-08-19
Change what information magit-mode-get-buffer uses to find an appropriate buffer

tarsius:
This is very unfinished.


tarsius:
Don't forget 9a72947fd8bfa86985fe08cd82a583f2877af222.

tarsius:
This needs a more holistic approach.

Issue #2147, opened 2015-08-06
way to switch between different histories of given file?


blaenk:
When I read through the magit manual I vaguely remember seeing something about this, but now I can't find it. I'm new to emacs and something vim-fugitive had was the ability to cycle through the history of a file. I think the git-timemachine package enables something like this, but I would absolutely love it if I could use existing magit constructs to do this, especially because it seems (or perhaps I'm mistaken) that magit is most of the way there.

I found out about `magit-log-buffer-file` which opens a split window of the git log of commits that affected the current file. What I'm envisioning is the ability to enable some sort of 'follow mode' so that when I traversed this list it would automatically set the contents of the buffer to the version of the file at that commit. If no follow mode ability, then at least some way to set the buffer contents via a binding?

tarsius:
This is not currently possible and I recommend you just use `git-timemachine` for now.

> but I would absolutely love it if I could use existing magit constructs to do this, especially because it seems (or perhaps I'm mistaken) that magit is most of the way there.

You are kinda mistaken. The current implementation is a bit of a hack (like those in vc and git-timemachine). I intend to implement a proper file-handler `git-handler.el` and not add any related functionality until I have done that.

blaenk:
Thanks for the response. I meant in terms of UI mainly, with the log window opening underneath the file, since I'm not at all familiar with the implementation of magit.

I look forward to this! But in the meantime I'll go ahead and use git-timemachine then :)

tarsius:
I have implemented basic blob navigation in #2195. This is merged into `pu` and will likely make it into `v2.3`.

PR #2195, opened 2015-08-21
Basic blob navigation

tarsius:
This allows navigating blobs similar to how `git-timemachine` does it. From a buffer visiting a tracked file use `C-c p` to go to the latest blob for that file, then `p` and `n` to navigate between blobs. Unlike `git-timemachine` which uses a some sort of session and displays all blobs in the same buffer, this simple implementation uses a separate buffer for each blob and there is no command to quit the session (i.e. kill all the created buffers). Therefore it is better to move "forward" using `q` instead of `n`, as the former kills the currently visited blob.

### DIFF
--- a/Documentation/RelNotes/2.3.0.txt
+++ b/Documentation/RelNotes/2.3.0.txt
@@ -98,7 +98,8 @@ THIS IS NOT COMPLETE.
 * Renamed `magit-file-buffer-mode' to `magit-file-mode'.  Related
   symbols were renamed accordingly.
 
-* Added new minor mode `magit-blob-mode'.  #2195
+* Added new minor mode `magit-blob-mode' and new commands
+  `magit-blob-previous' and `magit-blob-next'.  #2195
 
 Authors
 -------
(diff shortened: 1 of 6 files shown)


blaenk:
This is great work! I'll have to check it out, but I'm wondering what you thought about what I was saying in #2147? In particular, I really like how `magit-log-buffer-file` shows the log of commits that affected the file in a split. It would be very nice if we could (if not already possible with this) keep that around and in fact use it for easier navigation of the history, so we could press some key on one of those commits to load a buffer of that point in history, and I believe it would be even more awesome if it had support for next-error-follow-mode like compilation mode and certain search modes have, so that it would automatically switch to the buffer represented by the commit when pressing `n` or `p` for example. I feel like it would be very natural and seamless.

Perhaps this is just preliminary work, but I was interested to see what you thought about that.

tarsius:
Yes, I'll consider that too.

blaenk:
Great!

tarsius:
The mode names `magit-file-buffer-mode` and now `magit-blob-buffer-mode` are a bit silly. Should I use `magit-file-mode` and `magit-blob-mode` instead?

blaenk:
Sounds good to me.

tarsius:
Or maybe `magit-{blob,file}-minor-mode`.

What do you think @kyleam?

kyleam:
I prefer magit-{blob,file}-mode.

PR #2201, opened 2015-08-24
magit-blame fixes: narrowed and indirect buffers, untracked files

kyleam:


jinnovation:
I believe commit 8fba616 breaks `magit-blame` over Tramp.

kyleam:
Please open a separate issue with details.

jinnovation:
Done. See #2206.

PR #2209, opened 2015-08-26
magit-toplevel: always follow links to repo subdirectories

tarsius:
This is intended to be merged before v2.2.2 is released.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -290,10 +290,9 @@ call function WASHER with no argument."
   (declare (indent 1) (debug (form body)))
   `(catch 'unsafe-default-dir
      (let ((default-directory
-             (let ((file ,file))
-               (file-name-as-directory (if file
-                                           (expand-file-name file)
-                                         default-directory)))))
+             (file-name-as-directory (--if-let ,file
+                                         (expand-file-name it)
+                                       default-directory))))
        (while (not (file-accessible-directory-p default-directory))
          (when (string-equal default-directory "/")
            (throw 'unsafe-default-dir nil))
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -302,9 +302,11 @@ call function WASHER with no argument."
        ,@body)))
 
 (defun magit-git-dir (&optional path)
-  "Return absolute path to the GIT_DIR for the current repository.
-If optional PATH is non-nil it has to be a path relative to the
-GIT_DIR and its absolute path is returned."
+  "Return absolute path to the control directory of the current repository.
+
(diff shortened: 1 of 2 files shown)


tarsius:
@kyleam does this look sane to you? I have added tests and everything, but cannot hurt if you review too. Thanks.

kyleam:
> @kyleam does this look sane to you?

Yes.  I tried to reproduce issue #2220 (git 2.5.0, linux) and cannot.  Perhaps it's the git version.

kyleam:
Hmmm, over tramp 

```
(while (not (file-symlink-p
                           (directory-file-name default-directory)))
                (when (string-equal default-directory "/")
                  (error "Unexpected error while determining toplevel"))
                (setq default-directory
                      (file-name-directory
                       (directory-file-name default-directory))))
```

can get into a infinite loop beause "/" is not the boundary case.  Perhaps this should also check if default-directory has changed over the iteration.

kyleam:
Also I think the

    (equal (expand-file-name cdup) topdir)

might be off by a trailing slash.

lisp/magit-git.el
```diff
@@ -303,19 +302,47 @@ call function WASHER with no argument."
...
+              topdir
+            (if (equal cdup "")
+                ;; `default-directory' links to toplevel, don't follow.
+                default-directory
+              ;; A directory above `default-directory' is a link...
+              (while (not (file-symlink-p
+                           (directory-file-name default-directory)))
+                (when (string-equal default-directory "/")
+                  (error "Unexpected error while determining toplevel"))
+                (setq default-directory
```
kyleam: Over tramp this can get into a infinite loop beause a directory is no longer changes at the root and "/" is not the boundary case. Perhaps this should also check if default-directory has changed over the iteration.

lisp/magit-git.el
```diff
@@ -303,19 +302,47 @@ call function WASHER with no argument."
...
-        (magit-expand-git-file-name
-         (file-name-as-directory (expand-file-name cdup)))
+    (-if-let (topdir (magit-rev-parse-safe "--show-toplevel"))
+        (let (cdup)
+          (setq topdir (file-name-as-directory topdir))
+          (if (or find-file-visit-truename
+                  ;; ^ User wants to follow all links.
+                  ;; v There are no relevant links.
+                  (and (setq cdup (magit-rev-parse-safe "--show-cdup"))
+                       (equal (expand-file-name cdup) topdir)))
```
kyleam: Think this can be off by a trailing slash.

kyleam:
[Moved code comments inline.  Sorry for the noise.]

PR #2216, opened 2015-08-27
Refactor how the revision buffer is created and looks

tarsius:
This isn't quite finished but I have merged into `pu` anyway.


kyleam:
Minor comments (which might change after I use this for more than a few
hours) about the new appearance:

* I'd prefer a new line before the subject line so that there is a
  little separation from the header.

* Perhaps the hash should be on a separate line than refs.

kyleam:
After using this a bit longer, I miss having the author information / date before the commit message.  In addition to expecting it because git log's format and email headers, I think it is nice to have that information (which is a fixed number of lines) in the same place when moving between revisions.

tarsius:
I've changed the order.

kyleam:
The last magit-delete-line in magit-diff-wash-diffstat fails for
some revisions because `git show` seems to vary in the number of new lines.

For example

```
$ git show -p --numstat 332a020c161421f0bbce7e52ab9b4eda577c8e6b^
commit ef6e8f4877fdf97ab62a7b24888a2896039375f9
Merge: 4eb1147 8b26328
Author: Jonas Bernoulli <jonas@bernoul.li>
Date:   Mon Aug 24 20:06:44 2015 +0200

    Merge branch 'jb/spinoff-and-reset-branch' into pu

26      0       Documentation/magit.org
30      0       Documentation/magit.texi
63      1       lisp/magit.el


```

does not fail, but

```
$ git show -p --numstat 332a020c161421f0bbce7e52ab9b4eda577c8e6b
commit 332a020c161421f0bbce7e52ab9b4eda577c8e6b
Merge: ef6e8f4 be69262
Author: Jonas Bernoulli <jonas@bernoul.li>
Date:   Mon Aug 24 20:06:53 2015 +0200

    Merge branch 'jb/quit-window' into pu

38      32      lisp/magit-mode.el

```

contains one fewer new lines and fails when magit-delete-line tries to
delete beyond point-max.

tarsius:
Actually there is `core.quotepath`. Don't know why I missed that until now.

lisp/magit-git.el
```diff
@@ -814,9 +814,17 @@ Return a list of two integers: (A>B B>A)."
...
-(defun magit-rev-format (format &optional rev)
-  "Return output of `git show -s --format=FORMAT [REV]' --."
-  (magit-git-string "show" "-s" (concat "--format=" format) rev "--"))
+(defun magit-rev-format (format &optional rev args)
+  (let ((str (magit-git-string "show" "--no-patch"
+                               (concat "--format=" format) args
+                               (concat rev "^{commit}") "--")))
+    (unless (string-equal str "")
+      str)))
+
```
kyleam: Why not return the full output here instead of just the first line?  The format is already specified, so it shouldn't break any calling code.

tarsius: The full output would then include at least one newline.

kyleam: doh, right

tarsius:
> Actually there is `core.quotepath`. Don't know why I missed that until now.

Not that this is relevant here...

tarsius:
> Perhaps the hash should be on a separate line than refs.

I think that's okay. Sure the line might get rather long and then part of the hash might be cut of, but that's only for users who don't know about `M-w` anyway.

PR #2232, opened 2015-09-02
test user, in case not set in chroot during testing

barak:
Set user.name and user.email in repo created during "make test".
This is necessary on, e.g., autobuilders using chroots, where these are in general not set.


tarsius:
I did something similar recently, but just for Travis. What you are proposing works for the current tests, but some tests cooking in feature branches do not use this macro. Instead I am setting the environment variables in `magit-with-test-directory`, see e02359fc4a9bbfc911b317ea7a3597294d1a9cc5.

PR #2233, opened 2015-09-02
Add --cover-letter to magit-patch-popup

rootzlevel:


tarsius:
I don't intend to make any changes here until we get started with #1800. This argument makes sense now but it might make less sense once Magit has gained less bare-bone support. I would like to avoid adding this now and then "take it away" from users later.

Please add an entry to https://github.com/magit/magit/wiki/Additional-proposed-infix-arguments-and-suffix-commands.

PR #2235, opened 2015-09-03
r

barak:


tarsius:
Applied with a tiny change in the commit message. Thanks.

Issue #2135, opened 2015-08-03
Symlinks to subdirectories of repos are not followed


chep:
HI,

I don't know if this is a regression or a feature but magit does not follow
symbolic links anymore.

Example:

/home/chep
|
|---- folder1 -> gitRepo/folder1
|
|---- gitRepo
|     |
|     |---- folder1
|     |
|     |---- folder2
...

When I use C-u M-x magit-status RET /home/chep/folder1 RET I have the message:

/home/chep/ is a repository.  Create another in /home/chep/folder1/? (y or n)

It used to notice that folder1 is in gitRepo and was opening gitRepo status but
since a couple of weeks it does not work anymore.

Regards.

tarsius:
Works for me. What does `M-x magit-version` say?

```
/tmp/test
├── extern
│   ├── .git
│   └── subdir
└── outer
    ├── .git
    └── inner -> ../extern
```

Entering `/tmp/test/outer/inner/` shows the status for that repository. `/tmp/test/outer/inner/subdir/` asks whether I want to create a new repository in there.

chep:
M-x magit-version:
Magit 20150802.1333, Git 2.4.6, Emacs 25.0.50.1

tarsius:
Indeed I get that behavior too.

We now just ask where the repository is, doing the equivalent of:

```
cd /home/chep/folder1
git rev-parse --git-dir
```

I suppose we could add an option analogous to `find-file-visit-truename` which would cause `/home/chep/gitRepo/folder1` to be substituted for `/home/chep/folder1`. But I don't think this sort of symlink is common, so I am not sure this is worth it.

tarsius:
Hm... actually this should work without such a hack. I think. Will check later.

tarsius:
What I wrote before was all non-sense. Will have to check later.

Just so much for now: the relevant commit is 65787d92910a77f0528fc4623c41af0e0b8e0e91.

tarsius:
* We cannot use `--show-toplevel` because it always follows symlinks (anywhere leading up to pwd).
* `--show-cdup` doesn't work properly when a symlink points to a *subdirectory* of repository.
* We currently use `--show-toplevel` because that also respects `--work-tree`, `GIT_WORK_TREE`, and `core.worktree`.

Looks like we have to implement `--show-toplevel` in elisp and make it respect `find-file-visit-truename`.

tarsius:
[...]

PR #2242, opened 2015-09-05
Refactor magit-toplevel

tarsius:
The idea is to only follow symlinks if absolutely necessary.  The
previous implementation was broken in that it usually failed when
a symlink to some sub-directory of a repository was involved (and
when that succeeded, then only by accident).

This issue was reported in #2135.  A previous attempt (#2209)
failed to address this issue, as it caused other cases to fail.

### DIFF
--- a/Documentation/RelNotes/2.3.0.txt
+++ b/Documentation/RelNotes/2.3.0.txt
@@ -86,5 +86,9 @@ THIS IS NOT COMPLETE.
   the same index as the previous hunk, but that often was very far
   away from where point was before the refresh.  #2227
 
+* The function `magit-toplevel' was refactored and many tests were
+  added.  It now returns the correct value when a symlink to a
+  sub-directory of a working tree is involved.  #2242
+
 Authors
 -------
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -314,6 +314,24 @@ absolute path is returned."
       (if path (expand-file-name (convert-standard-filename path) it) it))))
 
 (defun magit-toplevel (&optional directory)
+  "Return the absolute path to the toplevel of the current repository.
+
+From within the working tree or control directory of a repository
+return the absolute path to the toplevel directory of the working
+tree.  As a special case, from within a bare repository return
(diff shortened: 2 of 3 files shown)


lisp/magit-blame.el
```diff
@@ -218,52 +218,49 @@ only arguments available from `magit-blame-popup' should be used.
...
            (user-error "Buffer isn't visiting a file"))))))
-  (let ((toplevel (or (magit-toplevel)
-                      (user-error "Not in git repository"))))
+  (magit-with-toplevel
     (if revision
         (magit-find-file revision file)
-      (let ((default-directory toplevel))
-        (find-file file)))
+      (find-file file))
     (widen)
```
kyleam: Avoiding magit-with-toplevel was intentional here: b640c6f (magit-blame: fix calls from indirect buffers, 2015-08-23).

tarsius: 597a4687bcc2e37ba53b76ee6d88307394125377 should do the trick, right?

kyleam: Yep.  Thanks.

kyleam:
The magit-toplevel rewrite looks reasonable to me.

tarsius:
Also have a look at the tests. (Yes they are ugly, maybe I don't write more tests because I haven't figured out yet how to do it cleanly.) Have I again forgotten any cases which should be checked?

lisp/magit-blame.el
tarsius: It seems that by now it is already impossible to get the rev for this out of github. However it also looks like having mentioned a commit which is still on a branch has prevented this from being discarded completely.

If that is the case, then it would probably be a good idea to add an explicit link from the commit to the pull request when first commenting on a commit. Just add it at the very end.

#2242

kyleam:
> Have I again forgotten any cases which should be checked?

I can't think of any.

lisp/magit-blame.el
kyleam: > It seems that by now it is already impossible to get the rev for this
> out of github.

Welp, time to move development to a mailing list :)

PR #2243, opened 2015-09-06
avoid `git log --graph' inefficiency

tarsius:
`git log --graph --max-count=N` calculates the graph for the complete
history; the `--max-count=N` only affects the output phase.

`git log --graph HEAD~N..HEAD` only calculates the graph for
HEAD~N..HEAD but the graph is not the same; commits that are reachable
from HEAD but do not descend from HEAD~N will appear as root commits.

Address this by using `--max-count` and REV~M..REV to together when
possible.  More precisely do so when showing the log for a single rev
in the log buffer.  M is usually twice as large as N but at least 1024,
which greatly reduces the risk of the graphs for REV~M..REV and REV to
differ.

In the status buffer also limit the range for the log section inserted
by `magit-insert-recent-commits`, but without avoiding that non-root
commits appear as root commits (because that happens for the sections
inserted by `magit-insert-{unpulled and unpushed}-commits` too).

Also see http://www.spinics.net/lists/git/msg232230.html.

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -668,6 +668,16 @@ Type \\[magit-reset] to reset HEAD to the commit at point.
     (setq args (remove "--follow" args)))
   (when (--any-p (string-match-p magit-log-remove-graph-re it) args)
     (setq args (remove "--graph" args)))
+  (when (and magit-log-cutoff-length
+             (= (length revs) 1)
+             (setq revs (car revs))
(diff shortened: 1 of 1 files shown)


npostavs:
This currently has the effect that `l a` (`magit-log-all`) shows no commits.

tarsius:
Fixed that.

npostavs:
> Fixed that.

Looks good :)

PR #2244, opened 2015-09-07
magit-rebase-interactive-1: don't drop rebase args

kyleam:
I'm not sure if this is a good way to fix this.

### DIFF
--- a/lisp/magit-sequence.el
+++ b/lisp/magit-sequence.el
@@ -368,8 +368,9 @@ START has to be selected from a list of recent commits."
                                  (unless (member "--root" args) commit)
                                  (magit-rebase-arguments)))
     (magit-log-select
-      `(lambda (commit) (magit-rebase-interactive-1 commit ,message ,editor
-                                                    (list ,@args)))
+      `(lambda (commit)
+         (magit-rebase-interactive-1 commit ,message ,editor
+                                     (list ,@args ,@(magit-rebase-arguments))))
       message)))
 
 ;;;###autoload


tarsius:
> I'm not sure if this is a good way to fix this.

What are you trying to fix? Also, why is the "backquote construct fix" needed?

kyleam:
> What are you trying to fix?

The popup switches are dropped when the commit is selected through
magit-log-select.

For example, running 'r-ae' when point is not on a commit ends up
running

      git … rebase -i commit^

instead of

     git … rebase -i commit^ --autosquash

> Also, why is the "backquote construct fix" needed?

To prevent things like 'rf' failing with

    magit-rebase-interactive-1: Invalid function: "--autosquash"

npostavs:
I would prefer writing it like this
```elisp
`(lambda (commit)
   (magit-rebase-interactive-1 commit ,message ,editor
                               ',(append args (magit-rebase-arguments))))
```

or with lexical-binding:
```elisp
(let ((rebase-args (append args (magit-rebase-arguments))))
  (lambda (commit)
    (magit-rebase-interactive-1 commit message editor rebase-args)))
```

tarsius:
My intention was to get the value of `magit-rebase-arguments` inside `magit-rebase-interactive-1`, right before actually needing it: in the "we have a commit branch", and we do. We don't need this in the "user needs to select a commit branch" because that ultimately calls the other branch.

The `args` argument is still needed to pass variant-specific arguments. Maybe I should have called it `additional-args`, but you know how I feel about long variable/argument names :-)

Usually I try hard to get prefix and infix arguments inside `interactive` but here I have chosen otherwise because the noise that it would cause is not worth it in this case. This makes it impossible to pass arbitrary arguments when using these commands non-interactively, but "that's a feature" and the commands are properly marked as only intended for interactive use.

The unquoting was indeed broken.

And yeah, we should switch to lexical scope soon.

kyleam:
> We don't need this in the "user needs to select a commit branch"
> because that ultimately calls the other branch.

But by that point the popup is no longer active and
(magit-rebase-arguments) returns the value of the magit-rebase-arguments
variable, dropping any switches activated in the popup.

tarsius:
Oh boy, yes. I have applied a slightly different fix (but yours would have worked too and the arguments would not have been duplicated as I originally though).

Needless to day I am no longer so happy with my original design decision and might go back to getting the arguments inside `interactive`.

tarsius:
Hm, and my solution also doesn't work then the commands are invoked directly instead of through the popup.

Issue #2049, opened 2015-07-11
magit causes git to hog CPU by running "git update-index" with tons of input files


mgalgs:
It looks like magit is calling `git update-index`, passing what appears to be every file in my repository.  Based on some observations of the offending sub-processes being created (see below) the offending code appears to be `(magit-update-files)`.  While working on the Linux kernel this completely hangs emacs for up to a minute.

I've noticed this while staging changes after resolving merge conflicts and also while stashing.  Looking at the call sites for `(magit-update-files)` I don't see how this can possibly happen...  Any ideas?

Here's the (truncated) command line scraped from `ps` during one such hang while trying to stage a file where I had just finished resolving some merge conflicts:

```
ps -ef | grep git
[...]

PR #2253, opened 2015-09-09
magit-stash: #2049 fix git diff taking up 100% cpu

bsuh:
Per my comments in #2049, I believe this is the correct fix.


tarsius:
See #2049.

PR #2256, opened 2015-09-11
magit-log-refresh-buffer: handle non-existing rev

kyleam:


tarsius:
I suppose that the motivation is that `(empty)` being shown in the buffer is better than a type error when the user provided invalid input. I think it would be even better if we raised a `user-error` saying "your input ain't no good".

kyleam:
> I think it would be even better if we raised a `user-error`
> saying "your input ain't no good".

Hmm I left it at "(empty)" because user input isn't the only way this
can happen.  Another way (and how I came across the error) is if you
have a log buffer sitting around for a branch that you delete.  Do you
think it's worth dealing with these two things separately?

tarsius:
I have applied your fix. This is certainly better than a type error.

> I left it at "(empty)" because user input isn't the only way this can happen.

I haven't thought of that. That just makes it more important that we apply the fix.

But we should also improve this. We should probably figure out a way to print the error messages when an error occurs when running git in order to refresh a `magit-diff-mode` or `magit-log-mode` buffer. (But we should have done that a long time ago, so there's no rush just because of that other error.)

tarsius:
... but that's on my todo-list anyway. And this commit has been cherry-picked. So I am closing.

PR #2260, opened 2015-09-15
In terminal C-return is sent as C-j due to some terminal limitations

mar-kolya:
It looks like it is safe to allow C-j work as C-return for diff-related commands


tarsius:
Generally speaking you're probably right, but given my personal configuration (I use <kbd>C-{jikl}</kbd> for basic navigation) it means I have to do these bindings. I have cherry-picked you commit, and made some minor changes to alignment and the commit message. Thanks.

PR #2263, opened 2015-09-15
Create a separate status section for unmerged changes

tarsius:
This is an naive early implementation, for demonstration purposes.

* Can we assume that `git diff --cached --diff-filter=U` always includes all unmerged changes? Or do you have to *also* use `git diff --diff-filter=U`?
* For easier parsing we should also use `--name-status`.
* Many commands that behave differently based on the "diff type" of a file have to be adjusted due to this.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -1734,7 +1734,8 @@ or a ref which is not a branch, then it inserts nothing."
   (magit-insert-section (unstaged)
     (magit-insert-heading "Unstaged changes:")
     (magit-git-wash #'magit-diff-wash-diffs
-      "diff" magit-diff-section-arguments "--no-prefix"
+      "diff" magit-diff-section-arguments
+      "--diff-filter=ACDMRTXB" "--no-prefix"
       "--" magit-diff-section-file-args)))
 
 (defvar magit-staged-section-map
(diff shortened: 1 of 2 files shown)


kyleam:
> Can we assume that `git diff --cached --diff-filter=U` always includes
> all unmerged changes?

I believe so.

kyleam:
One odd thing about having a separate "Unmerged changes" section is
that, for cases other than deletions, it can result in entries that
are already covered by entries in "Unstaged changes".

For example, with a plain merge conlfict

    Unmerged changes (1)
    unmerged   two

    Unstaged changes (1)
    unmerged   two
    @@@ -1,1 -1,1 +1,7 @@@
    ++<<<<<<< HEAD
     +ours
    ++||||||| merged common ancestors
    ++changetwo
    ++=======
    + other
    ++>>>>>>> plain

or a merge conflict involving an added file in both histories

    Unmerged changes (1)
    unmerged   five (both added)

    Unstaged changes (1)
    unmerged   five
    @@@ -1,1 -1,1 +1,6 @@@
    ++<<<<<<< HEAD
     +five
    ++||||||| merged common ancestors
    ++=======
    + other-five
    ++>>>>>>> fv

Issue #2115, opened 2015-07-25
"Push Current" always asks where to push


mrkkrp:
It seems like `Push Current` <kbd>P P</kbd> now always asks where to push, every time I push I see IDO interface with only one option `origin/master`. I've tried to set upstream again (although it's set already), but it doesn't work.

I believe it's due to some recent change in Magit. I can't say I'm happy with this new behavior because pushing is a bit harder now. Can you clarify whether it's new feature or a bug?

If this is on purpose, how can I return to old-fashioned `Push Current`?

mrkkrp:
I've just read #2076, so I see it's on purpose. This works:

```emacs-lisp
(setq magit-push-always-verify nil)
```

I would add this to Magit FAQ.

tarsius:
Indeed I recently made such a change to protect the innocent (see #2115). I did not only change the behavior, I also added a note about it to the prompt:

```
Push somebranch to [also see option magit-push-always-verify]
```

Not the part in brackets. Doing that will answer your questions.

tarsius:
Ah, you've seen it.

syl20bnr:
@tarsius  I'm not sure about what is different in Spacemacs but when I push the  `develop` branch I got a helm popup with `origin/develop` as the default selection which is different than OP. Moreover the helm buffer is populated with all the existing branches which again is different than OP.

PR #2264, opened 2015-09-15
Document magit-push-always-verify in functions.

Peeja:
The functions which change behavior based on `magit-push-always-verify` still only describe the case where it is nil, which it is not by default. I don't know if the wording is great, but I'd love to have something in there that makes this clear.

This should help with confusion like #2115.


tarsius:
Thanks. I have cherry-picked.

PR #2271, opened 2015-09-17
Allow applying multiple hunks at once

tarsius:
Also allow applying multiple files at once in the "regular apply" variant, which was oddly missing.

* I am considering passing the value of `magit-region-sections` around for consistency.
* Note that the `&rest args` argument is being flattened in most cases, so an `apply` might be missing in places where you would expect one.


lisp/magit-apply.el
```diff
@@ -61,27 +61,58 @@ With a prefix argument and if necessary, attempt a 3-way merge."
...
+      (`(untracked ,(or `file `files))
+       (magit-am-popup))
+      (`(,_ region) (magit-apply-region it args))
+      (`(,_   hunk) (magit-apply-hunk   it args))
+      (`(,_  hunks) (magit-apply-hunks     args))
+      (`(,_   file) (magit-apply-diff   it args))
+      (`(,_  files) (magit-apply-diffs     args)))))
+
+(defun magit-apply-diffs (&rest args)
+  (let ((sections (magit-apply--get-diff-sections (magit-region-sections))))
```
kyleam: Discarding a staged file goes down this path, resulting in a type
error when magit-apply-patch is called with nil for section:s.

PR #2274, opened 2015-09-18
Add magit-file-mode to dired buffers

mar-kolya:


tarsius:
Not all features used by that mode work in dired buffers. I don't intend to add something like a `magit-dired-mode` anytime soon, and would like to delay further work in this area until I have implemented a file handler for blobs and trees.

Issue #2276, opened 2015-09-18
Can't reclaim authorship if there's nothing staged


mgalgs:
When I accidentally commit something with the wrong authorship information the way I usually fix it is with `c -R e` (i.e. `pop-up commiting` => `Claim authorship and reset author date` => `Extend`), which is equivalent to:

    git commit --amend --no-edit --reset-author

However, this recently started error'ing out with:

    user-error: Nothing staged (or unstaged)

This check seems to have been introduced by 00aec0a082f8f55cd40ebfb4bc8cc7dd40596b93 so I'm not sure why I'm only noticing it now... Maybe I just haven't used this feature for a few months and suddenly I've needed to use it multiple times this week :)

---

    M-x magit-version
    Magit 20150916.450, Git 2.5.2, Emacs 25.0.50.1

npostavs:
This reminds me that I was trying to fix this for `--author=` and I noticed this too.

https://github.com/magit/magit/compare/master...npostavs:np/extend-author

tarsius:
You should use the "vanilla amend" using <kbd>c-Ra</kbd> and then <kbd>C-cC-c</kbd> without editing the message.

"Extend" is for adding more changes to HEAD and I don't think it would make sense for it to behave differently when these arguments are set.

While you can use the "reword" variant to do this you shouldn't because you would otherwise end up with a commit which was "committed before it was authored". The same would apply to the "extend" variant if we added a special case to allow such usage.

npostavs:
> You should use the "vanilla amend" using <kbd>c-Ra</kbd> and then <kbd>C-cC-c</kbd> without editing the message.

I was under the impression that the whole point of "extend" is to skip "then <kbd>C-cC-c</kbd> without editing the message". The manual's describes it as

> - Key: c e, magit-commit-extend

>  Amend the last commit, without editing the message.

But now you seem to be saying that preserving the date is the more (or at least equally) important feature. It seems a bit obscure, how often does anyone compare the commiter and author dates anway?

tarsius:
> Amend the last commit, without editing the message.

I think the name and date changing arguments weren't available from the popup when I wrote that, so what I meant was:

* extend: add more changes to HEAD without editing the message
* reword: change HEAD's message without adding more changes

Both are optimized for the "I just made a mistake very recently and don't think it has to be recorded in the history" use case by not changing the committer date which allows you to pretend that there was no typo in what was first committed. The options `magit-commit-{extend and reword}` can be used to change that.

[...]

PR #2278, opened 2015-09-19
magit-popup.org: Add example for `magit-define-popup-action'

mgalgs:
Requests for customizing the default magit popups are a large source of
support requests on the issue tracker (see #2276, for example).  Add an
example for `magit-define-popup-action` to enable more users to
customize things on their own rather than requesting changes to magit
core.


tarsius:
Could you please add that to https://github.com/magit/magit/wiki/Additional-proposed-infix-arguments-and-suffix-commands instead.

tarsius:
I have added a link to that page in 25e055a7f44ebfd1637534b0a6eb6870f6beb0d1.

PR #2281, opened 2015-09-20
Use the "noglob" option on Cygwin and MSYS2

### LINKED ISSUE #2246: Magit no longer works with cygwin ?
It seems magit version 20150907.504 doesn't work with cygwin on windows 

When I try to open magit-buffer with magit-status command :
magit-insert-head-header: Wrong type argument: stringp, nil

It was working fine with previous versions (with magit 2.1.0). Did you change something concerning how git command is running ?

Thanks

johnmastro:
By default, Cygwin and MSYS2 executables attempt to expand globs when
called by a program other than a Unix shell.  The intent is to provide
globbing when, for instance, using Cygwin's grep from the Windows
command prompt.

However, this causes problems when calling the Cygwin or MSYS2 git from
Magit in a native Windows Emacs.  The solution is to detect this
behavior and, when present, add "noglob" to the "CYGWIN" and "MSYS"
environment variables to disable it.

See https://cygwin.com/cygwin-ug-net/using-cygwinenv.html

Fixes #2246.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -152,21 +152,12 @@ pass arguments through this function before handing them to Git,
 to do the following.
 
 * Flatten ARGS, removing nil arguments.
-* Prepend `magit-git-global-arguments' to ARGS.
-* Quote arguments as required when using Powershell together
-  with Cygwin Git.  See #816."
-  (setq args (-flatten args))
-  (when (and (eq system-type 'windows-nt)
-             (let ((case-fold-search t))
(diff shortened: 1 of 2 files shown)


tarsius:
Looks good to me. @npostavs?

npostavs:
Yup, looks good to me too.

johnmastro:
Thanks @tarsius and @npostavs!

tarsius:
And thanks to you!

PR #2282, opened 2015-09-21
Give users more control over how buffers are displayed

tarsius:
This isn't quite ready yet.


tarsius:
That should be the last big change for a while. Now I intend to concentrate on bug fixing and polish.

Issue #2094, opened 2015-07-18
In hunk sections, when region is valid, next-line moves to column 0


npostavs:
Make some random change so that there will be an easily accessible change showing in the status buffer, then `emacs -Q -L ..\dash -L lisp -l magit -f magit-status` 
Navigate into the the hunk: `j u 4 C-5 C-n`
Move out of column 0: `C-f`.
Activate the region (should be a "valid" magit region): `C-SPC`
Move down 1 line: `C-n`
Point ends up in column 0 instead of staying in column 1 as expected. 

Repeating the experiment without activating region also stays in column 1.
Using an "invalid" region (starting from outside the hunk) also stays column 1.
Moving up (`C-p`) instead of down also stays in column 1.
Moving down from a point above the starting point of region (where it was initially activated) also stays in column 1 (i.e. doing `C-p` first, up to an equal amount of `C-n` stays in column 1).

Tested with Emacs 24.4, 24.5, 25 (July 3 snapshot) on Windows. Latest magit `master`: 3a758463980b2cd9a4e6b40539469a37032e51ef

---

I'm not sure where exactly the region stuff is implemented. Setting `post-command-hook` to `nil` has no effect.

tarsius:
It appears that something inside `magit-diff-update-hunk-region` causes this, but simply wrapping it's complete body with `save-excursion` doesn't appear to do the trick.

npostavs:
Seems like `(debug)` is disabled during these functions. Putting some `(message "xx %d" (current-column)` calls in, I see that point isn't moved during the function itself, but commenting out the bottom line overlay in this function does "fix" it:
```elisp
          (ov rend (1+ rend) 'after-string
              (propertize (concat (propertize "\s" 'display '(space :height (1)))
                                  (propertize "\n" 'line-height t))
                          'face 'magit-diff-lines-boundary)))
```

npostavs:
I put a similar overlay using the following function:
```elisp
(defvar np-overlays ())

(defun np-put-overlay (beg end)
  (interactive "r")
  (if np-overlays
      (progn (mapc #'delete-overlay np-overlays)
             (setq np-overlays nil))
    (let* ((rbeg (save-excursion (goto-char (region-beginning))
                                 (line-beginning-position)))
           (rend (save-excursion (goto-char (region-end))
                                 (line-end-position)))
           (ov (make-overlay rbeg (1+ rbeg) nil 'front-advance nil)))
      (overlay-put ov 'evaporate t)
[...]

PR #2293, opened 2015-09-25
Underline using :align-to

tarsius:
Open pr so that we have a place to talk about it. Edit: for previous discussions see #2094.

If Eli says one should not use newlines in overlays then we should try to follow his advise but unfortunately I have run into several issues with his approach too, which are just as much beyond our control as the issue with the old newline-approach, which I fixed for us. We might end up having to ask him for workarounds to be able to follow his advice.

Since the current approach works and I want to get a release out, I see no reason to rush this. But I would like to work with Eli to address the issues that are coming up here - they have bothered me in different contexts before.

There also were two issues with you implementation which I addressed in v2. When both the overline and the underline overlays were on the same line, then the underlining did not work. And you didn't account for pixel-wise resizing of windows.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -2021,11 +2021,11 @@ are highlighted."
       (when magit-diff-unmarked-lines-keep-foreground
         (setq face (list :background (face-attribute face :background))))
       (cl-flet ((ov (start end &rest args)
-                  (let ((ov (make-overlay start end nil t)))
-                    (overlay-put ov 'evaporate t)
-                    (while args (overlay-put ov (pop args) (pop args)))
-                    (push ov magit-region-overlays)
-                    ov)))
+                    (let ((ov (make-overlay start end nil t)))
+                      (overlay-put ov 'evaporate t)
+                      (while args (overlay-put ov (pop args) (pop args)))
+                      (push ov magit-region-overlays)
+                      ov)))
         (ov sbeg cbeg 'face 'magit-diff-lines-heading
             'display (concat (magit-diff-hunk-region-header section) "\n"))
         (ov cbeg rbeg 'face face 'priority 2)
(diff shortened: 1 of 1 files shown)


npostavs:
Not sure exactly what you meant by "blows up", but assuming it's the addition and removal of the newline following the header it should be fixed by b30aec92a64cd277ae61315cfef2de8be815a1b6.

I've noticed another glitch though: doing `C-e` (`move-end-of-line`) goes to the window's end instead of the line's end. Or perhaps more accurately, the end of line is pushed to rightward to the end of window, causing various changes with point movement around there.

tarsius:
I wonder, despite Eli's arguments, if we only want to change how things look, wouldn't it then maybe be best if we did just that ;-)

tarsius:
Or kindly ask him to implement `(space :align-to 'the-visible-edge-of-the-window)`.

npostavs:
> I've noticed another glitch though: doing C-e (move-end-of-line) goes to the window's end instead of the line's end.

Fixed that, but there's still some weirdness with `C-n` when point is at the end of line.

npostavs:
Eli fixed a bug where truncated lines with the overlay would cause `end-of-line` to skip a line, but there is still a minor glitch where the temporary-goal-column gets set to the windows' right edge instead of eol, when point is at eol. Eli won't fix that.

>This is not a bug: vertical-motion doesn't know about the 'cursor'
property, so it works as if the property didn't exist.
[...]
> I think a casual mismatch in a column while moving
through a buffer, and in marginal cases at that, is not a big deal.

He also says the overlay with newline approach will likely cause more, as yet uncovered, problems

> I'm quite sure problems will surface sooner or later.  Overlays with
newlines hit on fragile parts of the display engine code, which was
never really designed for that kind of stuff.

---

So we have to decide if a little glitch (or 2 on Emacs 24) is worth (hopefully) avoiding potential future problems. I'm leaning towards yes, because I don't think the glitches will be triggered in typical uses of the region.

tarsius:
I lean toward let's keep what is working now until it starts breaking. We should however have a patch ready for when it does start breaking.

tarsius:
Lets keep it ready (branch) but not pending (pull request).

Issue #2290, opened 2015-09-25
Cannot bisect run?


Silex:
Hello,

`bisect run` seems a bit weird at the moment... it's directly available on the bisect popup instead of being available *after* you did bisect start. If you try to use it it (rightly) complains that bisect hasn't started yet.

Atm that forces me to use bisect run from the command line.

tarsius:
`run` doesn't have to be invoked manually at every step I believe. I have never actually used this, but the documentation makes it sound that way in my ears. It talks about exit codes and such, not printing "this is bad/good". So I would assume that this automatically calls `good` or `bad`, and that causes the search to continue (unless we already found the first bad commit).

So it seems that I assumed that `run` could also be used to begin the search, without actually testing it. Assuming that the above is correct, we can implement that by teaching `magit-bisect-run` to first do what `magit-bisect-start` does if no sequence is in progress yet. I think it would make sense if one could specify the test script along with a known good and a known bad commit up front.

tarsius:
Maybe you could look into that some more?

Silex:
The workflow is like this:

* You start a bisect: `git bisect start GOOD_REV BAD_REV`
* You then have two choices:
  * Mark a revision as good/bad (`git bisect good`)
  * Run `git bisect run` which will mark revisions as good/bad for you.

Silex:
I think it makes more sense to mimic the natural git workflow... that is the first popup only offers "start" and then the 2nd popup offers reset/good/bad/run/visualize/etc

That said, maybe on the first popup there could be an option like "Start & Run" which would do what you suggested.

Silex:
See the examples section of http://git-scm.com/docs/git-bisect to see how to use `run`.

tarsius:
Since the discussion has move to #2294, I am closing this.

PR #2294, opened 2015-09-25
Add `magit-bisect-start-and-run`

Silex:
See #2290 

Quickly pulled together, but seems to work.

### DIFF
--- a/lisp/magit-bisect.el
+++ b/lisp/magit-bisect.el
@@ -50,11 +50,12 @@
   'magit-commands 'magit-popup-sequence-mode
   :man-page "git-bisect"
   :actions            '((?B "Start" magit-bisect-start)
-                        (?a "Run"   magit-bisect-run))
+                        (?R "Start & Run" magit-bisect-start-and-run))
   :sequence-actions   '((?r "Reset" magit-bisect-reset)
                         (?b "Bad"   magit-bisect-bad)
                         (?g "Good"  magit-bisect-good)
-                        (?k "Skip"  magit-bisect-skip))
+                        (?k "Skip"  magit-bisect-skip)
+                        (?a "Run"   magit-bisect-run))
   :sequence-predicate 'magit-bisect-in-progress-p)
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


Silex:
Btw, why not `s` for `skip`? Maybe as an alias?

Also, `a` for run is pretty bad... maybe `n`?

tarsius:
> Also, `a` for run is pretty bad...

I am guessing it stands for "automatically".

tarsius:
I think I would prefer to do it like this: https://github.com/magit/magit/compare/jb/bisect-run?expand=1. That's more in line with what I intended to do all along.

Silex:
Hum, what if I already started a session? I think your commit just lacks the binding for `magit-bisect-run` in the 2nd popup and otherwise looks nice.

tarsius:
Ah so my original intention must have been to *only* offer `run` when starting to bisect. It's not how `git bisect` does it, but I feel it makes sense.

npostavs:
> Ah so my original intention must have been to *only* offer `run` when starting to bisect. It's not how `git bisect` does it, but I feel it makes sense.

I'd prefer to have `run` available after starting too. Sometimes I need a few interactive "practice" runs before I'm ready to let the script run by itself. And sometimes I need to a few interactive runs before I realize I should really be using a script. It would be nice if these scenarios didn't require restarting the bisect session.

tarsius:
Makes sense.

So what should the key bindings and the descriptions be?

npostavs:
I think @Silex's descriptions are fine, but `Start & Run` and `Run` should be on the same key. I don't have a strong preference for `R` vs `a`, I guess `R` is a bit more memorable.

tarsius:
We use `B` for the default action because `BB` is nicer than `Bb`. We do the same for other popups too. But for `run` we should use `Br` (or `Ba`), again like we do for other popups.

Silex:
I agree that `Start & Run` and `Run` should be on the same key. I think `BR` makes more sense because it's easier to type & remember, and it implies that it's an action a bit different (more "final"), just like `bB` is for branches. 

I looked at few popups and uppercase/lowercase consistency didn't seem to be such a strong rule (e.g `PQ` `TS` `bB`), but tbh I don't care really much, whatever bindings I'll be happy with as long as it's fixed :)

npostavs:
> but tbh I don't care really much, whatever bindings I'll be happy with as long as it's fixed :)

Yeah, it's not a thing one uses often enough to develop a muscle memory for.

Anyway, I pushed commits [jb/bisect](https://github.com/magit/magit/compare/jb/bisect-run?expand=1) to use `Ba` both before and during bisection. `Br` could work but then we have to decide what "Reset" should be.

Silex:
Looks good to me.

tarsius:
I have pushed my (improved) version to `master` and am closing this. If you disagree with the descriptions and keys I have picked we can still change them.

Silex:
Thanks!

PR #2299, opened 2015-09-27
Show default file in the interactive prompt for magit-am-apply-patches

phil-s:
read-file-name doesn't display the DEFAULT-FILENAME value, so we ought to generate a prompt which does so.


npostavs:
> read-file-name doesn't display the DEFAULT-FILENAME value

Wouldn't it make more sense to fix this in Emacs, rather than magit?

phil-s:
Perhaps. Emacs bug #12443 seems relevant.

tarsius:
Cherry-picked. Thanks.

Issue #2297, opened 2015-09-26
reflog shows the "wrong" dates


michael-heerdegen:
Hello,

I recently wondered why what I got with r l was not in chronological order.  Then I recognized that it actually was.  But unlike vanilla git reflog, Magit shows the dates of the referenced commits, instead of the dates of the reference update.

Not the end of the world, but not very useful and confusing.  Not so urgent anyway.

Thanks in advance,

Michael.

tarsius:
> But unlike vanilla git reflog, Magit shows the dates of the referenced commits, instead of the dates of the reference update.

`git reflog show REF` doesn't show any dates at all. As far as I can tell the only way to do that is to use `--format`. Which is what Magit does - it uses `--format="%h [%aN] %ct %gd %gs"`, i.e. `%ct`, the committer date. There's no "update reflog entry date" different from the committer date, the files in `.git/log/refs/` record the committer date I think.

michael-heerdegen:
Jonas Bernoulli <notifications@github.com> writes:

> git reflog show REF doesn't show any dates at all. As far as I can
> tell the only way to do that is to use --format. Which is what Magit
> does - it uses --format="%h [%aN] %ct %gd %gs", i.e. %ct, the
> committer date. There's no "update reflog entry date" different from
> the committer date, the files in .git/log/refs/ record the committer
> date I think.

As far as I learned yesterday - no.  Try "git reflog --date=relative".
git reflog does accept all args git log does (says the manpage).

Have a look at the logs in .git/logs.  There are timestamps, dunno what
format this is, but they look chronological.


Michael.

tarsius:
You can confirm that the "update the reflog entry date" is the same thing as the committer date using this:

```
$ git init .
Initialized empty Git repository in /tmp/testibus/.git/
$ git commit -m "first" --allow-empty
[master (root-commit) 596be23] first
$ git commit -m "second" --allow-empty
[master f70e657] second
$ GIT_COMMITTER_DATE="Thu Jan 1 00:00:00 2015 +0000" git commit -m "time travel" --allow-empty
[master 0082fb5] time travel
$ git reflog --date=relative          
0082fb5 HEAD@{9 months ago}: commit: time travel
f70e657 HEAD@{20 seconds ago}: commit: second
596be23 HEAD@{29 seconds ago}: commit (initial): first
```

[...]

PR #2300, opened 2015-09-28
Show update date in reflog margin, don't show author name

tarsius:
Suggested in #2297.


michael-heerdegen:
About not showing the author: git also logs the "author" of the reference changes, maybe consider showing that in the last column.

tarsius:
But isn't that always "me"? (Unless the computer and user account is shared with somebody else, *but* that other person does bother to change `user.name` before committing. I.e. highly unlikely, one would just use another user in such cases I would assume.)

michael-heerdegen:

> But isn't that always "me"?

I wonder, too.  Probably it is.

PR #2304, opened 2015-09-29
magit-mode-get-buffer: set dir in %less formats

npostavs:
Otherwise we might execute git commands in the wrong repository.

---

Or maybe we should just drop support for this now? It does seem fairly likely that I'm the only the one who uses it.


tarsius:
In #2282 I am removing support for %-less formats. But that isn't quite ready yet (i.e. it still removes support for *any* user defined format, which clearly isn't what we want.

But applying this change now would cause a lot of work when later rebasing #2282, so I would rather not merge it (it has been like this for a while and, as you sais, you might be the only one actually using a %-less format).

Issue #2309, opened 2015-10-04
Magit doesn't work well with git credential.helper


ghost:
Sometimes I use pubkeys, but other times https auth is needed.

https://help.github.com/articles/caching-your-github-password-in-git/

github has baked in support for an authentication cache, so you don't
have to enter your credentials at every push, if you'd like to avoid putting
them in the remote url.

From the command line, once you've activated the helper with 
```
git config --global credential.helper 'cache --timeout=3600'
```

You are prompted for credentials once, and subsequent pushes
work without entering them again.

In magit, I get asked for credentials every time even with this turned on.
The above sets a key in config so It doesn't look like missing envars are
the issue (unlike ssh-agent).

If it's relevent, I have SSH_ASKPASS="", which I believe is used as a 
pass prompt whatever the mechanism if set.

npostavs:
magit, OS, and git versions?

> The above sets a key in config so It doesn't look like missing envars are
> the issue (unlike ssh-agent).

On Windows, mismatching `HOME` values can cause config to be skipped.

> If it's relevent, I have SSH_ASKPASS="", which I believe is used as a
> pass prompt whatever the mechanism if set.

I think that's only if you use ssh, otherwise `GIT_ASKPASS` is relevant.

MikeSanMichel2:
linux, magit.20151003, git 2.4.3, emacs 24.5.1, 

>>    If it's relevent, I have SSH_ASKPASS="", which I believe is used as a
>>    pass prompt whatever the mechanism if set.

> I think that's only if you use ssh, otherwise GIT_ASKPASS is relevant.

`(setenv `SSH_ASKPASS` ""` stopped the GUI pass prompt from appearing when
pushing to an https remote, so you may not be right there. 
But likely GIT_ASKPASS has precedence in that case. It's just not set in my shell.

npostavs:
Works for me. 

What's do you get from evaluating `(magit-git-string "config" "credential.helper")` and is it different from running `git config credential.helper` in your shell where it works?

---

> [...] But likely GIT_ASKPASS has precedence in that case.

Ah right, it does fall back to SSH_ASKPASS, it's explained here: http://git-scm.com/docs/gitcredentials#_requesting_credentials

npostavs:
Okay after l seeing [this stackexchange question][1] I notice that entering the password from magit doesn't get cached, although magit does use the existing cached credentials when they have been entered from the shell.

[...]

PR #2311, opened 2015-10-04
Try to find the password before asking user

DamienCassou:
When git asks the user for a password, this patch first asks the auth-source library. If no password is found, prompt the user.

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -101,10 +101,30 @@ When this is nil, no sections are ever removed."
     "^\\(Enter \\)?[Pp]assword\\( for '\\(?99:.*\\)'\\)?: ?$"
     "^.*'s password: ?$"
     "^Yubikey for .*: ?$")
-  "List of regexps matching password prompts of Git and its subprocesses."
+  "List of regexps matching password prompts of Git and its subprocesses.
+Also see `magit-process-find-password-functions'."
   :package-version '(magit . "2.1.0")
   :group 'magit-process
   :type '(repeat (regexp)))
+
+(defcustom magit-process-find-password-functions nil
+  "List of functions to try in sequence to get a password.
+
+These functions may be called when git asks for a password, which
+is detected using `magit-process-password-prompt-regexps'.  They
+are called if and only if matching the prompt resulted in the
+value of the 99th submatch to be non-nil.  Therefore users can
+control for which prompts these functions should be called by
+putting the host name in the 99th submatch, or not.
+
+If the functions are called, then they are called in the order
+given, with the host name as only argument, until one of them
+returns non-nil.  If they are not called or none of them returns
+non-nil, then the password is read from the user instead."
+  :package-version '(magit . "2.3.0")
+  :group 'magit-process
+  :type 'hook
+  :options '(magit-process-password-auth-source))
 
 (defcustom magit-process-username-prompt-regexps
   '("^Username for '.*': ?$")
@@ -623,11 +643,11 @@ tracked in the current repository are reverted if
           string)
          "\n"))))))
 
-(defun magit-process-password-auth-source (host)
-  "Use `auth-source-search' to get a password for HOST.
+(defun magit-process-password-auth-source (key)
+  "Use `auth-source-search' to get a password.
 If found, return the password.  Otherwise, return nil."
   (require 'auth-source)
-  (let ((secret (plist-get (car (auth-source-search :max 1 :host host))
+  (let ((secret (plist-get (car (auth-source-search :max 1 :host key))
                            :secret)))
     (if (functionp secret)
         (funcall secret)
@@ -635,19 +655,18 @@ If found, return the password.  Otherwise, return nil."
 
 (defun magit-process-password-prompt (process string)
   "Find a password based on prompt STRING and send it to git.
-If STRING corresponds to a password prompt, get a suitable
-password with `magit-process-password-auth-source' and send it to the
-git process."
+First try the functions in `magit-process-find-password-functions'.
+If none of them returns a password, then read it from the user
+instead."
   (--when-let (magit-process-match-prompt
                magit-process-password-prompt-regexps string)
     (process-send-string
      process (magit-process-kill-on-abort process
-               (concat
-                ;; 99 is a match-group in magit-process-password-prompt-regexps
-                (or (--when-let (match-string 99 string)
-                      (magit-process-password-auth-source it))
-                    (read-passwd it))
-                "\n")))))
+               (concat (or (--when-let (match-string 99 string)
+                             (run-hook-with-args-until-success
+                              'magit-process-find-password-functions it))
+                           (read-passwd it))
+                       "\n")))))
 
 (defun magit-process-username-prompt (process string)
   "Forward username prompts to the user."
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -97,7 +97,8 @@ When this is nil, no sections are ever removed."
 
 (defcustom magit-process-password-prompt-regexps
   '("^\\(Enter \\)?[Pp]assphrase\\( for \\(RSA \\)?key '.*'\\)?: ?$"
-    "^\\(Enter \\)?[Pp]assword\\( for '.*'\\)?: ?$"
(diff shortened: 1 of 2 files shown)


tarsius:
Please fix indentation, don't change line wrapping of existing code, and make sure the compiler has nothing to complain about. Also I am wondering why there appears a literal regexp in here and whether it's really necessary/preferable to split this across three functions.

DamienCassou:
Here is a new version. First, I want to say I'm sorry about the bad quality of the previous commit. I was in a hurry and decided to use github's edit interface instead of properly cloning your repository locally. That was clearly a bad decision. Regarding your comments:

- the compiler has nothing to say about my code
- regarding the 2nd regexp match: `auth-source` depends on the notion of `:host`: a host must be given as input to get the corresponding password. But `magit-process-password-prompt-regexps` matches against more than host password requests: this variable matches for yubikey queries, passphrases... If you have a better idea, I'd be happy to implement it. Something I could do would be to split `magit-process-password-prompt-regexps` into several regular expressions.
- regarding the split across 3 functions: I like my decomposition because each function has a precise role. Each function can then be understood separately, patched separately, advised separately. But you are the maintainer so please tell me what you prefer and I will implement it.

npostavs:
> But magit-process-password-prompt-regexps matches against more than host password requests: this variable matches for yubikey queries, passphrases... If you have a better idea, I'd be happy to implement it.

We could require that the `magit-process-password-prompt-regexps` put host name (if any) in match group 99 (some arbitrarily large number to avoid a user-added regexp hitting it by accident).

```elisp
(let ((str "Enter password for 'foo.com':"))
  (when (string-match "^\\(Enter \\)?[Pp]assword\\( for '\\(?99:.*\\)'\\)?: ?$" str)
    (match-string 99 str))) ; => "foo.com"
```

tarsius:
I haven't looked very closely at the code the first time because of the formatting issues and just mentioned some things that popped into my head. After having looked at the new revision I have no issues with splitting the thing across multiple functions, but if you had not done that I would have been fine with it too. So no need to change that.

I'll have to look into what exactly `auth-source` does though, I wasn't aware of its existence.

npostavs:
Another possibility is to read netrc from git:
http://stackoverflow.com/a/6031266/319698
https://github.com/git/git/blob/master/contrib/credential/netrc/git-credential-netrc

DamienCassou:
@npostavs ok for me. Should `99` be stored in a constant to avoid magic numbers? @tarsius what's your opinion? I guess this would avoid duplicating part of the regexp.

@tarsius [`auth-source`](https://www.gnu.org/software/emacs/manual/html_mono/auth.html) is an Emacs library to find passwords. This is how Emacs reads the `.netrc` and `.authinfo` files. `auth-source` can also read the GNOME Keyring, the KDE Wallet, and a [password store](https://github.com/DamienCassou/auth-password-store).

@npostavs please don't limit magit to `.netrc` because `auth-source` already abstracts over it.

npostavs:
> please don't limit magit to .netrc because auth-source already abstracts over it.

`contrib/credential/netrc/git-credential-netrc` is not limited to `.netrc`:
```
  -f|--file AUTHFILE : specify netrc-style files.  Files with the .gpg extension
                       will be decrypted by GPG before parsing.  Multiple -f
                       arguments are OK.  They are processed in order, and the
                       first matching entry found is returned via the credential
                       helper protocol (see below).
                       When no -f option is given, .authinfo.gpg, .netrc.gpg,
                       .authinfo, and .netrc files in your home directory are used
                       in this order.
```

DamienCassou:
@npostavs sorry, but this is not enough. `auth-info` handles:

- the [Secret Storage API](http://www.freedesktop.org/wiki/Specifications/secret-storage-spec/) which is implemented by Gnome Keyring and KDE Wallet
- the macos keychain
- [password-store](http://www.passwordstore.org/) (through a [plugin](https://github.com/DamienCassou/auth-password-store))

npostavs:
Ah, well under https://github.com/git/git/tree/master/contrib/credential there is also `gnome-keyring` and `osxkeychain`. For KDE Wallet, https://blog.sleeplessbeastie.eu/2012/08/12/git-how-to-avoid-typing-your-password-repeatedly/ suggests setting `GIT_ASKPASS`. I can't find anything for password-store though.

tarsius:
> Should 99 be stored in a constant to avoid magic numbers?

It has to be hard-coded in one place anyway, the doc-string, so you might as well hard-code it in the other two (?) places also.

DamienCassou:
@npostavs @tarsius after looking at it more seriously, I think following @npostavs idea to use match groups will not work because `magit-process-match-prompt` uses `string-match` several times which erase previous `match-string` results:

```emacs
(defun magit-process-match-prompt (prompts string)
  (when (--any? (string-match it string) prompts)
    (let ((prompt (match-string 0 string)))
      (cond ((string-match ": $" prompt) prompt)
            ((string-match ":$"  prompt) (concat prompt " "))
            (t                           (concat prompt ": "))))))
```

DamienCassou:
I've just added a unit-test in case you want one.

npostavs:
> magit-process-match-prompt uses string-match several times which erase previous match-string results:

we can change the latter 2 `string-match` to `string-match-p` (or even just `string-suffix-p`), which should solve the problem.

npostavs:
I think `magit-process-match-prompt` should get a docstring/comment noting that it preserves the match-data from a successful match.

Maybe you should use `--when-let` to avoid saying `(match-string 99 string)` twice.

Otherwise looks good to me.

tarsius:
I had reservations about this, but considering #2309 that is fading.

I would still prefer if everyone just used public keys and got of my lawn, but that just won't happen. We maintainers have to often help users who cannot figure out how to configure credentials. Our usual approach was "configure Git then Magit can do it to", and if the follow up was "it works in Git but still not in Magit", then "you are on Windows - make 'git bash', ssh, and Magit agree on where $HOME is". @npostavs even wrote a [package](https://github.com/magit/ssh-agency) that prevents users from having to configure ssh themselves.

We also have a wiki page [Pushing with Magit from Windows](https://github.com/magit/magit/wiki/Pushing-with-Magit-from-Windows), but that isn't quite accurate anymore.

So I now think we should merge this, but it is even more important that we once more go over our instructions on the issue and document what doesn't work and why.

npostavs:
> We also have a wiki page Pushing with Magit from Windows, but that isn't quite accurate anymore.

I just updated it, the situation with Git 2.x is actually quite a bit better (actual error messages instead of hanging, and SSH_ASKPASS works)

DamienCassou:
@npostavs done. Anything else I can do before you merge it? I can rebase the commits and get either 1 commit or several semantic ones:

1) additions of tests to `match-prompt`
2) fix of `match-prompt` to preserve match groups + tests
3) addition of auth-source authentification + tests

Tell me what you want

DamienCassou:
Here it is.

tarsius:
I think I'll merge that tomorrow or so. @npostavs any objections?

npostavs:
> any objections?

Nope, all good from my end.

tarsius:
Do you intend to make changes? I just fetched this and was rather surprised that it had already been rebased onto changes I pushed less than an hour ago.

tarsius:
Oh, and one commit is fine, but if you prefer three, then go with that.

DamienCassou:
@tarsius I just wanted to help you integrate it :-). Here is one commit. I don't plan to do anything anymore on it (except if you ask me)

tarsius:
I made some minor changes. But then I also remembered just in time, that I wanted to make this optional. I have implemented that myself, sorry. But I would still appreciate your help with this, because I have not actually tested it (I still have to get around to learn to use `auth-source`). Thanks.

Any objections to not enabling this by default?

See https://github.com/magit/magit/compare/dc/auth-source?expand=1.

tarsius:
Ups wait a moment...

DamienCassou:
@tarsius I gave you write access to my `magit` fork if you need

tarsius:
Okay, so I probably should have at least run the tests once.... Done that now. Please have a look.

DamienCassou:
It seems to work fine. Thanks

Issue #2317, opened 2015-10-06
magit use hard-coded mount prefix


PierreTechoueyres:
if an user customised the mount prefix in his cygwin / MINGW install, he is unable to use last magit version.

In function magit-expand-git-file-name (lisp/magit-git.el) the moun prefix is hard coded with cygdrive.

here is an attempt to fix this (this work for me with /mnt prefix) :
Sorry I was unable to attach it with a file.
```text
[PATCH] * lisp/magit-git.el magit-cygwin-prefix
 magit-expand-git-file-name

Use `magit-cygwin-prefix' in `magit-expand-git-file-name' instead of
hard-coded cygdrive prefix.
---
 lisp/magit-git.el | 11 ++++++++++-
 1 file changed, 10 insertions(+), 1 deletion(-)

diff --git a/lisp/magit-git.el b/lisp/magit-git.el
index 22327f6..add2739 100644
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -509,11 +509,20 @@ range.  Otherwise, it can be any revision or range accepted by
         (setq pos (point)))
       status)))
 
+(defconst magit-cygwin-prefix
+  (with-temp-buffer
+    (call-process-shell-command "mount -p | tail -1 | cut -d ' ' -f1" nil t)
+    (forward-line -1)
+    (setq cygwin-prefix (buffer-substring-no-properties
+                         (+ 1 (point)) (line-end-position)))))
+
 (defun magit-expand-git-file-name (filename)
   (unless (file-name-absolute-p filename)
     (setq filename (expand-file-name filename)))
   (if (and (eq system-type 'windows-nt) ; together with cygwin git, see #1318
-           (string-match "^/\\(cygdrive/\\)?\\([a-z]\\)/\\(.*\\)" filename))
+           (string-match
+            (concat "^/\\(" magit-cygwin-prefix "/\\)?\\([a-z]\\)/\\(.*\\)")
+            filename))
       (concat (match-string 2 filename) ":/"
               (match-string 3 filename))
     filename))
-- 
2.5.2

```

tarsius:
Please look in https://cygwin.com/cygwin-ug-net.html for a command which can more reliably return the prefix.

PierreTechoueyres:
Sorry I don't understand what you expect.
The -p switch of mount return the prefix. This switch work with cygwin, MINGW, MSYS2.
May be add a test to check if system-type eq windows-nt before the call for building magit-cygwin-prefix ?

[...]

PR #2319, opened 2015-10-06
* lisp/magit-git.el magit-cygwin-prefix magit-expand-git-file-name

PierreTechoueyres:
```
Use `magit-cygwin-prefix' in `magit-expand-git-file-name' instead of hard-coded cygdrive prefix.
```
See issue #2317 for complement.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -509,11 +509,20 @@ range.  Otherwise, it can be any revision or range accepted by
         (setq pos (point)))
       status)))
 
+(defconst magit-cygwin-prefix
+  (with-temp-buffer
+    (call-process-shell-command "mount -p | tail -1 | cut -d ' ' -f1" nil t)
+    (forward-line -1)
+    (setq cygwin-prefix (buffer-substring-no-properties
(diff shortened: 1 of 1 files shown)


npostavs:
I think we're in danger of reimplementing cygwin-mount.el.

http://emacswiki.org/emacs/cygwin-mount.el

PierreTechoueyres:
Yes but actually magit-expand-git-file-name doesn't work if the prefix is different of cygdrive ...

npostavs:
> Yes but actually magit-expand-git-file-name doesn't work if the prefix is different of cygdrive ...

Right, so I'm suggesting instead of magit redoing all the mount point parsing, maybe we should suggest cygwin users install cygwin-mount.el and then we can call `cygwin-mount-convert-file-name` from `magit-expand-git-file-name`.

PierreTechoueyres:
Yes, but may be cygwin-mount is overkill for this isn't it ?

I think the patch could be improved by adding tests on system-type etc.
I've also open bug #2317 and tarsius addressed some other questions.

PierreTechoueyres:
See PR #2324.
Unable to update this one.

PR #2323, opened 2015-10-08
docs: U is now magit-unstage-all, not magit-reset-index

warner:
Commit 4c16736 changed the binding of "U" from `magit-reset-index` to `magit-unstage-all`, but the docs weren't updated. I *think* this fixes them.. I don't know texinfo very well, so please render this to .info and eyeball the results before landing in case I screwed up the formatting somehow.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -3060,7 +3060,7 @@ Explicitly enabling ~--interactive~ won't have an effect on the
 following commands as they always use that argument anyway, even if it
 is not enabled in the popup.
 
-- Key: r e, magit-rebase-interactive
+- Key: r i, magit-rebase-interactive
 
   Start an interactive rebase sequence.
 
@@ -3077,7 +3077,7 @@ is not enabled in the popup.
 
   Combine squash and fixup commits with their intended targets.
 
-- Key: r s, magit-rebase-edit-commit
+- Key: r e, magit-rebase-edit-commit
 
   Edit a single older commit using rebase.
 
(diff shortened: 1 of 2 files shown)


npostavs:
Actually, I think the source document is the .org one.

warner:
Oh, yeah. But the .texi is checked in? That's confusing.

I'll fix the PR to change the .org instead. Whoever lands this should squash the branch first.

tarsius:
@warner you can just edit the org files and I'll regenerate the texi files and amend that to your commit. Or you can use `ox-texinfo+` and export yourself. We need to track the texi files because Melpa requires them.

warner:
ok, take a look at that, squash it all, then amend the .texi change. Thanks!

warner:
BTW, would you be interested in a patch that adds some "how to do common tasks" prose docs? I've been studying the new (magit-2.x) rebase commands this week, trying to unlearn the old 1.x "rewrite" ones, and it might be a good time to write down what I've learned.

* how to split an old commit into two separate pieces
* how to combine two old commits into a single one
* how to add a file to an old commit
* how to change the comments on an old commit

Stuff that's obvious if you've been using `git rebase -i` for a while, but not so obvious to the rest of us :).

tarsius:
Actually some of that is already covered in https://github.com/magit/magit/wiki/History-Manipulation.

tarsius:
Contributions are very welcome, though!

Not sure whether we should add such tutorials to the manual though. Maybe we should.

tarsius:
I have squashed the changes and regenerated the magit.texi. Thanks!

PR #2331, opened 2015-10-12
.travis.yml: never send email on success

npostavs:
```
The default is to send on change, but this includes the 1st success of a
new branch (i.e. a change from nothing -> success).  Generally success
is not surprising/interesting anyway.
```


tarsius:
Picked.

Issue #2330, opened 2015-10-12
What is v2.x equiv of v1's "rewrite" feature?


MikeSanMichel2:
Note: *My OP was too ZOMG, rewritten for productive discussion.*

After much "spirited" discussion in #966, the "rewrite" feature of v1 was removed.
I depended on "rewrite" heavily for rewriting PRs before submission. Where the primary task is moving multiple hunks between multiple commits.

I'm aware of new and improved features of v2 and use them extensively, including the interactive rebase, cherry picking, amending, fixup, extend, etc' . They are excellent, but do not perform nearly as well when I need to rewrite a dozen already completed commits. I find myself going through this rewriting stage for most every PR I submit. It  sometimes feels like have a dozen balls in the air so having as little additional complexity introduced by the tool really matters.

Several alternative ways to do such rewriting with v2 have been suggested:

1. **Using interactive-rebase instead**: again, great for commit-level changes, but not for moving hunks between commits.
2. **Use reset-HEAD** and reconstruct the commits from the working directory: - forces you to reenter all commit messages, and restage every commit from scratch, only now the working directory contains the aggregate of *all* the changes you've made in all the commits. This is laborious and error prone.
3. **The [wiki]((https://github.com/magit/magit/wiki/History-Manipulation))** describes a flow for moving hunks between commits. Which involves two reverses (a feature I haven't internalized yet),  a jump to a revision buffer and back, per commit. This seems like a **lot** more work than what the "rewrite" mechanism required.

The advantages of the "rewrite" feature as I see it were:

1. Gave you control over **individual hunks** by reusing the familiar stage/unstage mechanism
in the status view.
2. Allowed you to perform the whole thing (which can be complicated) in a **single screen** (status buffer) without switching views.
3. Allowed you to progress incrementally, with the **current state always visible**, showing you a fully tab-expandable history of every commit you've done so far, and what remains to be done.

I realize I'm very late to raise an objection (I'm returning to emacs from a long hiatus, partly because I missed magit), but this was my favorite part of magit and I miss it dearly. Is there something in v2 which provides the same convenience, or did perhaps the suggestion that the "rewrite" feature would be broken out into a new package materialize into something I can `package-install`?
 
"[Rewriting](http://magit.vc/manual/1.4/magit/Rewriting.html#Rewriting)" - From the 1.4 manual.

MikeSanMichel2:
[...]

Issue #2296, opened 2015-09-26
magit-reset-index resets HEAD


michael-heerdegen:
Hello,

my magit version is the current Melpa one.

Whenever I use U on a commit (magit-reset-index), the HEAD points to the same commit afterwards (which it should not).  If I hit U on a wip index backup, the HEAD even points to the wip branch!  I have to use the reflog to restore the previous state.

Magit uses

    git reset <commit> --

which should not change the head according to its doc, but it does.  Is this even a bug in git reset, or am I missing something?

Thanks,

Michael.

michael-heerdegen:
BTW, what works for me (as I expect) is

     reset && read-tree <commit>

kyleam:
The docstring for magit-reset-index is wrong (except for the special but
frequent case when commit is HEAD).  'git reset <commit> --' falls under
the third form described in its man page, which involves HEAD being
moved to the commit.

michael-heerdegen:
Kyle Meyer <notifications@github.com> writes:

> The docstring for magit-reset-index is wrong (except for the special but
> frequent case when commit is HEAD).

Mmh, ok, though, the functionality described in docstring seemed useful
to me.  E.g to restore a wip save of a previous index.  And it seems the
name is wrong as well?  I wonder which behavior was originally intended.

> 'git reset <commit> --' falls under the third form described in its
> man page, which involves HEAD being moved to the commit.

Ahh - because no paths are given?  I don't like these man page pseudo
grammers...

Thanks, Michael.

kyleam:
> Mmh, ok, though, the functionality described in docstring seemed useful
> to me.  E.g to restore a wip save of a previous index.  And it seems the
> name is wrong as well?  I wonder which behavior was originally intended.

I suspect you're right that it describes the intended behavior,
especially since magit-reset-head currently does the same thing as
magit-reset-index.  Perhaps a "." was intended for the path.

>> 'git reset <commit> --' falls under the third form described in its
>> man page, which involves HEAD being moved to the commit.
>
> Ahh - because no paths are given?  I don't like these man page pseudo
> grammers...

Yep.  In the first form, paths aren't optional, and, in the second,
paths are optional, but --patch or -p must be given.

tarsius:
> I wonder which behavior was originally intended.

Me too :-)

tarsius:
Will have to look into that.

tarsius:
I'll have to look into the past, but since I am currently working on pickaxe support, I'll delay that until I am done with that.

[...]

PR #2332, opened 2015-10-12
Reverse in index

npostavs:
As discussed in #2330, the workflow for splitting commits requires a double reversal that can be done in a single step, which is quicker and simpler. The main difficulty is how to name and bind the command.

Initially I thought of putting it in the revert popup.
Another idea is to bind to `u` for committed changes (instead of user-error "can't unstage committed changes"). In patch here I've co-opted the unstage command, I couldn't quite see how to change the binding, would we need an unstage-thing command and then remapping in the section keymaps (but committed hunks are the same section type as uncommitted ones, so we migh have a problem there)?

For naming I started with `magit-reverse-index`. I also thought of `magit-anti-stage` (too cryptic? although @MikeSanMichel2 did independently come up with "Stage Anti-Hunk" as a label).


npostavs:
Continued from #2330.

> I can't think of a scenario where my next step wouldn't be "C e".

You might want `C a` to change the commit message to reflect the changed you just removed. Also, if the change you want to split out is spread over a few hunks it's nicer if you're not forced to commit the reversal of each one.

> "Revert Hunk"/"Stage Hunk for Removal"/"Stage Anti-Hunk"?

We already have "Revert", which is slightly different. I did think of "Anti Stage" which is a bit like your "Stage Anti-Hunk" (note that this command is not confined to whole hunks, it's it's just a modified call to `magit-reverse` so it works on files and regions (parts of hunks) too).

MikeSanMichel2:
I think `magit-reverse` should definitely be represented in `magit-revert-popup`, as a command (or several).

From just a few hours of playing with this, it seems very useful to have at least a variant of the command that "C e" immediately. You can always use the smart commit commands
to perform extra steps in the familiar way. One advantage of bundling the "C e" into the action is that it makes for a simpler concept. It is simply "Uncommit".  That's just my (strong) preference
based on how I work, but It simply be an option with whatever default you prefer (Just as long as I can override with what **I** prefer).

Also, I understand (now) that `magit-reverse` isn't limited to hunks, I just think that's the coolest part about it  (except perhaps uncommitting range selections). Today, after reading the docs again and what the wiki had to say about it (utterly incomprehensible) I feel this is a kickass feature that most people will have little chance of discovering with going to great lengths.
That's why a name like "Uncommit hunk" seems so attractive. But, I admit it's not wise to limit a general command to a specific case in the popup for just that reason. Still, I think something like
that would really benefit users, maybe you have a better idea how to go about this?

npostavs:
I added `magit-uncommit-extend` which does anti-stage and then commit-extend, bound to `V e`. It occurred to me that the other revert commands also commit directly, so bundling the commit action makes it closer to them.

> Otherwise, explaining how the commit really isn't affected, but you now have an electron in your working directory and its positron in the index, and that you need to "C e", to annihalate with part of the cherry-picked commit, so it can become the new, modified, cherry-picked commit? I mean, that's just pedagogically cruel.

But also hilarious :D

MikeSanMichel2:
Cool. I think it only makes sense to have duals (turtles?) all the way down:
```
edit file
    stage change
        commit change
        uncommit change  <--- only this is/was missing, and it deserves to be a first-class operation
    unstage change
undo edit
```

tarsius:
I haven't read up on all the above yet, but let me start by saying that I think this should be an action, not a switch. As far as I can tell so far, you seem to agree and the switch-idea was... well just an idea.

MikeSanMichel2:
Glad to hear it. I'm still in favor of "Uncommit (foo)".
`commit` moves things from index to history, so moving something from history to index is logically an `Uncommit`. That's what "extract and extend" does, without the implementation detail of the anti-hunk  and amend step.

tarsius:
> Initially I thought of putting it in the revert popup. Another idea is to bind to `u` for committed changes (instead of user-error "can't unstage committed changes").

This is only my tentative opinion: Adding to the popup seems like a good idea, but we should probably *also* provide a top-level binding. I don't like the idea of overloading unstaging, I would like to keep the "apply variants" pure. Users are already having problems telling some of the variants apart in their pure form - overloading is only going to make matters worse. In the past "reverse" did fallback to "discard" in places where "reverse" did not make sense. I removed that because "reverse" and "discard" are very different actions and while "reverse" is harmless, "discard" it quite dangerous. We still get complains about the removal of that fallback. I don't want something similar to happen with "unstage" and "'anti-stage'".

So maybe we could use `K` or something.

npostavs:
> commit moves things from index to history, so moving something from history to index is logically an Uncommit

`commit` doesn't remove things from the index though, whereas your proposed Uncommit remove things from the history so it's not quite symmetric (not necessarily a problem with that though).

> I don't like the idea of overloading unstaging, I would like to keep the "apply variants" pure

Actually I don't really like it either, I wanted to overload the *key* `u`, but that requires splitting the "hunk" section type into "committed hunk" and "uncommitted hunk". Even overloading the key could cause confusion though.

> So maybe we could use K or something.

`K` is `magit-file-untrack` (though that command would not be used often anyway, does it even deserve a 1 key binding?).
Unused bindings: `C` (unCommit?), `H`, `N`, `O`, `Q`.

npostavs:
> uncommit change <--- only this is/was missing, and it deserves to be a first-class operation

There are many more possible operations than in your diagram. I was attempting to enumerate them, when I noticed that your uncommit is actually very similar to mixed reset, the only difference being that uncommit works on hunks, whereas reset is restricted to whole commits. Perhaps this can inform our decision with respect to binding and naming. (it made me notice I missed `X` (and a few others) in my list of free bindings (which I've edited now)).

MikeSanMichel2:
>  > commit moves things from index to history, so moving something from history to 
> > index is logically  an Uncommit
>
> commit doesn't remove things from the index though, whereas your proposed Uncommit remove 
> things from the history so it's not quite symmetric (not necessarily a problem with that though).

commit doesn't remove anything from the *working directory*, but it sure enough clears the index. 
You are right in that what I'm calling "Uncommit" should also stage the extracted hunk after
the commit is extended with the anti-hunk (In the special case where magit-reverse was applied to a hunk and not a commit).

To be clear, let's spell "uncommit" out. Defining "state" as `working directory`+`index`+`history/log`:

```
touch foo
git add foo  <- let's call the state after this step S1
git commit -m "foo"
uncommit <- let's call the state after this step S2
```

The operation I'm calling Uncommit would `assert(S1==S2)`, when `magit-reverse` is performed
at the granularity of a whole commits. If a single hunk (From several) was uncommited, it dwim's.

I agree more operations are possible, but this feels fundamental to me. The rest are just
that + (some other steps).

npostavs:
> commit doesn't remove anything from the working directory, but it sure enough clears the index.

This may be an overly implementation centric perspective, but the index is not cleared: index stays the same, while it's changes are added to HEAD so that the *difference* between the index and HEAD becomes nil.

> To be clear, let's spell "uncommit" out.

Right, I was slightly confused in my previous post, it's like soft reset, not mixed.

MikeSanMichel2:
Right. we mean the same thing and you're technially correct. The outcome nonetheless is that index contains no pending (staged) changes.

npostavs:
Okay, so we have "uncommit" which is like a partial soft reset to HEAD, and "reverse in index" (aka "anti-stage") which is like a partial `(magit-reset-index HEAD)`.

How about `magit-reset-{soft,index}-dwim` commands which do the partial variants when on a committed change, and ask for a commit (as a reset target) otherwise? `magit-reset-soft-dwim` could be sensibly bound to `X`. We could leave `magit-reset-index-dwim` unbound, like `magit-reset-index`:

https://github.com/magit/magit/issues/2296#issuecomment-143867557

> The default should be what most users would expect, they shouldn't have to change a key binding to get that, and advanced users can get the advanced variant by changing a key binding.

MikeSanMichel2:
Well, I've been using git for 5 years? a decade? and I've never thought of `reset --soft` as `uncommit` before.  Much easier to grok, but I suppose remaining true to git is more important given how
many concepts it introduces by itself. Forking the names is a bad long-term strategy.

npostavs:
We could provide `uncommit` by itself, and bind that to `X`. `magit-reset-soft` would remain unbound, as it is now, so nothing is lost. The dwim type can be tricky, maybe they're better off avoided.

`reset --soft` isn't *exactly* the same as `uncommit` in all cases, but they do share some common ground.

MikeSanMichel2:
You need to be conservative about rechristening things when there are so many, but this is about the revert popup and the reset commands don't live here. So I guess this is legitimate. the user doesn't need to know how it is implemented or what it's equivalent to.

If this was a `reset` popup, *then* calling `--soft` by any another name would be a no-no.

npostavs:
> No, seriously, why isn't rebase continue r c?

`r r` is easier to type.

Silex:
>> No, seriously, why isn't rebase continue r c?

> r r is easier to type.

... and it follows the "most common operation" logic, like `c c`, `F F`, `P P` and all the others.

MikeSanMichel2:
> >   No, seriously, why isn't rebase continue r c?
> r r is easier to type.

so why isn't skip `r t`?

Silex:
> so why isn't skip r t?

Because that is not easy to type.

MikeSanMichel2:
okok, makes no sense but not what this issue is about.

Silex:
> The c in c c stands for Commit. The P in push stands for Push. I have no idea what F does.

So you never `pull`?

> okok, makes no sense but not what this issue is about.

Well, you're the one over-focusing on these keybindings :) They make perfect sense when you consider classic workflows.

MikeSanMichel2:
There aren't many pieces of software I care anough to argue about keystorkes. magit is pretty special.
Believe me, I can argue this with you for a long time, but let's not kill this thread. erasing in 5...4..

npostavs:
> the user doesn't need to know how it is implemented or what it's equivalent to.

Hmm the current implementation would also commit any unstaged changes.

tarsius:
I am not so sure the proposed commands are a good idea - they could be rather confusing. Feel free to keep experimenting if you disagree and I might change my mind when I see the end result.

But we should now concentrate on the remaining bugs, such as #2309. I would like to release v2.3.0 soon and hope that we can fix all known bugs before then.

MikeSanMichel2:
@Silex, I removed my offtopic comments would you mind doing the same for your resopnses? I don't  to hijack @npostavs's thread.

MikeSanMichel2:
@silex, since you won't remove your comments I might as well repond.

> >    The c in c c stands for Commit. The P in push stands for Push. I have no idea what F does.

> So you never pull?

That's right, not with magit. I somehow naturally always switch to shell when I want to pull.  It's mostly unconcious (I haven't actually noticed that I don't use magit for that), but I realize now this is because of #1485,  and why I always specify the remote I'm pulling from.

> >    okok, makes no sense but not what this issue is about.

> Well, you're the one over-focusing on these keybindings :) They make perfect sense when you 
> consider classic workflows (ease of continuing with the most common scenario).

That's a matter of taste. All other keybindings generally match the first letter of their operation and I find it easier to think of "(r)ebase (c)ontinue" as "r c" just as I think of "(r)ebase (a)bort" as "r a". But since magit lets you customize it however you want (I now have), this isn't a problem.

tarsius:
So this doesn't appear to be in popular demand.

npostavs:
I do still intend to implement this at some point, IMO just looking at [History-Manipulation][h] it's clearly needed.

[h]: https://github.com/magit/magit/wiki/History-Manipulation

tarsius:
Alright.

tarsius:
For the time being I have added these commands to the `magit-rockstar` package.

PR #2335, opened 2015-10-13
WIP Log washing fixes

kyleam:
I've run into a few issues with log washing.

* Calling log with --stat and without --graph deletes the first
  character in the hash of all entries after the first.

* Diff washing is log is not restricted to an entry.  To see this, run
  log with --stat and -h but not --graph.

*  Merge commits, which do not lead to magit-diff-wash-diffstat being
   called, are not separated from the next commit by a space.

These commits should fix the first two, but they are only lightly
tested at the moment.  An easy fix for the third issue hasn't jumped
out at me, but it is minor compared to the other two.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -1268,10 +1268,10 @@ Staging and applying changes is documented in info node
           "\\(contains \\(?:modified\\|untracked\\) content\\)\\|"
           "\\([^ :]+\\)\\( (rewind)\\)?:\\)$"))
 
-(defun magit-diff-wash-diffs (args)
+(defun magit-diff-wash-diffs (args &optional limit)
   (when (member "--stat" args)
     (magit-diff-wash-diffstat))
-  (when (re-search-forward magit-diff-headline-re nil t)
+  (when (re-search-forward magit-diff-headline-re limit t)
     (goto-char (line-beginning-position))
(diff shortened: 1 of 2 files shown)


tarsius:
I have tested and applied these commits. Thanks.

> Merge commits, which do not lead to magit-diff-wash-diffstat being called, are not separated from the next commit by a space.

What arguments do I have to use to reproduce that?

kyleam:
> What arguments do I have to use to reproduce that?

Sorry, meant to include that above.  I see it in the logs made with --stat but without --graph.

tarsius:
I suppose we could add a section body which consists of nothing but an empty line, but I don't think it is worth the trouble.

tarsius:
If you find an easy way to add that extra line, then please do. But it might not be as easy as it sounds - not without an *additional* empty line being added in those other cases which you forgot to test :-)

kyleam:
I agree that it's not worth the trouble.

The reason I said "An easy fix for the third issue hasn't jumped out at
me ..."  was because I didn't see a trivial way to insert a space
without messing up the other cases, so I did enough tests to realize
that :)

tarsius:
I am pretty sure that in one of my last commits that touched this function I said that "it is a monstrosity" :-)

That will be fixed eventually, but there's no hurry.

PR #2324, opened 2015-10-09
magit use hard-coded mount prefix

PierreTechoueyres:
* lisp/magit-git.el magit-cygwin-prefix magit-expand-git-file-name

Use `magit-cygwin-prefix` in `magit-expand-git-file-name` instead of hard-coded cygdrive prefix.
Use `defcustom` for `magit-cygwin-prefix` and add extra checks for default value.
Set return value to `nil` when the system isn't `windows-nt`.

PR #2348, opened 2015-10-18
handle custom cygwin mounts

### LINKED ISSUE #2317: magit use hard-coded mount prefix

npostavs:
```
Rely on the output from mount to tell us how to map cygwin paths into
native Windows paths.

This is also needed in git-commit; avoid a hard dependency on magit by
only loading magit-git in the case where we receive a bad path to
foo_EDITMSG (and falling back to the standard "/cygdrive" prefix if we
fail to load).
```
---
fix #2317
replace/ close #2324

### DIFF
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -358,15 +358,19 @@ usually honor this wish and return non-nil."
 (defun git-commit-setup ()
   ;; cygwin git will pass a cygwin path (/cygdrive/c/foo/.git/...),
   ;; try to handle this in window-nt Emacs.
-  (when (and (eq system-type 'windows-nt)
-             (not (file-accessible-directory-p default-directory))
-             ;; Emacs prepends a "c:".
-             (string-match "\\`[a-z]:/\\(cygdrive/\\)?\\([a-z]\\)/\\(.*\\)"
-                           buffer-file-name))
-    (let ((w32name (concat (match-string 2 buffer-file-name) ":/"
-                           (match-string 3 buffer-file-name))))
-      (when (file-accessible-directory-p (file-name-directory w32name))
-        (find-alternate-file w32name))))
+  (--when-let
+      (and (eq system-type 'windows-nt)
+           (not (file-accessible-directory-p default-directory))
+           (if (require 'magit-git nil t)
+               ;; Emacs prepends a "c:".
+               (magit-expand-git-file-name (substring buffer-file-name 2))
+             ;; Fallback if we can't load `magit-git'.
+             (and (string-match "\\`[a-z]:/\\(cygdrive/\\)?\\([a-z]\\)/\\(.*\\)"
+                                buffer-file-name)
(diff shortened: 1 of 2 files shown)


PierreTechoueyres:
Maybe you should have a look at `magit-clone` too. I've an spurious directory `d` created when cloning to my `~/projets` (which expand to `d:/home/ptechoueyres/projets`) directory.
I guess the directory path should be converted to an cygwin / msys path with a mount point.

tarsius:
Thanks a lot!

npostavs:
> Maybe you should have a look at magit-clone too. I've an spurious directory d created when cloning to my ~/projets (which expand to d:/home/ptechoueyres/projets) directory.
> I guess the directory path should be converted to an cygwin / msys path with a mount point.

I think cygwin programs should be able to take windows paths, so I don't understand why there should be a problem here.

PierreTechoueyres:
> I think cygwin programs should be able to take windows paths, so I don't understand why there should be a problem here.

Yes. You're right.
After some investigations, the clone part work flawlessly. I must investigate more to find what goes wrong after the `git clone` part.
Thanks for your patience and for this patch.

PR #2349, opened 2015-10-19
Use a slightly lighter grey on light background for sections and highlighting

NicolasPetton:
This PR changes the background color of the section and context highlights to a slightly lighter grey on light backgrounds, giving a more contrast with the content of the section.

screenshot from 2015-10-19 12-36-41

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -407,7 +407,7 @@ and https://debbugs.gnu.org/cgi/bugreport.cgi?bug=7847."
 
 (defface magit-diff-context-highlight
   '((((class color) (background light))
-     :background "grey85"
+     :background "grey95"
      :foreground "grey50")
     (((class color) (background dark))
      :background "grey20"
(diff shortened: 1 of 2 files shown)


tarsius:
I have cherry picked your commit.

I originally used the darker grays, among other things, because I defined the "light 'default' theme" as something like:

```lisp
(set-face-attribute 'default nil :background "gray90")
(set-face-attribute 'default nil :foreground "gray15"))
```

But (a) that's not actually the default and (b) it's also dead ugly. When the default background is "gray90" then I feel using "gray95" for highlighting is quite okay. On the other hand if I didn't use a theme and instead just tweaked the "light default", then I would probably use "gray95" as the background of default instead of "white", i.e. the same as is now used for highlighting. That's why I originally picked a darker gray for highlighting: so that it would "just work" for users who like me who don't want to stare at "#FFFFFF" all day.

So I am a bit worried that this change might cause some protests.

NicolasPetton:
Thanks for taking my changes!

I understand your concern, but I think it improves how Magit looks with vanilla Emacs, and otherwise themes override it anyway.

PR #2360, opened 2015-10-24
cache daemon hack

npostavs:
Work around for #2309

---

I think this is basically ready to merge now. Note that `magit-credential-hook` is no longer a defcustom; the `--socket` argument needs to be customizable anyway, and this neatly avoids the `add/remove-hook` vs `defcustom` problem too.

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -88,6 +88,41 @@ When this is nil, no sections are ever removed."
   :group 'magit-process
   :type '(choice (const :tag "Never remove old sections" nil) integer))
 
+(defcustom magit-credential-cache-daemon-socket
+  (--some (-let [(prog . args) (split-string it)]
+            (if (string-match-p
+                 "\\`\\(?:\\(?:/.*/\\)?git-credential-\\)?cache\\'" prog)
+                (or (cl-loop for (opt val) on args
+                             if (string= opt "--socket")
+                             return val)
+                    (expand-file-name "~/.git-credential-cache/socket"))))
+          ;; Note: `magit-process-file' is not yet defined when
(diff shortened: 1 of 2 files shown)


tarsius:
I rebased and added some cleanup. Can you agree to these changes?

> Some Git implementations (e.g. Windows) won't have this program; if we fail the first time, stop trying.

So we probably shouldn't try even just once on "e.g." Windows.

npostavs:
> I rebased and added some cleanup. Can you agree to these changes?

Yes, looks good. I just added a comment about the `process-lines` thing (which I should have done before to save you from learning about it the hard way, sorry about that).

> So we probably shouldn't try even just once on "e.g." Windows.

I was thinking it's similar to how feature tests are preferable over explicit version checks: here we test for failure rather than explicitly checking the OS.

tarsius:
> oh and the socket var check in magit-maybe-start-credential-cache-daemon was backwards

But wouldn't `not` make more sense than `null`? Personally I just always use `not`.

> I was thinking it's similar to how feature tests are preferable over explicit version checks: here we test for failure rather than explicitly checking the OS.

Makes sense.

Could you please finish up for merge. One commit would be fine with me but if you want more then do that.

npostavs:
> But wouldn't not make more sense than null? Personally I just always use not.

I had this idea to only use `not` for boolean variables. But maybe that doesn't make sense in elisp. Whatever, I changed to use `not`.

> Could you please finish up for merge. One commit would be fine with me but if you want more then do that.

I made 1 for the hook, and 1 for the rest. Which reminds me of https://github.com/magit/ssh-agency/issues/3#issue-93218666 Why do you think it's a bad idea to check for git argument "push", "pull", or "fetch" instead of going by magit function?

PR #2363, opened 2015-10-25
magit-toggle-buffer-lock: fix locked diff name

npostavs:
```
Before:

   *magit-diff(72 69 65 68 94 --cached): magit

After:

   *magit-diff(HEAD^ --cached): magit
```

I'm not entirely sure whether I've got this right, because it's a bit unclear to me what the format of `magit-refresh-args` is supposed to be.

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -664,8 +664,8 @@ latter is displayed in its place."
                                     magit-stashes-mode))
                  (car magit-refresh-args))
                 ((eq major-mode 'magit-diff-mode)
-                 (append (or (car magit-refresh-args) '(unstaged))
-                         (cadr magit-refresh-args)))))
+                 (-let [(from to . _) (-flatten magit-refresh-args)]
+                   (list (or from 'unstaged) to)))))
     (if magit-buffer-locked-p
         (rename-buffer (funcall magit-generate-buffer-name-function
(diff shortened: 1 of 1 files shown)


kyleam:
Here's the description of diff's magit-refresh-args from e82fd3c:

    Now the buffer-local value of `magit-refresh-args' in diff-derived
    buffers has the form (REVS CONSTS ARGS FILES).  CONSTS are the
    command-specific arguments.  They can only be changes using a diff
    command, but not from the refresh popup.

I think all that is needed to make the original code work as intended
is to s/append/cons/.  However, this still isn't quite right because a
(car magit-refresh-args) of nil can be a diff for either unstaged or staged
changes, depending on whether a "--cached" follows.

tarsius:
I think a566492efeb2f524cda22795fa1d87aff4f2bdb3 should do the trick.

npostavs:
> Here's the description of diff's magit-refresh-args from e82fd3c:

Great, can we put that in a docstring or comment?

> I think a566492 should do the trick.

One more corner case: locking a diff buffer made by `d p` (`magit-diff-paths`), shows `*magit-diff(unstaged): magit`.

tarsius:
If added a doc-string to `magit-diff-refresh-buffer` and fixed the `--no-index` case.

PR #2365, opened 2015-10-25
magit-git-command: drop --literal-pathspecs

### LINKED ISSUE #2354: Executing aliased git commands that start with '!'
I have a git alias called "fixup"

    $ git config alias.fixup
    !git add -uv && git commit --amend -CHEAD

Magit supports executing arbitrary git commands with `magit-git-command` (bound to the key `:`). Trying that on my `fixup` macro results in the following output in the `*magit-process: ...`  buffer:

    128 git … fixup
    fatal: pathspec ':/' did not match any files

`M-x magit` version returns:

    Magit 20151019.443, Git 2.4.4, Emacs 24.5.1

Maybe there is even another shortcut in Magit that achieves something  similar like my `fixup` alias?

npostavs:
As a side effect, this fixes #2354.

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -2169,7 +2169,11 @@ repository, otherwise in `default-directory'."
     (setq args (mapcar 'eval (eshell-parse-arguments (point-min)
                                                      (point-max))))
     (setq default-directory directory)
-    (magit-run-git-async args))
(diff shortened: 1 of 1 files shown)


tarsius:
> A human will want globbing by default.

:-)

PR #2366, opened 2015-10-28
fix magit-clone with cygwin git

### LINKED ISSUE #2357: magit-clone create extra dir on cygwin / mingw
`magit-clone` on cygwin create extra an directory named `c:` (or any letter mapped to the drive).

I think the directory used for the call of `magit-call-git` should be rewritten as a cygwin path.

I propose to add an `magit-shrink-git-file-name` for translating an Windows path to an cygwin one.
The change on `magit-expand-git-file-name` allow the translation of `~/`.  As `(file-name-absolute-p "~/")` is evaluated as `t` and so isn't translated.

``` patch
2 files changed, 10 insertions(+), 3 deletions(-)
lisp/magit-git.el    | 11 +++++++++--
lisp/magit-remote.el |  2 +-

modified   lisp/magit-git.el
@@ -534,14 +534,21 @@ Sorted from longest to shortest CYGWIN name."
   :type '(alist :key-type string :value-type directory))
 
 (defun magit-expand-git-file-name (filename)
-  (unless (file-name-absolute-p filename)
-    (setq filename (expand-file-name filename)))
+  (setq filename (expand-file-name filename))
   (-if-let ((cyg . win)
             (cl-assoc filename magit-cygwin-mount-points
                       :test (lambda (f cyg) (string-prefix-p cyg f))))
       (concat win (substring filename (length cyg)))
     filename))
 
+(defun magit-shrink-git-file-name (filename)
+  (setq filename (expand-file-name filename))
+  (-if-let ((cyg . win)
+            (cl-rassoc filename magit-cygwin-mount-points
+                       :test (lambda (f win) (string-prefix-p win f))))
+      (concat cyg (substring filename (length win)))
+    filename))
+
 (defun magit-decode-git-path (path)
   (if (eq (aref path 0) ?\")
       (string-as-multibyte (read path))
modified   lisp/magit-remote.el
@@ -47,7 +47,7 @@ Then show the status buffer for the new repository."
                        (match-string 1 url))))))))
   (make-directory directory t)
   (message "Cloning %s..." repository)
-  (when (= (magit-call-git "clone" repository directory) 0)
+  (when (= (magit-call-git "clone" repository (magit-shrink-git-file-name directory)) 0)
     (message "Cloning %s...done" repository)
     (magit-status-internal directory)))
```

npostavs:
Based on the [strace output][1] it looks like git is already making the needed directories so the `make-directory` call in magit is redundant. Can't find it in the release notes, but https://github.com/git/git/commit/8e21d63b02f1b26f7695ca515e51e4622a995af2 seems to confirm this.

I decided it's a bit "safer" to use absolute paths, so I dropped the `file-relative-name` part from `magit-shrink-git-file-name` and called it `magit-convert-git-file-name` instead (by analogy with `convert-standard-filename`; should we call it `magit-convert-git-filename` to be consistent with Emacs, or keep `-file-name` to be consistent with `magit-expand-git-file-name`?).

Since `magit-clone`'s `interactive` form already calls `expand-file-name`, I think the `~` case should be covered. @PierreTechoueyres does this work ok for you?

Should fix #2357.

[1]: https://github.com/magit/magit/issues/2357#issuecomment-150979312

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -542,6 +542,13 @@ Sorted from longest to shortest CYGWIN name."
       (concat win (substring filename (length cyg)))
     filename))
 
+(defun magit-convert-git-file-name (filename)
+  (-if-let ((cyg . win)
+            (cl-rassoc filename magit-cygwin-mount-points
+                       :test (lambda (f win) (string-prefix-p win f))))
+      (concat cyg (substring filename (length win)))
+    filename))
+
 (defun magit-decode-git-path (path)
   (if (eq (aref path 0) ?\")
       (string-as-multibyte (read path))
--- a/lisp/magit-remote.el
+++ b/lisp/magit-remote.el
@@ -45,7 +45,6 @@ Then show the status buffer for the new repository."
                   "Clone to: " nil nil nil
                   (and (string-match "\\([^./]+\\)\\(\\.git\\)?$" url)
                        (match-string 1 url))))))))
-  (make-directory directory t)
   (message "Cloning %s..." repository)
   (when (= (magit-call-git "clone" repository directory) 0)
     (message "Cloning %s...done" repository)
--- a/lisp/magit-remote.el
+++ b/lisp/magit-remote.el
@@ -46,7 +46,9 @@ Then show the status buffer for the new repository."
                   (and (string-match "\\([^./]+\\)\\(\\.git\\)?$" url)
                        (match-string 1 url))))))))
   (message "Cloning %s..." repository)
-  (when (= (magit-call-git "clone" repository directory) 0)
+  (when (= (magit-call-git "clone" repository
+                           ;; Stop cygwin git making a "c:" directory.
+                           (magit-convert-git-file-name directory)) 0)
     (message "Cloning %s...done" repository)
     (magit-status-internal directory)))
 


tarsius:
Thanks a lot!

Good commit message, I finally understand the issue. But probably I just didn't get it earlier due to a somewhat reduced attention span lately (thus the need for a break). :-)

I decided to go with `-filename`. Now both Magit functions are named consistently with regards to their approximate counterparts in Emacs. And like those functions the Magit functions are inconsistent named between each other. So the Magit functions are now named inconsistently, which is consistent with the inconsistency in Emacs. Given the inconsistent names in Emacs, I think that's to most consistency we can get out of this. :-)

PierreTechoueyres:
Sorry to be late.
Maybe you should be more "defensive" in the `magit-convert-git-filename` function by adding an`expand-file-name`. By this way , whatever the call we ensure filename will always be an absolute path.

FYI `w32-convert-standard-filename` has the same problem with cygwin prefix different than `/cygdrive`

my 0.002 €

PierreTechoueyres:
I think there is some other places where filename or directories should be converted with `magit-convert-git-filename`.
A grep on `magit-call-git` shows us some call with a file or directory argument :
```text
-*- mode: grep; default-directory: "~/projets/emacs-magit/lisp/" -*-
Grep started at Wed Oct 28 14:13:36

grep --color=always -niH magit-call-git *el
magit.el:1065:          (magit-call-git "update-index" "--cacheinfo"
magit.el:1127:  (magit-call-git "init" (expand-file-name directory))
magit.el:1217:        (magit-call-git "checkout" args "-b" branch current)
magit.el:1328:            (`detach (magit-call-git "checkout" "--detach"))
magit.el:1329:            (`master (magit-call-git "checkout" "master"))
magit.el:1502:         (magit-call-git "checkout" arg "--" file)
magit.el:1739:    (magit-call-git "tag" "-d" tags))
magit.el:1866:    (magit-call-git "config" "--add" global "notes.displayRef" ref))
magit-apply.el:401:        (magit-call-git "reset"  "--" files)
magit-apply.el:402:      (magit-call-git "checkout" "--" files))))
magit-apply.el:415:                (magit-call-git "rm" "--cached" "--" file))
magit-apply.el:422:                (magit-call-git "rm" "--cached" "--force" "--" file))
magit-apply.el:423:            (?D (magit-call-git "checkout" "--" file)
magit-apply.el:425:                (magit-call-git "rm" "--cached" "--force" "--" file))))))))
magit-apply.el:436:            (magit-call-git "mv" file orig)
magit-apply.el:437:          (magit-call-git "rm" "--cached" "--" file)
magit-apply.el:438:          (magit-call-git "reset" "--" orig))))))
magit-apply.el:446:          (magit-call-git "checkout" "--" files)
magit-apply.el:448:          (magit-call-git "add"   "--" new-files)
magit-apply.el:449:          (magit-call-git "reset" "--" new-files))
magit-apply.el:462:              (magit-call-git "reset" "--" (cadr binaries)))
magit-process.el:253:    (magit-call-git args)))
magit-process.el:270:  (magit-call-git args)
magit-process.el:291:(defun magit-call-git (&rest args)
magit-remote.el:49:  (when (= (magit-call-git "clone" repository
magit-sequence.el:94:      (progn (magit-call-git "reset" "--hard")
magit-stash.el:149:  (if (= (magit-call-git "stash" "apply" "--index" stash) 0)
magit-stash.el:159:  (if (= (magit-call-git "stash" "apply" "--index" stash) 0)
magit-stash.el:173:    (magit-call-git "reflog" "delete" "--updateref" "--rewrite" stash))
magit-stash.el:226:                (magit-call-git "checkout" "--" ".")
magit-stash.el:227:              (magit-call-git "reset" "--hard" "HEAD"))
magit-stash.el:229:              (magit-call-git "clean" "-f" (and (eq untracked 'all) "-x")))))
magit-wip.el:190:                     (magit-call-git "add" "--" files)
magit-wip.el:192:                     (magit-call-git "add" "-u" ".")))
magit-wip.el:210:      (magit-call-git "update-ref" wipref "-m" start-msg
magit-wip.el:215:    (magit-call-git "update-ref" wipref "-m" msg

Grep finished (matches found) at Wed Oct 28 14:13:37
```

My 0.002 €

npostavs:
> So the Magit functions are now named inconsistently, which is consistent with the inconsistency in Emacs.

:)

> Maybe you should be more "defensive" in the magit-convert-git-filename function by adding an expand-file-name. By this way , whatever the call we ensure filename will always be an absolute path.

But that would make it inconsistent with `convert-standard-filename`.

> I think there is some other places where filename or directories should be converted with magit-convert-git-filename.
> A grep on magit-call-git shows us some call with a file or directory argument :

I think only the `(magit-call-git "init" (expand-file-name directory))` might need it.

tarsius:
Should we use

```diff
diff --git a/lisp/magit-remote.el b/lisp/magit-remote.el
index 8f5119a..7a9144b 100644
--- a/lisp/magit-remote.el
+++ b/lisp/magit-remote.el
@@ -48,7 +48,9 @@ (defun magit-clone (repository directory)
   (message "Cloning %s..." repository)
   (when (= (magit-call-git "clone" repository
                            ;; Stop cygwin git making a "c:" directory.
-                           (magit-convert-git-filename directory)) 0)
+                           (magit-convert-git-filename
+                            (expand-file-name directory)))
+           0)
     (message "Cloning %s...done" repository)
     (magit-status-internal directory)))
 
diff --git a/lisp/magit.el b/lisp/magit.el
index 96f557a..5e25150 100644
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -1124,7 +1124,8 @@ (defun magit-init (directory)
      (list directory)))
   (make-directory directory t)
   ;; `git init' does not understand the meaning of "~"!
-  (magit-call-git "init" (expand-file-name directory))
+  (magit-call-git "init" (magit-convert-git-filename
+                          (expand-file-name directory)))
   (magit-status-internal directory))
 
 ;;;; Branch
```

for the time being?

PierreTechoueyres:
> > Maybe you should be more "defensive" in the magit-convert-git-filename function by adding an expand-file-name. By this way , whatever the call we ensure filename will always be an absolute path.
>
> But that would make it inconsistent with convert-standard-filename.
OK. then the code posted by @tarsius seems the thing to do.

> I think only the (magit-call-git "init" (expand-file-name directory)) might need it.
And for `magit-apply.el:448:          (magit-call-git "add"   "--" new-files)`  ?
is `new-files` an relative path name ? (I can't find my way in calls)

npostavs:
I guess the `expand-file-name` in `magit-clone`'s `interactive` form should be removed since it's redundant now?

PierreTechoueyres:
I also think it could be removed  and maybe the call to `file-name-as-directory` because the call to `read-directory-name` already return an directory.
But that don't hurt, doesn't it ?

npostavs:
> maybe the call to file-name-as-directory because the call to read-directory-name already return an directory

No, `file-name-as-directory` adds a `/` (if there isn't one), and `read-directory-name` may return a the name without `/`. Actually, maybe we should move it out of the `interactive` form too.

PierreTechoueyres:
Does git really need an ending / in the directory path ?

npostavs:
> Does git really need an ending / in the directory path ?

No, Emacs does.
```elisp
(with-temp-buffer
  (list (let ((default-directory "/tmp"))
          (call-process "pwd" nil '(t t) nil)
          (delete-indentation) (delete-horizontal-space)
          (buffer-string))
        (let ((default-directory "/tmp/"))
          (erase-buffer)
          (call-process "pwd" nil '(t t) nil)
          (delete-indentation)(delete-horizontal-space)
          (buffer-string)))) ;=> ("/" "/tmp")
```

npostavs:
> I guess the expand-file-name in magit-clone's interactive form should be removed since it's redundant now?

I pushed to https://github.com/magit/magit/compare/np/cygwin-clone (was trying to see if github would let me reopen the pull request with a new branch of the same name, it doesn't)

PierreTechoueyres:
@npostavs OK. I see now what I've missed. Thanks for the explanations.
Did you have time to check the call `(magit-call-git "add" "--" new-files)` ?

npostavs:
> Did you have time to check the call (magit-call-git "add" "--" new-files) ?

`git add` doesn't make any directories, so it shouldn't hit this problem (and I'm pretty sure that any repository files would be referenced relative to the repository root).

Issue #2364, opened 2015-10-25
.git file not recognized?


kccai:
I am using the latest magit version in melpa 
```
    Archive: melpa
    Version: 20151024.220
```

I have a .git file in my working directory which is pointing to a .git folder in another directory. The content of the .git file is in the following
```
gitdir: /home/me/Documents/myprj/.git/
```

When I stage the changed files and press `c`, I see the following errors in `*magit-process...*` buffer. And I cannot commit using magit. But committing in commandline works fine.

```
*ERROR*: Not inside a Git repository
error: There was a problem with the editor '/usr/bin/emacsclient --socket-name=/tmp/emacs1000/server'.
Please supply the message using either -m or -F option.
```

Can anyone help? Thanks!

tarsius:
I am using such ".git files" myself and am not having any problems with it.

E.g. `~/.emacs.d/lib/macrostep/.git`:

```
gitdir: ../../.git/modules/lib/macrostep
```

Try if you can fix this by using a relative file name. So far I have only encountered relative file names in these files. How did you create that file? It's possible that absolute file names are not handled correctly yet, at least I did not explicitly handle that case and there also are no tests for it.

tarsius:
What OS are you using? If Windows, then what Git build and what Emacs build?

kccai:
Hi, @tarsius , my system information is in the following. 

```
- OS: Arch linux (x86_64 Linux 4.2.3-1-ARCH)
- Emacs: 24.5.1
- magit: Version: 20151024.220
```
I changed the path described in the .git file to a relative path, but I still get that `Not inside a Git repository` error. Now I installed the magit stable version 2.2.2 and everything is fine. I think maybe some commits broke magit in the master branch.

kyleam:
I can reproduce this with

    gitdir: ../magit/.git/

in a directory at the same level as the magit repo.  I've bisected it
to 4d035b5 (magit-toplevel: return correct wtree from within
.git/worktrees/ID, 2015-10-15).

kyleam:
1156e6d37 seems to do the trick, although I'm not sure whether .git/gitdir, if it exists, always contains just ".git".

kccai:
@kyleam  Your fix works! Thanks!

tarsius:
[...]

PR #2367, opened 2015-10-28
magit-toplevel: ignore .git/gitdir

kyleam:
4d035b5 (magit-toplevel: return correct wtree from within
.git/worktrees/ID, 2015-10-15) taught magit-toplevel to look in
.git/worktrees/<name>/gitdir to find the worktree path.  However,
magit-toplevel now also reads .git/gitdir if it exists, which contains
".git" rather than a linked worktree.  In this case, magit-toplevel
fails to identify the top-level despite being inside a repo.

These .git/gitdir files are created in a repo when a .git file specifies
a gitdir path that points to the repo.  They result from a bug [1] that
is present in at least the version 2.6.2 release and that was likely
introduced in version 2.5.2.

Add a check to prevent this file from being processed as a worktree
gitdir file.

[1] http://thread.gmane.org/gmane.comp.version-control.git/280307/focus=280313

Re: #2364

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -386,10 +386,15 @@ returning the truename."
                         (expand-file-name gitdir))))
         (if (magit-bare-repo-p)
             gitdir
-          (let ((link (expand-file-name "gitdir" gitdir)))
-            (if (file-exists-p link)
+          (let* ((link (expand-file-name "gitdir" gitdir))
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks a lot. Also for taking this to the Git mailing list.

I have cherry-picked.

PR #2374, opened 2015-10-31
Fix typo

ZSVG:
Change "overriden" to "overridden"


tarsius:
Thanks. I have squashed the commits into one. At least the last two definitely belong in just one, and three pull requests is also quite excessive.

ZSVG:
Thanks for mentioning how to submit better requests in the future, I'll be sure to do so.

PR #2375, opened 2015-10-31
Add abbreviated hash to magit-blame-heading

pradhanpk:
Similar to `%H`, adding the `%h` to `magit-blame-heading-format` for an abbreviated hash, whose length can be customized through `magit-blame-abbreviated-hash-length`.

### DIFF
--- a/lisp/magit-blame.el
+++ b/lisp/magit-blame.el
@@ -42,6 +42,7 @@
 The following placeholders are recognized:
 
   %H    hash
+  %h    abbreviated hash
   %s    summary
   %a    author
   %A    author time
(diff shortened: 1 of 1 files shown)


tarsius:
I would want the abbreviated hashes to be `core.abbrev` long (or for hashes for which that substring would be ambiguous a string as long as necessary). But it's not possible to tell `git blame --incremental` to print abbreviated hashes instead of / in addition to the full hash. Sure one could set the length to 9 instead of to the default of 7, that makes ambiguous abbreviations less likely, but not impossible. To get the properly abbreviated hashes it would be necessary to call git for each of the relevant commits, which is way to expensive.

That's the primary reason I have not implemented support for `%h` myself. The other reason is that showing the (abbreviated or full) hash, is not actually very useful. (I am offering `%H` anyway because it comes "for free", but I would suggest putting that at the very end of the format.)

Showing the hash is not very useful because (1) it's not possible to mark it with the mouse in order to copy it, but (2) it *is* possible to copy the hash to the kill ring by pressing <kbd>M-w</kbd>, regardless of whether the hash or part of the hash is visible.

pradhanpk:
@tarsius Oh brilliant! I didn't know about `M-w`. That solves my problem way better than this attempted patch. Closing this PR. Thanks!

PR #2379, opened 2015-11-03
Fix elisp error.

Andersbakken:
So I really don't know if this is the correct fix but I don't seem to have a --some function defined in my environment. I use dash.el from trunk and GNU Emacs 24.5.1 (x86_64-apple-darwin13.4.0, NS apple-appkit-1265.21) of 2015-04-10 on builder10-9.porkrind.org [2 times].


npostavs:
> I use dash.el from trunk

An old trunk maybe? It's present in 2.12.1: https://github.com/magnars/dash.el/blob/2.12.1/dash.el#L504

PR #2398, opened 2015-11-18
make: fix compilation with Cygwin Emacs

### LINKED ISSUE #2392: compilation under cygwin fails
in default.mk:
LOAD_PATH ?= -L $(TOP)/lisp -L $(shell cygpath --mixed $(DASH_DIR))

doesn't work and generates windows sytle paths.  So compilation fails because it can't find dash.

Removing the "--mixed" option allows compilation to finish.

johnmastro:
Adjust default.mk so that `cygpath` is only called when compiling on Cygwin but with NT Emacs. Using `cygpath` is unnecessary when compiling with Cygwin Emacs, and can cause errors when used with the `--mixed` argument.

Fixes #2392

### DIFF
--- a/default.mk
+++ b/default.mk
@@ -71,7 +71,10 @@ ifeq "$(DASH_DIR)" ""
   DASH_DIR = $(TOP)../dash
 endif
 
-CYGPATH := $(shell cygpath --version 2>/dev/null)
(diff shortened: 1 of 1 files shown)


tarsius:
Cherry-picked.

PR #2400, opened 2015-11-18
Optimize refresh for large repos

ewstearns:
Making a list of tracked files for a refresh operation can take a long
time on large repos on remote paths.  Check file under top directory
indivitually.

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -826,22 +826,15 @@ When called interactively then the revert is forced."
   (interactive (list t))
   (when (or force (and magit-revert-buffers (not inhibit-magit-revert)))
     (-when-let (topdir (magit-toplevel))
-      (let* ((tracked (magit-revision-files "HEAD"))
-             (buffers
-              (if (> (length tracked)
-                     (length (buffer-list)))
-                  (--filter
+      (let* ((buffers
+              (--filter
                    (let ((file (buffer-file-name it)))
                      (and file
                           (equal (file-remote-p file)
                                  (file-remote-p topdir))
                           (file-in-directory-p file topdir)
-                          (member (file-relative-name file topdir) tracked)))
-                   (buffer-list))
-                (--mapcat
-                 (--when-let (find-buffer-visiting (expand-file-name it topdir))
-                   (list it))
-                 tracked))))
+                          (magit-file-relative-name file t)))
+                   (buffer-list))))
         (when (and buffers
                    (or force
                        (not (eq magit-revert-buffers 'ask))


tarsius:
I haven't looked at the patch in detail yet, but this sounds like a good idea.

tarsius:
Actually `(magit-revision-files "HEAD")` is really fast. In the Emacs repository, which contains quite a few files, it took 0.000001393 seconds. That function runs git exactly once (`git ls-tree -z -r --name-only rev`) so it also shouldn't make a difference whether this is run locally or via Tramp.

Your patch might still lead to a speedup because it drops `(member each-file huge-list-of-files)`. But that might also lead to a slowdown because you instead call git once for each file (and over Tramp that is expensive).

So there might be room for improvement here, but I doubt it's this.

ewstearns:
For my use case (large repo over tramp using scp protocol), the call to `(magit-revision-files "HEAD")` takes about 15 seconds. With this patch and 20 buffers open, refresh takes about 2-3 seconds. This patch makes refresh usable for me. Any solution would require eliminating the call to `magit-revision-files`.

When profiling, I found that almost all the time in my case was taken up with `tramp-check-for-regexp`, so the slow down I see may only be while using tramp.

tarsius:
How causes `tramp-check-for-regexp` to be called?

ewstearns:
Snipped from the profiler report:

````
  - magit-refresh                                              42802  89%
    - magit-revert-buffers                                      42209  88%
     - magit-revision-files                                     42209  88%
      - magit-toplevel                                          42204  88%
       - magit-rev-parse-safe                                   42203  88%
        - apply                                                 42203  88%
         - magit-git-str                                        42203  88%
          - apply                                               42202  88%
           - magit-process-file                                 42202  88%
            - apply                                             42202  88%
             - process-file                                     42202  88%
              - apply                                           42202  88%
               - tramp-file-name-handler                        42202  88%
                - apply                                         42202  88%
                 - tramp-sh-file-name-handler                   42202  88%
                  - apply                                       42202  88%
                   - tramp-sh-handle-process-file               42202  88%
                    - apply                                     42202  88%
                     - ad-Advice-tramp-sh-handle-process-file              42202  88%
                      - apply                                   42202  88%
                       - #<compiled 0x42e3d751>                 42202  88%
                        - tramp-send-command-and-check              42193  88%
                         - tramp-send-command                   42193  88%
                          - tramp-wait-for-output               42158  88%
                           - tramp-wait-for-regexp              42158  88%
                              tramp-check-for-regexp              40375  84%
 ````

tarsius:
Before I apply this or a similar change, I want to know *why* this is slow. Any help figuring that out is appreciated. I myself won't get to it until I have finished some work related to pushing and branching. Once I get to it (late December), I will likely also look into other ways in which reverting could be optimized.

ewstearns:
As an experiment, I tried to call `git ls-tree -r -z --name-only HEAD` without `magit` but using TRAMP. My methodology is `M-! (shell-command)` from a `dired` buffer open to the top directory of my repo (`/scp:[user]@[machine]:/home/[user]/[repo path]`. This took even longer than `magit-refresh`, closer to 60 seconds rather than 15 seconds. This points further to TRAMP as the cause of my slow down.

I also tried calling `M-x shell` while in the TRAMP `dired` buffer and tried the same command. This took significantly longer than the `shell-command` method and eventually there was a warning about exceeding the undo limit for the `*shell*` buffer.

I tried calling `time git ls-tree -r -z --name-only HEAD` in a ssh connection on a regular termal and it took 0.039 seconds, so git itself is very fast.

I cloned my repo on to my local machine and did the same tests. `magit-refresh` took about a third of the time, still a significant wait but more usable.

There are >18000 files in my repo.

My emacs version is `GNU Emacs 25.0.50.1 (x86_64-apple-darwin15.0.0, NS appkit-1404.13 Version 10.11.1 (Build 15B42)) of 2015-11-11`, a couple weeks off the head of the development repo installed using `brew`.

A possible compromise is to run the command `git ls-tree -r  --name-only HEAD | wc -l` to get a count of the files in the repo and if it is greater than 5000 (or a configurable value) use my method of checking each file individually instead of getting the whole list.

npostavs:
> I cloned my repo on to my local machine and did the same tests. magit-refresh took about a third of the time, still a significant wait but more usable.

A third of 15 (= 5)? How does your patch affect this case?

ewstearns:
With my patch and no source buffers open, `magit-refresh` takes less than 3 seconds on my local machine (my local machine is slow, which is why I use the remote build server) and about 0.6 seconds over tramp.

With 20 source buffers open, about 7 seconds on my local machine, 2-3 seconds over tramp.

I am using approximations for all of these because I see a lot of variations of time for both the local case and the remote case.

A typical call to `time git ls-tree -r -z --name-only HEAD` takes about 3 seconds on my local machine in a terminal.

My patch speeds up `magit-refresh` even with many buffers open on the fast remote device, but I start to see a significant slow down with many buffers open on a slow local device, although the difference between 5 seconds and seven is maybe not big enough to matter in practice.

The advantage to your original method is that it takes about the same amount of time regardless of how many buffers are open.

Another possible compromise may be to use my method only for remote machines with large repos and when there are relatively few buffers open (size of repo and number of buffers configurable).

tarsius:
Please figure out what the regex in question is and why it is slow. I suspect it has something to do with `git ls-tree -z ...` printing a single, very long line.

ewstearns:
I used the following code to profile this:

````elisp
(defun profile-magit ()
  (elp-restore-all)
  (elp-instrument-package "magit")
  (elp-reset-all)
  (magit-revision-files "HEAD")
  (elp-results))
````
For current network conditions (at home, over VPN; worst case conditions for tramp) with current head (46143e4) over four runs: 52.785337, 35.618167, 25.768841, 53.236029 seconds.

With the following change:

````
index 1bbeaaa..912ea63 100644
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -490,7 +490,7 @@ tracked file."
 
 (defun magit-revision-files (rev)
   (magit-with-toplevel
-    (magit-git-items "ls-tree" "-z" "-r" "--name-only" rev)))
+    (magit-git-lines "ls-tree" "-r" "--name-only" rev)))
 
 (defun magit-changed-files (rev-or-range &optional other-rev)
   "Return list of files the have changed between two revisions.
````
Over four runs:
7.412256, 5.294956, 6.816468, 6.138879 seconds.

With my patch under the same network conditions, `(magit-refresh)` with no buffers open takes, over 4 runs:
3.24654, 3.042895, 2.940799, 3.137578 seconds.

Using `git-ls-tree` without the `-z` option saves considerable time and should be used, but I think my patch should still be considered also for large repos because refresh operations would still be slow (about 10 seconds).

tarsius:
> Using `git-ls-tree` without the `-z` option saves considerable time and should be used,

Without  `-z` git would quote file-names containing non-ascii characters and we would have to pass all files through `magit-decode-git-path`.

tarsius:
I am pretty sure now that the reason is the very long line. I looked at the tramp code a bit and did not see an easy fix. But Michael, tramp's maintainer, might. Could you please bring this issue up on [`tramp-devel`](https://lists.gnu.org/mailman/listinfo/tramp-devel). Before you do that you should try with the latest standalone tramp release and maybe even try the development version.

tarsius:
I have finally decided to do this after all because determining tracked files can also be slow when *not* using Tramp. See #2458.

PR #2411, opened 2015-11-27
add backtab binding for magit-section-cycle-global

mosic:
On my machine (Ubuntu 14.04 running Emacs 24.5.1) pressing `<S-tab>` gets registered as `<S-iso-lefttab>` which gets translated to `<backtab>`, so keybinding for `magit-section-cycle-global` doesn't work.

This change fixes it by also binding `<backtab>`.


tarsius:
Thanks

PR #2412, opened 2015-11-27
made git-rebase-move-line-{up,down} prefixable

markshep:
Useful for moving commits multiple lines in an interactive rebase by using prefix args.


kyleam:
Thanks.  I haven't yet looked at the changes in detail, but I too find
this behavior useful.  I implemented a similar patch some time
ago, which I've now pushed to km/rebase-n, just for reference when
discussing this PR.

npostavs:
My impression (looking at both changesets) is that these commands might be simpler if reimplemented using `delete-and-extract-region` + `forward-line` + `insert` instead of trying to reuse `transpose-lines`.

kyleam:
Yes, probably true.

tarsius:
It would also be nice if one could multiple-lines (the region) around as a unit.

tarsius:
Speaking of things that would be nice - @kyleam and @npostavs, please consider setting an avatar. I have learned by now which of the generated avatars stands for whom, but users with less contact might have difficulties remembering who is who. ;-)

tarsius:
Work on this an related features is happening in #2419.

PR #2418, opened 2015-11-29
Include all info files in the elpa package

vanicat:
In debian, we now use `make elpa` to generate elpa package we use to create the debian package.

The magit elpa package generated by `make elpa` does only contain `magit.info` but not `magit.info-1` and `magit.info-2`. This patch fix this problem.


tarsius:
Already fixed: 018c263067e56a754867e2e6bcc97ccf8a6dc517. I'll create a new release shortly after merging #2414.

tarsius:
When the two extra files first appeared I did not realize that makeinfo had decided to split the output file for the fun of it but thought these were new temporary files.

PR #2419, opened 2015-11-30
rebase mode: move line extensions

kyleam:
See #2412.

### DIFF
--- a/Documentation/RelNotes/2.4.0.txt
+++ b/Documentation/RelNotes/2.4.0.txt
@@ -64,6 +64,12 @@ Changes since v2.3.0
   apply to diff-related faces because diff sections are highlighted
   differently from other sections.  #2387
 
+* The commands `git-rebase-move-line-up' and `git-rebase-move-line-down'
+  have been extended.  When a region is active, they now operate on
+  all lines touched by the region, not just the current line.  In
+  addition, a numeric prefix can now be given to specify how many
+  lines the current line or selection is moved across.
+
 This release also contains the changes described in the v2.3.1 release
 notes, as well as various other documentation fixes, code clean-ups,
 bug fixes, and other small to medium improvements.
--- a/lisp/git-rebase.el
+++ b/lisp/git-rebase.el
@@ -214,12 +214,33 @@
           (forward-line)))
     (ding)))
 
+(defun git-rebase-line-p (&optional pos)
+  (save-excursion
+    (when pos (goto-char pos))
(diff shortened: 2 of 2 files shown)


npostavs:
The multiple lines selection feels a bit weird, should we add magit's special region-composed-of-lines highlighting here too?

kyleam:
Thanks for the fixups.

> should we add magit's special region-composed-of-lines highlighting here too?

I think it's worth trying out.  I'll have a look.

kyleam:
OK.  I've added a commit that uses magit-section-heading-selection
instead of the normal region.  I'm sure there are places that could be
cleaned up, but it seems to be working.  WDYT?  Worth it?

tarsius:
I came here to say "please keep the lines selected after moving them", but that already works!

Please merge once you feel this is ready. Looks good to me.

npostavs:
> OK. I've added a commit that uses magit-section-heading-selection
> instead of the normal region.

Shouldn't the body of `git-rebase-update-highlight` go in `git-rebase-highlight-region` and then a `post-command-hook` is not needed?

kyleam:
> Shouldn't the body of git-rebase-update-highlight go in git-rebase-highlight-region and then a post-command-hook is not needed?

Yeah, I guess I was trying to follow magit proper too closely.  Your suggestion seems to work well.

kyleam:
Updated.  I'll merge after waiting a bit for any other comments.

tarsius:
Please mention this in the release notes.

npostavs:
I added a bit more simplification fixup.

My only remaining concern is that I think `magit-section-heading-selection` is a bit hard to see, at least with default (`emacs -Q`) colours.

kyleam:
> I added a bit more simplification fixup.

Thank you.

> My only remaining concern is that I think
> magit-section-heading-selection is a bit hard to see, at least with
> default (emacs -Q) colours.

Hmm, is there another magit selection face that you think we should
switch it to?  I went with magit-section-heading-selection because
it's the one used for log lines, which seems the closest match for
lines in the rebase buffer.

kyleam:
> Please mention this in the release notes.

Done.

tarsius:
> Hmm, is there another magit selection face that you think we should switch it to?

I would say you are using the right face. But maybe that face's default value ain't no good. I don't know, I haven't looked at the default theme for quite a while.

kyleam:
> I would say you are using the right face. 

OK, I've merged this then, since changing the default for that face is a separate issue.

PR #2423, opened 2015-12-03
pass dirname to magit--with-safe-default-directory

### LINKED ISSUE #2408: magit incorrectly treats file as untracked if file has executable bit set
Magit 20151123.1028, Git 2.6.3, Emacs 24.5.1 OS X 10.11.1

It's very easy to reproduce the bug.
Create a new repo with one bash script in it, set executable bit on the script and commit it to the repo. Then try magic-log-buffer-file on the script file.

npostavs:
This works around/ fixes #2408.

Also, I propose separating `magit--with-safe-default-directory` into macro and function, because having code only in a macro makes it impossible to step through in the debugger.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -300,21 +300,22 @@ call function WASHER with no argument."
 
 ;;; Files
 
+(defun magit--safe-default-directory (&optional file)
+  (catch 'unsafe-default-dir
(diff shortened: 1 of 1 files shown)


tarsius:
Cherry-picked. Thanks.

PR #2425, opened 2015-12-04
Add :require property to `auth-source-search' call

thisirs:
If :host is not required, the first item not having the :host property
or matching it is selected. This is not a problem for netrc files where
all entries have a :host (or :machine) property. Other backends like
secrets do not.

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -681,7 +681,8 @@ tracked in the current repository are reverted if
   "Use `auth-source-search' to get a password.
 If found, return the password.  Otherwise, return nil."
   (require 'auth-source)
-  (let ((secret (plist-get (car (auth-source-search :max 1 :host key))
+  (let ((secret (plist-get (car (auth-source-search :max 1 :host key
+                                                    :require '(:host)))
                            :secret)))
     (if (functionp secret)
         (funcall secret)


tarsius:
I know to little about this to know whether this is necessary (it *sounds* reasonable though). @DamienCassou what do you think?

DamienCassou:
@thisirs this sounds reasonable, but the patch would be better with a test case. For the moment, I fail to see how an entry could have a matching `host` and not `host`. What about requiring `:secret` as well?

tarsius:
Lol. I was looking at this *right now* and coming to the conclusion that I should probably just merge!

But since you ask a question too, it's back to @thisirs. And I agree, a test would be nice too.

thisirs:
It should be more clear with the following snippet:

   (progn
      (with-temp-buffer
        (insert "\
    user blah password foo
    machine github.com user blah2 password foo2")
        (write-file (expand-file-name ".netrc" temporary-file-directory)))
      (let ((auth-sources `((:source ,(expand-file-name ".netrc" temporary-file-directory)))))
        (message "%S" (auth-source-search :max 1 :host "github.com"))
        (message "%S" (auth-source-search :max 1 :host "github.com" :require '(:host)))))

In the first call of `auth-source-search` the second line of the netrc file is shadowed by the first one because there is no `machine` property`... This is fixed by requiring it in the second call!

The `secret` property could be required too, but it leads to the same result (i.e. nil is returned).

I could turn the above snippet into a test but is that really necessary? This is a very specific behaviour that only occurs when the user edits by himself the netrc file and does not provide a machine field...

What do you think?

tarsius:
Yes, a test isn't necessary (would have been nice, but many other tests, which would be more important, are also missing). I have cherry-picked this now.

tarsius:
@thisirs You might want to set `user.name`.

PR #2431, opened 2015-12-06
Automatically add cloned repos to projectile

JoshTGreenwood:
When cloning a repo, it would be nice if it showed up automatically in my projectile-known-projects list.

This PR adds that feature.  To turn it on, set `magit-add-project-to-projectile-after-clone` to `t`.  I also added a bit of documentation for that setting but was not sure where it should go, let me know if there is a better spot.  I wasn't sure whether `defvar` or `defcustom` was more appropriate, let me know if you want me to change that as well.

### DIFF
--- a/lisp/magit-remote.el
+++ b/lisp/magit-remote.el
@@ -31,7 +31,12 @@
 
 (require 'magit)
 
-(defvar magit-add-project-to-projectile-after-clone)
+(defcustom magit-add-project-to-projectile-after-clone nil
+  "when non-nil, automatically adds magit-cloned repos
+to projectile's known project list"
+  :group 'magit-clone
+  :options '(nil t)
+  :type 'boolean)
 
 ;;; Clone
 
--- a/lisp/magit-remote.el
+++ b/lisp/magit-remote.el
@@ -52,7 +52,7 @@ Then show the status buffer for the new repository."
                            (magit-convert-git-filename directory))
            0)
     ;; Automatically add project to projectile
-    (when (eq (magit-add-project-to-projectile-after-clone) t)
+    (when magit-add-project-to-projectile-after-clone
       (if (eq (fboundp 'projectile-add-known-project) t)
           (projectile-add-known-project directory)
         (message "projectile-add-known-project is not defined")))
--- a/lisp/magit-remote.el
+++ b/lisp/magit-remote.el
@@ -31,6 +31,8 @@
 
 (require 'magit)
 
+(defvar magit-add-project-to-projectile-after-clone)
+
 ;;; Clone
 
 ;;;###autoload
@@ -49,6 +51,11 @@ Then show the status buffer for the new repository."
                            ;; Stop cygwin git making a "c:" directory.
                            (magit-convert-git-filename directory))
            0)
+    ;; Automatically add project to projectile
+    (when (eq (magit-add-project-to-projectile-after-clone) t)
+      (if (eq (fboundp 'projectile-add-known-project) t)
+          (projectile-add-known-project directory)
+        (message "projectile-add-known-project is not defined")))
     (message "Cloning %s...done" repository)
     (magit-status-internal directory)))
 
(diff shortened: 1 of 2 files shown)


lisp/magit-remote.el
```diff
@@ -49,6 +51,11 @@ Then show the status buffer for the new repository."
                            ;; Stop cygwin git making a "c:" directory.
                            (magit-convert-git-filename directory))
            0)
+    ;; Automatically add project to projectile
+    (when (eq (magit-add-project-to-projectile-after-clone) t)
```
npostavs: `magit-add-project-to-projectile-after-clone` is not a function, I think it should be
```elisp
(when magit-add-project-to-projectile-after-clone
```

And the enabling value documented as non-nil, rather than `t`.

npostavs:
> I wasn't sure whether defvar or defcustom was more appropriate, let me know if you want me to change that as well.

Options provided for the user to change should be defcustom, unless there is some particular reason not to.

kyleam:
I'm not sure this is worth adding.  Won't the project quickly be registered anyway (for example, when finding a file or going to a directory)?  Also, it seems a bit odd to add projectile-specific code to magit.  For users who want this, perhaps advising magit-clone would be a better option.

tarsius:
I think that `(add-hook 'magit-mode-hook 'projectile-cache-projects-find-file-hook)` would be more useful.

JoshTGreenwood:
> Options provided for the user to change should be defcustom, unless there is some particular reason not to.

Updated to use `defcustom` and allow anything non-nil @npostavs 

> Won't the project quickly be registered anyway (for example, when finding a file or going to a directory)? 

Yes, but I find it easier to `magit-clone` then immediately jump to the project via the name vs. trying to navigate to it.  

> Also, it seems a bit odd to add projectile-specific code to magit. For users who want this, perhaps advising magit-clone would be a better option.

@kyleam I was hesitant for this reason but thought it might be widely applicable.  If this doesn't get merged, here is what I've been using for future readers:

```
(defun add-project-to-projectile-known-projects (_ directory)
  (projectile-add-known-project directory))

(advice-add 'magit-clone :after 'add-project-to-projectile-known-projects)
```

@tarsius what does the `'projectile-cache-projects-find-file-hook` buy us that `projectile-add-known-project` doesn't? I'm indifferent as long as the end result is the same.

kyleam:
> but I find it easier to magit-clone then immediately jump to the project via the name vs. trying to navigate to it.

I'm confused.  Why do you need to navigate to it?  Doesn't magit-clone go to the status buffer after cloning the project?

JoshTGreenwood:
> Doesn't magit-clone go to the status buffer after cloning the project?

It does, but that doesn't trigger projectile registering the project.

tarsius:
> what does the `'projectile-cache-projects-find-file-hook` buy us that `projectile-add-known-project` doesn't? I'm indifferent as long as the end result is the same.

It's the function provided by projectile to do this sort of thing. So we should use it. In addition to remember to project for this Emacs instance only, it also remembers it for future sessions.

> It does, but just being in the status buffer doesn't trigger projectile registering the project. (although @tarsius suggestion might?)

It does.

kyleam:
> It does, but that doesn't trigger projectile registering the project.

But then it's not "jump to the project via the name vs. trying to
navigate to it" since magit-clone already navigates to it by putting
you in the status buffer, right?

tarsius:
1. Your declared goal is that the status buffer is shown right after cloning. That already happens without your patch.
2. Your patch adds a cloned repository to projectile's project list. But only for the current session. And only when using `magit-clone`, not when using `magit-init` or simply showing the status buffer (or another magit buffer) for an existing repository.

In other words your patch is almost entirely useless. It doesn't do what you want it to do, but luckily that's not necessary because that happens anyway. And as a general magit - projectile bridge it's also of no use, because it only covers one particular special case.

Instead you should use the snippet I posted above. Or nothing at all if you don't actually care about all git repositories being added to projectile's project list when they are first visited in magit.

PR #2436, opened 2015-12-10
Support git-subtree

tarsius:
This is unfinished.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -4188,6 +4188,46 @@ Also see [[info:gitman#git-submodule]].
 
   Unregister the submodule at PATH.
 
+** Subtree
+
+Also see [[info:gitman#git-subtree]].
+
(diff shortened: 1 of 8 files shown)


zachlatta:
Subtree support would be greatly appreciated! Thanks for doing this :smile:.

tarsius:
I'll open a new issue when this is mostly ready. The branch is still available.

npostavs:
I've been looking at this (added a missing `interactive` form d7656233c41c0b86b2b0aa400681ff79ee8ea39f).

One thing I noticed when trying to use `subtree pull` from `magit-git-command` is that it needs `$EDITOR` (git confusingly tried to run `nano` without a terminal, which gave somewhat confusing results).

Also `subtree split` takes a *long* time, so it would be nice to run it asynchronously.

Anything else missing?

PR #2444, opened 2015-12-18
Various improvements to pushing and branching

tarsius:
**Update**: I merged most of the changes originally proposed here with #2447.

`magit-push-implicitly` still needs some work.


tarsius:
I'll open a new pr for what remains.

PR #2447, opened 2015-12-21
Make it less likely that the push-remote is undefined

tarsius:
and make it easier to define it if it is.


lisp/magit-remote.el
```diff
@@ -254,28 +319,77 @@ Then show the status buffer for the new repository."
...
+   (list (magit-push-arguments)
+         (and (magit--push-current-set-upstream-p current-prefix-arg)
+              (magit-read-remote (format "Set upstream of %s and push there"
+                                         (magit-get-current-branch))))))
   (--if-let (magit-get-current-branch)
-      (-if-let (target (magit-get-upstream-branch it))
-          (magit-git-push it target args)
-        (user-error "No upstream is configured for %s" it))
+      (progn (when upstream
+               (magit-call-git "branch" (concat "--set-upstream-to=" upstream)))
```
kyleam: Shouldn't the `upstream` passed to `--set-upstream-to` consist of the branch name in addition to the remote?  Also, this will fail if the branch doesn't exist on the remote.  Perhaps we should just set `branch.<name>.remote` and `branch.<name>.merge` directly.

tarsius: All spot on. I have pushed a fix.

lisp/magit-remote.el
```diff
@@ -254,28 +319,77 @@ Then show the status buffer for the new repository."
...
+                 (if (member remote (magit-list-remotes))
+                     (magit-git-push it (concat remote "/" it) args)
+                   (user-error "Remote `%s' doesn't exist" remote))
+               (user-error "No push-remote is configured for %s" it)))
     (user-error "No branch is checked out")))
 
+(defun magit--push-current-set-pushremote-p (&optional change)
+  (and (or change magit-push-current-set-remote-if-missing)
+       (magit-get-current-branch)
+       (not (magit-get-push-remote))))
```
kyleam: Won't this last form prevent the prefix argument from changing a remote that is already set?  Same for `magit--push-current-set-upstream-p`.

lisp/magit-remote.el
kyleam: Great, thanks for the fix.

lisp/magit-remote.el
tarsius: Yes. Thanks. I've pushed a fix.

PR #2449, opened 2015-12-22
Update git-popup dependency

kidd:
I updated magit today and it seems git popup dependency wasn't updated to the latest version. 
Manually updating worked so here it's a PR to update the dependency


tarsius:
Yep, I forgot to adjust that. Thanks for the fix, I have cherry-picked it.

PR #2452, opened 2015-12-23
magit-pull-or-fetch-popup: new popup

tarsius:
```
Combine `magit-pull-popup' and `magit-fetch-popup' into one common
popup.  By default the old popups are still bound in `magit-mode-map'.

Add this to your init file to use the new popup:

  (with-eval-after-load 'magit-remote
    (define-key magit-mode-map "f" 'magit-pull-and-fetch-popup)
    (define-key magit-mode-map "F" nil))

One advantage of the new popup is that it is once more possible to
press the same key twice to fetch.  However that binding now is "f f",
not "F F", and it fetches all remotes not just the current remote.

The new popup also lacks the commands for fetching from only a single
remote.  You can add them to the new popup by adding something like this
to your init file:

  (with-eval-after-load 'magit-remote
    (magit-define-popup-action 'magit-pull-and-fetch-popup ?U
      'magit-get-upstream-branch
      'magit-fetch-from-upstream-remote ?F)
    (magit-define-popup-action 'magit-pull-and-fetch-popup ?P
      'magit-get-push-branch
      'magit-fetch-from-push-remote ?F))
```

### DIFF
--- a/Documentation/RelNotes/2.4.0.txt
+++ b/Documentation/RelNotes/2.4.0.txt
@@ -46,6 +46,10 @@ Changes since v2.3.0
   Likewise many commands were renamed and their behavior was adjusted.
   Some new commands, related to the push-remote, were added.  #2414
 
+* The popup command `magit-pull-and-fetch-popup' was added as a
+  possible replacement for the separate `magit-push-popup' and
+  `magit-fetch-popup'.  #2452
+
 * The status buffer now features up to four logs listing unpulled and
   unpushed commits.  Two for the upstream and two for the push-remote.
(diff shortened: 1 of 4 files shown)


npostavs:
It should be `(with-eval-after-load 'magit-mode...` because `magit-mode-map` is defined in `magit-mode.el` (though possibly makes no difference depending on load order, haven't tested).

tarsius:
Once `magit-remote` is loaded, everything involved in this, including the keymap (which, granted, is the only thing that really matters) is also loaded.

PR #2458, opened 2015-12-30
No longer limit reverting to tracked files

tarsius:


tarsius:
Actually we have gone back to only reverting only tracked file *by default*.

PR #2507, opened 2016-01-20
Add log ordering options

smithzvk:
This exports an interface for accessing --topo-order, --date-order, and
--author-date-order.  These are options rather than switches because
they are mutually exclusive.  The --reverse option is left out because
it can be combined with the others.

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -292,6 +292,7 @@ the upstream isn't ahead of the current branch) show."
                (?D "Simplify by decoration"  "--simplify-by-decoration")
                (?f "Follow renames when showing single-file log" "--follow"))
     :options  ((?n "Limit number of commits" "-n")
+               (?o "Ordered by commit date"  "--*-order" magit-log-read-order)
                (?f "Limit to files"          "-- " magit-read-files)
                (?a "Limit to author"         "--author=")
                (?g "Search messages"         "--grep=")
@@ -322,6 +323,7 @@ the upstream isn't ahead of the current branch) show."
                (?D "Simplify by decoration"  "--simplify-by-decoration")
                (?f "Follow renames when showing single-file log" "--follow"))
     :options  ((?n "Limit number of commits" "-n")
+               (?o "Ordered by commit date"  "--*-order" magit-log-read-order)
                (?f "Limit to files"          "-- " magit-read-files)
                (?a "Limit to author"         "--author=")
                (?g "Search messages"         "--grep=")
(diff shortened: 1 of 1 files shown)


smithzvk:
I often find that it is useful for understanding what is going on in the repo if I reorder using the --date-order switch.  Here is a patch that adds this ability to magit-log.  I'm sure that this could be done better and would like any guidance any can give on how this could be better implemented.  I'm just trying to get the ball rolling on implementing this feature that I would like to use (and not needing to have my own patch set).

One issue I see is that I was expecting that setting the option would be 'sticky', i.e. it would persist at least when doing log refreshes.  This isn't the case, however.  I'm not sure how to do this and am still learning the Magit internals.

smithzvk:
....and I see that I didn't do due diligence before issuing this pull request.  There is a merged pull request that deals with this (#617) and it has a different interface.  It is quite old and perhaps what I'm seeing is a regression in capability.  If it isn't, I don't see that option any longer.  Perhaps it is hidden away somewhere?

Sorry if this is noise.

npostavs:
Some of the more "advanced" options were removed (see b9d9ba52e1ad3028a81d64273cfcf221dcc5f0fe) to keep things manageable. But you can add them for yourself, see https://github.com/magit/magit/wiki/Additional-proposed-infix-arguments-and-suffix-commands.

smithzvk:
npostavs, that link is very informative and, quite frankly, more useful for me than anything I was trying to do here.  I'll leave this to the maintainers.  There may be an argument that ordering by commit date should be within the default behavior Magit just because it can be valuable when doing a --graph log.  The topological sort can be confusing at times.

Close it, leave it open while you ponder, tell me to fix it (and I will to the best of my ability), but I'll be using a bit of customization in my .emacs file over my patch here, or at least until such time that Magit provides this.  At the very least, you have a commit to add to your wiki page for this suggestion.

Thanks.

tarsius:
I don't think this actually works. You would end up calling `git log --*ordertopo ...` I believe.

In the long run I would like to allow subclassing instead of providing a fixes set of events (switches, options, git-arguments, and actions). That would, among many other things, allow defining arguments such as `--[no-]track` (which could be on, off, or undefined). Or switches with even more states such as `--[date|author-date|topo]-order`. This would be similar to the new `git-variable` type, and pressing the corresponding key would cycle between the "values", or rather alternative arguments.

Since I have recently added support for such `git-variables`, I am more open to hack in something similar for switches. (I wanted to delay support for `git-variables` until after rewriting `magit-popup`, but I had to rush support because that was necessary for the changes I have made to how branching and pushing work.)

If it turns out to be fairly simple and not too ugly to support this, then I might add it.

smithzvk:
Yeah, I forgot to add the actual code to the commit.

I'll look into variables.

tarsius:
I have written my own variant of this: #2524. And I have also experimented with "cycle switches": https://github.com/magit/magit/compare/jb/multi-switch?expand=1. The latter seems to hacky (and also isn't quite finished).

PR #2512, opened 2016-01-23
magit-remote.el: Add more switch and options to git-format-patch

namhyung:
I usually use --cover-letter and --subject-prefix options to include
version numbers like "[PATCH v2]".  So it'd be better to have it in magit.


kyleam:
Thanks for the patch.  We're waiting to see what comes of #1800 before
extending the format-patch popup (see #2233).  But it's easy to add
these on your end:

https://github.com/magit/magit/wiki/Additional-proposed-infix-arguments-and-suffix-commands

PR #2513, opened 2016-01-24
Try to limit reverts to buffers for current repository

tarsius:
```
Add new option `auto-revert-buffer-list-filter' and redefine
`auto-revert-buffers' from `autorevert.el' to use it.  Add new
predicates `magit-auto-revert-buffer-p' and the more restrictive
`magit-auto-revert-buffer-p', which are both intended to be used as
potential value of the option.

For now the option defaults to `nil', but that might change in a future
release.  When Magit explicitly calls `auto-revert-buffers' (as opposed
to when that is called due to a file notification event or by a timer),
and `auto-revert-buffer-list-filter' is `nil', then it is let-bound to
`magit-auto-revert-repository-buffer-p'.

Users who use Tramp and experience delays, should consider setting
the option to `magit-auto-revert-repository-buffer-p'.
```

### DIFF
--- a/Documentation/RelNotes/2.4.2.txt
+++ b/Documentation/RelNotes/2.4.2.txt
@@ -4,6 +4,21 @@ Magit v2.4.2 Release Notes (draft)
 Updates since v2.4.1
 --------------------
 
+* Added new option `auto-revert-buffer-list-filter' and redefined
+  `auto-revert-buffers' from `autorevert.el' to use it.  Added new
+  predicates `magit-auto-revert-buffer-p' and the more restrictive
+  `magit-auto-revert-repository-buffer-p', which are both intended
+  to be used as potential value of the option.
+
+  For now the option defaults to `nil', but that might change in a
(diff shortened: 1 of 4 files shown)


tarsius:
@npostavs Please comment.

npostavs:
> @npostavs Please comment.

Did you mean @ndimiduk? My only comment is to wonder whether using advice would be better than redefining the whole function.

tarsius:
> whether using advice would be better than redefining the whole function.

I don't think that is possible here.

npostavs:
> I don't think that is possible here.

Hmm, isn't f4dc04c equivalent to some `:before` advice that does 
```elisp
(when auto-revert-buffer-list-filter
  (setq auto-revert-buffer-list
        (--filter auto-revert-buffer-list-filter auto-revert-buffer-list)))
```

tarsius:
Seems you are right... that would certainly make things simpler.

PR #2514, opened 2016-01-24
Don't busy wait for subprocesses

npostavs:
Re: https://github.com/magit/magit/issues/2454#issuecomment-171090747

We can send input to synchronous subprocesses with `call-process-region`, but there is no file-handler equivalent (like `process-file` for `call-process`), so this may have problems when using magit over tramp.

### DIFF
--- a/lisp/magit-apply.el
+++ b/lisp/magit-apply.el
@@ -145,10 +145,10 @@ With a prefix argument and if necessary, attempt a 3-way merge."
       (magit-wip-commit-before-change files (concat " before " command)))
     (with-temp-buffer
       (insert patch)
-      (magit-run-git-with-input nil
-        "apply" args "-p0"
-        (unless (magit-diff-context-p) "--unidiff-zero")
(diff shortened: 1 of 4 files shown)


tarsius:
First commit seems good to go. For `magit-run-git-with-input` we could keep using the old code when `file-remote-p` is non-nil, and else use the new code you propose.

npostavs:
> For magit-run-git-with-input we could keep using the old code when file-remote-p is non-nil, and else use the new code you propose.

Done. I also noticed callers of `magit-run-git-with-input` don't need the `INPUT` parameter, can we drop it?

tarsius:
> I also noticed callers of `magit-run-git-with-input` don't need the `INPUT` parameter,

I haven't checked yet, but memory told me that this would be the case :-)

> can we drop it?

Yes please.

tarsius:
If the removal of `INPUT` should turn out to be a mistake, then we could go with `(or (if (bufferp (car args)) (pop args) (current-buffer))`.

PR #2523, opened 2016-01-27
magit-refresh-file-visiting-buffers-hook: new hook

tarsius:
WIP but hoping for comments.

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -681,11 +681,14 @@ window."
 
 (defvar inhibit-magit-refresh nil)
 
-(defun magit-refresh ()
+(defun magit-refresh (&optional head)
   "Refresh some buffers belonging to the current repository.
 
 Refresh the current buffer if its major mode derives from
-`magit-mode', and refresh the corresponding status buffer."
+`magit-mode', and refresh the corresponding status buffer.
+
+Run hooks `magit-pre-refresh-hook', `magit-post-refresh-hook',
+and `magit-refresh-file-visiting-buffers-hook'."
   (interactive)
   (unless inhibit-magit-refresh
     (magit-run-hook-with-benchmark 'magit-pre-refresh-hook)
@@ -697,8 +700,10 @@ Refresh the current buffer if its major mode derives from
       (with-current-buffer it
         (magit-refresh-buffer)))
     (magit-auto-revert-buffers)
+    (magit-refresh-file-visiting-buffers head)
     (magit-run-hook-with-benchmark 'magit-post-refresh-hook)))
 
+;; TODO run other hooks too
 (defun magit-refresh-all ()
   "Refresh all buffers belonging to the current repository.
 
@@ -709,7 +714,8 @@ tracked in the current repository."
   (interactive)
   (dolist (buffer (magit-mode-get-buffers))
     (with-current-buffer buffer (magit-refresh-buffer)))
-  (magit-auto-revert-buffers))
+  (magit-auto-revert-buffers)
+  (magit-refresh-file-visiting-buffers "force"))
 
 (defvar-local magit-refresh-start-time nil)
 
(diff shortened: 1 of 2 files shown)


lisp/magit-mode.el
```diff
@@ -772,6 +778,39 @@ tracked in the current repository."
...
+               (setq topdir (magit-toplevel)))
+      (with-local-quit
+        (when magit-refresh-verbose
+          (setq start (current-time))
+          (message "Running magit-refresh-file-visiting-buffers-hook..."))
+        (run-hook-with-args
+         'magit-refresh-file-visiting-buffers-hook
+         (if (member head (list nil (magit-rev-parse "HEAD")))
+             (--keep (find-buffer-visiting (expand-file-name it topdir))
+                     (magit-modified-files))
```
dgutov: `magit-modified-files` only returns the files that aren't currently up-to-date, right?

After a user makes a commit in Magit, we want to call `diff-hl-update` in their buffers, to remove the outdated highlightings. But in the simplest case, all of the involved files are up-to-date after the commit is done. Will this function include them anyway somehow?

lisp/magit-mode.el
```diff
@@ -772,6 +778,39 @@ tracked in the current repository."
...
+                     (file-in-directory-p buffer-file-name topdir)
+                     (not (file-in-directory-p buffer-file-name gitdir))))
+              (buffer-list)))))
+        (when magit-refresh-verbose
+          (message "Running %s...done (%.3fs)"
+                   'magit-refresh-file-visiting-buffers-hook
+                   (float-time (time-subtract (current-time) start))))))))
+
+(add-hook 'magit-refresh-file-visiting-buffers-hook 'diff-hl-update-buffers)
+(defun diff-hl-update-buffers (buffers)
```
dgutov: It would be a bit more trivial to integrate if the hook ran in each of the buffers, instead of being passed a list of buffers. Not a big deal either way, though.

dgutov: This is just an example of integration, right? You're not going to keep `diff-hl-update-buffers` in Magit?

You probably shouldn't, although if Magit continues to be the less stable of the two, it might become a good choice.

lisp/magit-mode.el
tarsius: > After a user makes a commit in Magit, ...

... `HEAD` doesn't point to the same revision anymore, so `(eq head (magit-rev-parse "HEAD")` is be `nil` and the other code path is be used.

Also `magit-refresh`'s `HEAD` argument is only `nil` when it is called interactively. Calling that command interactively is supposed to refresh the current buffer only. So `magit-refresh-file-visiting-buffers` should probably not be called at all in this case. I might change that. I am mentioning it here because the actual test is `(member head (list nil (magit-rev-parse "HEAD")))`, not `(eq...)` as implied above. The command `magit-refresh-all` however refreshes all magit buffers and by passing `"force"` (which is not equal to any sha1) to `magit-refresh-file-visiting-buffers` also all file-visiting buffers.

After using `C-u x` to do `git reset --hard HEAD`, then there are no modified files, but the buffers still get re-diffed because of `after-revert-hook`.

lisp/magit-mode.el
tarsius: I can turn the hook into a per-buffer hook. I didn't do so before because `(dolist (elt list) (run-hook-with-args 'hook))` looks a bit odd, but using a per-buffer hook probably makes more sense, so I'll consider that.

Yes, the `diff-hl` code is only included for demonstration purposes.

lisp/magit-mode.el
tarsius: By the way, I am not sure that I have gotten this 100% right yet. So if something seems fishy, then let me know.

npostavs: It would be better to write the test as `(or (null head) (equal head (magit-rev-parse "HEAD")))` to avoid an extra call to git in the interactive case.

tarsius: Yes.

dgutov: > so (eq head (magit-rev-parse "HEAD") is be nil and the other code path is be used.

Oh, I see. That's what the other code path is for.

> Calling that command interactively is supposed to refresh the current buffer only.

Shouldn't `diff-hl-update` be called in the current buffer as well?

> After using C-u x to do git reset --hard HEAD, then there are no modified files, but the buffers still get re-diffed because of after-revert-hook.

I think you're ignoring the index here. I can call `git reset` (without the revision argument, and without `--hard`), and see the file status change, without any changes to `HEAD` or file contents on disk. `diff-hl` also tries to indicate when the file's status is `added` or `deleted`. If it won't work here, it's not that big of a deal, but if you are targeting a general solution, this one might not be enough.

When thinking about it, I imagined a hook that's tied to Magit status buffers more. Like before every command the status buffer would save the list of non-up-to-date files (just from the buffer contents, or accessing the internal data structures), then do the same after the command completes, and run the hook for all files that were in either of the lists. This shouldn't impose additional process calls, because the status buffer needs to produce the lists of files to render itself (right?), so it should be even more efficient than this solution. But I suppose it might require significant changes in the code.

(I haven't tried this code yet, so I might be saying something unusually stupid).

PR #2524, opened 2016-01-27
log: add ++order pseudo option

tarsius:
```
Before `git log' is called, `++order=TYPE' is converted to
`--TYPE-order', one of `--topo-order', `--date-order', and
`--author-date-order'.  We do that instead of adding the
individual arguments to the popup because they are mutually
exclusive.
```

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -294,6 +294,7 @@ the upstream isn't ahead of the current branch) show."
     :options  ((?n "Limit number of commits" "-n")
                (?f "Limit to files"          "-- " magit-read-files)
                (?a "Limit to author"         "--author=")
+               (?o "Order commits by"        "++order=" magit-log-select-order)
                (?g "Search messages"         "--grep=")
                (?G "Search changes"          "-G")
                (?S "Search occurences"       "-S")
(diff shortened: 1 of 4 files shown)


tarsius:
I am not sure I want to add this. A compromise would be to add everything except for the actual bindings.

lisp/magit-log.el
```diff
@@ -364,6 +367,12 @@ the upstream isn't ahead of the current branch) show."
         (concat trace (or (match-string 2 trace) ":") file)
       (user-error "Trace is invalid, see man git-log"))))
 
+(defun magit-log-select-order (&rest _ignored)
+  (magit-read-char-case "Order commits by" t
```
kyleam: Prompt needs a trailing space.

lisp/magit-log.el
```diff
@@ -364,6 +367,12 @@ the upstream isn't ahead of the current branch) show."
...
+    (?t "[t]opography"     "topo")
+    (?a "[a]uthor date"    "date")
+    (?c "[c]ommitter date" "author-date")))
```
kyleam: Author and commit date are swapped.

kyleam:
I'd be in favor of adding this.

PR #2529, opened 2016-01-29
[doc] remove remaining references to with-editor

dunn:


tarsius:
Thanks.

I changed the prefixes of the commit messages and dropped the added quotes you around the yaml variable value.

dunn:
The quotes are necessary; otherwise `CURL` is set simply to 'curl': https://travis-ci.org/magit/magit#L109

tarsius:
Could you please create a new commit which says so. Thanks.

dunn:
No problem; sorry for not explaining fully at first!

PR #2530, opened 2016-01-29
magit-update-uncommitted-buffers: new function

tarsius:
Re @dgutov https://github.com/magit/magit/pull/2523#discussion_r51051022

> I imagined a hook that's tied to Magit status buffers more.

Magit commands can be used even when no status buffer exists. So that's not an option.

> before every command [...] save the list of non-up-to-date files [...], then do the same after the command completes, and run the hook for all files that were in either of the lists.

That sounds like a good idea and is what I have implemented here.

Running `find-buffer-visiting` (twice!) continues to be a potential bottleneck. But I think the main bottleneck is whatever functions are being added to the hook.

To use this, you have to

1. Add this to `diff-hl-mode`: `(add-hook 'magit-update-uncommitted-buffer-hook #'diff-hl-update)`.
2. Instruct users to opt-in by adding this to their init file: `(add-hook 'magit-post-refresh-hook #'magit-update-uncommitted-buffers)`.

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -777,6 +777,37 @@ Run hooks `magit-pre-refresh-hook' and `magit-post-refresh-hook'."
                                                 beg (line-end-position))))
                                 (t t)))))))))
 
+;;; Update File-Visiting Buffers
+
+(defvar magit--uncommitted-buffers nil)
+
+(defun magit-uncommitted-buffers ()
+  (let ((topdir (magit-toplevel)))
+    (--keep (find-buffer-visiting (expand-file-name it topdir))
+            (magit-modified-files t))))
+
+(defun magit-maybe-cache-uncommitted-buffers ()
+  "Maybe save a list of file-visiting buffers with uncommitted changes.
+That list is later used by `magit-update-uncommitted-buffers',
+provided it is a member of `magit-post-refresh-hook'.  If it is
+not, then don't save anything here."
+  (when (memq 'magit-update-uncommitted-buffers magit-post-refresh-hook)
+    (setq magit--uncommitted-buffers (magit-uncommitted-buffers))))
+
+(add-hook 'magit-pre-refresh-hook #'magit-maybe-cache-uncommitted-buffers)
+(add-hook 'magit-pre-call-git-hook #'magit-maybe-cache-uncommitted-buffers)
+(add-hook 'magit-pre-start-git-hook #'magit-maybe-cache-uncommitted-buffers)
+
+(defun magit-update-uncommitted-buffers ()
+  "Update file-visiting-buffers belonging to the current repository.
+Run `magit-update-uncommitted-buffer-hook' for each buffer
+which visits a file inside the current repository that had
+uncommitted changes before running the current Magit command
+and/or that does so now."
+  (dolist (buf (cl-union magit--uncommitted-buffers
+                         (magit-uncommitted-buffers) :test 'eq))
+    (run-hook-with-args 'magit-update-uncommitted-buffer-hook buf)))
+
 ;;; Save File-Visiting Buffers
 
 (defvar disable-magit-save-buffers nil)


lisp/magit-mode.el
```diff
@@ -777,6 +777,37 @@ Run hooks `magit-pre-refresh-hook' and `magit-post-refresh-hook'."
...
+
+(defun magit-update-uncommitted-buffers ()
+  "Update file-visiting-buffers belonging to the current repository.
+Run `magit-update-uncommitted-buffer-hook' for each buffer
+which visits a file inside the current repository that had
+uncommitted changes before running the current Magit command
+and/or that does so now."
+  (dolist (buf (cl-union magit--uncommitted-buffers
+                         (magit-uncommitted-buffers) :test 'eq))
+    (run-hook-with-args 'magit-update-uncommitted-buffer-hook buf)))
```
dgutov: Why not `(with-current-buffer buf (run-hook 'magit-update-uncommitted-buffer-hook))`?

`diff-hl-update` does not expect an argument. But once I used a different function that does, it started working, thanks.

dgutov:
> Running find-buffer-visiting (twice!) continues to be a potential bottleneck. But I think the main bottleneck is whatever functions are being added to the hook.

Calling git twice to get the list of files is not a bottleneck? Anyway, you could merge the lists (as strings) before calling `find-buffer-visiting` on each element. That would mean a sort and a `delete-consecutive-dups` call. Probably a bit better for long lists of files NlogN + N + M*N, rather than 2*M*N.

dgutov:
> Magit commands can be used even when no status buffer exists. So that's not an option.

Too bad, I rather liked the idea of zero overhead. I'd even be fine with only updating after the user presses `g` inside a status buffer. But it's your call.

Another thing I've noticed is that `magit-update-uncommitted-buffer-hook` gets called a bit more often than I'd expect: e.g., when I do `c c`, it's called both before I enter the commit message, and after.

If you'd like to return to the previous approach (for its performance benefits), I think I figured out how to make it work: even though `magit-modified-files` only returns the currently "outdated" files, `diff-hl` can walk all buffers by itself anyway, and for any buffer that contains highlighting but is not in the list returned by Magit, simply remove the highlighting (or conservatively call `diff-hl-update`, to be really sure).

tarsius:
> Calling git twice to get the list of files is not a bottleneck?

```
$ git ls-files | wc -l                                  
3723
$ git diff-files --name-only --ignore-submodules | wc -l
1428
$ git diff-files -z --name-only --ignore-submodules  0.00s user 0.01s system 100% cpu 0.008 total
```

Seems okay. (Actually `-z` is problematic with that many files over Tramp because it looks at the whole line using `looking-back`.)

> Anyway, you could merge the lists (as strings)

Good idea. I'll do that.

tarsius:
I think it would be best if you implemented all of this directly in `diff-hl`.

You initially requested a list of "affected buffers", presumingly in the hope that Magit could provide that for free. It can't. Then we discussed various ways to determine the set of buffers, what buffers exactly should be in the set, and how Magit should hand over that set to `diff-hl`. One thing is certain: there are multiple ways to do "it". And what might be right for `diff-hl` might not be right for other packages.

You also made it clear, that you were hoping that whatever approach we come up now would not change later again. The only two possible sets of files/buffers that I could imagine committing to are: (a) all tracked files (b) all modified files. But that's not where the discussion is headed:

> If you'd like to return to the previous approach (for its performance benefits), I think I figured out how to make it work: even though `magit-modified-files` only returns the currently "outdated" files, `diff-hl` can walk all buffers by itself anyway, and for any buffer that contains highlighting but is not in the list returned by Magit, simply remove the highlighting (or conservatively call `diff-hl-update`, to be really sure).

That would work for `diff-hl` but it might not for other packages. And since you are prepared to do what you are describing here in `diff-hl`, you might just as well get the list of modified files yourself.

The discussion above was useful, but I don't see how we can come up with a set of files/buffers which are both cheap to calculate and make sense to anyone but the two of us. I would have to pick one of the two and could not promise that the set would never change. Once other packages started using it it would very likely turn out that the set is, depending I chose the "cheap" or "sane" approach now, either insane or to expensive.

If you implement it in `diff-hl`, then there is no "interface" I have to commit too.

So I would suggest you just use `magit-post-refresh-hook` and then do whatever is necessary.

----

Some things we haven't talked about yet which you should also take into account:

* `diff-hl-mode` is a local mode which works by adding `diff-hl-update` to the buffer-local value of various hooks. The various magit hooks we have discussed are not buffer local, and so you will likely have to add a check which prevents `diff-hl-update` from being called in buffers in which `diff-hl-mode` is not enabled.

* A buffer which is in the set of buffers that have to be updated as discussed above, might also be among the buffers that are being reverted. Since `diff-hl-update` is also on `after-revert-hook`, that might mean that `diff-hl-update` is called twice in a given buffer. I recommend that you make sure that the same buffer is not needlessly updated twice using something like:

```lisp
(defvar diff-hl-updated-buffers nil)

(defun diff-hl-pre-command-hook ()
  (setq diff-hl-updated-buffers nil))

(add-hook 'pre-command-hook #'diff-hl-pre-command-hook)

(defun diff-hl-update ()
  (unless (member buffer-file-name buffer diff-hl-updated-buffers)
     (push buffer-file-name diff-hl-updated-buffers)
     ...))

```

dgutov:
Fair enough. Thank you for the discussion!

> check which prevents diff-hl-update from being called in buffers in which diff-hl-mode is not enabled

Naturally.

> (add-hook 'pre-command-hook #'diff-hl-pre-command-hook)

Yes, that's what I had in mind with "debouncing". Didn't get around to it yet, because normally `diff-hl-update` is pretty fast. 2 times "pretty fast" usually isn't slow either.

tarsius:
Thanks to you too!

dgutov:
Here's a minor problem I've encountered: when all changes in the current working directory are staged (no unstaged changes), `(magit-modified-files t)` returns nil, but `(magit-unstaged-files t)` returns the files with staged changes. Oxymoron?

It seems I should use the latter.

PR #2531, opened 2016-01-29
travis: quote CURL variable

dunn:
Otherwise everything after 'curl' is lost.  See https://travis-ci.org/magit/magit/builds/105755778#L109


tarsius:
Thanks! (This time I inserted a newline into the commit message :-)

PR #2537, opened 2016-02-03
fetch-popup: Add magit-fetch-branch

namhyung:
It seems there's no way to fetch from a specific branch in a remote
repository.  This is sometimes useful to review/test patchsets.

### DIFF
--- a/lisp/magit-remote.el
+++ b/lisp/magit-remote.el
@@ -183,6 +183,7 @@ Delete the symbolic-ref \"refs/remotes/<remote>/HEAD\"."
               (?u magit-get-remote         magit-fetch-from-upstream)
               (?e "elsewhere"              magit-fetch)
               (?a "all remotes"            magit-fetch-all)
+              (?b "specific branch"        magit-fetch-branch)
               "Fetch"
               (?m "submodules"             magit-submodule-fetch))
   :default-action 'magit-fetch
@@ -225,6 +226,15 @@ Delete the symbolic-ref \"refs/remotes/<remote>/HEAD\"."
   (interactive (list (magit-fetch-arguments)))
   (run-hooks 'magit-credential-hook)
   (magit-run-git-async "remote" "update" args))
+
+;;;###autoload
+(defun magit-fetch-branch (remote branch args)
+  "Fetch from the specific branch."
+  (interactive (list (magit-read-url "Fetch remote or url")
+                     (magit-read-branch-or-commit "Branch" "master")
+                     (magit-fetch-arguments)))
+  (run-hooks 'magit-credential-hook)
+  (magit-run-git-async "fetch" remote branch args))
 
 ;;;###autoload
 (defun magit-fetch-all-prune ()


tarsius:
What's wrong with just fetching everything from the respective remote?

tarsius:
Ah, didn't notice that this also allows specifying the url of the remote.

tarsius:
While I generally agree that this would be a useful addition, this still needs some work. I am currently [taking a break](http://emacsair.me/2016/02/10/magit-2.5). If that weren't so, then I would probably try to find solutions myself to the issues I mention below. That might not be the best solution anyway; instead I should discuss the issues I have found so that the person who opened the pr can fix them himself. In the past I often did not do so because it would often have been more work than just making the changes myself. But I should depart from that behavior, because it makes it less likely that we gain a contributor if I just do everything myself.

*As I said, I am on a break, and that includes a break from code review. In this one case I am doing it anyway. I mainly do so to have something to point to in other trivial pull requests, i.e. so that I can say "sure this is trivial and I could just merge it as is, but checkout this other trivial pr for how a trivial change isn't actually all that trivial".*

* ` (magit-read-url "Fetch remote or url")`
  * The configured remotes should be available as completion candidates.
  * Maybe some common prefixes such as `git@github.com:` and `git@github.com:<me>` should be offered as completion candidates.
* `(magit-read-branch-or-commit "Branch" "master")`
   * Why default to `master`? In most cases it will be something like `fix-some-issue`.
   * This should offer the available branches (or all refs?) from the remote.
     * If an existing remote was picked earlier, then use the local tracking information.
     * If an url was given, then query the remote.
   * Maybe it should be possible to insert multiple, comma-separated, branches and/or refspecs.
* Maybe there should be a single prompt expecting something like "<remote>/<branch>".
  * Elsewhere we use this form (instead of git's "REMOTE BRANCH"), i.e. one argument instead of two.
  * But here that doesn't work so well because we might want to allow multiple branches.
  * Also `REMOTE/FIRST-BRANCH,SECOND-BRANCH` would be strange and `URL/BRANCH` very strange and additionally ambiguous.
  * But we haven't decided yet, and we would want to avoid picking one approach now and later switching to the other. (Because users are complaining about "things changing to often".)

Anyway, I am on a break and will now go watch the Fc Basel play instead of listing other things to think about.

----

Also you haven't actually mentioned why you need this in the first place. Well, you mentioned that you need this to "review/test patchsets". That's pretty vague.

If you only need this in combination with Github for example, then `magit-branch-pull-request` from the `magit-rockstar` might do the trick. There is a reason that command is in `magit-rockstar`, not `magit` itself: I don't know exactly I want this to work, so I only put it in the "staging-ground package" (see its library commentary).

Also since your current implementation doesn't implement any goodies like completion candidates, using it like

```
f b git@example.com/foo.git RET feature RET
```

isn't actually more convenient than

```
: fetch git@example.com/foo.git feature RET
```

namhyung:
Thank you for the detailed reply.  I've read your article and it's sad to hear that you're considering finishing magit work.  Magit is a great tool and I've been using it with great satisfaction.  I hope you can keep working on it professionally.

Returning to the PR, yes this is a very basic implementation and provides no goodies yet.  Actually I'm not familiar with the elisp and had no enough time to digging the source code.  What you proposed is what I had in mind mostly but I just didn't know how to implement them.  It'd be great if you give me some pointers.

My use case is somewhat non-standard.  I work with some coworkers who have many branches.  I don't want to fetch all of them.  So I added the remote and manually edit the 'fetch' pattern in the .git/config file.  But I sometimes want to fetch new branch that's not in the fetch pattern.  In this case it can be useful to fetch from a single branch.  Also sometimes I need to play with patches from a randome contributor.

tarsius:
If started working on this myself now (I really shouldn't do that, but its *hard* breaking a habit), so I am closing this. I am afraid it would be more work for me to coach you then to just do it myself.

PR #2538, opened 2016-02-03
add --first-parent switch to magit-log

thepickle:


kyleam:
Thank you for the patch.  I do find this switch useful, but I'm not sure if it is useful enough to most users to be in the popup by default.  

I've added it to the wiki.

https://github.com/magit/magit/wiki/Additional-proposed-infix-arguments-and-suffix-commands#add-argument---first-parent

thepickle:
I beg to differ, but I use it a lot, especially since a great deal of my commit-history involves merges, which can make the (important) history difficult to look at.  I don't mind putting that snippet in my `.emacs` but I feel this is a feature most would find useful, and welcome.

thepickle:
Also it should be noted that this switch _was_ in the popup in the past.

kyleam:
> Also it should be noted that this switch was in the popup in the past.

It was removed in b9d9ba5 (magit-log-popup: remove most expert arguments, 2014-11-09).

thepickle:
Hmm can't say I'm a fan of removing it but ill yield.. as I'm sure you removed it for good reason.

PR #2547, opened 2016-02-11
merge-preview: recognize other merge-tree headers

kyleam:


lisp/magit-diff.el
```diff
@@ -1226,7 +1226,8 @@ is set in `magit-mode-setup'."
 
 (defconst magit-diff-headline-re
   (concat "^\\(@@@?\\|diff\\|Submodule\\|"
-          "\\* Unmerged path\\|merged\\|changed in both\\)"))
+          "\\* Unmerged path\\|merged\\|changed in both\\|"
+          "added in remote\\|removed in remote\\)"))
```
npostavs: I think there are enough alternatives that it would read better to use `regexp-opt` here.

kyleam: The first item uses a regexp.  I'm guessing that the group is used,
making it necessary to match all three `@`s rather than just a leading
two, but I haven't checked.

npostavs: > The first item uses a regexp.

Ah, missed that, sorry.

tarsius:
Looks good to me.

PR #2548, opened 2016-02-11
Small fix in documentation for `f u`

dato:


tarsius:
Thanks.

I have regenerated `magit.texi` and amended to result to your commit.

PR #2556, opened 2016-02-17
Added function magit-diff-buffer-file.

dov:
* Allows diffing the current buffer with another commit.
* Code based on magit-diff and magit-log-buffer-file.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -761,6 +761,20 @@ be committed."
                               (expand-file-name b))))
 
 ;;;###autoload
+(defun magit-diff-buffer-file (rev-or-range)
+  "Show changes between commits for current file
+
+REV-OR-RANGE should be a range or a single revision.  If it is a
+revision, then show changes in the working tree relative to that
+revision.  If it is a range, but one side is omitted, then show
+changes relative to `HEAD'."
+  (interactive
+   (list (magit-diff-read-range-or-commit "File diff for range" nil current-prefix-arg)))
+  (-if-let (file (magit-file-relative-name))
+      (magit-diff-setup rev-or-range nil magit-diff-arguments (list file))
+    (user-error "Buffer isn't visiting a file")))
+
+;;;###autoload
 (defun magit-show-commit (rev &optional args files module)
   "Show the revision at point.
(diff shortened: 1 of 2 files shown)


kyleam:
Thanks for the PR.

For the commit message:
s/Added/Add/ and drop period.  Or "magit-diff-buffer-file: new command"

Also, I'd be ok dropping the message body.

Other comments inline.

lisp/magit-diff.el
```diff
@@ -761,6 +761,20 @@ be committed."
                               (expand-file-name b))))
 
 ;;;###autoload
+(defun magit-diff-buffer-file (rev-or-range)
+  "Show differences between the buffer file and the same file in another commit.
+
```
kyleam: Maybe "Show changes between commits for current file."  Like
magit-diff, that's incomplete because it's between two commits or
between a commit and the working tree, but that gets a bit wordy and
is explained in the next paragraph of the doc-string.

#2556

lisp/magit-diff.el
```diff
@@ -761,6 +761,20 @@ be committed."
...
+REV-OR-RANGE should be a RANGE or a single revision.  If it is a
```
kyleam: "range" shouldn't be capitalized here because it isn't an argument.
(I'll fix this in magit-diff's doc-string.)

#2556

lisp/magit-diff.el
```diff
@@ -761,6 +761,20 @@ be committed."
...
+revision, then show changes in the working tree relative to that
+revision.  If it is a range, but one side is omitted, then show
+changes relative to `HEAD'."
+  (interactive
+   (list (magit-diff-read-range-or-commit "File diff for range" nil current-prefix-arg)))
```
kyleam: You can drop current-prefix-arg as an argument because that argument
of magit-diff-read-range-or-commit is only relevant for commit lines
selected in a region.  Also, I think "HEAD" might be a good default to
pass as the SECONDARY-DEFAULT argument.

kyleam: Ignore the bit about passing "HEAD" as SECONDARY-DEFAULT.
magit-diff-read-range-or-commit will fall back to using the name of
current branch.

PR #2562, opened 2016-02-22
Add function magit-diff-buffer-file

dov:
Fixed comments from #2556

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -761,6 +761,20 @@ be committed."
                               (expand-file-name b))))
 
 ;;;###autoload
+(defun magit-diff-buffer-file (rev-or-range)
+  "Show changes between commits for current file
+
+REV-OR-RANGE should be a range or a single revision.  If it is a
+revision, then show changes in the working tree relative to that
+revision.  If it is a range, but one side is omitted, then show
+changes relative to `HEAD'."
+  (interactive
+   (list (magit-diff-read-range-or-commit "File diff for range" nil current-prefix-arg)))
+  (-if-let (file (magit-file-relative-name))
+      (magit-diff-setup rev-or-range nil magit-diff-arguments (list file))
+    (user-error "Buffer isn't visiting a file")))
+
+;;;###autoload
 (defun magit-show-commit (rev &optional args files module)
   "Show the revision at point.
 If there is no revision at point or with a prefix argument prompt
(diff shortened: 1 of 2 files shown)


lisp/magit-diff.el
```diff
@@ -761,6 +761,20 @@ be committed."
...
 ;;;###autoload
+(defun magit-diff-buffer-file (rev-or-range)
+  "Show changes between commits for current file
+
+REV-OR-RANGE should be a range or a single revision.  If it is a
+revision, then show changes in the working tree relative to that
+revision.  If it is a range, but one side is omitted, then show
+changes relative to `HEAD'."
+  (interactive
+   (list (magit-diff-read-range-or-commit "File diff for range" nil current-prefix-arg)))
```
kyleam: Please see my [comment][1] in #2556 about current-prefix-arg.

#2562

[1]: https://github.com/magit/magit/pull/2556#discussion_r53424522

kyleam:
Thanks for the update.  There's no need to open a new PR.  Please just
force push.

tarsius:
Maybe, instead of reading a rev or range, this should just show the unstage changes and with a prefix argument the staged changes. Or maybe just always the same as `magit-diff-working-tree`. Currently it is not possible to show the staged changes, which doesn't seem good.

Maybe with two arguments it could read a rev or range. Theoretically we could even add the same four variants (staged, unstaged, working-free, rev-or-range) to this popup as for the regular popup, but this seems like to much.

Which diff do you actually usually want to show?

tarsius:
A completely different approach would be to replace the last `nil` in `magit-diff-arguments` with `(list (magit-file-relative-name))` and bind `d` in `magit-file-popup` to `magit-diff-popup`.

dov:
The use case that I had in mind when I wrote magit-file-diff is manually picking changes from another branch. I.e. lets say that I made a series of changes to the Develop branch and I now want to copy these to the Integ branch. So I checkout Integ and visit the files to where I want to import changes, do the equivalent of `git diff Develop foo.cc` and then use the revert-hunk command in the diff buffer to import the requested changes. 

So my use case is a comparing with a different branch. 

Note that for viewing the changes vs the last commit there is also `vc-ediff`. But typically I just do magit-status and use tab expand to view the changes to the files.

tarsius:
I've taken a quick stab at "a completely different approach": https://github.com/magit/magit/compare/jb/file-diff-popup?expand=1. Please give it a try.

kyleam:
> I've taken a quick stab at "a completely different approach":

Seems to work nicely.  Like @dov, I was only interested in being able
diff an explicit range and didn't mind not being able to diff the
index, but getting the whole diff popup limited to a file is great.

tarsius:
I've pushed 230ab992fbbb6d0dbb7164ce87d64834c5276719 instead of this.

PR #2563, opened 2016-02-22
Replace setenv with push on process-environment

npostavs:
Let-binding process-environment does not nessarily protect from changes
made by setenv, because setenv may destructively modify the list object.

See https://github.com/magit/magit/issues/2246#issuecomment-139842749 until https://github.com/magit/magit/issues/2246#issuecomment-139882266

---

Unfortunately, I forgot about this until after `with-editor` was split from the magit repo, so the fix is split over the repos now, see magit/with-editor#3.


tarsius:
Thanks for coming back to this. I have cherry-picked the commit.

PR #2568, opened 2016-02-28
Fixed magit-current-file for magit-log-mode

killdash9:
The structure of magit-refresh-args has apparently changed.


tarsius:
Thanks!

I have applied your commit with some changes to the commit message.

> The structure of magit-refresh-args has apparently changed.

I am not sure that's true. I am guessing `3` was always wrong. But that element probably was an atom at one point.

It also appeared there was something wrong with the commit object. A newline was missing at the end of the commit message, I think.

PR #2569, opened 2016-02-28
Allow falling back from Ido to regular completion

tarsius:
It has annoyed me for a long time that it was not possible to fall back from Ido completion to regular completion (except when using the file and buffer completion functions defined in `ido` itself).  In the case of Magit that meant that in some cases it was not possible to enter a value which was not among the completion candidates.

I've fixed that for Magit now. The implementation is quite a hack and seems to work reliably. It's also inspired by (and uses) code in `ido`, so it appears that's how this is done. I have also checked whether `ido-ubiquitous` provides a generalized variant of this but couldn't find anything.

@DarwinAwardWinner Did I overlook something? What do you think about this approach? And if you think this is the way to go, would you be willing to add a generalized variant of this to `ido-ubiquitous`?

### DIFF
--- a/Documentation/RelNotes/2.5.1.txt
+++ b/Documentation/RelNotes/2.5.1.txt
@@ -20,6 +20,9 @@ Updates since v2.5.0
 * Added new command `magit-reset-popup' featuring all the available
   reset variants.
 
+* When using Ido completion, then it's now possible to fall back to
+  regular completion using `C-x C-f' or `C-x C-b' if necessary.
+
 * When washing the diff that is to be displayed while writing a
   commit message takes too long, then it is now possible to abort that
   by pressing `C-g'.  Previously that would have aborted the commit.
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -231,17 +231,35 @@ results in additional differences."
                    choices predicate require-match
                    initial-input hist def))
 
+(defvar magit-ido-use-fallback nil)
+
+(defvar magit-ido-completion-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map "\C-x\C-b" 'magit-ido-fallback)
+    (define-key map "\C-x\C-f" 'magit-ido-fallback)
+    map)
+  "Keymap used by `magit-ido-completing-read'.
+`ido-common-completion-map' is used as parent.")
+
 (defun magit-ido-completing-read
   (prompt choices &optional predicate require-match initial-input hist def)
-  "Ido-based `completing-read' almost-replacement.
+  "Magit wrapper for `ido-completing-read+' function.
 
 Unfortunately `ido-completing-read' is not suitable as a
-drop-in replacement for `completing-read', instead we use
+drop-in replacement for `completing-read', instead we wrap
 `ido-completing-read+' from the third-party package by the
 same name."
   (if (require 'ido-completing-read+ nil t)
-      (ido-completing-read+ prompt choices predicate require-match
-                            initial-input hist def)
+      (if magit-ido-use-fallback
+          (progn (setq magit-ido-use-fallback nil)
+                 (magit-builtin-completing-read
+                  prompt choices predicate require-match
+                  initial-input hist def))
+        (let ((ido-common-completion-map
+               (make-composed-keymap magit-ido-completion-map
+                                     ido-common-completion-map)))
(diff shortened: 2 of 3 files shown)


DarwinAwardWinner:
My ido-completing-read+ package actually already has a generalized fallback mechanism to standard completion, using the standard ido fallback triggers of C-f at the end of input or C-b at the beginning. If you're already using ido-completing-read+, just try C-f at the end of input and it should fall back.

DarwinAwardWinner:
I have to admit that I don't completely remember how my fallback code works, but that's ok, because it passes the [tests](https://github.com/DarwinAwardWinner/ido-ubiquitous/blob/master/test/ido-ubiquitous-test.el#L519-L547).

tarsius:
We are talking about `ido-magic-forward-char` here, right?

That doesn't fallback to "same command but with regular completion". It falls back to `find-file`, which isn't what I want when trying to complete, say, a branch :-)

What your code does is to teach `ido-magic-forward-char` to use `ido-fallback-command`. But that's not enough because that in turn only makes sure that the `ido-exit` method is `fallback`. Unfortunately `fallback` is ignored when it would be most useful (when completing a `list` item).

`ido-file-internal` and `ido-buffer-internal` know how to call `this-command` again when `ido-exit` is `fallback`, but generic `ido-completing-read` and basic `ido-read-internal` do not. Actually, since the same command might read multiple values with completion from the user, this wouldn't be good enough anyway.

After jumping around an hour in `ido.el` I've come up with the following. It shouldn't be to hard to do this in `ido-completing-read+` instead of `ido-completing-read`:

```lisp
(defun ido-completing-read (prompt choices &optional predicate require-match
                            initial-input hist def inherit-input-method)
  (let ((ido-current-directory nil)
	(ido-directory-nonreadable nil)
	(ido-directory-too-big nil)
	(ido-context-switch-command 'ignore)
	(ido-choice-list choices))
    ;; Initialize ido before invoking ido-read-internal
    (ido-common-initialization)
    (let ((value (ido-read-internal 'list prompt hist def
                                    require-match initial-input)))
      ;; ########### NEW FROM HERE
      (if (eq ido-exit 'fallback)
          (let ((completing-read-function
                 (or (bound-and-true-p ido-cr+-fallback-function)
                     'completing-read-default)))
            (completing-read prompt choices predicate require-match
                             initial-input hist def inherit-input-method))
        value))))

(defun ido-magic-forward-char (arg)
  (interactive "P")
  (cond
   ((or arg (not (eobp)))
    (forward-char (min (prefix-numeric-value arg)
		       (- (point-max) (point)))))
   ((memq ido-cur-item '(file dir list)) ; ########### LIST IS NEW
    (ido-fallback-command))
   (ido-context-switch-command
    (call-interactively ido-context-switch-command))
   ((eq ido-cur-item 'buffer)
    (ido-enter-find-file))))
```

It would also be a good idea to bind `C-x C-f` and `C-x C-b` in `ido-common-completion-map`, but that can be left to the user. I also noticed that, at least in some cases, I could just have used `ido-edit-input`, but the above is better.

npostavs:
> In the case of Magit that meant that in some cases it was not possible to enter a value which was not among the completion candidates.

Doesn't `C-j` always work for that?

DarwinAwardWinner:
> We are talking about `ido-magic-forward-char` here, right?

Correct, ido-completing-read+ sets up `ido-magic-forward-char` and `ido-magic-backward-char` to trigger falling back. See https://github.com/DarwinAwardWinner/ido-ubiquitous/blob/master/ido-completing-read%2B.el#L267-L279. But it also catches the case where `ido-exit` is set to `fallback` and responds appropriately, here: https://github.com/DarwinAwardWinner/ido-ubiquitous/blob/master/ido-completing-read%2B.el#L219-L226. The tests that I linked to verify that we get the behavior of regular `completing-read` after pressing C-f at end of input, and C-b at beggining of input. You can verify that the manual fallback is working by enabling `ido-cr+-debug-mode`. You should see a message like this when falling back: `ido-completing-read+: Falling back to `completing-read-default' because user manually triggered fallback.`

DarwinAwardWinner:
Also, C-f already works as expected for me to fallback to regular `completing-read` in `magit-read-branch`, for example.

tarsius:
> Also, C-f already works

Indeed. Sorry for the noise.

Issue #2575, opened 2016-03-03
Let magit-ediff-dwim behave more usefully over a stash


michael-heerdegen:
Hi,

currently, the command doesn't guess anything and uses the most general fallback ("can't read your mind" etc.).  The user will most likely  want to ediff against the stash's parent, but maybe against an arbitrary different commit.

I think it should be doable quite easily: in `magit-ediff-dwim` replace the

```
(commit . ,value)
```

pattern with

```
(,(or `commit `stash) . ,value)
```

to treat it the same as commits.  You may also want to change `magit-diff-read-range-or-commit` similarly for that purpose.

Thanks,

Michael.

kyleam:
Thank you.  I think both of those are good suggestions and you proposed changes should work.  Want to submit a PR?

tarsius:
It's a bit more complicated than that.

* Multiple files might have changed. Unless only one has changed we have to ask the user.
* A stash may consist of up to 3 commits (but let's ignore the "new files" commit)

I have written the boilerplate, but reading the file is still missing: https://github.com/magit/magit/compare/jb/ediff-stash?expand=1.

kyleam:
> Multiple files might have changed. Unless only one has changed we have to ask the user.

That's true for the commit too, but `magit-ediff-compare--read-files` takes care of it in `magit-ediff-show-commit`.  It seems to handle `stash@{N}` fine.

kyleam:
> A stash may consist of up to 3 commits (but let's ignore the "new files" commit)

I think the initial suggest would allow you to select a file for ediffing the working tree relative to `stash@{N}`, as with `git diff stash@{N}`.

michael-heerdegen:
Kyle Meyer <notifications@github.com> writes:

> Thank you. I think both of those are good suggestions and you proposed
> changes should work. Want to submit a PR?

I'm not sure if I have sufficing knowledge about the internals...  Let's
discuss the stuff a bit more for now.

michael-heerdegen:
Jonas Bernoulli <notifications@github.com> writes:

> It's a bit more complicated than that.
>
> * Multiple files might have changed. Unless only one has changed we have to ask the
>  user. 
> * A stash may consist of up to 3 commits (but let's ignore the "new
>  files" commit)

I guess you mean workspace vs. index changes?  (Where can I read about
how this works internally, btw?)

I'm not sure which behavior would be the most useful, but at the first
look, using ediff3 seems a bit of overkill to me.


[...]

PR #2581, opened 2016-03-06
Stash ediffing

kyleam:
Re: #2575

### DIFF
--- a/Documentation/RelNotes/2.6.0.txt
+++ b/Documentation/RelNotes/2.6.0.txt
@@ -33,6 +33,10 @@ Updates since v2.5.0
 * Added new command `magit-reset-popup' featuring all the available
   reset variants.
 
+* Added new `magit-ediff-show-stash', which is available in the Ediff
+  popup and will be used when `magit-ediff-dwim' is called with point
+  on a stash.
+
 * Added new section inserters
   `magit-insert-modules-unpulled-from-pushremote' and
   `magit-insert-modules-unpushed-to-pushremote', and renamed
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -2247,6 +2247,13 @@ without having to using one of the diff popups.
   ~magit-ediff-show-unstaged~ to show staged and unstaged changes,
   respectively.
 
+- User Option: magit-ediff-show-stash-with-index
+
+  This option controls whether ~magit-ediff-show-stash~ includes a
+  buffer containing the file's state in the index at the time the
+  stash was created.  This makes it possible to tell which changes in
+  the stash were staged.
+
 - User Option: magit-ediff-quit-hook
 
   This hook is run after quitting an Ediff session that was created
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -2235,6 +2235,10 @@ without having to using one of the diff popups.
 
   Show changes to a file introduced by a commit using Ediff.
 
+- Key: E z, magit-ediff-show-stash
+
+  Show changes to a file introduced by a stash using Ediff.
+
 - User Option: magit-ediff-dwim-show-on-hunks
 
   This option controls what command ~magit-ediff-dwim~ calls when
--- a/Documentation/magit.texi
+++ b/Documentation/magit.texi
@@ -3032,6 +3032,14 @@ point is on uncommitted hunks.  When nil, always run
 respectively.
 @end defopt
 
+@defopt magit-ediff-show-stash-with-index
+
+This option controls whether @code{magit-ediff-show-stash} includes a
+buffer containing the file's state in the index at the time the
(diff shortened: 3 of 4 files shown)


tarsius:
Looks good to me. I only did some limited testing, but that it passed at least.

kyleam:
Thanks for you comments.  I'll incorporate them and merge tomorrow unless @michael-heerdegen has any additional feedback.

michael-heerdegen:
Kyle Meyer <notifications@github.com> writes:

> Thanks for you comments. I'll incorporate them and merge tomorrow
> unless @michael-heerdegen has any additional feedback.

Hi, sorry, I could not yet have a closer look at the stuff due to the
flu in the family.  I'll try to have a look today.


Thanks so far,

Michael.

kyleam:
> Hi, sorry, I could not yet have a closer look at the stuff due to the
> flu in the family. I'll try to have a look today.

OK, no rush.  I'll wait for you to try it out before applying.

michael-heerdegen:
Kyle Meyer <notifications@github.com> writes:

> OK, no rush. I'll wait for you to try it out before applying.

I had a look now.  Looks very good so far, I like it, thanks!


The only little "discrepancy" I found:

The doc of magit-ediff-show-stash-with-index says:

| * stash@{N}^1 represents the HEAD commit at the time the stash
|   was created.

But when I create a stash with z W (aka "snapshot worktree"),
stash@{N}^1 is actually a "pre stash index" - dunno if it comes from git
itself or from Magit.  So you get an (two way) ediff stashed worktree
vs. index, which was a bit surprising to me.  But maybe this is
expected?


Btw, when you will have merged this PR, will this change appear
in Melpa?


Thanks again - regards,

Michael.

kyleam:
> I had a look now. Looks very good so far, I like it, thanks! 

Thanks for having a look.

> The only little "discrepancy" I found:
>
> The doc of magit-ediff-show-stash-with-index says:
>
> | * stash@{N}^1 represents the HEAD commit at the time the stash
> |   was created.
>
> But when I create a stash with z W (aka "snapshot worktree"),
> stash@{N}^1 is actually a "pre stash index" - dunno if it comes from git
> itself or from Magit.

It's in Magit's implementation of the snapshot functionality.

> So you get an (two way) ediff stashed worktree
> vs. index, which was a bit surprising to me.  But maybe this is
> expected?

It is expected/intentional in that `stash@{N}^1..stash@{N}` gives you
the changes that were in the working tree.  This is the same as
diffing `stash@{N}^^..stash@{N}` and filtering out changes that were
staged, unless a staged changed involved the same hunk as an unstaged
change.

However, as you point out, it does make the description wrong (though
right in spirit) in the case of a working tree snapshot.  I'm not sure
what the best way to clarify this is.

> Btw, when you will have merged this PR, will this change appear
> in Melpa?

Yes.

michael-heerdegen:
Kyle Meyer <notifications@github.com> writes:

> However, as you point out, it does make the description wrong (though
> right in spirit) in the case of a working tree snapshot. I'm not sure
> what the best way to clarify this is.

Are there more such "corner cases"?


>  Btw, when you will have merged this PR, will this change appear
>  in Melpa?
>
> Yes.

Great! :-)

kyleam:
> It's in Magit's implementation of the snapshot functionality.

Sorry, I said the wrong thing there.  It's Magit's implementation for
letting you make a stash of just unstaged changes (that is,
magit-stash-worktree and magit-snapshot-worktree).

kyleam:
> Are there more such "corner cases"?

I think the two worktree functions are it.

kyleam:
I suppose it's worth noting that I find `magit-stash-worktree` and
`magit-snapshot-worktree` confusingly named because they capture just
the unstaged changes, not all changes in working tree.

kyleam:
The pushed update incoporates @tarsius's suggestions.

michael-heerdegen:
Kyle Meyer <notifications@github.com> writes:

> I suppose it's worth noting that I find magit-stash-worktree and
> magit-snapshot-worktree confusingly named because they capture just
> the unstaged changes, not all changes in working tree.

Indeed, such things are very confusing especially to beginners.  And the
docstring doesn't explain this as well.

Would "magit-snapshot-unstaged" be a better name?

tarsius:
I agree that we should consider renaming these functions and/or clarifying the documentation, but that shouldn't block the merge of this.

michael-heerdegen:

> I agree that we should consider renaming these functions and/or
> clarifying the documentation, but that shouldn't block the merge of
> this.

Great, than - let's merge!

Shall I create a new issue for the confusing command names?

kyleam:
Merged.

> Shall I create a new issue for the confusing command names?

There are at least a couple other cases where worktree is used this
way, so I can take care of looking through those and opening an issue
or PR for discussion.

Issue #2583, opened 2016-03-09
Mishandling of files with CRLF line endings when staging regions/chunks


luismbo:
Staging parts of files with CRLF line endings ends up staging changes with LF line endings. If this is a known issue (couldn't find it in the FAQ), is there a workaround?

Here's how to reproduce:

```
$ git init
Initialized empty Git repository in C:/tmp/repo/.git/
$ git config core.autocrlf false
$ echo "line 1" > file
$ unix2dos file
unix2dos: converting file file to DOS format...
$ git add file
$ git commit -m 'initial commit'
[master (root-commit) 45305a8] initial commit
 1 file changed, 1 insertion(+)
 create mode 100644 file
$ echo "line 2" >> file
$ echo "line 3" >> file
$ unix2dos file
unix2dos: converting file file to DOS format...
```

At this point, `magit-status` displays something like this:

```diff
Head:     master initial commit

Unstaged changes (1)
modified   file
@@ -1 +1,3 @@
 line 1
+line 2
+line 3
```

So far so good, but if I stage the changed hunk (rather than the whole file), e.g. by placing the cursor on the `@@ -1 +1,3 @@` line and pressing <kbd>s</kbd>, I get the following result:

```diff
Head:     master initial commit

Unstaged changes (1)
modified   file
@@ -1,3 +1,3 @@
 line 1
-line 2
-line 3
+line 2
+line 3

Staged changes (1)
modified   file
@@ -1 +1,3 @@
 line 1
+line 2
+line 3

```

This happens because the staged file has LF lines instead of CRLF lines and the magit-status buffer's EOL type is CRLF. **Update:** turns out the EOL type of the buffer is irrelevant, what matters is `default-process-coding-system`. (See discussion in #2584 and #2587.)

Tested using GNU Emacs 24.5.1 (i686-pc-mingw32) and magit 20160305.2218.

luismbo:
I took a stab at fixing this, but my fix is bogus since I then get the same problem when editing files with LF endings. Perhaps the right the thing to do is to peek at the file to determine what the EOL type is? Comments/guidance would be much appreciated! Thanks. :-)

```diff
diff --git a/lisp/magit-apply.el b/lisp/magit-apply.el
index 919e55c..19242ff 100644
--- a/lisp/magit-apply.el
+++ b/lisp/magit-apply.el
@@ -92,16 +92,16 @@
                      (mapconcat
                       (lambda (s)
                         (concat (magit-diff-file-header s)
[...]

Issue #2403, opened 2015-11-24
Can not stage hunk in Windows


makble:
Before staging
<img width="452" alt="magit-stage" src="https://cloud.githubusercontent.com/assets/5926356/11357545/35cb605e-92a3-11e5-9600-f8a0606c5c55.png">
After staging
<img width="451" alt="magit-stage2" src="https://cloud.githubusercontent.com/assets/5926356/11357548/44abf9f8-92a3-11e5-8abc-46c422acab19.png">

Stage it again will freeze Emacs. 

The reason is the CR in diff patch is stripped off when sent to git.exe via following command
```javascript
git --no-pager --literal-pathspecs -c core.preloadindex=true apply --cached -p0 --ignore-space-change -
diff --git others.el others.el
index d68ec21..ef012d2 100644
--- others.el
+++ others.el
@@ -1,3 +1,5 @@
+; 
```

To fix it run following code in ielm
```lisp
(setq default-process-coding-system (cons 'utf-8-dos 'utf-8-dos))
```

To reproduce the situation in command line, save the diff patch as  file and delete all CR in the file and run
```javascript
git --no-pager --literal-pathspecs -c core.preloadindex=true apply --cached -p0 --ignore-space-change patch.txt
```
Not a bug of Magit itself, but confusing when it happens, better to add the issue to manual or display a warning message when staging.

npostavs:
Is this the same as #2401?

TomAndrews:
This looks very similar to the problem I was having in #2401.  

@makble Please could you give me some more details on your workaround?  I tried executing 

```
(setq default-process-coding-system (cons 'utf-8-dos 'utf-8-dos))
```
in `ielm` but it still wouldn't stage the hunk properly.

makble:
Check the ELISP code below, I added debug statements in important places. When you stage, the text that actually sent to git will be saved to a file, you can check the file, if your situation is the same as mine you will find there is no CR in that file.

If my solution doesn't work, you can trace the code to find out when and where the patch is modified.

```lisp
(defun magit-start-process (program &optional input &rest args)
  "Start PROGRAM, prepare for refresh, and return the process object.

If optional argument INPUT is non-nil, it has to be a buffer or
the name of an existing buffer.  The buffer content becomes the
processes standard input.

The process is started using `start-file-process' and then setup
to use the sentinel `magit-process-sentinel' and the filter
`magit-process-filter'.  Information required by these functions
is stored in the process object.  When this function returns the
[...]

Issue #2401, opened 2015-11-19
Staging hunk leaves an unstaged "no change" change


TomAndrews:
I have an intermittent problem with magit on Windows whereby I stage a hunk it leaves an unstaged "change" with no differences.  For example staging the first hunk in the file below:

![before](https://cloud.githubusercontent.com/assets/13886890/11277165/39b9ed08-8edd-11e5-97d6-ccf98626c148.png)

Gives this as a result:
![after](https://cloud.githubusercontent.com/assets/13886890/11277226/7eac1a58-8edd-11e5-92fe-a954543e9974.png)

Note that the hunk has staged (see the bottom of the picture) but the 'Unstaged changes' section is showing a diff but with no change.

If I unstage the staged hunk, it goes back to the state in the first picture (as you'd expect).  If I then try staging the hunk using `git add -i` from the command line and refresh the magit status buffer then everything seems to have worked as expected.

I'm afraid I haven't worked out how to reproduce this problem.  It only seems to happen on windows (I also use magit on linux but have never seen the problem there).  If I restart emacs then the problem goes away for a while.  I am using the latest master from github, compiled using `C-u 0 M-x byte-recompile-directory`.  I am using emacs 0.25.0.50.1 from http://sourceforge.net/projects/emacsbinw64/

I realise you don't really stand a chance of helping me fix this when I can't reliably reproduce the problem.  If you could let me know of anything I can try when the problem does occur to try and narrow down the problem I'm keen to try!

Thanks very much for your help (and for writing such a great piece of software!).

johnmastro:
@TomAndrews, is it possible you have any raw bytes in your file? Something like this happened to me once and it turned out to be because I'd ended up with a null byte in the file, causing `git` to detect it as a binary file. I deleted the null byte and things went back to normal - I didn't pursue it further than that.

Hm, however, I just tried to recreate the issue, unsuccessfully.

npostavs:
> Note that the hunk has staged (see the bottom of the picture) but the 'Unstaged changes' section is showing a diff but with no change.

Is the change gone or just collapsed? i.e. Does doing `M-4` restore it?

TomAndrews:
@johnmastro Thanks for the suggestion.  I tried this: http://www.emacswiki.org/emacs/FindingNonAsciiCharacters
but it didn't find any non-ascii characters in my file

TomAndrews:
@npostavs Sorry, I don't understand.  I'm trying to stage the first hunk in the first picture, beginning

[...]

PR #2584, opened 2016-03-09
magit-apply: add literal CR chars if necessary

luismbo:
A tentative fix for #2401, #2403 and #2583.

### DIFF
--- a/lisp/magit-apply.el
+++ b/lisp/magit-apply.el
@@ -86,22 +86,24 @@ With a prefix argument and if necessary, attempt a 3-way merge."
       (`(,_   file) (magit-apply-diff   it args))
       (`(,_  files) (magit-apply-diffs  it args)))))
 
+(defun magit-apply--section-content (section)
+  (buffer-substring (magit-section-start section)
+                    (magit-section-end section)))
+
 (defun magit-apply-diffs (sections &rest args)
   (setq sections (magit-apply--get-diffs sections))
   (magit-apply-patch sections args
                      (mapconcat
                       (lambda (s)
                         (concat (magit-diff-file-header s)
-                                (buffer-substring (magit-section-content s)
-                                                  (magit-section-end s))))
+                                (magit-apply--section-content s)))
                       sections "")))
 
 (defun magit-apply-diff (section &rest args)
   (setq section (car (magit-apply--get-diffs (list section))))
   (magit-apply-patch section args
                      (concat (magit-diff-file-header section)
-                             (buffer-substring (magit-section-content section)
-                                               (magit-section-end section)))))
+                             (magit-apply--section-content section))))
 
 (defun magit-apply-hunks (sections &rest args)
   (let ((section (magit-section-parent (car sections))))
@@ -109,19 +111,15 @@ With a prefix argument and if necessary, attempt a 3-way merge."
       (user-error "Cannot un-/stage resolution hunks.  Stage the whole file"))
     (magit-apply-patch section args
                        (concat (magit-section-diff-header section)
-                               (mapconcat
-                                (lambda (s)
-                                  (buffer-substring (magit-section-start s)
-                                                    (magit-section-end s)))
-                                sections "")))))
+                               (mapconcat 'magit-apply--section-content
+                                          sections "")))))
 
 (defun magit-apply-hunk (section &rest args)
   (when (string-match "^diff --cc" (magit-section-parent-value section))
     (user-error "Cannot un-/stage resolution hunks.  Stage the whole file"))
   (magit-apply-patch (magit-section-parent section) args
                      (concat (magit-diff-file-header section)
-                             (buffer-substring (magit-section-start section)
-                                               (magit-section-end section)))))
+                             (magit-apply--section-content section))))
 
 (defun magit-apply-region (section &rest args)
   (unless (magit-diff-context-p)
--- a/lisp/magit-apply.el
+++ b/lisp/magit-apply.el
@@ -67,6 +67,38 @@ information see command `magit-reverse-in-index'."
   :group 'magit-commands
   :type 'boolean)
 
+(defcustom magit-preserve-crlf-line-endings t
+  "Whether to try harder to preserve CRLF line endings.
+
+Magit buffers may contain diffs for more than one file and these
+files might use different line ending types, but Emacs normalizes
+file endings in a buffer to a single newline character. When the
+diffs have been constructing using a `process-coding-system' with
+LF line endings no information is lost: each sequence of CR and
+LF bytes in a diff is converted into a sequence of ?\r and ?\n
+characters; but, when `process-coding-system' has CRLF line
+endings, each CRLF sequence in the diff is converted into a
+single ?\n character. Because both LF and CRLF sequences are
+normalized into a single ?\n character, ambiguity ensues.
+
+This isn't an issue when staging (or otherwise applying) complete
+files, because the buffer contents are not used in that case.
+But when applying hunks, then a patch has to be built from parts
+of the buffer contents.
+
+If this option is non-nil (the default), then Magit peeks the
+file that is being changed to determine whether its line ending
+style is CRLF.  If so, then CR characters are inserted into the
+patch that is being sent to git.
+
+Only Windows users are likely to be affected by this variable.
+
+Also see the Git variables `core.eol', `core.safecrlf', and
+`core.autocrlf' for an alternative approach."
+  :package-version '(magit . "2.6.0")
+  :group 'magit
+  :type 'boolean)
+
 ;;; Commands
 ;;;; Apply
 
@@ -86,9 +118,25 @@ With a prefix argument and if necessary, attempt a 3-way merge."
       (`(,_   file) (magit-apply-diff   it args))
       (`(,_  files) (magit-apply-diffs  it args)))))
 
+(defun magit-apply--maybe-fix-up-eols (section string)
+  (if (and magit-preserve-crlf-line-endings
+           ;; diff was read using CRLF eol type
+           (eql 1 (coding-system-eol-type (car default-process-coding-system)))
+           ;; file has CRLF eol type, too
+           (with-temp-buffer
+             (insert-file-contents-literally (magit-section-value section)
+                                             nil 0 500)
+             (search-forward "\r" nil t)))
+      (replace-regexp-in-string "\n" "\r\n" string)
+      string))
+
 (defun magit-apply--section-content (section)
-  (buffer-substring-no-properties (magit-section-start section)
-                                  (magit-section-end section)))
+  (magit-apply--maybe-fix-up-eols
+   (pcase (magit-section-type section)
+     ('file section)
+     ('hunk (magit-section-parent section)))
+   (buffer-substring-no-properties (magit-section-start section)
+                                   (magit-section-end section))))
 
 (defun magit-apply-diffs (sections &rest args)
   (setq sections (magit-apply--get-diffs sections))
@@ -126,9 +174,11 @@ With a prefix argument and if necessary, attempt a 3-way merge."
     (user-error "Not enough context to apply region.  Increase the context"))
   (when (string-match "^diff --cc" (magit-section-parent-value section))
     (user-error "Cannot un-/stage resolution hunks.  Stage the whole file"))
-  (magit-apply-patch (magit-section-parent section) args
-                     (concat (magit-diff-file-header section)
-                             (magit-diff-hunk-region-patch section args))))
+  (let ((parent (magit-section-parent section))
+        (patch (magit-diff-hunk-region-patch section args)))
+    (magit-apply-patch parent args
+                       (concat (magit-diff-file-header section)
+                               (magit-apply--maybe-fix-up-eols parent patch)))))
 
 (defun magit-apply-patch (section:s args patch)
   (let* ((files (if (atom section:s)
--- a/lisp/magit-apply.el
+++ b/lisp/magit-apply.el
@@ -87,8 +87,8 @@ With a prefix argument and if necessary, attempt a 3-way merge."
       (`(,_  files) (magit-apply-diffs  it args)))))
 
 (defun magit-apply--section-content (section)
-  (buffer-substring (magit-section-start section)
-                    (magit-section-end section)))
+  (buffer-substring-no-properties (magit-section-start section)
+                                  (magit-section-end section)))
 
 (defun magit-apply-diffs (sections &rest args)
   (setq sections (magit-apply--get-diffs sections))
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -405,6 +405,14 @@ Unless optional argument KEEP-EMPTY-LINES is t, trim all empty lines."
       (insert-file-contents file)
       (split-string (buffer-string) "\n" (not keep-empty-lines)))))
 
+(defun magit--buffer-has-crlf-eol-style-p ()
+  (eql 1 (coding-system-eol-type buffer-file-coding-system)))
+
+(defun magit--file-has-crlf-eol-style-p (file)
+  (with-temp-buffer
+    (insert-file-contents-literally file nil 0 500)
+    (search-forward "\r" nil t)))
+
 ;;; magit-utils.el ends soon
 (provide 'magit-utils)
 ;; Local Variables:
--- a/Documentation/RelNotes/2.6.0.txt
+++ b/Documentation/RelNotes/2.6.0.txt
@@ -94,6 +94,13 @@ Fixes since v2.5.0
 
 * `magit-section-cycle-diffs' hung when hiding sections.
 
+* Staging hunks/regions belonging to files with CRLF line endings on
+  Windows (or, to be precise, when the Magit status buffer uses CRLF
+  end-of-line style) ended up erroneously staging changes with LF line
+  endings. Magit now tries harder to not mangle line endings. A new
+  variable `magit-preserve-crlf-line-endings' has been added to allow
+  disabling the new behavior, should it cause unforseen troubles.
+
 This release also contains documentation updates.
 
 Authors


lisp/magit-apply.el
```diff
@@ -86,22 +86,37 @@ With a prefix argument and if necessary, attempt a 3-way merge."
       (`(,_   file) (magit-apply-diff   it args))
       (`(,_  files) (magit-apply-diffs  it args)))))
 
+(defun magit-apply--section-file-name (section)
+  (expand-file-name (magit-section-value (magit-section-parent section))
+                    (magit-toplevel)))
```
npostavs: The relative filename should be enough, no? (or else the git command wouldn't work, I think).

lisp/magit-apply.el
```diff
@@ -111,8 +126,7 @@ With a prefix argument and if necessary, attempt a 3-way merge."
                        (concat (magit-section-diff-header section)
                                (mapconcat
                                 (lambda (s)
-                                  (buffer-substring (magit-section-start s)
-                                                    (magit-section-end s)))
+                                  (magit-apply--section-content-for-diff s))
```
npostavs: You could get rid of the `lambda` too.

lisp/magit-utils.el
```diff
@@ -405,6 +405,14 @@ Unless optional argument KEEP-EMPTY-LINES is t, trim all empty lines."
       (insert-file-contents file)
       (split-string (buffer-string) "\n" (not keep-empty-lines)))))
 
+(defun magit--buffer-has-crlf-eol-style-p ()
+  (eql 1 (coding-system-eol-type buffer-file-coding-system)))
+
+(defun magit--file-has-crlf-eol-style-p (file)
+  (with-temp-buffer
+      (insert-file-contents-literally file nil 0 500)
+    (and (re-search-forward "\r" nil t) t)))
```
npostavs: Should be fine to return a generalized boolean so the `and` isn't needed. Also, use `search-forward` since `"\r"` doesn't need to be a regexp.

npostavs:
I think you might need to set the `buffer-file-coding-system` explicitly as well. My understanding from #2401 and #2403 is that the patch can sometimes end up with CRs and sometimes not, depending on some setting that I'm unsure of.

Also would be nice to see some more explanatory comments/commit message.

luismbo:
I *think* `buffer-file-coding-system` is not relevant here since the patch isn't actually written to a file. I think this approach assumes that `(cdr default-process-coding-system) => utf-8-unix`. I'm not sure what's the right place to codify that requirement. Maybe `magit-run-git-with-input`?

npostavs:
> I think buffer-file-coding-system is not relevant here since the patch isn't actually written to a file. I think this approach assumes that (cdr default-process-coding-system) => utf-8-unix.

Ah, that makes sense.

> I'm not sure what's the right place to codify that requirement. Maybe magit-run-git-with-input?

Yeah, that sounds like the right place.

luismbo:
@npostavs I've just pushed a new commit that tweaks the code as you suggested and adds a comment, but I'm not feeling particularly inspired. :-)

I don't know how to tweak `magit-run-git-with-input`. (I argue that git *always* expects input to come with LF line endings, so that concern is orthogonal to this pull request. :-P)

npostavs:
> adds a comment, but I'm not feeling particularly inspired. :-)

Thanks, now that I see the overview of what's happening, I'm wondering if a better approach is to force the magit buffer to use LF style line endings?

> (I argue that git always expects input to come with LF line endings, so that concern is orthogonal to this pull request. :-P)

Possibly, but my impression is that this is precisely the cause for #2401 and #2403.

luismbo:
@npostavs I considered the option of forcing the magit buffer to have LF eol style, but then that would mean that when dealing with projects that use CRLF eol style I would always see the ugly `^M`s at the end of changed lines, and that would be rather inconvenient.

luismbo:
Anyway, I'm happy with the current state of this PR if you are too.

lisp/magit-apply.el
```diff
@@ -86,22 +86,40 @@ With a prefix argument and if necessary, attempt a 3-way merge."
...
 
+;;; Emacs buffers normalise the various possible ways of ending lines
+;;; to a single ?\n. Therefore, when the magit buffer's eol style is
+;;; CRLF, we lose information whether lines should have CRLF or LF
+;;; endings. To recover that information, we check if the file to be
+;;; patched has CRLF lines, in which case we insert literal CR chars
+;;; to ensure they make their way to the patch file fed to git. The
+;;; remainder of the patch will have LF eol style.
+(defun magit-apply--maybe-fixup-eols (section string)
+  (let ((file (magit-section-value (magit-section-parent section))))
```
npostavs: I suspect this expression isn't correct (i.e. doesn't return a filename) for all section types, though I haven't got around to testing.

luismbo:
I see your point but `magit-apply-patch` makes the same assumption further down in the file.

luismbo:
Oh, you're right. The various callers of `magit-apply-patch` pass it different kinds of sections. Sorry about that. Clearly I haven't tested all of those code paths. Will fix.

luismbo:
@npostavs should be fixed now! (Thanks for catching that.) I couldn't figure out in what context `magit-apply-diff` is invoked though so I didn't test that one (nor its plural variant).

tarsius:
This looks like it is a step in the right direction, but we don't know yet whether this is sufficient or whether this has to be further tweaked in the future. Because of that I think that this should be split up into two commits. The first should add `magit-apply--section-content` and adds/renames local `parent` bindings, i.e. things that on their own could be considered cosmetics. The second commit should be limited to fiddling with line endings. We might have to come back to this a few times and it would be nice if the changes were limited to the issue the commit is trying to solve, not also related refactorings and cleanup.

luismbo:
@tarsius done. The exercise you suggested helped me further simplify the code. (Nice!) BTW, rewriting the initial commit using magit is a joy. ;-)

luismbo:
Added a variable as suggested. (Naming is hard, writing documentation is hard... Uf.)

Regarding `core.autocrlf` et al., my personal experience tells me that doing that sort of automatic CRLF <-> LF conversion in source code is a recipe for disaster.

tarsius:
> (Naming is hard, writing documentation is hard... Uf.)

Yeah :-)

What about:

```
(defcustom magit-preserve-crlf-line-edings t
  "Whether to try harder to preserve CRLF line ending type.

Magit buffers may contain diffs for more than one file and these
files might use different line ending types, but Emacs normalizes
file endings in a buffer to just one type, usually LF.

This isn't an issue when staging (or otherwise applying) complete
files, because the buffer contents are not used in that case.
But when applying hunks, then a patch has to be build from parts
of the buffer contents.

If this option is non-nil (the default), then Magit peeks the
file that is being changed to determine whether its line ending
style is CRLF.  If so, then CR characters are inserted into the
patch that is being sent to git.

Also see the Git variables `core.eol', `core.safecrlf', and
`core.autocrlf' for an alternative approach."
  :package-version '(magit . "2.6.0")
  :group 'magit
  :type 'boolean)
```

Also could you please mention this in the release notes.

@npostavs Minor issues aside, do you think this is good to go? (I would like to create a new release after merging the two remaining pull requests. /cc @kyleam)

npostavs:
> Minor issues aside, do you think this is good to go?

I think so, but I haven't actually tested it.

lisp/magit-utils.el
```diff
@@ -405,6 +405,14 @@ Unless optional argument KEEP-EMPTY-LINES is t, trim all empty lines."
       (insert-file-contents file)
       (split-string (buffer-string) "\n" (not keep-empty-lines)))))
 
+(defun magit--buffer-has-crlf-eol-style-p ()
+  (eql 1 (coding-system-eol-type buffer-file-coding-system)))
+
+(defun magit--file-has-crlf-eol-style-p (file)
+  (with-temp-buffer
+      (insert-file-contents-literally file nil 0 500)
```
npostavs: This lines is mis-indented.

luismbo:
Ugh, sorry I missed the `defcustom` arguments. I've tweaked the documentation string as follows:

```diff
modified   lisp/magit-apply.el
@@ -72,11 +72,11 @@
 
 Magit buffers may contain diffs for more than one file and these
 files might use different line ending types, but Emacs normalizes
-file endings in a buffer to just one type, usually LF.
+file endings in a buffer to a single newline character.
 
 This isn't an issue when staging (or otherwise applying) complete
 files, because the buffer contents are not used in that case.
-But when applying hunks, then a patch has to be build from parts
+But when applying hunks, then a patch has to be built from parts
 of the buffer contents.
 
 If this option is non-nil (the default), then Magit peeks the
```

Added the following entry to the release notes in the fixes section:

```
* Staging hunks/regions belonging to files with CRLF line endings on
  Windows (or, to be precise, when the Magit status buffer uses CRLF
  end-of-line style) ended up erroneously staging changes with LF line
  endings. Magit now tries harder to not mangle line endings. A new
  variable `magit-preserve-crlf-line-endings' has been added to allow
  disabling the new behavior, should it cause unforseen troubles.
```

lisp/magit-utils.el
luismbo: Oops. Fixed.

luismbo:
Forgot to update the variable name in the commit message. Fixed in d46419e.

tarsius:
I just tried reproducing the issue on Linux, using both Emacs 24.5 and 25.0.50, but couldn't.

In my case each line (added, removed, and context) in the CRLF file ends with `^M` in the diff in the status buffer. Git only shows the `^M` for added and remove lines, but I think that is an optimization. If I stage a hunk in Magit the staged changes still end with `^M` and when I commit the eol style is also preserved, which I verified using both Magit and Git directly.

The coding system in the buffer visiting the file is `undecided-dos` and in the Magit buffer it is `utf-8-unix`. Does that match what you see?

In the past users have reported that they see the `^M` and we might look into *hiding* them, but I think this is correct. Magit buffers should always use `utf-8-unix` - since they might contain part from different files we should use a reasonable default.

However Magit never actually enforces that, it just works because that's what Emacs does by default. I am not sure though whether that is always the case, maybe on Windows it doesn't, which would explain why you don't see the `^M`, or maybe this depends on more than just the OS. @npostavs could you too please try on Windows?

If we come to the conclusion that this always happens on Windows, then we should set the option value based on the `system-type`. Otherwise it should be off by default, I think.

lisp/magit-apply.el
```diff
@@ -86,22 +109,43 @@ With a prefix argument and if necessary, attempt a 3-way merge."
       (`(,_   file) (magit-apply-diff   it args))
       (`(,_  files) (magit-apply-diffs  it args)))))
 
+;;; Emacs buffers normalise the various possible ways of ending lines
+;;; to a single ?\n. Therefore, when the magit buffer's eol style is
+;;; CRLF, we lose information whether lines should have CRLF or LF
+;;; endings. To recover that information, we check if the file to be
+;;; patched has CRLF lines, in which case we insert literal CR chars
+;;; to ensure they make their way to the patch file fed to git. The
+;;; remainder of the patch will have LF eol style.
```
tarsius: Let's drop that completely and concentrate on the option's doc-string (which the reader of the below code can easily find because the option is used). Also comments should begin with just two semicolons. Using more turns these lines into seven section headings when using `outline-minor-mode`.

lisp/magit-utils.el
```diff
@@ -405,6 +405,14 @@ Unless optional argument KEEP-EMPTY-LINES is t, trim all empty lines."
       (insert-file-contents file)
       (split-string (buffer-string) "\n" (not keep-empty-lines)))))
 
+(defun magit--buffer-has-crlf-eol-style-p ()
+  (eql 1 (coding-system-eol-type buffer-file-coding-system)))
+
+(defun magit--file-has-crlf-eol-style-p (file)
+  (with-temp-buffer
+    (insert-file-contents-literally file nil 0 500)
+    (search-forward "\r" nil t)))
```
tarsius: Instead of doing that, have you considered using `(coding-system-eol-type buffer-file-coding-system)` here too?

#2584

lisp/magit-apply.el
luismbo: Dropped in a23f473. Using three semicolons for top-level comments is a CL convention, sorry about that.

lisp/magit-utils.el
luismbo: I did consider that, but my emacs-fu was not strong enough to figure out how to quietly open the file using `find-file`. `with-temp-buffer` doesn't work for that because it will have some default `buffer-file-coding-system` to start with.

luismbo:
> I just tried reproducing the issue on Linux, using both Emacs 24.5 and 25.0.50, but couldn't.

If this is a surprise for you, then I have failed to convey what is
going on. (Perhaps the `magit-preserve-crlf-line-endings` docstring
needs further tweaking then!) The problem only occurs when Magit's
status buffer has CRLF encoding (and that should only occur on
Windows).

There is no problem when Magit status buffer has LF encoding because
in that case a change coming from file with lines ending with two
bytes (CR and LF) are translated into two character in the buffer:
`?\r` and `?\n`. However, when Magit's status buffer has CRLF
encoding, those same two bytes (CR and LF) are translated into a
single a character in the status buffer: `?\n`.

Does it make more sense now?

lisp/magit-utils.el
tarsius: > figure out how to quietly open the file using `find-file`.

You can use `find-file-noselect`.

> `with-temp-buffer` doesn't work for that because it will have some default `buffer-file-coding-system` to start with.

Ah, didn't realize that. But using `find-file-noselect` would be problematic because it runs a lot of stuff we don't want and if the file is already being visited in a buffer... So It's probably better to stick with what you have done. (But 500 seems a bit low.)

luismbo: How about something like this?

```elisp
(defun magit--file-has-crlf-eol-style-p (file)
  (with-temp-buffer
    (cl-loop with start = 0
             for (_ bytes-read) = (insert-file-contents-literally file nil start 500)
             until (zerop bytes-read)
             do (incf start bytes-read)
             when (re-search-forward "(\r)?\n" nil t)
             return (string= (match-string 1) "\r"))))
```

(It doesn't work because the regex is botched, but you get the general idea.)

tarsius:
> Does it make more sense now?

Yes! (But you explained that well enough before, it's just that I find it very unintuitive and appear to have over read the important parts. Why would it be an issue if both the buffer and the file used the *same* eol style?! But I get it now.)

> The problem only occurs when Magit's status buffer has CRLF encoding

Maybe we should figure out how to use LF encoding in Magit buffers? If I understand everything correctly that shouldn't cause any issues - except of the `^M`s everywhere. Which we could hide, but that might be *more* expensive than what you are suggesting, so...

> (and that should only occur on Windows).

In that case the option's default value should be set using `(and (memq system-type '(ms-dos windows-nt)) t)` (but not `cygwin`, right?). That way only those you might benefit have to pay the price.

I finally start realizing that this probably is the *best* approach :-) By the way, thanks for hanging in there! That's the sort of issue that only gets resolved if an affected user invests the necessary time (because my understanding is spotty and I also have no way of testing.)

> Perhaps the `magit-preserve-crlf-line-endings` docstring needs further tweaking then!

Yes, please incorporate the information from your last message.

lisp/magit-apply.el
```diff
@@ -86,22 +109,36 @@ With a prefix argument and if necessary, attempt a 3-way merge."
       (`(,_   file) (magit-apply-diff   it args))
       (`(,_  files) (magit-apply-diffs  it args)))))
 
+(defun magit-apply--maybe-fixup-eols (section string)
+  (let ((file (magit-section-value section)))
```
tarsius: Okay, now I start nitpicking (and it's totally a matter of personal preference), but since the value is only used once, I don't see a reason to use a variable here.

lisp/magit-apply.el
```diff
@@ -86,22 +109,36 @@ With a prefix argument and if necessary, attempt a 3-way merge."
...
+    (if (and magit-preserve-crlf-line-endings
+             (magit--buffer-has-crlf-eol-style-p)
+             (magit--file-has-crlf-eol-style-p file))
+        (replace-regexp-in-string "\n" "\r\n" string)
```
tarsius: It might be worth investigating whether this can be done more efficiently using `line-end-position`. But that would have to happen in `magit-apply--section-contents` and maybe `conat`enating many times is *less* efficient than just using `buffer-substring` and `replace-regexp-in-string`. And it would get a lot more complicated.

lisp/magit-apply.el
```diff
@@ -86,22 +109,36 @@ With a prefix argument and if necessary, attempt a 3-way merge."
...
+        string)))
+
+(defun magit-apply--section-content (section)
+  (magit-apply--maybe-fixup-eols
+   (pcase (magit-section-type section)
+     (`file section)
+     (`hunk (magit-section-parent section)))
+   (buffer-substring (magit-section-start section)
```
tarsius: But we should use `buffer-substring-no-properties` here. And since that isn't really related to the other changes, it should be done using a separate commit in between the other two.

lisp/magit-utils.el
tarsius: Couldn't that mean that we end up in the middle of some character? Probably just leave it as is.

lisp/magit-apply.el
tarsius: In other words, let's just leave this as is too.

lisp/magit-apply.el
luismbo: On Sat, Mar 12, 2016 at 10:45 AM, Jonas Bernoulli
<notifications@github.com> wrote:
> But we should use buffer-substring-no-properties here. And since that isn't
> really related to the other changes, it should be done using a separate
> commit in between the other two.

Done in 8450a3e.

lisp/magit-utils.el
luismbo: I think `re-search-forward` doesn't move the point if it doesn't find anything, so it'd be okay, but I agree it's probably better to leave the current, simpler, version as is.

lisp/magit-apply.el
luismbo: > Okay, now I start nitpicking (and it's totally a matter of personal preference), but since the value is only used once, I don't see a reason to use a variable here.


I don't have strong feelings either way. Changed as suggested in 5999c20.

luismbo:
> In that case the option's default value should be set using (and (memq system-type '(ms-dos windows-nt)) t) (but not cygwin, right?). That way only those you might benefit have to pay the price.

The price of peeking the file is payed if, and only if,
`magit--buffer-has-crlf-eol-style-p` returns t. That will only happen
on Windows. I think that check is more robust than a hard-coded list
of `system-type`s.


> > Perhaps the magit-preserve-crlf-line-endings docstring needs further tweaking then!
>
> Yes, please incorporate the information from your last message.

Done in 5999c20.

tarsius:
> I think that check is more robust than a hard-coded list of `system-type`s.

True. But that (me not realizing that we don't do anything expensive if we don't have to) reminds me that initially I thought that `magit--{buffer,file}-has-crlf-eol-style-p` should just be folded into `magit-apply--maybe-fixup-eols`. Since these functions are tiny and only used here I don't see much benefit in keeping them around. And if we stick the code directly into the fixup function, then one can immediately see that determining the eol style for the buffer is cheap while the same is more costly for the file. (I would have realized that anyway instead of making the previous suggestion, but still.)

luismbo:
My reasoning there was that

```elisp
(defun magit-apply--maybe-fixup-eols (section string)
  (if (and magit-preserve-crlf-line-endings
           (magit--buffer-has-crlf-eol-style-p)
           (magit--file-has-crlf-eol-style-p (magit-section-value section)))
      (replace-regexp-in-string "\n" "\r\n" string)
      string))
```

is less obscure than

```elisp
(defun magit-apply--maybe-fixup-eols (section string)
  (if (and magit-preserve-crlf-line-endings
           (eql 1 (coding-system-eol-type buffer-file-coding-system))
           (with-temp-buffer
             (insert-file-contents-literally (magit-section-value section)
                                             nil 0 500)
             (search-forward "\r" nil t)))
      (replace-regexp-in-string "\n" "\r\n" string)
      string))
```

Perhaps adding some comments helps, but when in doubt I try to make code self-readable rather than adding comments.

```elisp
(defun magit-apply--maybe-fixup-eols (section string)
  (if (and magit-preserve-crlf-line-endings
           ;; buffer has CRLF eol type
           (eql 1 (coding-system-eol-type buffer-file-coding-system))
           ;; file has CRLF eol type, too
           (with-temp-buffer
             (insert-file-contents-literally (magit-section-value section)
                                             nil 0 500)
             (search-forward "\r" nil t)))
      (replace-regexp-in-string "\n" "\r\n" string)
      string))
```

I don't have hard feelings for either of the three versions. Let me know which one you prefer.

npostavs:
(I don't have access to a Windows system with magit right at the moment)

>> I just tried reproducing the issue on Linux, using both Emacs 24.5 and 25.0.50, but couldn't.
>
> If this is a surprise for you, then I have failed to convey what is going on. (Perhaps the `magit-preserve-crlf-line-endings` docstring needs further tweaking then!) The problem only occurs when Magit's status buffer has CRLF encoding (and that should only occur on Windows).

(Using my GNU/Linux box) I tried setting `buffer-file-coding-system` to `utf-8-dos`, but I still see the `^M`s in the status buffer.

luismbo:
@npostavs `buffer-file-coding-system` needs to be set before the buffer is constructed. Perhaps setting `default-buffer-file-coding-system` does the trick.

npostavs:
Hmm, I tried `(setq-default buffer-file-coding-system 'utf-8-dos)` and `(setq default-buffer-file-coding-system 'utf-8-dos)` but no difference.

tarsius:
Go with the third. I'll merge once that is done.

luismbo:
@npostavs gosh. It's actually not `buffer-file-coding-system` that's at fault here. Try this instead:

```
(setq default-process-coding-system '(utf-8-dos . utf-8-unix))
```

I'll have to think a little bit about the implications of this.

luismbo:
Alright. So, the general idea of this pull request is still correct, except the information lossage is not due to `buffer-file-coding-system` but due to `process-coding-system` used when grabbing the diffs from git. I see two options to proceed:

  1. Tweak this pull request to reflect this new insight (i.e. check `(car default-process-coding-system)` instead of `buffer-file-coding-system`) and live with the fact that `default-process-coding-system` might change between grabbing the diffs and `magit-apply`ing. (Fairly unlikely, but still...)
  2. Change `magit-process` to ensure that `process-coding-system` has an LF eol style and (optionally) hide them in the Magit buffer using something like
```elisp
(setq buffer-display-table (make-display-table))
(aset buffer-display-table ?\^M [])
```

(I might need some help if we want to pursue the 2nd option, both for tweaking `magit-process` and for figuring out the best way to hide the `^M` chars.)

What do you guys think?

lisp/magit-apply.el
```diff
@@ -86,22 +118,40 @@ With a prefix argument and if necessary, attempt a 3-way merge."
...
+           (with-temp-buffer
+             (insert-file-contents-literally (magit-section-value section)
+                                             nil 0 500)
+             (search-forward "\r" nil t)))
+      (replace-regexp-in-string "\n" "\r\n" string)
+      string))
+
+(defun magit-apply--section-content (section)
+  (magit-apply--maybe-fix-up-eols
+   (case (magit-section-type section)
```
npostavs: should be `cl-case`

npostavs:
>  Change `magit-process` to ensure that `process-coding-system` has an LF eol style and (optionally) hide them in the Magit buffer

I like this method a bit better, it seems less likely to go wrong.

lisp/magit-apply.el
luismbo: Ups. Fixed in bceef47 to use `pcase` again. (Noticed that `pcase` is used extensively throughout magit, never `cl-case`.)

luismbo:
I've pushed bceef47, which implements the 1st option, just in case. I'll start exploring the 2nd option now.

luismbo:
Should I open another pull request to pursue that option?

tarsius:
> Should I open another pull request to pursue that option?

Whatever works best for you. You could also just use a temporary commit, to be later either squashed or discarded.

luismbo:
Closing in favor of #2587.

PR #2587, opened 2016-03-13
Another approach at solving issues #2401, #2403 and #2583

luismbo:
As discussed in pull request #2584, here's another approach:

  1. Diffs are always read using LF eol style to ensure CR bytes are not lost.
  2. When appropriate, CR characters (aka ^M) are hidden.  This is similar to what git does with `git config core.whitespace cr-at-eol`.

This approach seems to be more robust and removes the need for peeking files to guess end-of-line styles. However, it touches `magit-process-file` which makes me uncomfortable because (a) lots of functionality go through there and (b) git seems to be invoked through other means as well.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -164,6 +164,13 @@ many spaces.  Otherwise, highlight neither."
                                (integer :tag "Spaces" :value ,tab-width)
                                (const :tag "Neither" nil)))))
 
+(defcustom magit-diff-hide-trailing-cr-characters
+  (and (memq system-type '(ms-dos windows-nt)) t)
+  "Whether to hide ^M characters at the end of a line in diffs."
+  :package-version '(magit . "2.6.0")
+  :group 'magit-diff
+  :type 'boolean)
+
 ;;;; Revision Mode
 
 (defgroup magit-revision nil
@@ -1940,6 +1947,10 @@ are highlighted."
               (stage nil))
           (forward-line)
           (while (< (point) end)
+            (when (and magit-diff-hide-trailing-cr-characters
+                       (char-equal ?\r (char-before (line-end-position))))
+              (put-text-property (1- (line-end-position)) (line-end-position)
+                                 'invisible t))
             (put-text-property
              (point) (1+ (line-end-position)) 'face
              (cond
--- a/Documentation/RelNotes/2.6.0.txt
+++ b/Documentation/RelNotes/2.6.0.txt
@@ -101,6 +101,14 @@ Fixes since v2.5.0
 * `magit-show-stash' did not properly handle untracked files that were
   within subdirectories.
 
+* Staging hunks/regions belonging to files with CRLF line endings on
+  Windows (or, to be precise, when `default-process-coding-system' had
+  CRLF end-of-line conversion) ended up erroneously staging changes
+  with LF line endings. Magit now ensures line endings are preserved
+  by enforcing a `process-coding-system' with LF end-of-line
+  conversion. The new behavior may be disabled by setting
+  `magit-process-ensure-unix-coding-system' to nil.
+
 This release also contains documentation updates.
 
 Authors
--- a/Documentation/RelNotes/2.6.0.txt
+++ b/Documentation/RelNotes/2.6.0.txt
@@ -63,6 +63,10 @@ Updates since v2.5.0
   useful when using Magit as a library in another packages that runs
   git in many different repositories.
 
+* Added new variable `magit-diff-hide-trailing-cr-characters'. When it
+  is non-nil (the default on Windows), ^M characters at the end of
+  diff lines are hidden.
+
 Fixes since v2.5.0
 ------------------
 
--- a/lisp/magit-apply.el
+++ b/lisp/magit-apply.el
@@ -86,22 +86,24 @@ With a prefix argument and if necessary, attempt a 3-way merge."
       (`(,_   file) (magit-apply-diff   it args))
       (`(,_  files) (magit-apply-diffs  it args)))))
 
+(defun magit-apply--section-content (section)
+  (buffer-substring (magit-section-start section)
+                    (magit-section-end section)))
+
 (defun magit-apply-diffs (sections &rest args)
   (setq sections (magit-apply--get-diffs sections))
   (magit-apply-patch sections args
                      (mapconcat
                       (lambda (s)
                         (concat (magit-diff-file-header s)
-                                (buffer-substring (magit-section-content s)
-                                                  (magit-section-end s))))
+                                (magit-apply--section-content s)))
                       sections "")))
 
 (defun magit-apply-diff (section &rest args)
   (setq section (car (magit-apply--get-diffs (list section))))
   (magit-apply-patch section args
                      (concat (magit-diff-file-header section)
-                             (buffer-substring (magit-section-content section)
-                                               (magit-section-end section)))))
+                             (magit-apply--section-content section))))
 
 (defun magit-apply-hunks (sections &rest args)
   (let ((section (magit-section-parent (car sections))))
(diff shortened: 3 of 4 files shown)


npostavs:
> (b) git seems to be invoked through other means as well.

c84481aade652cc8058a8bdf7b81af7e3839421c should cover the other git calls as well.

luismbo:
@npostavs cool, thanks! Should I refactor those three points into a macro like `with-unix-process-coding-system`?

@tarsius would you like a `magit-process-ensure-unix-coding-system` variable to control this new behavior?

tarsius:
> Should I refactor those three points into a macro like `magit-process--with-unix-coding-system`?

No, that seems like an unnecessary complication. I would rather simplify the duplicated code to not use distructuring (as much as I usually like it):

```lisp
(cons (coding-system-change-eol-conversion
       (car default-process-coding-system) 'unix)
      (coding-system-change-eol-conversion
       (cdr default-process-coding-system) 'unix))))
```

is boilerplate, but is also self-contained enough to be used three times.

> would you like a `magit-process-ensure-unix-coding-system` variable to control this new behavior?

I am not sure yet, but if yes, then we probably should add a *function* (not a macro as suggested  above):

```lisp
(defun magit--process-coding-system ()
  (if magit-process-ensure-unix-coding-system
      (cons (coding-system-change-eol-conversion
             (car default-process-coding-system) 'unix)
            (coding-system-change-eol-conversion
             (cdr default-process-coding-system) 'unix))
    default-process-coding-system))
```

My biggest concern however is the use of the `invisible` text property to make the `^M`s invisible. Did you investigate other options? I had a quick look, but it seems "glyths" cannot be used. Hopefully I am wrong about that or there is some other option. But maybe the current approach actually isn't all that bad - since it uses text properties and not overlays, it should at least not affect performance (except for the "washing cost" of course).

lisp/magit-diff.el
```diff
@@ -164,6 +164,12 @@ many spaces.  Otherwise, highlight neither."
                                (integer :tag "Spaces" :value ,tab-width)
                                (const :tag "Neither" nil)))))
 
+(defcustom magit-diff-hide-trailing-cr-characters
+  (and (memq system-type '(ms-dos windows-nt)) t)
+  "Whether to hide ^M characters at the end of a line in diffs."
```
tarsius: `:package-version '(magit . "2.6.0")`

lisp/magit-diff.el
```diff
@@ -1940,6 +1946,12 @@ are highlighted."
               (stage nil))
           (forward-line)
           (while (< (point) end)
+            (when (and magit-diff-hide-cr-characters
+                       (string= (buffer-substring-no-properties
+                                 (1- (line-end-position)) (line-end-position))
+                                "\r"))
```
tarsius: There's a shorter test: `(= (char-before (line-end-position)) ?\r)`.

tarsius:
And sorry for the delay, I was quite busy with other things ;)

luismbo:
`magit--process-coding-system` is a much better idea than a macro. I think the code looks much cleaner that way. Pushed 5375afa based on that idea and @npostavs's commit.

luismbo:
> My biggest concern however is the use of the invisible text property to make the ^Ms invisible. Did you investigate other options?

@tarsius I considered using `buffer-display-table` but properties felt more natural. I'm not familiar with the performance implications of either option. Intuitively, since diffs already use many properties it seemed like this extra one wouldn't hurt too much, but intuition is not a good ally when it comes to assessing performance! :)

lisp/magit-diff.el
luismbo: Added in a62a913.

lisp/magit-diff.el
luismbo: Uh, nice. Pushed 4e86904 (but using `char-equal` instead of `=`).

luismbo:
Squashed the commits that needed squashing. Let me know what you decide regarding the need for a `magit-process-ensure-unix-coding-system` variable.

tarsius:
> Let me know what you decide regarding the need for a `magit-process-ensure-unix-coding-system` variable.

Let's do it.

tarsius:
I've been playing with display-tables a bit:

```lisp
(setq buffer-display-table (make-display-table))
(aset buffer-display-table ?\r [])
```

While the second line works as expected, the necessary first causes all ?\n to be displayed as an A with a bird sitting on top (a new display line is still started though). Any idea why that might be?

luismbo:
Pushed cce5f66 with `magit-process-ensure-unix-coding-system` added.

I don't know why `buffer-display-table` is doing that for you, sorry. It works for me on GNU Emacs 24.5.1 (x86_64-w64-mingw32). :-/

npostavs:
Not seeing any strange characters after setting `buffer-display-table` here (GNU Emacs 24.5.2 (x86_64-unknown-linux-gnu)) either.

tarsius:
Opened a question on stackexchange about the birds: http://emacs.stackexchange.com/questions/21054/setting-buffer-display-table-to-a-pristine-table-causes-strange-character-at-eol

tarsius:
The problem seems to be the `fill-column-indicator`, specifically the assumptions it makes in `fci-process-display-table`.

tarsius:
Okay then I would prefer to go with the display-table approach. Losing `fci-mode`, while unfortunate, is not such a big deal to me. There have been similar issue with that mode before in my usage. Eventually I would like to figure out how to make `fci` play nicer with other modes (or figure out what it expects other modes to do). If someone else would like to look into that, that would be nice of course :-)

tarsius:
Uh and since it wouldn't make much sense to turn on `fci-mode` in Magit buffers, there's not really an issue anyway.

luismbo:
@tarsius what's the advantage of using `buffer-display-table`?

tarsius:
It just seems cleaner to me. And Eli, the maintainer of Emacs' display engine has criticized Magit for abusing the `invisible` property. I also expect it to be slightly more efficient because it operates on lower level where the `(aref display-table N)` has to happen anyway, regardless of whether we put something non-nil there or not.

luismbo:
@tarsius OK. I'm leaving for a trip in a while, so I might not have time to implement that change in the upcoming days.

npostavs:
Using the display table does have the small drawback of hiding non-trailing CRs as well.

luismbo:
@npostavs indeed. I vote for keeping the properties. :)

tarsius:
Alright. Let's go with that then.

If there's still time before you leave, could you please mention the second option in the release notes and both options in the manual (just edit `magit.org`, I'll then regenerate `magit.texi). Otherwise I'll just do all that myself (but please give me a quick okay that you are fine with my modifying you commits like that (I might also move a definition and/or an entry in the release notes).

tarsius:
I quickly made these changes myself.

* Moved some things around. Most importantly this is now the fix that is listed first in the release notes, because it is the most important one :-)
* Renamed `magit-process-ensure-unix-coding-system` to `magit-process-ensure-unix-line-ending`.
* Documented `magit-diff-hide-trailing-cr-characters` in manual. But not `magit-process-ensure-unix-line-ending` because I could not find a good place for it. All the other `magit-process-*` options are not mentioned in the manual either. I'll fix that later.

I've pushed the result to `crlf-eol`. Please double-check and give your okay.

tarsius:
(I've split that out into fixup commits to make it easier to review.)

luismbo:
@tarsius looks great, thanks for finishing up my commits.

tarsius:
Merged now.

Thanks again Luís!
And Noam too for helping along the way.

tarsius:
@luismbo how would you like to be known in [AUTHORS.md](https://raw.githubusercontent.com/magit/magit/master/Documentation/AUTHORS.md)

So far we have committed as:
```
Luís Borges de Oliveira <lbo@siscog.pt>
Luís Oliveira <loliveira@common-lisp.net>
Luís Oliveira <luismbo@gmail.com>
```

luismbo:
@tarsius the gmail one, I guess. Thanks!

PR #2595, opened 2016-03-23
magit-read-{remote,upstream}-branch: remove duplicate choices

ilohmar:
In both cases, the collection to choose from is created from an
argument (local branch) turned into a remote ref, plus a list of all
remote or all branches.  The latter part can already contain the refs
constructed in the former, leading to duplicates.

Since we use dash anyway and the order in the collection is irrelevant,
replacing nconc by -union should get rid of these duplicates.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1265,12 +1265,12 @@ Return a list of two integers: (A>B B>A)."
     (prompt &optional remote default local-branch require-match)
   (let ((choice (magit-completing-read
                  prompt
-                 (nconc (and local-branch
-                             (if remote
-                                 (concat remote "/" local-branch)
-                               (--map (concat it "/" local-branch)
-                                      (magit-list-remotes))))
-                        (magit-list-remote-branch-names remote t))
+                 (-union (and local-branch
+                              (if remote
+                                  (concat remote "/" local-branch)
+                                (--map (concat it "/" local-branch)
+                                       (magit-list-remotes))))
+                         (magit-list-remote-branch-names remote t))
                  nil require-match nil 'magit-revision-history default)))
     (if (or remote (string-match "\\`\\([^/]+\\)/\\(.+\\)" choice))
         choice
@@ -1336,9 +1336,9 @@ Return a list of two integers: (A>B B>A)."
     (&optional (branch (magit-get-current-branch)) prompt)
   (magit-completing-read
    (or prompt (format "Change upstream of %s to" branch))
-   (nconc (--map (concat it "/" branch)
-                 (magit-list-remotes))
-          (delete branch (magit-list-branch-names)))
+   (-union (--map (concat it "/" branch)
+                  (magit-list-remotes))
+           (delete branch (magit-list-branch-names)))
    nil nil nil 'magit-revision-history
    (or (let ((r (magit-remote-branch-at-point))
              (l (magit-branch-at-point)))


tarsius:
I believe `completing-read` and friends actually remove duplicates. Can you provide instructions on how to demonstrate the negative effect of having duplicates.

```
(completing-read ": " '("a" "b" "c" "a" "d"))

Click on a completion to select it.
In this buffer, type RET to select the completion near point.

Possible completions are:
a 	b
c 	d
```

ilohmar:
Hmm, I see.  You're right, it's because I have been testing ivy instead of ido for the last two months.  

Do you know, off of the top of your head, if the duplicate-removal of the default completing-read and ido is well-documented and mandatory?  Or, more plainly: should callers rely on that feature or shouldn't they?  Reading the docs of `completing-read` and the info node `Basic Completion` leave me under the impression that the behavior with duplicates is not specified.

tarsius:
I don't remember where I got that information from I am afraid. If, after taking another look, you still cannot find any documentation on this, then you should probably ask on emacs-devel whether this behavior is documented and whether alternative completion methods are expected to do the same.

I am not totally opposed to the proposed patch, but let's first see what the Emacs developers think. (In case I miss that discussion, please report the result here.) I am closing this for now, if it turns out that alternative completion methods are not expected to behave the same way (or that we want to merge for another reason), then I'll reopen.

Thanks for your efforts.

ilohmar:
Discussion unfolding here: http://lists.gnu.org/archive/html/emacs-devel/2016-04/msg00068.html

*Very* preliminary impression: duplicate removal is accidental, so the caller should actually do it.

tarsius:
Reopening. (I have already forgotten about this.)

ilohmar:
Yeah, I was a bit sluggish asking.  So after three messages the thread has died.  But on rereading the answers and my question, I am quite convinced that it's "obviously" :) the caller that should remove whatever it considers as duplicates (Drew's answer has a good general point for this, although it does not apply to the concrete case at hand).

Would you be willing to reconsider the PR in this light?  I think I went through `magit-git.el` to have it consistent in this respect (I only found two instances where possible overlapping sets were `nconc`d), but did not check all of magit.

tarsius:
>> Is removing duplicates from the collection *expected* from completion
>> functions, or is the caller of the function responsible for providing a
>> collection without duplicates?

> It's the generic completion code that's responsible for that, yes.
>
>
>    Stefan

Since Stefan maintained Emacs for a long time I would consider this the authoritative answer, even though it is much shorter than Drews. Which I only skimmed I have to admit. But I did notice the "the order of candidates can be useful and sometimes even duplicates" argument. And I agree that can indeed sometimes be useful, but that's not the question here since we are talking about functions that are just like `completing-read`, just nicer.

These functions should not unintentionally behave different from `completing-read`. So let's first hear what @abo-abo has to say.

abo-abo:
> These functions should not unintentionally behave different from completing-read. So let's first hear what @abo-abo has to say.

Removing duplicates is an expensive operation, it's basically sorting. It should not be the responsibility of `completing-read` since the collection size can be arbitrarily large.  Completion is an O(N) operation, it should not be made O(NlogN).

It should always be the responsibility of the caller, since the caller actually knows the details about the collection and can sometimes be sure that a call to `delete-dups` is not necessary, e.g. for `(setq collection (number-sequence 1 100000))`.

tarsius:
That's a good (and expected) argument. Now that I know that this difference is intentional I will adjust Magit accordingly by merging this and, if necessary, make similar changes. I just wanted to avoid making this change and it then becoming unnecessary because that "bug" was later fixed in `ivy` and other completion methods, but now that I know that there is at least one method that will stick to this behavior I don't mind changing this.

tarsius:
I've cherry-picked.

ilohmar:
Thanks!

Drew's main argument IMO was that the concept of "duplicate" is not even well-defined inside the completion method.  In general, only the caller *can* decide what constitutes duplicates.

Please note that I did not want to dismiss Stefan's answer, I just read "they're responsible" as describing the "is-state", not describing what callers of an arbitrary completion method can rely upon.

Issue #2533, opened 2016-02-01
Improve order of GUI sections


floli:
Hello,

when I push b magit-branch-popup appairs:

```
Configure existing branches
 d branch.parallelRBF_i1.description unset
 u branch.parallelRBF_i1.merge       refs/heads/parallelRBF_i1
   branch.parallelRBF_i1.remote      origin
 r branch.parallelRBF_i1.rebase      [true|false|default:false]
 p branch.parallelRBF_i1.pushRemote  [origin|peter]

Configure repository defaults
 M-r pull.rebase        [true|false|default:false]
 M-p remote.pushDefault [origin|peter]

Configure branch creation
 U branch.autoSetupMerge  [always|true|false|default:true]
 R branch.autoSetupRebase [always|local|remote|never|default:never]

Actions
 c Create and checkout    b Checkout
 n Create                 m Rename
 s Create spin-off        x Reset
                          k Delete

Common Commands
 C-c C-c Set defaults       C-h i View popup manual    C-t Toggle this section
 C-x C-s Save defaults      ?     Popup help prefix    C-g Abort
```
for me the ordering appears not optimal.
On top are three configuration sections, something you probably do not need very often. On the very bottom is the common section, also not needed too frequently. Somewhere in the middle are the Actions, what is I am looking for most of the time. You usually start scanning a window from top. The same applies to a lot of other popups (merge, stash).
I would suggest moving the Action section to top.

kyleam:
I know that your main suggestion is about the top portion, but regarding the bottom part that you also mention, you can remove that by setting `magit-popup-show-common-commands` to nil.

npostavs:
I've been thinking that it might be better to move the branch configuration to a separate (nested?) popup, partly because all the options are a bit noisy, and also because on Windows querying the config options takes just long enough to make opening the branch popup annoying (around 50-100 ms, I guess).

floli:
This ordering (actions, i.e. what you mostly look at, are at the bottom or inbetween) are also at the popups for stash, commit and merge.

tarsius:
The actions in *all* popups are at the bottom and I think that makes sense because the minibuffer is at the bottom too. I do not intend to change that.

tarsius:
> I've been thinking that it might be better to move the branch configuration to a separate (nested?) popup,

Might be a good idea, but should probably be opt-in to avoid upsetting any muscle memory. If we decide to go with a nested popup, then we should probably also make it available from the push popup.

PR #2596, opened 2016-03-24
magit-branch-config-popup: new popup

npostavs:
As discussed in https://github.com/magit/magit/issues/2533#issuecomment-179128807 and https://github.com/magit/magit/issues/2533#issuecomment-182097038.

Still need to add a mention in the manual.

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -279,6 +279,15 @@ prefer the former, then you should add branches such as \"master\",
   :group 'magit-commands
   :type '(repeat string))
 
+(defcustom magit-branch-separate-config-popup nil
+  "Whether to have a separate popup for branch config options.
+
+This must be set to t before loading magit in order to have
+effect."
+  :package-version '(magit . "2.6.0")
+  :group 'magit
+  :type 'boolean)
+
 (defcustom magit-repository-directories nil
   "Directories containing Git repositories.
 Magit checks these directories for Git repositories and offers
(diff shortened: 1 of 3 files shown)


lisp/magit.el
```diff
@@ -1315,6 +1329,14 @@ Non-interactively DIRECTORY is (re-)initialized unconditionally."
     (local-set-key (vector (magit-popup-event-key ev))
                    'magit-invoke-popup-action)))
 
+(if magit-branch-separate-config-popup
+    (magit-define-popup-action 'magit-push-popup
```
kyleam: s/magit-push-popup/magit-branch-popup/ ?

kyleam: Oh, never mind. I see that was suggested in the comment you linked to.
Shouldn't the config popup also be available from the branch popup?

lisp/magit.el
```diff
@@ -279,6 +279,12 @@ prefer the former, then you should add branches such as \"master\",
   :group 'magit-commands
   :type '(repeat string))
 
+(defcustom magit-branch-separate-config-popup nil
+  "Whether to have a separate popup for branch config options."
```
kyleam: Should this mention that this needs to be set before Magit is loaded?

lisp/magit.el
npostavs: Oh right, I was using a loop to put it in several other popups, but then I decided to remove it, and I forgot to put the branch popup back.

npostavs:
Hmm, I added the a reference in the manual, but I'm not sure how to update the texi properly. When I just `ox-texinfo` there were a bunch of whitespace changes which I assume means I should be loading `ox-texinfo+`, but after loading I get `Lisp error: (void-function nil)` while trying to export to texi.

lisp/magit.el
npostavs: yeah, done now.

tarsius:
I intend to soon give my "lesser packages" some attention and deal with issues such as this. For now just edit the org file and I will generate the texi myself before merge.

Please add an entry in the release notes. Other than that this is ready for merge right?

npostavs:
> Please add an entry in the release notes.

Done.

> Other than that this is ready for merge right?

Yes, I think so.

tarsius:
Unfortunately re-evaluating `magit.el` after making changes to it, would result in a very repetitive `magit-branch-popup`. And we don't have to force users to set this before loading `magit.el`. Please have a look at my not-quite-finished approach at https://github.com/magit/magit/compare/jb/separate-branch-config.

npostavs:
> Please have a look at my not-quite-finished approach at https://github.com/magit/magit/compare/jb/separate-branch-config.

Oh neat, I didn't realize the popup values can use `lambda` like that.

tarsius:
Well, it's quite a hack and I don't think I bothered to document it.

tarsius:
Anyway, I'll try to clean up my version and merge it before I release v2.6 later this month.

PR #2607, opened 2016-04-02
move trailing newline into `Untracked Files` section

fritzgrabo:
This patch aims to streamline the placement of the trailing `\n` in the "Untracked Files" section with that of all the other sections.

Without this patch, the trailing newline for this section is inserted outside of it. As a consequence, it's not hidden when that section is closed, and a single empty lines between the section headers remains visible.


kyleam:
Thanks.

Issue #2426, opened 2015-12-04
"M-x magit-show-commit" errors


alezost:
Hello, if I run `M-x magit-show-commit` in a fresh emacs (before
running `M-x magit-status` or any other magit command), I get the
following error:

    Symbol's value as variable is void: magit-blame-mode

Checked on the latest master (commit 9ad32d573e5f909c7581d5d55ffaba5898ffdb48)

tarsius:
Loading `magit` also loads `magit-blame`, which in turn defined `magit-blame-mode`.  This sounds like there is something wrong with your installation. Do you have other installations of magit laying around - due to `load-path` issues you might be loading an old version.

npostavs:
Confirmed with `emacs -Q -L ..\dash -L lisp -l lisp/magit-autoloads.el -f magit-show-commit`

> Loading magit also loads magit-blame, which in turn defines magit-blame-mode

Running `magit-show-commit` loads `magit-diff` which loads `magit-core` which loads a few other magit things, but apparently not `magit-blame`.

tarsius:
It's ... a bit complicated. In short, internal dependencies are not strictly linear and that doesn't work well with autoloading.

I think for now we should just require `magit-blame` inside `magit-show-commit` and be done with it.

alezost:
Now (after commit b0c77f5b60b9fe360cb679ed7fdee81718ba6d4b), it
complains about the other thing.  After `M-x magit-show-commit`
it prompts for a commit and then errors:

    run-hooks: Symbol's function definition is void: magit-maybe-show-margin

tarsius:
So far I was operating under the assumption that `magit-status` would have been called at least once before other Magit commands are called. Other issues like this might bubble up when that is not the case.

alezost:
So if `M-x magit-show-commit` (and maybe other commands) wouldn't
work without `M-x magit-status`, then I think it shouldn't be
autoloaded.  Because if it is available right away, a user would expect
it to work, WDYT?

tarsius:
Yes. But because not many users do that, those issues have not been reported yet.

alezost:
Ah, sorry, I thought you meant `magit-show-commit` was not
supposed to be used before `magit-status`, so this issue still
presents then.

tarsius:
After the latest fix `emacs -Q -L ../dash -L lisp -l lisp/magit-autoloads.el -eval '(magit-show-commit "HEAD")'` works for me.

alezost:
[...]

PR #2628, opened 2016-04-23
magit-file-popup: autoload it

johnmastro:
Trivial addition to support cases where users have `magit-file-popup` bound to a key and may invoke it before e.g. `magit-status`.

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -2276,6 +2276,7 @@ the current repository."
     map)
   "Keymap for `magit-file-mode'.")
 
+;;;###autoload (autoload 'magit-file-popup "magit" nil t)
 (magit-define-popup magit-file-popup
   "Popup console for Magit commands in file-visiting buffers."
   :actions '((?s "Stage"     magit-stage-file)


kyleam:
Do any of the commands in this popup have the [problem] in #2426?

[problem]: https://github.com/magit/magit/issues/2426#issuecomment-162244161

johnmastro:
> Do any of the commands in this popup have the problem described in #2426?

Hm, I forgot about that.

I briefly tried each command in the popup (starting from `make emacs-Q` each time) and didn't encounter any problems. For `c` and `d`, which launch their respective popups, I only tried a couple commands each (specifically, `cc`, `ce`, `dd`, and `du`).

The benefit of having this autoloaded is admittedly minor, and I certainly won't object if you think it's not worth the potential hassle.

kyleam:
> I briefly tried each command in the popup (starting from make
> emacs-Q each time) and didn't encounter any problems. For c and d,
> which launch their respective popups, I only tried a couple commands
> each (specifically, cc, ce, dd, and du).

Thanks.  That was the only problem I could think of with this, so
looks good to me.  (Plus, I think most, if not all, of those commands
are autoloaded anyway, so I suppose it just makes running into the
pre-magit-status issue a bit more likely, which isn't necessarily a
bad thing.)

johnmastro:
Thanks!

PR #2635, opened 2016-05-03
Pass absolute file name to `make-temp-name`

### LINKED ISSUE #2633: Bad file descriptor error when creating stash
When creating a stash I got the following error.

```
magit-stash-create: Cannot create temporary name for prefix: bad file descriptor, index.magit.
```

When debugging the issue I found that a macro `magit-with-temp-index` is used that calls `make-temp-name` with a relative file prefix. According to the docs of `make-temp-name` the given prefix should be an absolute file name.

The offending line is at: https://github.com/magit/magit/blob/50a7f49b8f9e9f42f02aa3b2304f5c9e75267885/lisp/magit-git.el#L1165

Stashing works again when I rewrite this to

```
(make-temp-name (magit-git-dir "index.magit."))
```

As I'm unsure if that would impact working with remote repositories, I didn't file a PR. Please let me know if you'd like me to.

bjpbakker:
On OS X `make-temp-name` triggers a 'bad file descriptor' error when
passed a non-absolute file path. Moving `magit-git-dir` inside
`make-temp-name` ensures that an absolute file path is available.

This fixes #2633, where the above is also discussed.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1162,7 +1162,7 @@ Return a list of two integers: (A>B B>A)."
   (declare (indent 2) (debug (form form body)))
   (let ((file (cl-gensym "file")))
     `(let ((,file (magit-convert-git-filename
-                   (magit-git-dir (make-temp-name "index.magit.")))))
+                   (make-temp-name (magit-git-dir "index.magit.")))))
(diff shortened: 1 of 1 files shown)


kyleam:
Thanks

Issue #2634, opened 2016-05-03
Support for staging parts of a new file / --intent-to-add


cpaulik:
Adding a new file with magit is currently an all or nothing operation. But often it makes sense to commit a new file in parts. This can manually be done by using 

```shell
git add -N new_file
```

and then staging the parts you want in magit.

```
-N, --intent-to-add
    Record only the fact that the path will be added later.  An entry
    for the path is placed in the index with no content.  This is useful
    for, among other things, showing the unstaged content of such files
    with git diff and committing them with git commit -a.
```

Support for this would be great.

Solution from http://stackoverflow.com/questions/6436681/how-to-stage-only-part-of-a-new-file-with-git/6440500#6440500

kyleam:
I have a `magit-stage-file-intent` function in my [customization][si].
I don't use it that much (to the point where I'm fine calling `M-x
magit-stage-file-intent` instead of giving it a key binding).

Perhaps we could add a similar command but not expose it with a
binding. 

[si]: https://github.com/kyleam/emacs.d/blob/master/lisp/km-magit.el#L96

cpaulik:
Would it make sense to use the prefix argument for that? `C-u s` on an untracked file could add the file with -N enabled.

phil-s:
If `C-u s` wouldn't interfere with anything else, then that interface sounds like a good idea to me.

kyleam:
> Would it make sense to use the prefix argument for that? C-u s on an
> untracked file could add the file with -N enabled.

Yes, that's probably a better idea than a separate function.  Using a
prefix argument to `magit-stage` limits it to acting on the change at
point (versus completion), but I think that's fine.  I suppose the
main thing is whether we want to extend `magit-stage` directly for
what is probably a very uncommon operation for most users.

> C-u s sounds like a good interface to me, if it doesn't interfere
> with anything else.

I don't think it interferes with anything.  A draft using a prefix
argument with `magit-stage` is on the branch `km/intent`.

phil-s:
I've only done cursory testing, but that seems to work very nicely. I've wanted an easy way to do this on numerous occasions (but didn't know about `-N` otherwise I might have suggested it myself), so I think this is an excellent little enhancement.

cpaulik:
This works exactly as I envisioned it. So I would be very happy with this addition.

kyleam:
@phil-s @cpaulik Thanks for your feedback.  Closing in favor of #2638.

PR #2638, opened 2016-05-05
magit-stage: support --intent-to-add

kyleam:
Re: #2634


gomain:
Is this still working? This option seems not supported in "transient" pop-up.

kyleam:
> Is this still working? This option seems not supported in "transient" pop-up.

`--intent-to-add` is exposed via a prefix argument to `magit-stage`.  Try <kbd>C-u s</kbd> on an untracked file.

PR #2647, opened 2016-05-10
Make better use of git-apply's --reject and --3way arguments

tarsius:


lisp/magit-apply.el
```diff
@@ -72,7 +72,7 @@ information see command `magit-reverse-in-index'."
 
 (defun magit-apply (&rest args)
   "Apply the change at point."
-  (interactive)
+  (interactive list "--3way")
```
npostavs: Missed parens?

PR #2648, opened 2016-05-11
magit-clone: run asynchronously

### LINKED ISSUE #2639: Magit clone does not seem to work
`magit-clone` invocation results in a message “not inside a Git repository” and no new directory appears.

npostavs:
should fix #2639.

```
This allows intercepting password prompts, and is also nicer if the
clone happens to take a long time.
```

---

I'm having some trouble with the error reporting for when the clone happens to fail. See FIXME in code.

### DIFF
--- a/Documentation/RelNotes/2.7.0.txt
+++ b/Documentation/RelNotes/2.7.0.txt
@@ -34,6 +34,9 @@ Changes since v2.6.2
 
 * Added new section inserter `magit-insert-worktrees'
 
+* The command `magit-clone' now runs asynchronously, which avoid
+  blocking Emacs and allows handling password prompts.
+
 * The command `magit-stage' learned to stage an untracked file while
   leaving its content unstaged (i.e. `git add --intent-to-add') when
   called with a prefix argument.
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -811,10 +811,10 @@ as argument."
                             (--when-let (magit-section-content section)
                               (when (re-search-backward
                                      magit-process-error-message-re it t)
-                                (match-string 1))))))
+                                (match-string-no-properties 1))))))
                    "Git failed")))
       (if magit-process-raise-error
-          (signal 'magit-git-error (format "%s (in %s)" msg default-dir))
+          (signal 'magit-git-error (list (format "%s (in %s)" msg default-dir)))
         (--when-let (magit-mode-get-buffer 'magit-status-mode)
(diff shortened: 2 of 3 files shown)


tarsius:
> I'm having some trouble with the error reporting for when the clone happens to fail. 

Seems to work for me.

npostavs:
> Seems to work for me.

By "work" you mean if the `git clone` fails you get relevant error messages?

tarsius:
Yes. In the minibuffer and `*Messages*` I see `Could not read from remote repository. ... [Hit $ to see buffer *magit-process: magit for details]`.

npostavs:
Ah, I missed a condition: the non-relevant error message happens when calling `magit-clone` from a non-magit buffer.

npostavs:
Ok, I think I got it (looks I had the error messages reversed in the FIXME comment).

tarsius:
Please apply this for now.

But I think this kludge (the `condition-case` thing) should not be necessary. I'll investigate why it is necessary in this case and hope to be able to remove it once I have determined the underlaying cause. It probably has something to do with not being inside a Git repository.

npostavs:
> But I think this kludge (the condition-case thing) should not be necessary. I'll investigate why it is necessary in this case and hope to be able to remove it once I have determined the underlaying cause. It probably has something to do with not being inside a Git repository.

Binding `magit-process-raise-error` to `t` is needed if calling `magit-clone` from outside a git repository, because the `nil` branch in `magit-process-finish` assumes we're inside a repo.

I think the `condition-case` is needed due to a missing `list` call:
```diff
modified   lisp/magit-process.el
@@ -796,7 +796,7 @@ (defun magit-process-finish (arg &optional process-buf command-buf
                                 (match-string 1))))))
                    "Git failed")))
       (if magit-process-raise-error
-          (signal 'magit-git-error (format "%s (in %s)" msg default-dir))
+          (signal 'magit-git-error (list (format "%s (in %s)" msg default-dir)))
         (--when-let (magit-mode-get-buffer 'magit-status-mode)
           (setq magit-this-error msg))
         (message "%s ... [%s buffer %s for details]" msg
```
cf 
```
(signal ERROR-SYMBOL DATA)
[...]
DATA should be a list.
```

I'll check later today.

npostavs:
> I think the condition-case is needed due to a missing `list` call:

This was indeed the problem. Also needed to remove text properties or the error message comes with a whole bunch of `magit-section` data.

I added a RelNotes entry too.

PR #2649, opened 2016-05-11
magit-display-buffer-traditional: fix alist syntax

npostavs:
This possibly a bug reported in the chatroom:

> Here is what I got when I press `c c` in a magit buffer:
`*ERROR*: Wrong type argument: listp, inhibit-same-window error: There was a problem with the editor '/usr/bin/emacsclient --socket-name=/tmp/emacs1000/server'. Please supply the message using either -m or -F option.`

I made this fix because I think there is a missing layer of nesting around the alist, but committing is working for me either way (Emacs 24.5), so I'm not really sure if this fixes anything (hence I didn't commit this to `master` directly).


tarsius:
Thanks

Issue #2541, opened 2016-02-05
Option to make magit-status always open in current window


jimeh:
Since I updated magit, it always opens in a "other" window now. I recall earlier last year there were a few options around this behavior, and they changed names a few times, the most recent one I set in my config was `magit-status-buffer-switch-function` to `'switch-to-buffer` which made it open in the current buffer as I like.

But now there doesn't seem to be any options for it. And sadly my lisp skills aren't good enough to dig through the code and the find the solution in 5 minutes. Hence any possible of adding back such an option, or pointing me in the right direction to make magit behave that way?

Thanks :)

kyleam:
This is controlled by `magit-display-buffer-function`, which by
default is

```elisp
(defun magit-display-buffer-traditional (buffer)
  "Display BUFFER the way this has traditionally been done."
  (display-buffer
   buffer (if (and (derived-mode-p 'magit-mode)
                   (not (memq (with-current-buffer buffer major-mode)
                              '(magit-process-mode
                                magit-revision-mode
                                magit-diff-mode
                                magit-stash-mode
                                magit-status-mode))))
              '(display-buffer-same-window)
            nil))) ; display in another window
```

If you wanted Magit to open up every buffer in the current window, you
could set it to

```elisp
(setq magit-display-buffer-function
      (lambda (buffer)
        (display-buffer buffer '(display-buffer-same-window))))
```

However, that's not very convenient for things like showing revisions
from log buffers.  To behave like `magit-display-buffer-traditional`,
but open up the status in the current window, you use something like

```elisp
(setq magit-display-buffer-function
      (lambda (buffer)
        (display-buffer
         buffer (if (and (derived-mode-p 'magit-mode)
                         (memq (with-current-buffer buffer major-mode)
                               '(magit-process-mode
                                 magit-revision-mode
                                 magit-diff-mode
                                 magit-stash-mode
                                 magit-status-mode)))
                    nil
                  '(display-buffer-same-window)))))
```

jimeh:
@kyleam Thanks, your last code snippet seems to do the job.

Would it not make sense to have a more user friendly option for this stuff though? Cause I found it annoying that magit would pop up somewhere semi-random when I was jumping in and out of `magit-status` a lot.

fxleblanc:
[...]

PR #2656, opened 2016-05-21
add more magit-display-buffer-function options

kyleam:
Thoughts?  Other variants to consider including?

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -102,6 +102,9 @@ which in turn uses the function specified here."
   :package-version '(magit . "2.3.0")
   :group 'magit-modes
   :type '(radio (function-item magit-display-buffer-traditional)
+                (function-item magit-display-buffer-same-window-except-diff-v1)
+                (function-item magit-display-buffer-fullframe-status-v1)
+                (function-item magit-display-buffer-fullcolumn-most-v1)
                 (function-item display-buffer)
                 (function :tag "Function")))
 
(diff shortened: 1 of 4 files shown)


lisp/magit-mode.el
```diff
@@ -99,9 +99,12 @@ inside your function."
 All Magit buffers (buffers whose major-modes derive from
 `magit-mode') are displayed using `magit-display-buffer',
 which in turn uses the function specified here."
-  :package-version '(magit . "2.3.0")
+  :package-version '(magit . "2.7.1")
```
npostavs: Shouldn't this be left unchanged, since the default value is not changing?

kyleam: Doh, yep.

kyleam:
@Silex @jimeh @fxleblanc, if any of you have time, it'd be great to
get your feedback.  @Silex, `magit-display-buffer-fullframe-status-v1`
is the option for the full frame status buffer (#1953).  @jimeh
@fxleblanc, `magit-display-buffer-same-window-except-diff-v1` should
have a similar behavior to what was requested in #2541.

jimeh:
@kyleam The `magit-display-buffer-same-window-except-diff-v1` display function does exactly what I want, thanks :)

kyleam:
@jimeh Thanks for testing it out.

Applied  (fad19f5de2c964077758f4015efc856b5d64ab14).

tarsius:
I've it turns out that one of these functions still needs some tweaking *before* the next release, then we should just do it, instead of adding a new version. Just saying :-)

> Other variants to consider including?

Personally I haven't gone with fullframe-status yet, because I am so used to having the status (or log) buffer at the bottom but when a second window is created to display another Magit buffer, then *that* buffer is displayed at the bottom. So something that uses `display-buffer-above-selected` (doesn't exist yet) would be nice.

kyleam:
> I've it turns out that one of these functions still needs some tweaking before the next release, then we should just do it, instead of adding a new version. Just saying :-)

Right, my thinking too.

> Personally I haven't gone with fullframe-status yet, because I am so used to having the status (or log) buffer at the bottom but when a second window is created to display another Magit buffer, then that buffer is displayed at the bottom. So something that uses display-buffer-above-selected (doesn't exist yet) would be nice.

Interesting.  I'll look into that.

kyleam:
`km/display-topleft` adds a function that I hope gets close to the behavior you want.

tarsius:
That works very well. Thanks!

(Not so well when using `magit-mode-quit-window` as `magit-bury-buffer-function`, so I went back to the default `magit-restore-window-configuration` for now.)

kyleam:
Great.  I've added the function, with a few tweaks to make it play better
with `magit-mode-bury-buffer`.

PR #2658, opened 2016-05-22
magit-insert-worktrees: use relative filenames

npostavs:
IMO, relative path names that are lined up are nicer to look at. Should the padding width be configurable, or is it enough to just choose a "sufficiently" large number?

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -2194,27 +2194,24 @@ If there is only one worktree, then insert nothing."
     (when (> (length worktrees) 1)
       (magit-insert-section (worktrees)
         (magit-insert-heading "Worktrees:")
-        (let ((head-width 0))
-          (pcase-dolist
-              (`(,head . ,path)
-               ;; Decide what to insert, and check column width.
-               (mapcar
-                (-lambda ((path barep commit branch))
-                  (let ((relpath (file-relative-name path))
-                        (head (cond
-                               (branch (propertize branch
-                                                   'face 'magit-branch-local))
-                               (commit (propertize (magit-rev-abbrev commit)
-                                                   'face 'magit-hash))
-                               (barep  "(bare)")
-                               (t      "(detached)"))))
-                    (setq path (abbreviate-file-name path))
-                    (when (< (length relpath) (length path))
-                      (setq path relpath))
-                    (setq head-width (max head-width (length head)))
-                    `(,head . ,path)))
-                worktrees))
-            ;; Now actually insert the sections.
+        (let* ((section-data
+                (mapcar
+                 (-lambda ((path barep commit branch))
+                   (setq path (abbreviate-file-name path))
+                   (cons (cond
+                          (branch (propertize branch
+                                              'face 'magit-branch-local))
+                          (commit (propertize (magit-rev-abbrev commit)
+                                              'face 'magit-hash))
+                          (barep  "(bare)")
+                          (t      "(detached)"))
+                         (let ((relpath (file-relative-name path)))
+                           (if (< (length relpath) (length path))
+                               relpath path))))
+                 worktrees))
+               (head-width (-max (mapcar (-lambda ((head . _path)) (length head))
+                                         section-data))))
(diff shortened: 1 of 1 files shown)


tarsius:
> Should the padding width be configurable, or is it enough to just choose a "sufficiently" large number?

Neither. Let's determine the appropriate width dynamically.

tarsius:
Hm. I cannot try this out right now. v2.8.3's `git worktree list [--porcelain]` segfaults when called from Magit.

I think I prefer absolute paths here. When the various worktrees are not located inside the the same parent directory, then relative file paths can be strange. Maybe you should add an option for *this* (but not the padding).

tarsius:
Narrowed the segfault down to `git --no-pager worktree list` (also from command line).

tarsius:
> When the various worktrees are not located inside the the same parent directory, then relative file paths can be strange.

That looks okay/better than absolute:

```
Worktrees (2)
master               .
other                ../morlock2
```

But this is terrible:

```
Worktrees (2)
master               .
emacs-25             ../../../../../mnt/1/Repos/emacs/25.0.94
```

(Here the two worktrees are actually located in the same directory, but symlinks are involved and `git worktree` does not appear those consistently.

npostavs:
> Neither. Let's determine the appropriate width dynamically.

Makes sense.

> But this is terrible:

Yeah, that's pretty gross. What about using the absolute file name if it's shorter? (see the `squash!` commit I just pushed).

tarsius:
That's a good compromise.

lisp/magit.el
```diff
@@ -2194,18 +2194,31 @@ If there is only one worktree, then insert nothing."
...
+                                                   'face 'magit-branch-local))
+                               (commit (propertize (magit-rev-abbrev commit)
+                                                   'face 'magit-hash))
+                               (barep  "(bare)")
+                               (t      "(detached)"))))
+                    (setq path (abbreviate-file-name path))
+                    (when (< (length relpath) (length path))
+                      (setq path relpath))
+                    (setq head-width (max head-width (length head)))
+                    `(,head . ,path)))
```
tarsius: Yikes. Just use `cons`.

lisp/magit.el
```diff
@@ -2194,18 +2194,31 @@ If there is only one worktree, then insert nothing."
     (when (> (length worktrees) 1)
       (magit-insert-section (worktrees)
         (magit-insert-heading "Worktrees:")
-        (dolist (worktree worktrees)
-          (-let [(path barep commit branch) worktree]
+        (let ((head-width 0))
+          (pcase-dolist
+              (`(,head . ,path)
+               ;; Decide what to insert, and check column width.
```
tarsius: I would probably break this up instead of trying to do everything at once. That would make this more readable and wouldn't impact performance (I assume that nobody will ever need more than a few dozen worktrees).

lisp/magit.el
npostavs: But this way it corresponds directly with the `pcase-dolist` spec.

lisp/magit.el
npostavs: done (also replaces the backquoting with `cons`)

tarsius:
Okay so we could probably have done without my nitpicking. Anyway I just pushed a fixup with my final tweaks. Take it (possibly partially) if you like it.

npostavs:
Mmm, I don't like reusing the `worktrees` variable, so I dropped that part.

But before I merge, I wanted to ask about the `(t      "(detached)")` clause: doesn't the `(commit ...)` clause represent the detached case?

tarsius:
> Mmm, I don't like reusing the `worktrees` variable, so I dropped that part.

Actually I went back and forth between changing the value and using a new variable. That's one style question where I still haven't made up my mind ;-)

tarsius:
> doesn't the `(commit ...)` clause represent the detached case?

Logic and experimentation would suggest so. I think it is save to drop the `t` case. (I don't really think it is necessary to append "(detached)" in the commit case, but it wouldn't hurt either.)

PR #2661, opened 2016-05-24
Tweaks to branch create and rename UI

npostavs:
I have the vague memory that @tarsius rejected something like the branch rename change, but I can't recall why.

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -1533,7 +1533,8 @@ With prefix, forces the rename even if NEW already exists.
   (interactive
    (let ((branch (magit-read-local-branch "Rename branch")))
      (list branch
-           (magit-read-string-ns (format "Rename branch '%s' to" branch))
+           (magit-read-string-ns (format "Rename branch '%s' to" branch)
+                                 nil 'magit-revision-history)
            current-prefix-arg)))
   (unless (string= old new)
     (magit-run-git "branch" (if force "-M" "-m") old new)))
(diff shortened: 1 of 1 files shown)


tarsius:
> I have the vague memory that @tarsius rejected something like the branch rename change, but I can't recall why.

Looks good to me now ;-)

lisp/magit.el
```diff
@@ -1371,6 +1371,7 @@ changes.
            (setq branch (magit-read-string-ns
                          "Branch name"
                          (and (member start (magit-list-remote-branch-names))
+                              (not (member start (magit-list-local-branch-names)))
```
tarsius: That doesn't appear to make a difference. If start is a member of `magit-list-remote-branch-names` (e.g. "origin/feature"), then we already know that it cannot be a member of `magit-list-local-branch-names` (e.g. "feature"). Okay, except in broken cases such as when a ref `refs/heads/origin/feature` exists in a addition to `refs/remotes/origin/feature`. Am I missing something?

npostavs: > That doesn't appear to make a difference.

Hmm, you're right, now I'm unsure why I though it did.

> I cannot reproduce that.

Oh, it's actually "when creating a new topic branch  from `origin/master`". Maybe when I tested my "fix" I entered `master`.

npostavs:
I realized we can just use the existing `magit-revision-history` to provide the original branch name from `M-p` (like `rename-file`, not `rename-buffer`). 

Fixed my fix for the "create branch from `origin/master`" case.

Not merging immediately in case I've still left some silly mistakes.

npostavs:
picked to `master`: c6e3293e21dc7b1bcbbcec43ad90b2ceb6aba296, 7eac94f22cddc727142d80a277a4f8452e557fc1

PR #2662, opened 2016-05-26
Add option magit-mute-auto-revert-messages

gambyte:
This adds a customize option to mute the messages for
magit-auto-revert-mode starting, which can be annoying.

### DIFF
--- a/lisp/magit-autorevert.el
+++ b/lisp/magit-autorevert.el
@@ -110,6 +110,13 @@ seconds of user inactivity.  That is not desirable."
 (make-obsolete-variable 'magit-revert-buffers 'magit-auto-revert-mode
                         "Magit 2.4.0")
 
+(defcustom magit-mute-auto-revert-messages nil
+  "Whether `magit-auto-revert-mode' emits messages upon
+  starting."
+  :package-version '(magit . "2.7.2")
+  :group 'magit-auto-revert
+  :type 'boolean)
+
 ;;;###autoload
 (define-globalized-minor-mode magit-auto-revert-mode auto-revert-mode
   magit-turn-on-auto-revert-mode-if-desired
@@ -147,15 +154,17 @@ and code surrounding the definition of this function."
   ;; had been established, so consult the value of both variables.
   (if (and magit-auto-revert-mode magit-revert-buffers)
       (let ((start (current-time)))
-        (message "Turning on magit-auto-revert-mode...")
+        (unless magit-mute-auto-revert-messages
+          (message "Turning on magit-auto-revert-mode..."))
         (magit-auto-revert-mode 1)
-        (message
-         "Turning on magit-auto-revert-mode...done%s"
-         (let ((elapsed (float-time (time-subtract (current-time) start))))
-           (if (> elapsed 0.2)
-               (format " (%.3fs, %s buffers checked)" elapsed
-                       (length (buffer-list)))
-             ""))))
+        (unless magit-mute-auto-revert-messages
+          (message
+           "Turning on magit-auto-revert-mode...done%s"
+           (let ((elapsed (float-time (time-subtract (current-time) start))))
+             (if (> elapsed 0.2)
+                 (format " (%.3fs, %s buffers checked)" elapsed
+                         (length (buffer-list)))
+               "")))))
     (magit-auto-revert-mode -1)))
 (if after-init-time
     ;; Since `after-init-hook' has already been


tarsius:
Without this option you see this *once* per Emacs session:

```
Turning on magit-auto-revert-mode...done
```

I don't think it is worth adding an option to prevent that message from briefly appearing in the echo area and being logged in the `*Messages*` buffer. Adding an option has a cost too: users who go through the list of available options to find optional features and alternatives to the default behavior, would now also have to think about this option.

gambyte:
Yes it is only once, but I find it extremely annoying. I wrote this because it very often clobbers a message which I wanted, instead I get this completely useless (to me) message forcing me to go to the message buffer to get the information I wanted. The alternative is that I and anyone else who doesn't want this arguably debugging information is to write some hackish customization around the problem which could break with future changes.

As for the the supposed cost, anyone who does `M-x customize-group RET magit-auto-revert RET` will find a whopping five options with this change. Each of them include a doc string describing exactly what they do. How stupid do you think the end users are that it is some great burden to parse what

>Whether `magit-auto-revert-mode` emits messages upon starting.

might mean? Besides, the cost of not having it is much greater since everyone who doesn't want it will have to re-implement a solution themselves.

Really this is a deficiency of core emacs which has long needed a facility to change the verbosity level of messages. This isn't the first package I've come across which has had this problem, so I know I'm not the only one who dislikes this sort of thing.

tarsius:
> This isn't the first package I've come across which has had this problem, so I know I'm not the only one who dislikes this sort of thing.

This statement is logically equivalent to "I see many clouds, therefore I am not the only one who dislikes rain." The latter might very well be true, but that doesn't follow from the former.

Likewise you jump to an unfounded conclusion here too:

>> users [...] would now also have to think about this option.

> How stupid do you think the end users are that it is some great burden to parse [...]

I did in no way imply that users are stupid or that the proposed doc-string is hard to understand. What I meant is that I do not want to add yet another option because there are already very many options and it is time consuming to make an informed choice in every case. It's time consuming even for extremely intelligent people.

There exist some options which change Magit's behavior in fundamental ways. Burying those options in a see of options that change the behavior in (for most users) rather insignificant ways, does not serve users well. So yes, I am making things a bit more inconvenient for you, for the benefit of other users. As the maintainer of a free software project, one sometimes has to make such unpopular decisions.

Magit currently has ~140 options, so granted adding one more option doesn't make much of a difference. But this isn't the first time that I am not adding a requested option. If I had added every option ever suggested, then there would be several dozen additional options.

I do suspect that there are extremely few users who are inconvenienced by this message. But I am aware that I might be wrong about that. (Likewise you cannot, at this time, know the opposite to be true either.) However if it turns out that other users are also inconvenienced by this message, then I will certainly reconsider. It wouldn't be the first time that I change my opinion, once additional information becomes available.

gambyte:
If you don't want to add this option that's fine, however

>This statement is logically equivalent to "I see many clouds, therefore I am not the only one who dislikes rain." 

is incorrect. You've misunderstood my claim due to my poor wording. I've seen other packages which have had spurious messages which resulted in people complaining. Typically the solution is that the messages are just removed since they usually don't do anything useful. I thought an option might be preferable since it appears to offer some minute amount of useful information. However, outside of debugging, I'm not sure what practical use you see these messages as having.

In terms of evidence, [here](https://lists.gnu.org/archive/html/emacs-devel/2015-04/msg01061.html) [are](https://lists.gnu.org/archive/html/emacs-devel/2011-02/msg00659.html) [some](https://lists.gnu.org/archive/html/emacs-devel/2009-01/msg00305.html) threads from emacs-devel where users complain about messages which annoy them. Additionally, in Emacs core there is the option `find-file-suppress-same-file-warnings` and the [infamous](https://lists.gnu.org/archive/cgi-bin/namazu.cgi?query=inhibit-startup-echo-area-message&submit=Search%21&idxname=emacs-devel&max=20&result=normal&sort=score) `inhibit-startup-echo-area-message`, both of which only exist to satiate some users' desire to not have messages they don't want.

As for the several dozen other options I don't see why you shouldn't add them. I doubt many people bother to go through all of a package's customization options one-by-one, particularly for magit since it is an excellent example of good defaults. If you're worried about clutter you could just create a new customization group like `magit-superfluous`. I don't think I've ever heard an Emacs user who complained of too many options.

tarsius:
Thanks for this thoughtful response. I'll let this rest a little and decide some other time.

---

The message is not displayed for debugging purposes only. The main purpose is to give users a change to notice that Magit enables some kind of automatic reverts. I think that reverting buffers when the files on disk change is an extremely useful feature, which should be enabled by default. But there also is a very small risk of data-loss, and that's the reason I added this message and [this explanation](https://magit.vc/manual/magit/Risk-of-Reverting-Automatically.html) to the manual.

> If you're worried about clutter you could just create a new customization group like `magit-superfluous`.

A reasonable suggestion. I've also been thinking about something like a `magit-no-chatter` option, similar to `magit-no-confirm`.

tarsius:
I have added option `magit-no-message`.

gambyte:
Thanks!

PR #2667, opened 2016-05-28
support showing and copying bisect-output section

### LINKED ISSUE #2655: magit-bisect result section improvement
At the end of a `magit-bisect` at the `<hex> is the first bad commit` section, it would be great if:

* `RET` would go to the offending commit
* `magit-copy-section-value` and/or `magit-copy-buffer-revision` would copy the commit revision
(currently it copies `t` and HEAD's revision respectively)

`Magit 20160517.838, Git 2.5.5, Emacs 24.5.1, gnu/linux`

kyleam:
Closes #2655.

### DIFF
--- a/lisp/magit-bisect.el
+++ b/lisp/magit-bisect.el
@@ -135,18 +135,20 @@ bisect run'."
 (defun magit-insert-bisect-output ()
   "While bisecting, insert section with output from `git bisect'."
   (when (magit-bisect-in-progress-p)
-    (let ((lines
-           (or (magit-file-lines (magit-git-dir "BISECT_CMD_OUTPUT"))
-               (list "Bisecting: (no saved bisect output)"
-                     "It appears you have invoked `git bisect' from a shell."
-                     "There is nothing wrong with that, we just cannot display"
-                     "anything useful here.  Consult the shell output instead.")))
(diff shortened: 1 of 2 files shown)


lisp/magit-bisect.el
```diff
@@ -135,18 +141,19 @@ bisect run'."
...
+            (or (magit-file-lines (magit-git-dir "BISECT_CMD_OUTPUT"))
+                (list "Bisecting: (no saved bisect output)"
+                      "It appears you have invoked `git bisect' from a shell."
+                      "There is nothing wrong with that, we just cannot display"
+                      "anything useful here.  Consult the shell output instead.")))
+           (done-re "^\\([a-z0-9]\\{40\\}\\) is the first bad commit$")
+           (bad-line (or (and (string-match done-re (car lines))
+                              (pop lines))
+                         (--first (string-match done-re it) lines))))
+      (magit-insert-section (bisect-output (and bad-line (match-string 1 bad-line)))
```
tarsius: Since currently nothing actually uses the `bisect-output` section type, I think we could just use `commit` in the "found bad" case.

```lisp
(magit-insert-section ((eval (if bad 'commit 'bisect-output)) ...) ...)
```

kyleam: agreed, thanks

PR #2670, opened 2016-05-30
cache git config more aggressively

npostavs:
It shaves a couple 100 ms or so from refreshing the status

With `master`:
```
Refreshing magit...done (1.107s, cached 27/53)
Refreshing magit...done (1.107s, cached 27/53)
Refreshing magit...done (1.108s, cached 27/53)
```

With the patch:
```
Refreshing magit...done (0.921s, cached 33/53)
Refreshing magit...done (0.936s, cached 33/53)
Refreshing magit...done (0.951s, cached 33/53)
```

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1426,18 +1426,34 @@ Return a list of two integers: (A>B B>A)."
 
 ;;; Variables
 
+(defun magit-config-slurp ()
+  (magit--with-refresh-cache 'config
+    (let ((configs (make-hash-table :test 'equal)))
+      (dolist (conf (magit-git-items "config" "--list" "-z"))
+        (let* ((nl-pos (cl-position ?\n conf))
+               (key (substring conf 0 nl-pos))
+               (val (substring conf (1+ nl-pos))))
+          (puthash key (cons val (gethash key configs)) configs)))
+      configs)))
+
 (defun magit-get (&rest keys)
   "Return the value of Git config entry specified by KEYS."
-  (magit-git-str "config" (mapconcat 'identity keys ".")))
+  (if magit--refresh-cache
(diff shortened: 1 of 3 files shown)


tarsius:
Yes, let's do that.

I am not a fan of the order of `magit-set`s arguments. To avoid mistakes I would recommend using `(magit-set value "single.variable.string")` when all parts of the variable name are constant. Or maybe we should deprecate `magit-{get,set}` and add `magit-config-{get,set,unset}` like:

```lisp
(defun magit-config-set (&rest args)
  "Set the value of the Git VARIABLE to VALUE.
\n(fn VARIABLE [VARIABLE-PART...] VALUE)"
  (magit-git-string "config"
                    (mapconcat #'identity (butlast args) ".")
                    (car (last args))))

(defun magit-config-unset (variable &rest variable-part)
  (magit-git-string "config" "--unset" (cons variable variable-part)))
```

npostavs:
While I was writing this code, I had the idea that we could somehow use the `gv` machinery so that `magit-set` could be done with `(setf (magit-get <var>) <value>)`.

tarsius:
I like your idea better than what I previously suggested.

npostavs:
Dropping the changes to `magit-set`, it's not really related to this; it just happened to come up when I grepped for `git config` calls. I'll do the `setf` thing later.

---

Oh, by the way, the original timings were on Windows, on GNU/Linux there's not really a difference:

`master`
```
Refreshing magit...done (0.123s, cached 24/52)
Refreshing magit...done (0.124s, cached 24/52)
Refreshing magit...done (0.160s, cached 24/52)
```

with patch
```
Refreshing magit...done (0.161s, cached 29/52)
Refreshing magit...done (0.103s, cached 29/52)
Refreshing magit...done (0.108s, cached 29/52)
```

Issue #2657, opened 2016-05-22
submodule updating


npostavs:
I don't use submodules a lot, basically my only use is when working on yasnippet to update the yasnippet-snippets submodule. The current interface seems suboptimal for my minimal use cases:

#### I need get the changes made in the upstream yasnippet-snippets recorded into the submodule.

The git command for this is `git submodule update --remote` (which I have to lookup pretty much every time). Perhaps `--remote` should be added to the submodule popup? Although my feeling is that the actions should be renamed/reorganized because I can never remember the difference between `update`, `sync`, and `fetch` (and in this particular use case they are all nops).

#### I've switched branches and so the submodule has been left in a modified state.

It's a bit annoying that `k` (`magit-discard`) does nothing on submodule sections. Actually, it looks like magit doesn't even distinguish submodule section as a separate type (even in case of `diff.submodule = log` (this is way too obscure by the way)). We could do that by checking for the special file mode `160000`, according to https://git-scm.com/book/en/v2/Git-Tools-Submodules. 

But maybe it's better to pass `--ignore-submodules` to `git diff` and just rely on `magit-insert-submodules` instead? Speaking of...

#### Looking at submodule status

The submodules that I work with don't have tags, so `magit-insert-submodules` doesn't give very useful info (just a list of directory names). What about using `git submodule status` output for this?

---

Thoughts?

tarsius:
@npostavs @kyleam Now that v2.7 is out I intend to give the "take a break" thing another try. Which means I will try very hard not to implement new features *myself*. But you are more than welcome to implement new features or improve features you feel are lacking. I'll try to comment in a useful time frame, but in some cases my response will essentially be "looks good in principal, I haven't looked at the details, go ahead".

#2656 is an instance of that. We had a discussion about that earlier and I think we pretty much agreed on what to do, just didn't get around to doing it. Now that you have done it, you can probably just go ahead and merge it. Or you could first invite the users who previously showed interest in such variants and ask them to give your implementations a try before you merge.

But it is a somewhat big change, so I appreciate that you opened a pull request and gave my the opportunity to raise any objects. I don't have any in this case - now it is up to you to decide how to proceed.

----

This issue (#2657) is different. It's much more open ended. Please develop these ideas further and then open pull requests. In this case me trying to take a break will probably lead to some delays.

tarsius:
Okay so back from the meta-level...

[...]

PR #2674, opened 2016-05-31
Improve submodule commands

tarsius:
I am opening a pr for your branch so that I can comment on some things without risking that those comments are lost.

(Ref #2657)

### DIFF
--- a/lisp/magit-submodule.el
+++ b/lisp/magit-submodule.el
@@ -276,6 +276,38 @@ These sections can be expanded to show the respective commits."
                               'magit-get-push-branch
                               "%s..HEAD"))
 
+(defun magit-submodule-visit (module &optional other-window)
+  "Visit MODULE by calling `magit-status' on it.
+Offer to initialize MODULE if it's not checked out yet."
+  (interactive (list (or (magit-section-when module)
+                         (user-error "No submodule at point"))
+                     current-prefix-arg))
+  (let ((path (expand-file-name module)))
+    (if (or (file-exists-p (expand-file-name ".git" module))
+            (not (y-or-n-p (format "Setup submodule '%s' first?"
+                                   module))))
+        (magit-diff-visit-directory path other-window)
+      (magit-submodule-setup module)
+      (set-process-sentinel
+       (lambda (process event)
+         (when (memq (process-status process) '(exit signal))
+           (let ((magit-process-raise-error t))
+             (magit-process-sentinel process event)))
+         (when (and (eq (process-status process) 'exit)
+                    (= (process-exit-status process) 0))
+           (magit-diff-visit-directory path other-window)))))))
+
+(defvar magit-submodule-section-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map [C-return] 'magit-submodule-visit)
+    (define-key map "\C-j"     'magit-submodule-visit)
+    (define-key map [remap magit-visit-thing]  'magit-submodule-visit)
+    (define-key map [remap magit-delete-thing] 'magit-submodule-deinit)
+    (define-key map "K" 'magit-file-untrack)
+    (define-key map "R" 'magit-file-rename)
+    map)
+  "Keymap for `submodule' sections.")
+
 (defun magit--insert-modules-logs (heading type fn format)
   "For internal use, don't add to a hook."
   (-when-let (modules (magit-get-submodules))
--- a/lisp/magit-submodule.el
+++ b/lisp/magit-submodule.el
@@ -74,12 +74,27 @@ an alist that supports the keys `:right-align' and `:pad-right'."
   "Popup console for submodule commands."
   'magit-commands nil nil
   :man-page "git-submodule"
-  :actions  '((?a "Add"    magit-submodule-add)
-              (?b "Setup"  magit-submodule-setup)
-              (?u "Update" magit-submodule-update)
-              (?f "Fetch"  magit-submodule-fetch)
+  ;:max-action-columns 2
+  :actions  '("Managing"
+              (?a "Add"    magit-submodule-add)
+              (?s "Setup"  magit-submodule-setup)
               (?d "Deinit" magit-submodule-deinit)
-              (?C "Configure..." magit-submodule-config-popup)))
+              (?C "Configure..." magit-submodule-config-popup)
+              "Updating"
+              (?u "Update" magit-submodule-update)
+              (?U "Update all" magit-submodule-update-all)
+              (?F "Pull" magit-submodule-update-remote)
+              (?R "Pull all" magit-submodule-update-all-remote)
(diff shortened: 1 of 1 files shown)


tarsius:
Is it no longer possible to comment on a *commit* itself instead of on changed lines?!

> eec3a3dbd439338b2a5b79586a52ee497edd2609 magit-submodule-add: fix interactive form

The things that seem dubious to you are all due to my desire to add an existing repository (which is already located inside an outer repository) as a submodule. I.e. when I create a new Emacs package, then I first create a new local repository, and then create and push an initial version. Only after that I add the repository as a submodule to `~/.emacs.d`. (I don't think you can add a repository that doesn't have an initial commit yet as a submodule.)

> removed let-binding of default-directory around (magit-get "remote.origin.url") due to reordering of evaluation.  But I don't understand what that was good for anyway.

I just configured "origin" inside the inner repository. Now that I add it as a submodule I don't want to type it again. Changing the minibuffer order breaks that (and that might be okay, see below). But if we do change the order, then we probably should drop the `(magit-get "remote.origin.url")` altogether. (It would be nice to get some "templates" like `git@github.com:my/.git` as completion candidates, but that's something that should be solved in a more general way).

> default value for submodule name has [...code...]

If you add a submodule and then unstage it but later want to add it as a submodule again, then it is nice not to have to type the url again. This happened a lot to me during the creation of [`emacs.g`](https://github.com/emacscollective/emacs.g). I.e. I added a package submodule early on, but later decided that this package should not be part of the proto-starter-kit I publish but should be added later in the private part which is built on top of `emacs.g`, but not public.

---

It probably makes sense to change the *command* `magit-submodule-add` as you suggest. At least the order in which values are read - the existing order doesn't make much sense if you don't know about the above use-case. And I simply use <kbd>s</kbd> to do this now.

innocence_195

However, `magit-stage` ultimately adds a submodule by calling `magit-submodule-add` interactively, so all code removed from the latter has to be moved somewhere else.

lisp/magit-submodule.el
```diff
@@ -131,10 +131,15 @@ For each section insert the path and the output of `git describe --tags'."
                   (expand-file-name (file-name-as-directory module))))
             (magit-insert-section (file module t)
               (insert (format "%-25s " module))
-              (--when-let (magit-git-string "describe" "--tags")
-                (when (string-match-p "\\`[0-9]" it)
-                  (insert ?\s))
-                (insert it))
+              (if (not (file-exists-p ".git")) (insert " (unitialized)")
```
tarsius: s/unitialized/uninitalized/

tarsius:
> a434b0f06870e6fdcb3dd64179389e8a8541915f Add new module section type

This has the potential of breaking many commands.

`magit-submodule-visit` is unnecessary because I already taught `magit-diff-visit-file` about the submodule special case, just press <kbd>RET</kbd>. Well, `magit-submodule-visit` does a little more, but `magit-diff-visit-file` could be changed to call this function without having to add a new section type and then having to make adjustments in many places.

(In the long run I want to use Eieio, among other things so that a section type can derive from another. But in the short run adding a new section type is risky and likely to cause more work than expected.)

npostavs:
> This has the potential of breaking many commands.

Speaking only of module sections under the "Modules" heading, not staged/unstaged changes to submodules (I've not looked into changing the latter yet):

Currently `a` (`magit-apply`) and `v` (`magit-reverse`) throw errors on submodule sections.
`s` (`magit-stage`), `u` (`magit-unstage`), and `k` (`magit-discard`) silently do nothing.

As far I can tell, `RET` (`magit-diff-visit-file`) is the only non-broken command anyway.

tarsius:
> Speaking only of module sections under the "Modules" heading, not staged/unstaged changes to submodules (I've not looked into changing the latter yet):

I didn't realize that, and retract my objection.

> Currently `a` (`magit-apply`) and `v` (`magit-reverse`) throw errors on submodule sections.
> `s` (`magit-stage`), `u` (`magit-unstage`), and `k` (`magit-discard`) silently do nothing.

And your changes result in more meaningful errors? That would be an improvement.

npostavs:
> And your changes result in more meaningful errors? That would be an improvement.

With my changes `a`, `v`, `s`, `u` all prompt for a file/commit, i.e. the submodule section no longer affects these commands. `k` prompts for submodule to `deinit` (defaulting to submodule at point).

---

> The things that seem dubious to you are all due to my desire to add an existing repository (which is already located inside an outer repository) as a submodule[...]

Oh, I hadn't even considered that possibility at all.

> However, `magit-stage` ultimately adds a submodule by calling `magit-submodule-add` interactively, so all code removed from the latter has to be moved somewhere else.

I think it should go into a separate command `magit-submodule-add-from-nested-repo`. And `magit-submodule-add` should be changed to run async like the other `magit-submodule-*` commands (`magit-submodule-add-from-nested-repo` would stay synchronous because `magit-stage` is synchronous).

tarsius:
>> However, `magit-stage` ultimately adds a submodule by calling `magit-submodule-add` interactively, so all code removed from the latter has to be moved somewhere else.

> I think it should go into a separate command `magit-submodule-add-from-nested-repo`. And `magit-submodule-add` should be changed to run async like the other `magit-submodule-*` commands (`magit-submodule-add-from-nested-repo` would stay synchronous because `magit-stage` is synchronous).

I think this should be kept as it is, i.e. <kbd>s</kbd> on a nested repository prompts to add that as a submodule instead of calling `git add <repo>`, because that would add the files inside `<repo>` to the outer repository and they would then be tracked in two repositories which wouldn't make sense.

npostavs:
> I think this should be kept as it is, i.e. <kbd>s</kbd> on a nested repository prompts to add that as a submodule instead of calling `git add <repo>`, because that would add the files inside `<repo>` to the outer repository and they would then be tracked in two repositories which wouldn't make sense.

Yes agreed, my comment was only regarding where to put the code, not to change the behaviour of `s`.

tarsius:
If cherry-picked "magit-submodule-setup: only setup uninitialized" with some cosmetics, and pushed the rest after rebasing onto `master`.

tarsius:
We still have to take care of https://github.com/magit/magit/issues/2657#issuecomment-237976396

> `k` on a *staged* submodule attempts `delete-file` on the directory resulting in a `file-error` ("is a directory").

npostavs:
> `k` on a *staged* submodule attempts `delete-file` on the directory resulting in a `file-error` ("is a directory").

I think fixing this would require making submodule changes to be a new type of section though.

tarsius:
I've applied the branch column patch and implemented the suggested dynamic column width in a separate commit.

Then I have rebased this branch and changed the name of the new section type to `submodule` and moved the respective commit before the others.

I have some doubts about the configuration subpopup and the edit file commands. I haven't yet looked at the other new command.

PR #2675, opened 2016-05-31
Make sure the right default-directory value is set to avoid the "Not …

wrn:
…in a git repository" error.

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -1177,9 +1177,11 @@ commit as argument."
 (defun magit-log-select-quit ()
   "Abort selecting a commit, don't act on any commit."
   (interactive)
-  (kill-buffer (current-buffer))
-  (when magit-log-select-quit-function
-    (funcall magit-log-select-quit-function)))
+  (let ((start-default-directory default-directory))
+    (kill-buffer (current-buffer))
+    (when magit-log-select-quit-function
+      (let ((default-directory start-default-directory))
+        (funcall magit-log-select-quit-function)))))
 
 ;;; Cherry Mode
 
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -1168,9 +1168,11 @@ Call `magit-log-select-pick-function' with the selected
 commit as argument."
   (interactive)
   (let ((fun magit-log-select-pick-function)
-        (rev (magit-commit-at-point)))
+        (rev (magit-commit-at-point))
+        (start-default-directory default-directory))
     (kill-buffer (current-buffer))
-    (funcall fun rev)))
+    (let ((default-directory start-default-directory))
+      (funcall fun rev))))
 
 (defun magit-log-select-quit ()
   "Abort selecting a commit, don't act on any commit."


wrn:
I am new to git and github contribution (only grabbed stuff from here before), so please forgive me if I don't appear to know what I am doing.

I do believe the pull request here fixed a bug (try magit-rebase-interactive and most likely you will bump into this bug).

Another related bug, which I don't know how to fix, is if you do M-x magit-rebase-interactive, choose the revision and then abort from the "git-rebase-todo" buffer by "C-c C-k", I will get the "error in process sentinel: Not inside a Git repository" message. I think this again is related to the wrong default-directory, but not sure where the code is.

npostavs:
I don't quite understand the scenario. Do you mean you're doing `M-x magit-rebase-interactive RET` from outside the git repository?

wrn:
I did do M-x magit-rebase-interactive from inside a git repository in a single window, then the system will split the window into two, with the second window showing the logs for me to choose a revision from. After I do "C-c C-c" on a revision line, the code will kill the log buffer, and now current buffer becomes whatever next buffer in the ring, which may have a different default-directory, then the git function will run in the wrong default-directory.

npostavs:
> I did do M-x magit-rebase-interactive from inside a git repository in a single window, then the system will split the window into two,

This part doesn't happen for me: the log buffer shows up in the same window. Do you see this from `emacs -Q`?

kyleam:
> This part doesn't happen for me: the log buffer shows up in the same
> window. Do you see this from emacs -Q?

I don't think this will be there with `emacs -Q` because
`magit-display-buffer-traditional` (as well as all the current
`magit-display-buffer-*` variants, I think) display modes that are
derived from `magit-log-mode` in the same buffer.  However, we should
account for the possibility of the `magit-log-select-mode` buffer
showing up in the other window since the user may supply a custom
function to `magit-display-buffer-function` or modify
`display-buffer-alist`.

I think ideally, before calling `FUN`, we'd return to the buffer from
which `magit-log-select-mode` was called, rather than just let-binding
`default-directory`.

wrn:
Even with "emacs -Q" I still have log buffer in a different window. I am using magit-20160522.1346 (from elpa) and Emacs 24.5.1, on CentOS.

I agree with kyleam on we should return to the originating buffer (maybe with "with-current-buffer").

I am not that familiar with the code here. I hope this pull request exposes some problems we need to fix:

1. magit-log-select-pick  and magit-log-select-quit need to have the right current buffer before calling the custom function
2. when we abort from git-rebase-todo buffer with "C-c C-k", the right current buffer or/and default directory need to be set to avoid "Not in git repository" error

But maybe somebody else can come up with a more proper and elegant fix? Thanks.

kyleam:
@wrn I don't understand why you're seeing the buffer displayed in the
other window with the default configuration, but either way this
shouldn't give an error.  Please give 7ceca0a a try.

wrn:
7ceca0a works! For some reason, it even fixed "Not in git repository" error from the "C-c C-k" abort in git-rebase-todo buffer. Thanks a lot!

kyleam:
> For some reason, it even fixed "Not in git repository" error from
> the "C-c C-k" abort in git-rebase-todo buffer.

Hmm, I was only considering `magit-log-select-quit` and
`magit-log-select-pick`.  I don't see how those changes would affect
`with-editor-cancel`, so I suspect there's something else changing on
your end.  I'll apply 7ceca0a.  Please open a new issue if the
`with-editor-cancel` problem resurfaces.

PR #2677, opened 2016-06-01
Make magit-get setf'able

npostavs:
As discussed in https://github.com/magit/magit/pull/2670#issuecomment-222482218.

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -1660,9 +1660,9 @@ already set.  When nil, then always unset."
                        (magit-read-upstream-branch)))))
   (if upstream
       (-let (((remote . merge) (magit-split-branch-name upstream)))
-        (magit-call-git "config" (format "branch.%s.remote" branch) remote)
-        (magit-call-git "config" (format "branch.%s.merge"  branch)
-                        (concat "refs/heads/" merge)))
+        (setf (magit-get (format "branch.%s.remote" branch)) remote)
+        (setf (magit-get (format "branch.%s.merge"  branch))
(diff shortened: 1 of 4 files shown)


tarsius:
I like it but I don't think you can use `(setf (magit-get ...) ...)` in `magit-popup.el`. Even with 

```lisp
(declare-function magit-set 'magit-git)
(declare-function magit-get 'magit-git)
```

we still get

```
magit-popup.el:1277:1:Warning: the function ‘(setf magit-get)’ is not known to
    be defined.
```

lisp/magit.el
```diff
@@ -1660,9 +1660,9 @@ already set.  When nil, then always unset."
                        (magit-read-upstream-branch)))))
   (if upstream
       (-let (((remote . merge) (magit-split-branch-name upstream)))
-        (magit-call-git "config" (format "branch.%s.remote" branch) remote)
-        (magit-call-git "config" (format "branch.%s.merge"  branch)
-                        (concat "refs/heads/" merge)))
+        (setf (magit-get (format "branch.%s.remote" branch)) branch)
```
tarsius: Should be `... remote)`.

npostavs:
> I like it but I don't think you can use `(setf (magit-get ...) ...)` in `magit-popup.el`.

Oh, because the expander definition is used during macro expansion, so we would need `(require 'magit-git)`.

tarsius:
But `magit-popup` is a separate package, so we can't.

tarsius:
I've added a new commit which makes all the changes to `magit-popup-set-variable` in one go.

Issue #2673, opened 2016-05-31
magit rebase do not obey core.commentchar


franjozen:
I have changed my core.commentchar to $ as we are required to add a tracker number to the commit message, prepended with #.

When I rebase interactively and drop (c-k) a commit I expect that the process succeed with the message 

> Successfully rebased and updated refs/heads/master.

Instead I get this error message:

> Warning: the command isn't recognized in the following line:
>  - #pick cd9d1ee temp
> 
> You can fix this with 'git rebase --edit-todo'.
> Or you can abort the rebase with 'git rebase --abort'.

This error can be reproduced by changing `core.commentchar` to `$` and adding a commit you don't want to keep, and run `git rebase -i HEAD~2`.
After that you go to the line with the commit you want to drop and press `c-k`. This inserts a `#` as the first char on the line. Pressing `c-c c-c` causes the error message to show.

I have confirmed that changing `#` to `$` in function `git-rebase-kill-line` in file `git-rebase.el` and compiling it solves the problem. Unfortunately my lisp-skills stops there. This is not the right solution of course, the `core.commentchar` should be what is inserted.

Thank you for a great program!

kyleam:
Thanks for the report.  git-rebase.el hardcodes '#' as the comment
character throughout, so it should be reworked to respect
`core.commentchar`.

tarsius:
Agreed, but that could be quite some work.

@franjozen Could you avoid setting that variable for the time being?

PR #2680, opened 2016-06-04
Use `core.commentchar` in Git Rebase mode

kyleam:
Re: #2673


tarsius:
Please use camelCase (`core.commentChar`), like in the Git documentation and like we do in other places.

tarsius:
Thanks. (I've applied with the change I suggested above.)

PR #2681, opened 2016-06-06
magit-branch-description: Fix for detached head

marsam:
When `magit-show-refs-popup` is called in a detached HEAD throws an error because `(magit-get-current-branch) => nil`


tarsius:
Thanks!

kyleam:
@marsam Could you please provide a valid email address that we can add
as a .mailmap entry?

Issue #2684, opened 2016-06-08
magit-get fails to retrieve branch with upcase letters


ptrv:
Hi,

since commit c3d55c81d24fb6694b155bad74566e81beb30914 `magit-get` retrieves the information from cache if activated.
However it fails to get remote branch if the branch name consists upper-case letters because `mapconcat` is used with `'downcase`. Why `downcase`?

The observed behavior is that in `magit-status` buffer in the header there is nor `Merge:` line anymore, even with upstream branch configured. Also pushing works and upstream branch is shown.

Repro steps:
Create branch with uppercase letters and configure as `branch.*name*.merge`.

Thanks

npostavs:
> because `mapconcat` is used with `'downcase`. Why `downcase`?

Because `git config --list` outputs the keys downcased, except for branch names (and possibly some others?) apparently.

npostavs:
Okay, I got it, git lowercases the outer components of the key, i.e. only `prefix` and `suffix` in `prefix.Key.suffix`.

PR #2685, opened 2016-06-08
Don't downcase keys when getting values from cache

ptrv:
See #2684

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1439,20 +1439,20 @@ Return a list of two integers: (A>B B>A)."
 (defun magit-get (&rest keys)
   "Return the value of Git config entry specified by KEYS."
   (if magit--refresh-cache
-      (car (gethash (mapconcat 'downcase keys ".") (magit-config-slurp)))
+      (car (gethash (mapconcat 'identity keys ".") (magit-config-slurp)))
     (magit-git-str "config" (mapconcat 'identity keys "."))))
 
 (defun magit-get-all (&rest keys)
   "Return all values of the Git config entry specified by KEYS."
   (let ((magit-git-debug nil))
     (if magit--refresh-cache
-        (reverse (gethash (mapconcat 'downcase keys ".") (magit-config-slurp)))
-    (magit-git-lines "config" "--get-all" (mapconcat 'identity keys ".")))))
+        (reverse (gethash (mapconcat 'identity keys ".") (magit-config-slurp)))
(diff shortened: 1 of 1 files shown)


npostavs:
This doesn't work because *most* keys are downcased by `git config --list`.

ptrv:
The outer keys are anyways always lower case, aren't they?

npostavs:
No, typically the last key is camelCased.

ptrv:
Ok, what should we do about this PR. Do you have some suggestions for me to apply here or should I close the PR?

npostavs:
Well if you can change it to downcase the first and last keys as mentioned in https://github.com/magit/magit/issues/2684#issuecomment-224567876, then it could work.

ptrv:
What I don't get is why then `magit-git-str` uses `'identity` for mapconcat when it gets the config value?
Is only `git config --list` downcasing? Can you point me to the docs?

npostavs:
`man git-config(1)` says

> --get-regexp
>
>    Like --get-all, but interprets the name as a regular expression and writes out the key names. Regular expression matching is currently case-sensitive and done against a canonicalized version of the key in which section and variable names are lowercased, but subsection names are not.

It's not specified for `--list`, but it seems to work the same.

ptrv:
Thanks for the pointer to the docs. Yes, `git config --list` works the same. I could reproduce it myself.

I wonder if we add code to downcase only the key and the section part, would it make the performance benefit of using the hashtable void?

Can't we just use the `identity` function in mapconcat and require the user of the function to pass downcased key and section?

Or just use `magit-git-str` again?

What do you think?

npostavs:
> I wonder if we add code to downcase only the key and the section part, would it make the performance benefit of using the hashtable void?

As I mentioned in #2670, the performance benefits are only seen on Windows, but it's more than 100ms there. I doubt a couple of string transformations will be noticable either way.

> Can't we just use the `identity` function in mapconcat and require the user of the function to pass downcased key and section?

Well it would simplify things a bit, but on the other hand `magit-get` calls would be slightly harder to read. @tarsius what do you think?

> Or just use `magit-git-str` again?

Nah, I don't want to give up even a few hundred ms towards making magit usable on Windows.

npostavs:
Closing in favour of #2688.

ptrv:
@npostavs Thanks for your answer. I agree that the 100 ms benefit on windows is quite important to keep.

PR #2688, opened 2016-06-09
Fix cached config --list lookup

### LINKED ISSUE #2684: magit-get fails to retrieve branch with upcase letters

npostavs:
Fixes #2684.

```
git config --list outputs the key names with the first and last
components downcased.  Therefore we have to canonicalize the key value
before looking it up in the hash table.
```

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1426,34 +1426,43 @@ Return a list of two integers: (A>B B>A)."
 
 ;;; Variables
 
-(defun magit-config-slurp ()
-  (magit--with-refresh-cache 'config
-    (let ((configs (make-hash-table :test 'equal)))
-      (dolist (conf (magit-git-items "config" "--list" "-z"))
-        (let* ((nl-pos (cl-position ?\n conf))
-               (key (substring conf 0 nl-pos))
-               (val (if nl-pos (substring conf (1+ nl-pos)) "")))
-          (puthash key (cons val (gethash key configs)) configs)))
-      configs)))
+(defun magit-config-get-from-cached-list (key)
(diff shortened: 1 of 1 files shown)


ptrv:
I can confirm that this fixes #2684

tarsius:
I had to revert the config cache because it has an additional problem. The `default-directory` also has to be taken into account or the values from the super-repository end up being used inside submodules. This might turn out to be a bit complicated, and there are probably multiple ways of implementing this, so I have reverted for now.

npostavs:
I've squashed the config cache commits and pushed to `np/config-cache`, I'll take a look at the submodule config thing later.

npostavs:
> The default-directory also has to be taken into account or the values from the super-repository end up being used inside submodules.

Where does this come into play?

tarsius:
In `magit-insert-submodules` for example.

npostavs:
> In magit-insert-submodules for example.

That doesn't call `magit-get`, does it?

tarsius:
Maybe not. `magit-insert-modules-unpulled-from-pushremote` does. It uses `magit-get-push-branch` which uses `magit-get-push-remote`, which uses `magit-get`.

PR #2691, opened 2016-06-10
add magit-log-move-to-parent command

### LINKED ISSUE #2686: move to 1st/2nd parent in log buffer
It would be great if in the log buffer you could navigate to the 1st or 2nd parent directly, when on a merge commit for instance.

kyleam:
Closes #2686.

### DIFF
--- a/Documentation/RelNotes/2.7.1.txt
+++ b/Documentation/RelNotes/2.7.1.txt
@@ -33,6 +33,8 @@ Changes since v2.7.0
 * Added support for globbing in file arguments of diff and log
   commands.
 
+* Added new command `magit-log-move-to-parent'.
+
 Fixes since v2.7.0
 ------------------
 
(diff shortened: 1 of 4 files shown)


tarsius:
You  should use `magit-section-goto` to move to the section so that the `magit-section-movement-hook` is run. And if you do that anyway, it might be easier to find the appropriate section by looking in `(magit-section-siblings section 'next)` instead of doing a regexp search.

kyleam:
> it might be easier to find the appropriate section by looking in
> (magit-section-siblings section 'next)

Ooh, thanks.  Will update with that (my) tonight.

tarsius:
When there isn't an n-th parent, then this should probably raise a user-error.

kyleam:
> When there isn't an n-th parent, then this should probably raise a user-error.

The updated version now raises distinct user errors in two cases:
1) The commit has no n-th parent.
2) The commit has an n-th parent, but it isn't visible in the current log.

tarsius:
Nice

PR #2692, opened 2016-06-10
allow magit-repository-directories to specify depth per element

tarsius:
An element can now be a cons-cell whose cdr is the depth.  For elements
that are strings the value of `magit-repository-directories` is still
used.

---

Suddenly I can no longer export the manuals and have yet to figure out what changed. @kyleam could you please regenerate the manual and amend to this commit for me.

### DIFF
--- a/Documentation/RelNotes/2.7.1.txt
+++ b/Documentation/RelNotes/2.7.1.txt
@@ -18,6 +18,11 @@ Changes since v2.7.0
 * Added support for showing and copying bad commit identified by git
   bisect.
 
+* In addition to the directory, each member of the value of option
+  `magit-repository-directories' can now specify the depth to look for
+  repositories inside that directory, overriding the default depth
+  specified using the option `magit-repository-directories-depth'.
+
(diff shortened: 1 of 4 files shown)


npostavs:
> For elements
> that are strings the value of `magit-repository-directories` is still
> used.

You meant "`magit-repository-directories-depth` is still" used, right?

tarsius:
Yes oO

kyleam:
> @kyleam could you please regenerate the manual and amend to this
> commit for me.

Done.

Originally when I used ox-texinfo-plus, I believe the output only had
changes concerning the actual edits I had made to magit.org.  However,
for some time, I've been getting minor differences (e.g., number of
@'s and whitespace [1]).  So there is some sort of discrepancy between
how I have it set up and how you do (I mean, before yours was
completely broken), but I haven't taken the time to investigate.  I've
just been discarding the changes that were clearly not related to my
magit.org edits.

[1] As of a27a195, I'm getting one major difference, which is your
texi output doesn't have "\* Repository list::" under `@menu` and mine
does.

PR #2704, opened 2016-06-21
magit-ediff-stage: use same encoding for all texts

### LINKED ISSUE #2701: eol convention sometimes ignored
Hello,

magit-version => 20160614.1158

I'm working on a file that is encoded with utf-8-dos.  This one, to be precise: https://github.com/michael-heerdegen/hyper-german-trans/blob/master/language-de.cpp.

So, I'm on Debian using Unix style eol (end of line)  convention, and the file has the dos style eol convention.  Normally, this works transparently when working with Emacs, because it auto-detects the coding system.

But Magit has problems: if I hit E s, the ~HEAD~ buffer shows a ^M at each line's end (i.e. uses wrong coding), but ~{index} and the file itself don't.  As a result, it's nearly impossible to work with this (stage).

AFAIK the reason for the difference is that the index is opened as a file generated by git, so Emacs cares about encoding when opening it.

File version OTOH are got from git directly as a string (cat-file).  Dunno how much coding system aware Git is, anyway, Emacs receives the string with the "^M" chars in effect.

As a workaround, piping the string through `detect-coding-string` seems to help (fix the issue), but I'm not sure if it's the right thing to do.

There is a part number 2: magit-update-index must do the same thing in reverse: currently, we do
```
(with-temp-file index
            (insert-buffer-substring buffer))
```
which ignores eol convention as well, removing all ^M chars.  It seems that binding `coding-system-for-write` to `buffer-file-coding-system` (which is bound in the index buffer) fixes this.

But note I'm not sure if my fixes are correct as I don't know how many decodings and encodings are happening in the communication between Magit and Git.

More Magit commands may be concerned by this issue.

TIA, Michael.

npostavs:
```
When using ediff to stage, the working tree file is visited by Emacs in
the normal way, so Emacs applies normal encoding detection on it.  We
must apply the same encoding the output of `git cat-file' (used by
magit-find-file-{,index}-noselect) so that ediff can apply diffs
sensibly.

When writing back the changes back to git (via magit-update-index), we
have to use the same encoding to write as the one used to read, or else
we may ending changing the file recorded in git in a completely
unintended way.
```
fixes #2701.

We might want make encoding match across buffer for a few other ediff functions, `magit-ediff-show-working-tree` and `magit-ediff-show-stash` perhaps? To let the user choose different encodings, I think the best way is to set the `revert-buffer-function` so that <kbd>C-x C-m r</kbd> (`revert-buffer-with-coding-system`) can be used.

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -1195,68 +1195,59 @@ existing one."
 
 (defun magit-get-revision-buffer (rev file &optional create)
   (funcall (if create 'get-buffer-create 'get-buffer)
-           (format "%s.~%s~" file (subst-char-in-string ?/ ?_ rev))))
+           (format "%s.~%s~" file (if (equal rev "") "index"
+                                    (subst-char-in-string ?/ ?_ rev)))))
 
 (defun magit-get-revision-buffer-create (rev file)
   (magit-get-revision-buffer rev file t))
 
+(defun magit-find-file-noselect-1 (rev file hookvar &optional revert)
+  "Read FILE from REV into a buffer and return the buffer.
+FILE must be relative to the top directory of the repository.
+An empty REV stands for index."
+  (let ((topdir (magit-toplevel)))
+    (when (file-name-absolute-p file)
+      (setq file (file-relative-name file topdir)))
+    (with-current-buffer (magit-get-revision-buffer-create rev file)
+      (when (or (not magit-buffer-file-name)
+                (if (eq revert 'ask-revert)
+                    (y-or-n-p (format "%s already exists; revert it? "
+                                      (buffer-name))))
+                revert)
+        (let ((inhibit-read-only t))
+          (erase-buffer)
+          (magit-git-insert "cat-file" "-p" (concat rev ":" file)))
+        (setq magit-buffer-revision
+              (if (string= rev "") "{index}" (magit-rev-format "%H" rev))
+              magit-buffer-refname rev
+              magit-buffer-file-name (expand-file-name file topdir))
+        (setq default-directory (file-name-directory magit-buffer-file-name))
+        (let ((buffer-file-name magit-buffer-file-name)
+              (after-change-major-mode-hook
+               (remq 'global-diff-hl-mode-enable-in-buffers
+                     after-change-major-mode-hook)))
+          (normal-mode t))
+        (setq buffer-read-only t)
+        (set-buffer-modified-p nil)
+        (goto-char (point-min))
+        (run-hooks hookvar))
+      (current-buffer))))
+
 (defvar magit-find-file-hook nil)
+(add-hook 'magit-find-file-hook #'magit-blob-mode)
 
 (defun magit-find-file-noselect (rev file)
   "Read FILE from REV into a buffer and return the buffer.
 FILE must be relative to the top directory of the repository."
-  (let ((topdir (magit-toplevel)))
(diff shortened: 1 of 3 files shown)


kyleam:
This approach seems reasonable to me.

npostavs:
> To let the user choose different encodings, I think the best way is to set the `revert-buffer-function` so that <kbd>C-x C-m r</kbd> (`revert-buffer-with-coding-system`) can be used.

While implementing this I realized that `magit-find-{index,file}-noselect` could both use `cat-file`, and it would nicer to just let-bind `coding-system-for-read` rather than pass down explicit parameters. I forced-pushed the new approach, old version is still at `np/encoding-v1`.

lisp/magit.el
```diff
@@ -1190,76 +1190,66 @@ existing one."
...
+                                    (subst-char-in-string ?/ ?_ rev)))))
 
 (defun magit-get-revision-buffer-create (rev file)
   (magit-get-revision-buffer rev file t))
 
+(defun magit-find-file-noselect-1 (rev file hookvar &optional revert)
+  "Read FILE from REV into a buffer and return the buffer.
+FILE must be relative to the top directory of the repository.
+An empty REV stands for index."
+  (let ((default-directory (magit-toplevel)))
```
kyleam: I think always having the revison buffer use the top directory as its
`default-directory` makes sense, but this change in behavior should
probably be mentioned in the commit message.

lisp/magit.el
```diff
@@ -1190,76 +1190,66 @@ existing one."
...
+                     after-change-major-mode-hook)))
+          (normal-mode t))
+        (setq buffer-read-only t)
+        (set-buffer-modified-p nil)
+        (goto-char (point-min))
+        (run-hooks hookvar))
+      (current-buffer))))
+
 (defvar magit-find-file-hook nil)
+(add-hook 'magit-find-file-hook #'magit-blob-mode)
```
kyleam: Why not just add this to the `defvar`?

lisp/magit.el
```diff
@@ -1196,6 +1196,31 @@ existing one."
 (defun magit-get-revision-buffer-create (rev file)
   (magit-get-revision-buffer rev file t))
 
+(defun magit-revert-rev-file-buffer (_ignore-auto noconfirm)
+  (when (or noconfirm
```
kyleam: Extra `noconfirm`?

lisp/magit.el
npostavs: Hooks should initialized be to `nil` so that users customizing with `add-hook` before loading won't break it. But on the other hand, this is a `defvar` not a `defcustom` so maybe it doesn't matter?

lisp/magit.el
npostavs: Hmm, actually this happened by accident, I picked it up from `magit-find-index-noselect`. I think it makes more sense to use `default-directory` of the file.

lisp/magit.el
npostavs: oops

lisp/magit.el
kyleam: `default-directory` of the file, maybe, but it used to be whatever
subdirectory the call was from.

lisp/magit.el
kyleam: Right, thanks for reminding me.  I re-read a5798b2 (define hooks like
we are supposed too, 2015-07-27) as well.

PR #2706, opened 2016-06-24
magit-bisect-visualize-log: new command

### LINKED ISSUE #2689: Bisect Rest as branches (like log)
When bisecting, the `Rest` part is just a list of commits.
I think it would be must easier to understand if we could have the log (like `lb` or something) so we can see in the tree where the bad commits and good commits are.

npostavs:
Resolves #2689. Not sure if I chose the best the way to determine args to pass to `magit-log`.

### DIFF
--- a/lisp/magit-bisect.el
+++ b/lisp/magit-bisect.el
@@ -56,7 +56,7 @@
                         (?k "Skip"         magit-bisect-skip)
                         (?r "Reset"        magit-bisect-reset)
                         (?s "Run script"   magit-bisect-run)
-                        (?L "Show log"     magit-bisect-visualize-log))
+                        (?l "Show log"     magit-bisect-visualize-log))
   :sequence-predicate 'magit-bisect-in-progress-p)
 
 ;;;###autoload
--- a/lisp/magit-bisect.el
+++ b/lisp/magit-bisect.el
@@ -55,7 +55,8 @@
                         (?g "Good"         magit-bisect-good)
                         (?k "Skip"         magit-bisect-skip)
                         (?r "Reset"        magit-bisect-reset)
-                        (?s "Run script"   magit-bisect-run))
+                        (?s "Run script"   magit-bisect-run)
+                        (?L "Show log"     magit-bisect-visualize-log))
   :sequence-predicate 'magit-bisect-in-progress-p)
 
 ;;;###autoload
@@ -75,6 +76,17 @@ other actions from the bisect popup (\
 (defun magit-bisect-start-read-args ()
   (let  ((b (magit-read-branch-or-commit "Start bisect with bad revision")))
     (list b (magit-read-other-branch-or-commit "Good revision" b))))
+
+;;;###autoload
+(defun magit-bisect-visualize-log ()
+  "Show log of remaining commits to bisect."
+  (interactive)
+  (magit-mode-setup-internal
+   #'magit-log-mode
+   (list '("--bisect") (car (magit-log-arguments))
+         (magit-split-and-unsq-quote-string
+          (magit-file-line (magit-git-dir "BISECT_NAMES"))))
+   t))
 
 ;;;###autoload
 (defun magit-bisect-reset ()
--- a/lisp/magit-bisect.el
+++ b/lisp/magit-bisect.el
@@ -175,6 +175,15 @@ bisect run'."
         "bisect" "visualize" "git" "log"
         "--format=%h%d %s" "--decorate=full"))))
 
+(defun magit-insert-bisect-rest-graph ()
+  "While bisecting, insert section visualizing the bisect state."
+  (when (magit-bisect-in-progress-p)
+    (magit-insert-section (bisect-view)
+      (magit-insert-heading "Bisect Rest:")
+      (magit-git-wash (apply-partially 'magit-log-wash-log 'log)
+        "bisect" "visualize" "git" "log" "--graph"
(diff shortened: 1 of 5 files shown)


kyleam:
> Not sure if I chose the best the way to determine args to pass to magit-log

What alternatives are you considering?  Adding
`magit-bisect-visualize-log` as an action to the log popup and then
aborting if there isn't a bisect in progress?  If so, I prefer the way
you've done it.  And they can be adjusted later by the user from the
`magit-log-refresh-popup`.

lisp/magit-bisect.el
```diff
@@ -55,7 +55,8 @@
                         (?g "Good"         magit-bisect-good)
                         (?k "Skip"         magit-bisect-skip)
                         (?r "Reset"        magit-bisect-reset)
-                        (?s "Run script"   magit-bisect-run))
+                        (?s "Run script"   magit-bisect-run)
+                        (?L "Show log"     magit-bisect-visualize-log))
```
kyleam: Why not lowercase?

lisp/magit-bisect.el
```diff
@@ -77,6 +78,14 @@ other actions from the bisect popup (\
     (list b (magit-read-other-branch-or-commit "Good revision" b))))
 
 ;;;###autoload
+(defun magit-bisect-visualize-log ()
+  "Show log of remaining commits to bisect."
+  (interactive)
+  (magit-log '("--bisect") (default-value 'magit-log-arguments)
+             (split-string (magit-file-line (magit-git-dir "BISECT_NAMES"))
```
kyleam: 4bbc8d6eb9e0afa54d15422167d49e4eea2a7f41

lisp/magit-bisect.el
npostavs: Because I was already holding down shift for `B` I guess?. Didn't really think much about it.

npostavs:
> What alternatives are you considering?

I didn't have anything else in mind, this is just the first thing that seemed to work after I realized that just calling `(magit-log-argument)` wasn't right. If it looks reasonable to you, I guess that's a good sign. :)

kyleam:
I wonder if it'd be better to use a locked buffer so that it doesn't
interfere with the main log buffer. `magit-bisect-visualize-log`
could be changed to something like

    (magit-mode-setup-internal
           #'magit-log-mode
           <same args>
           t)

kyleam:
> split-string isn't enough either to quote cover names with internal single quotes though

Shoot, I should've stuck to just pointing out the issue :)

npostavs:
> I wonder if it'd be better to use a locked buffer so that it doesn't
> interfere with the main log buffer.

Yeah I think it makes sense. I also changed the arg grabbing to be like `magit-log-buffer-file`.

>> split-string isn't enough either to quote cover names with internal single quotes though
>
>Shoot, I should've stuck to just pointing out the issue :)

Added a function to handle it correctly. It feels a bit like overkill though...

npostavs:
> Added a function to handle it correctly. It feels a bit like overkill though...

The overkill was annoying me, I changed to shelling out to git instead.

kyleam:
nice

npostavs:
Ok, I've rebased this on master, only question mark left for me is the keybinding of `magit-bisect-visualize-log`, `L` or `l`?

kyleam:
As my comment suggests, I prefer `l`, but I don't feel strongly about it.

pvinis:
besides `l` and `L`, is there something else keeping that from merging? i think i would also prefer `l`, since `g`, `b` etc are lower case.

npostavs:
> besides `l` and `L`, is there something else keeping that from merging?

Now that I about it, documentation in the manual.

> i think i would also prefer `l`, since `g`, `b` etc are lower case.

Ok, I'll change to `l` then.

tarsius:
Nice job!

However I am not sure we really need all of this.

A simpler approach would be to simply teach `magit-insert-bisect-rest` to use "--graph" if it is a member of `magit-log-section-arguments`. Or do you expect that there would be situations where the use of a graph would make it harder to reason about the ongoing bisect? If so then it might be better to add a separate `magit-bisect-show-graph` option and a toggle command instead of depending on `magit-log-section-arguments`.

pvinis:
that sounds nice. a toggle in the bisect popup, or just getting the graph option from log, would be great. i think @npostavs just put it in a different function so we can try it out.

npostavs:
Ah, I had forgotten about `magit-log-section-arguments`. Yeah, probably makes sense to use that.

> Or do you expect that there would be situations where the use of a graph would make it harder to reason about the ongoing bisect?

Probably the opposite, i.e. the graph should make it easier to see what's going on. But maybe one might want to avoid adding the `--graph` to `magit-log-section-arguments` for performance reasons, while still wanting it during bisection?

tarsius:
I went with adding `magit-bisect-show-graph` and defaulting to `t`.

lisp/magit-utils.el
```diff
@@ -412,6 +412,11 @@ If optional NUM is specified only delete that subexpression."
   (delete-region (match-beginning (or num 0))
                  (match-end (or num 0))))
 
+(defun magit-split-and-unsq-quote-string (str)
```
tarsius: Might this still be useful elsewhere?

npostavs: In theory yes, though I don't know of any place that would currently use it.

PR #2709, opened 2016-06-24
support globbing in diff and log file arguments

kyleam:


tarsius:
It *might* be nicer if we remove "--literal-pathspecs" from `magit-git-global-arguments`'s global value and instead only added it when needed. But it would be a lot of work to find all those instances and then we would always have to think of this when adding something that deals with files. So I think your approach is more reasonable, at least for now.

PR #2710, opened 2016-06-29
Include subtrees in the dispatch pop-up

swflint:
Until I looked at the manual, I didn't realize that magit supports subtree commands.  Mostly because I don't check the manual, figuring the dispatch pop-up will tell me what I want to know.  This adds subtrees to the dispatcher.


kyleam:
Thanks

PR #2716, opened 2016-07-07
minor ui fixes

### LINKED ISSUE #2712: Harmless error when saving defaults in magit-branch-config-popup: setting-constant nil
I received the following error when attempting to save a change to the `branch.autoSetupRebase` setting in magit-branch-config-popup, which I access from the push command (P):

```
Debugger entered--Lisp error: (setting-constant nil)
  set-default(nil nil)
  customize-save-variable(nil nil)
  magit-popup-save-default-arguments(nil)
  call-interactively(magit-popup-save-default-arguments nil nil)
  command-execute(magit-popup-save-default-arguments)
```

However, the setting does in fact get saved, so I think it's complaining about another setting but it does still work.

### LINKED ISSUE #2715: Cannot stage file through magit-dispatch-popup
- Make changes to a file
- `C-x M-g`, which runs `magit-dispatch-popup`
- `s`, to stage the file

I then get this error message:
> Wrong type argument: arrayp, nil

Running `M-x magit-stage-file` does work.
Magit version is `20160704.1629`
Emacs version is `24.4.1`

npostavs:
- `magit-popup-{save,set}-default-arguments: don't use when useless` fixes #2712.
- `magit-stage: work outside of magit-mode buffer` fixes #2715 .


lisp/magit-apply.el
```diff
@@ -196,18 +196,22 @@ so causes the change to be applied to the index as well."
...
+        (`(unstaged    file) (magit-stage-1 "-u" (list (magit-section-value it))))
+        (`(unstaged   files) (magit-stage-1 "-u" (magit-region-values)))
+        (`(unstaged    list) (magit-stage-1 "-u"))
+        (`(staged        ,_) (user-error "Already staged"))
+        (`(committed     ,_) (user-error "Cannot stage committed changes"))
+        (`(undefined     ,_) (user-error "Cannot stage this change")))
+    (magit-with-toplevel
+      (magit-stage-1 "-u" (list (magit-read-file-choice
+                                 "Stage file" (nconc (magit-list-files)
+                                                     (magit-untracked-files))))))))
```
kyleam: You included untracked files as candidates, but they can't be added
with the `-u` flag, right?

kyleam: Is this worth adding when there's already `magit-stage-file`?

npostavs: > Is this worth adding when there's already magit-stage-file?

No, it is not.  Let's call that instead.

kyleam:
The `magit-stage` update looks fine to me.

PR #2720, opened 2016-07-18
magit-get: return all lines of config value

### LINKED ISSUE #2719: magit-insert-branch-description discards description body
As per the `git branch` [documentation](https://git-scm.com/docs/git-branch), branch descriptions may be multi-line, with a header and a body. [`magit-insert-branch-description`](https://github.com/magit/magit/blob/master/lisp/magit.el#L865) used to produce a section with a header and body to match, but this regressed in [a recent commit](https://github.com/magit/magit/commit/4394f5f5333c1f3ae408fc03d3a29d4ddb6c3ed4) such that the body is not present. (This is caused by `magit-get` and `magit-git-str`'s behavior of discarding all output lines except the first.)

To reproduce, create a multi-line description for any branch, check out that branch, and open the ref manager. You can see the section produced by `magit-insert-branch-description` at the top. It will not include the body when expanded.

Observed using Magit 20160713.1658, Git 2.9.0.windows.1, Emacs 24.5.1, windows-nt.

npostavs:
```
New function magit-get-line gets only first line, as previous behaviour
of magit-get.  This is used only for user.name and user.email.

Also make sure to include newlines in values returned by magit-get-all.
```
Fixes #2719.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1428,12 +1428,16 @@ Return a list of two integers: (A>B B>A)."
 
 (defun magit-get (&rest keys)
   "Return the value of Git config entry specified by KEYS."
-  (magit-git-str "config" (mapconcat 'identity keys ".")))
+  (car (magit-git-items "config" "-z" (mapconcat 'identity keys "."))))
+
+(defun magit-get-line (&rest keys)
+  "Return the first line of Git config entry specified by KEYS."
+  (car (split-string (apply #'magit-get keys) "\n")))
 
 (defun magit-get-all (&rest keys)
   "Return all values of the Git config entry specified by KEYS."
   (let ((magit-git-debug nil))
-    (magit-git-lines "config" "--get-all" (mapconcat 'identity keys "."))))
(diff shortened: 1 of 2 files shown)


tarsius:
> New function magit-get-line gets only first line, as previous behaviour of magit-get.  This is used only for user.name and user.email.

Why do these two variables need special treatment?

> Also make sure to include newlines in values returned by magit-get-all.

"... instead of incorrectly treating each line as a separate value". Right?

npostavs:
> Why do these two variables need special treatment?

Possibly only name needs it, but somehow, `user.name` seems to end up with an extra newline at the end. Perhaps this is a bug in an older version of git, because I'm not able to reproduce with a fresh `HOME` right now (it just tells me to run `git config`, didn't it use to autofill those variables?).

> "... instead of incorrectly treating each line as a separate value". Right?

Right.

tarsius:
I decided to forgo handling `user.name` differently. We only use that variable to display its value, so the worst that can happen is that to output is messed up a bit.

PR #2722, opened 2016-07-20
Use higher contrast for magit-blame with dark themes.

Wilfred:
Currently, magit uses black text for both light and dark themes. For
dark themes, this means there's very little contrast and commit messages
are hard to read.

Instead, use a constrasting color for dark themes.

Before:

screen shot 2016-07-20 at 14 44 45

After:

screen shot 2016-07-20 at 14 45 24


kyleam:
Thanks.

PR #2725, opened 2016-07-26
Update magit-popup.org

shoe:
Fix typo s/change/chance/


kyleam:
Thanks.  Applied your commit in 5834d0d with a slightly modified
commit message and adding the regenerated magit-popup.texi.

PR #2727, opened 2016-08-02
Fix "safe the buffer" typo in documentation

ivanbrennan:
Before:
> ...safe the buffer...

After:
> ...save the buffer...


kyleam:
Thanks.  In b924cdd I've merged your magit.org changes from here and
#2728 into one commit and regenerated magit.texi.

PR #2730, opened 2016-08-03
Recenter hunk after movement/refresh

### LINKED ISSUE #2683: recenter-top after staging a hunk
Oftentimes I'll go through different things to stage and I do so by pressing `C-l` to set the beginning of the hunk at the very top of the window so that I can see the most of the hunk that I can. However, then I press `s` to stage it and it correctly moves the cursor to the beginning of the next hunk, but the window seems to reposition so that the cursor is now at the bottom of the window, which is very disorienting and breaks my flow because I have to press `C-l` three times to get it back to the top.

Here's some before and after screenshots of what I mean in case that's not clear enough:

Before: position the cursor on the beginning of the hunk (most likely by pressing `n` to jump to the correct position), then press `C-l` twice so that the window repositions so that the top of the hunk is at the top of the window. Then press `s` to stage the hunk.

before

After: the cursor correctly jumped to the beginning of the next hunk, but the window repositioned so that it's now at the bottom of the window, forcing one to press `C-l` _three more times_ to get it back at the top of the window.

after

The reason I want to position the hunk at the top of the window is so that I can see the most of it. If this were to happen automatically then I could very quickly go over the hunks to ensure that they are the changes I want to include in a given commit.

If this isn't behavior you'd like to include by default, I would greatly appreciate it if it could be provided as an opt-in behavior, or even if there were some hook exposed or something so that I could call the appropriate function to do this.

It looks like `recenter-top-bottom`, the function that `C-l` is bound to, has this relevant section:

``` elisp
    (let ((this-scroll-margin
	   (min (max 0 scroll-margin)
		(truncate (/ (window-body-height) 4.0)))))
      (cond ((eq recenter-last-op 'middle)
	     (recenter))
	    ((eq recenter-last-op 'top)
	     (recenter this-scroll-margin))     ;; <---- code to set cursor top of window
	    ((eq recenter-last-op 'bottom)
	     (recenter (- -1 this-scroll-margin)))
	    ((integerp recenter-last-op)
	     (recenter recenter-last-op))
	    ((floatp recenter-last-op)
	     (recenter (round (* recenter-last-op (window-height))))))))))
```

tarsius:
Closes #2683.

This isn't enabled by default (yet?). Use to enable:

```lisp
(add-hook    'magit-section-movement-hook 'magit-hunk-maybe-recenter)
(remove-hook 'magit-section-movement-hook 'magit-hunk-set-window-start)
```

### DIFF
--- a/lisp/magit-section.el
+++ b/lisp/magit-section.el
@@ -58,7 +58,7 @@ That function in turn is used by all section movement commands."
   :group 'magit-section
   :type 'hook
   :options '(magit-section-set-window-start
-             magit-hunk-maybe-recenter
+             magit-diff-maybe-recenter
              magit-hunk-set-window-start
              magit-status-maybe-update-revision-buffer
              magit-status-maybe-update-blob-buffer
@@ -276,15 +276,15 @@ It the SECTION has a different type, then do nothing."
   (when (eq (magit-section-type section) 'hunk)
     (magit-section-set-window-start section)))
 
-(defun magit-hunk-maybe-recenter (section)
-  "Ensure the beginning of the `hunk' SECTION is visible and maybe the end too.
+(defun magit-diff-maybe-recenter (section)
+  "Ensure beginning and maybe end of the diff-related SECTION are visible.
 
 First make the beginning visible. Then, if the end isn't visible
 yet, attempt to scroll the window to make the end visible, while
 still keeping the beginning visible.
 
-It the SECTION has a different type, then do nothing."
-  (when (eq (magit-section-type section) 'hunk)
+It the SECTION's type isn't `hunk' of `file', then do nothing."
+  (when (memq (magit-section-type section) '(hunk file))
     (magit-section-set-window-start section)
     (unless (pos-visible-in-window-p (magit-section-end section))
       (recenter 0))))
@@ -925,8 +925,8 @@ invisible."
                         (siblings (magit-section-siblings section 'prev))
                         (previous (nth (length siblings) children)))
                    (if (not arg)
-                       (--when-let (or previous (car (last children)))
-                         (magit-section-goto it)
+                       (-when-let (sibling (or previous (car (last children))))
+                         (magit-section-goto sibling)
                          t)
                      (when previous
                        (magit-section-goto previous))
@@ -939,6 +939,16 @@ invisible."
                        (while (looking-at "^ ")    (forward-line -1))
                        (while (looking-at "^[-+]") (forward-line -1))
                        (forward-line))))))
+          (and (eq (magit-section-type section) 'file)
+               (-when-let (sibling
+                           (or (--when-let
+                                   (car (magit-section-siblings section 'next))
+                                 (magit-get-section (magit-section-ident it)))
+                               (--when-let
+                                   (car (magit-section-siblings section 'prev))
+                                 (magit-get-section (magit-section-ident it)))))
+                 (magit-section-goto sibling)
+                 t))
           (goto-char (--if-let (magit-section-goto-successor-1 section)
                          (if (eq (magit-section-type it) 'button)
                              (point-min)
--- a/lisp/magit-section.el
+++ b/lisp/magit-section.el
@@ -57,7 +57,9 @@ That function in turn is used by all section movement commands."
   :package-version '(magit . "2.3.0")
   :group 'magit-section
   :type 'hook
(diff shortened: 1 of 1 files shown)


lisp/magit-section.el
```diff
@@ -57,7 +57,9 @@ That function in turn is used by all section movement commands."
   :package-version '(magit . "2.3.0")
   :group 'magit-section
   :type 'hook
-  :options '(magit-hunk-set-window-start
+  :options '(magit-section-maybe-recenter
```
kyleam: `magit-section-maybe-recenter` is added here but not defined.

kyleam:
Hmm, does "movement/refresh" mean I should see recentering if I run
`magit-refresh`?  When I move to the beginning of a hunk with
<kbd>C-n</kbd> and press <kbd>g</kbd>, I don't see the same
recentering as if I go there with <kbd>n</kbd>.

kyleam:
Neat.  After playing with this a bit, I'm still not sure if I'd want
this behavior, but I think it's worth having as an option.

One thing I noticed is that it feels inconsistent to me that the hunk
line recenters but the the file line doesn't. That is, I think I'd
prefer this:

```diff
diff --git a/lisp/magit-section.el b/lisp/magit-section.el
index 1c847af..1a21b65 100644
--- a/lisp/magit-section.el
+++ b/lisp/magit-section.el
@@ -284,7 +284,7 @@ (defun magit-hunk-maybe-recenter (section)
 still keeping the beginning visible.

 It the SECTION has a different type, then do nothing."
-  (when (eq (magit-section-type section) 'hunk)
+  (when (memq (magit-section-type section) '(hunk file))
     (magit-section-set-window-start section)
     (unless (pos-visible-in-window-p (magit-section-end section))
       (recenter 0))))
```

tarsius:
> Hmm, does "movement/refresh" mean I should see recentering if I run `magit-refresh`?

No, not unless that causes the previously current hunk section to disappear and a sibling hunk section to become the current section. That normally only happens when a refresh happens because you staged/unstaged/discarded a hunk.

In that case the user did not use a movement command, but the refresh triggers a "move" to a section that still exists, which is why I think it is appropriate to use the movement hook.

> One thing I noticed is that it feels inconsistent to me that the hunk line recenters but the the file line doesn't.

Yeah, I've been thinking the same.

tarsius:
... but after I have tried it I don't like it.

When you move to a hunk, then you most likely want to see it in full. When moving to a file that does not necessarily have to be the case. File changes can be much longer than a hunk because they can contain more than one hunk, so the recentering is much more likely to happen and make much more previously visible text invisible, possibly including text that would help when trying to understand the changes in the first hunk of the file that has been recentered.

kyleam:
OK, fair enough.  I don't think I played around long enough with
either varaint to have a strong preference.

tarsius:
There is still a lot of fine-tuning required and somehow this doesn't work correctly when moving upward. I'll let this rest for a while.

tarsius:
`magit-hunk-set-window-start' already does more or less what the new `magit-hunk-maybe-recenter' does. So the latter is not needed. I have merged " 	magit-section-goto-successor: use magit-section-goto for sibling hunks".

The problem that moving upward causes the beginning of the section to be recentered instead of being "recentered at the top" still exists, but that isn't really related to the feature request this pr was supposed to address.

blaenk:
Hey I'm the one that requested a feature like this and just noticed you implemented this. Thanks a lot! I'm curious how to go about trying it out. It seems like the original post is no longer up to date. My understanding is that I should use add `'magit-section-goto-successor` to `magit-section-movement-hook` but I'm no longer sure if I should remove `magit-hunk-set-window-start` like you said to do in the original comment. It seems like you ultimately decided that it already does what `magit-hunk-maybe-recenter` would do so I get the impression that I shouldn't remove it after all.

This leaves me with simply:

``` elisp
(add-hook 'magit-section-movement-hook #'magit-section-goto-successor)
```

Though I tried that and I got a wrong arguments error. Looking at the source I found `magit-section-goto-successor-1` which does seem to be the correct arity and taking the correct parameter (section). Am I supposed to use that?

tarsius:
You shouldn't have to do anything. The `magit-section-movement-hook` is now called automatically when a refresh moves to a sibling of a hunk that has disappeared.

blaenk:
Ah okay thanks. I'll have a look out for that!

PR #2733, opened 2016-08-05
add feature: stage hunk in magit-diff buffer

gadmyth:
in magit-diff buffer, if the file of the diff hunk is managed by git, press "s" will cause "Cannot  stage committed changes".

### DIFF
--- a/lisp/magit-apply.el
+++ b/lisp/magit-apply.el
@@ -292,6 +292,7 @@ ignored) files.
       (`(unstaged      ,_) (user-error "Already unstaged"))
       (`(staged    region) (magit-apply-region it "--reverse" "--cached"))
       (`(staged      hunk) (magit-apply-hunk   it "--reverse" "--cached"))
+      (`(committed   hunk) (magit-apply-hunk   it "--reverse" "--cached"))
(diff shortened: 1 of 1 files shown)


npostavs:
Isn't this just duplicating the existing functionality a few lines down?
```elisp
      (`(committed     ,_) (if magit-unstage-committed
                               (magit-reverse-in-index)
                             (user-error "Cannot unstage committed changes")))
```

tarsius:
@gadmyth Yes, if that doesn't work, then set `magit-unstage-comitted` to `t` (though that is the default).

gadmyth:
@tarsius thanks, I find magit-unstage-committed worked in linux(fedora), but doesn't on Mac OSX (emacs 24.5, magit 2.4.1).  Emacs was downloaded from https://emacsformacosx.com/

npostavs:
> magit-unstage-committed worked in linux(fedora), but doesn't on Mac OSX (emacs 24.5, magit 2.4.1).

What do you mean by "doesn't work"? Any errors? What git version? Does it happen starting from `

> press "s" will cause "Cannot stage committed changes".

Note that you should presss "u", not "s".

PR #2739, opened 2016-08-08
magit-show-commit: use magit-buffer-refname

marsam:
This allows `magit-show-commit` work with buffers opened with `magit-diff-visit-file`


tarsius:
Thanks!

PR #2740, opened 2016-08-15
Add popup menu for rebasing onto origin/master.

Fuco1:
In many workflows it is very common to fetch the master branch and rebase local branches on top of that, so that when we do a pull request the history stays as clean as possible.  I routinely do this with all my branches (1 to 10 at a time), and so a quick popup option saves a lot of time.

I've implemented this for rebase only because that is my workflow, but I imagine this can be added for merge as well (I'm happy to add it).

I can also extract the code into functions, but for now I'm just throwing this out there for a review, or if such a feature is even wanted.

Feel free to comment, Cheers :)

### DIFF
--- a/lisp/magit-sequence.el
+++ b/lisp/magit-sequence.el
@@ -308,6 +308,15 @@ This discards all changes made since the sequence started."
               (?u (lambda ()
                     (--when-let (magit-get-upstream-branch) (concat it "\n")))
                   magit-rebase-onto-upstream)
+              (?M (lambda ()
+                    (-when-let* ((remote (magit-get-upstream-remote))
+                                 (master (concat remote "/master")))
+                      (when (magit-branch-p master) master)))
+                  (lambda (args)
+                    (interactive (list (magit-rebase-arguments)))
+                    (magit-rebase (concat
+                                   (magit-get-upstream-remote)
+                                   "/master") args)))
               (?e "elsewhere"               magit-rebase)
               "Rebase"
               (?i "interactively"      magit-rebase-interactive)


tarsius:
> In many workflows it is very common to fetch the master branch and rebase local branches on top of that

I completely agree, but I don't think the proposed change is not required for that.

A while ago I added support for a "push-remote" in addition to the upstream. Usually the upstream branch should be `master` or `origin/master` and the push-remote should be `my-fork`. Then you would push `my-feature` to `my-fork/my-feature` using <kbd>p p</kbd>, and rebase it onto the upstream `master` using <kbd>r u</kbd>.

In your case `master` probably isn't configured as the upstream of your feature branches, leading to the conclusion that `from/to/onto master` variants of existing commands would be desirable.

But doing it that way has disadvantages. If you *configure* the upstream, then you also get log sections listing the commits that haven't merged into the upstream and which haven't been pulled from there yet. Same for the "push-remote". Up to four such log sections can be shown (and they are all relevant), two for each of the "two related branch" - but only if you have configured both remotes.

These two related remotes/branches are for the most part setup automatically. See https://magit.vc/manual/magit/Branching.html. There are a few options to fine-tune this.

One issue remains, when someone else creates a new feature branch and you then branch your local copy of that, then the upstream is set to e.g. `origin/feature` instead of `origin/master`. But I have already started working on that in #2742.

Fuco1:
Thank you, this seems like it might work well for me.  Since I always work from "upstream head" (origin/master) those log sections will be empty most of the time.

I'll try to set this up and report back how I feel about the flow :)

Fuco1:
The last issue you mentioned is also not a problem since I don't do that much.

I completely misunderstood the new settings and I was using `P u` for the "merge" update.  Thanks again for the explanation, this is fully satisfactory for my needs.

Awesome as always :)

PR #2745, opened 2016-08-21
Support duet and other alternative commit commands

totherme:
Git duet is a tool to support a pair-programming workflow with git:

https://github.com/git-duet/git-duet

It provides a handy interface for setting one of the pair as the commit
author, and the other of the pair as the committer. In this way both
members of the pair are attributed on github, and so on. Using git-duet
requires committing not with the command `git commit`, but with the
command `git duet-commit`.

This patch introduces the custom variable `magit-commit-command`. By
default the commit command is `commit`, so everything works exactly like
we're used to. But if we do `(setq magit-commit-command "duet-commit")`,
then magit will commit using git duet instead.

### DIFF
--- a/lisp/magit-commit.el
+++ b/lisp/magit-commit.el
@@ -43,6 +43,11 @@
 
 ;;; Options
 
+(defcustom magit-commit-command "commit"
+  "The command used when committing."
+  :group 'magit-commands
+  :type '(string :tag "Command"))
+
 (defcustom magit-commit-arguments nil
   "The arguments used when committing."
   :group 'magit-commands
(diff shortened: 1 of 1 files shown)


tarsius:
I would rather not merge that. But you can accomplish the same using this advice:

```lisp
(defcustom git-duet-enabled nil
  ""
  :group 'git-duet
  :type 'boolean)

(advice-add 'magit-run-git-with-editor :around
            'magit-run-git-with-editor--git-duet)

(defun magit-run-git-with-editor--git-duet (fn &rest args)
  (if (and git-duet-mode
           (equal (car args) "commit"))
      (apply fn "duet-commit" (cdr args))
    (apply fn args)))
```

You might want to consider packaging this feature. The package should  probably provide some command which reads the two authors and enables `git-duet` for them. It would also be nice if the status buffer indicated that `git-duet` is in use, so that you notice if you forget to turn it off.

totherme:
Thanks, I'll use that in my configs.

PR #2747, opened 2016-08-22
set show-trailing-whitespace nil

zonuexe:
I have `(setq-default show-trailing-whitespace t)` in my `init.el`. (see [GNU Emacs Manual: Useless Whitespace](http://www.gnu.org/software/emacs/manual/html_node/emacs/Useless-Whitespace.html))

### DIFF
--- a/lisp/magit-popup.el
+++ b/lisp/magit-popup.el
@@ -999,6 +999,7 @@ restored."
   :mode 'magit-popup
   (setq truncate-lines t)
   (setq buffer-read-only t)
+  (setq show-trailing-whitespace nil)
   (setq-local scroll-margin 0)
   (setq-local magit-popup-show-common-commands magit-popup-show-common-commands)
   (hack-dir-local-variables-non-file-buffer))


tarsius:
> I have `(setq-default show-trailing-whitespace t)` in my `init.el`.

I don't think you are supposed to do that. The manual sais (emphasis mine):

> You can make trailing whitespace at the end of a line visible by setting the **buffer-local** variable show-trailing-whitespace to t.

So you might want to use:

```lisp
(add-hook 'prog-mode-hook (lambda () (setq show-trailing-whitespace t))
(add-hook 'text-mode-hook (lambda () (setq show-trailing-whitespace t))
```

instead of adding such a kludge to all non-file modes which are a bit careless about unnecessary whitespace.

npostavs:
I have this in my init.el:

```elisp
;; only show trailing whitespace in files
(define-and-add-hook find-file-hook
  (setq show-trailing-whitespace t))
```

tarsius:
I don't have `define-and-add-hook`. Seems like a good idea thought. What package provides that?

npostavs:
Oh, sorry, it's in my init files:
```elisp
(defmacro define-and-add-hook (add-to &rest body)
  "defun a hook function (named my-ADD-TO) and add it to hook
variable ADD-TO"
  (let ((hook-name (intern (concat "my-" (symbol-name add-to)) obarray)))
    `(progn
       (defun ,hook-name ()
         ,@body)
       (add-hook ',add-to #',hook-name))))
```

There was [a thread about adding a package, hook-helpers](https://lists.nongnu.org/archive/html/emacs-devel/2016-06/msg00153.html) (which has a similar macro), to GNU ELPA, though I don't see it in [the list](https://elpa.gnu.org/packages/) right now.

Issue #2396, opened 2015-11-16
error: Required feature `magit-autoloads' was not provided


targzeta:
Hi, 
I use to load magit with these code lines:
```lisp
(require 'magit-version)
(require 'magit-autoloads)
```
now I receive the error on subject. I guess that *magit-autoloads.el* leaks to **provide** magit-autoloads :)

Emanuele

npostavs:
`magit-version` is autoloaded now, so you don't need to `(require 'magit-version)`.

targzeta:
Oh, thank you npostavs (I have removed the line)! Anyway, the problem is another ;)

Thank you again,
Emanuele

npostavs:
Can you explain exactly how you are installing magit? Which of the methods described here: http://magit.vc/manual/magit/Installation.html

npostavs:
> Can you explain exactly how you are installing magit?

Uh, never mind. Looking at it now, I see `lisp/Makefile` indeed does not put a `(provide 'magit-autoloads)` in magit-autoloads.el, so I guess you could use `(load "magit-autoloads")` instead.

tarsius:
... and I think that's the correct behavior. `package.el` (or as far as I know anything else in Emacs) also does not insert a `provide` form into `*-autoload.el` files. And since these files don't provide any functionality, that makes sense.

PR #2754, opened 2016-08-26
Provide 'magit-autoloads feature

dmsh:


kyleam:
According to [this comment], the lack of a `provide` is intentional.

[this comment]: https://github.com/magit/magit/issues/2396#issuecomment-157439339

tarsius:
But we should still update the manual to include concrete steps on how one does "load the autoloads file" (by using `load`).

dmsh:
What does it have to do with package.el or "providing any functionality"?

The reason I'm adding this is to make it work nicely with use-package:
```
(use-package magit-autoloads)
(use-package magit
  :defer t
  :config
  (blah blah ....))
```
Even without use-package `(require 'magit-autoloads)` is still better than primitive `(load "/path/to/magit/lisp/magit-autoloads.el")`.

Here is another example: https://github.com/slime/slime/blob/master/slime-autoloads.el

kyleam:
I don't have a strong opinion about whether providing it is more
appropriate, but with regard to `use-package`,

```
(use-package magit
  :defer t
  :init
  (load "magit-autoloads.el")
  ...))
```

should work, no?

dmsh:
@kyleam 
Not really since it defeats the whole purpose of use-package i.e. to provide a way of building simplified, uniform, fast (lazy loading) and **robust** init.el. What happens if magit is nowhere to be found on your system and you're trying to `(load "magit-autoloads")`? The whole init process stops in the case of toplevel `(load ...)` and you get an annoying warning in the case of `(use-package ... :init (load ...))`. Yes, I know about `NOERROR` but that, again, defeats the simplicity and uniformity use-package is supposed to bring.

npostavs:
`(use-package "magit-autoloads")` should work fine.

tarsius:
Maybe this should be brought up on emacs-devel.

> `(use-package "magit-autoloads")` should work fine.

I didn't know about that, nice.

Personally I don't use `package.el` but I do use `use-package`. As package manager I use my own [`borg`](https://emacsair.me/2016/05/17/assimilate-emacs-packages-as-git-submodules/). Like `package.el` it does take care of loading the autoload files, so I don't have to do that manually in my `use-package` forms.

I think any package manager should do that. If you don't want to use any package manager, then you have to do it yourself and I recommend you do it like so:

```lisp
(dolist (dir (directory-files "~/.emacs.d/lib/" t "^[^.]"))
  (let (file (expand-file-name (concat drone "-autoloads.el") dir)))
  (when (file-exists-p file)
    (with-demoted-errors "Error loading autoloads: %s"
      (load file nil t))))
```

Also see [`borg-initialize`](https://gitlab.com/tarsius/borg/blob/master/borg.el#L123) for a more robust implementation, but the above should do.

dmsh:
@npostavs 
Fair enough, but my point about uniformity still stands, it's really annoying having to special-case anything especially when there is no good reason for it. The only potential downside I can think of is proliferation of `*-autoloads` features.

@tarsius 
You're suggesting every user should invent/copy-paste some kind of  horrible spaghetti to handle something completely trivial instead of fixing it once and for all. Did I get it right?

And again, what does it have to do with package managers? Most packages don't seem to provide their own autoloads files, so package.el takes care of that as best as it can to make the whole thing work. Magit is different in that it does provide its own autoload file and even talks about it in the documentation (albeit it talks about explicit `load`ing which is more of a disservice than anything but that's an orthogonal issue). Why not take it one tiny step further and make it actually useful/usable? The only apparent cost so far is that one's `features` list will be one item longer.

tarsius:
> Why not take it one tiny step further and make it actually useful/usable?

It is already useful. You have to put this in your init file:

```lisp
(use-package "magit-autoloads")
```

You would rather put that there instead:

```lisp
(use-package magit-autoloads)
```

Yes, this is a bit inconvenient because it diverges from what you usually do. But this very minor inconsistency does not justify characterizing what Magit currently does as not useful/usable. You can already accomplish what you want to do, literally by adding two additional characters to your init file, "unusable" means somethings else.

If you want autoload files to provide features, then you should discuss this on the emacs-devel mailing list. If `package.el` starts providing autoload features, then I will too.

> And again, what does it have to do with package managers?

As the author of an alternative package manager, I do take not of what the default package manager does, and try to not needlessly diverge from what it does. This also affects my decisions as the maintainer of packages.

PR #2763, opened 2016-09-07
remove "e" binding for magit-ediff-dwim

jeffersoncarpenter:
The reason I submit this pull request, with all due respect, is that occasionally I press the "e" key accidentally, which rearranges my buffers and opens several copies of the file I've selected.

The function can still be reached from the ediff popup.  We could have the "e" key open the popup ("E" does currently).

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -2626,7 +2626,6 @@ Currently this only adds the following key bindings.
              (?c "Committing"      magit-commit-popup)
              (?d "Diffing"         magit-diff-popup)
              (?D "Change diffs"    magit-diff-refresh-popup)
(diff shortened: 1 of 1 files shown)


tarsius:
> occasionally I press the "e" key accidentally

Many more users regularly press "e" intentionally. Removing this binding would inconvenient them. I suggest that you remove this binding locally instead.

npostavs:
Also consider using `winner-undo` (https://www.emacswiki.org/emacs/WinnerMode)

jeffersoncarpenter:
It's definitely correct behavior to require two key presses before destructively modifying buffers and opening an additional two copies of the file.

PR #2766, opened 2016-09-11
Update README.md

vermiculus:
Fix donations link


tarsius:
Thanks a lot! We can't have that be a dead link of course ;-)

PR #2771, opened 2016-09-15
git-rebase: disable save check when showing commits

### LINKED ISSUE #2770: git-rebase-show-commit asks to save git-rebase-todo
1) `emacs -Q`
2) `magit-status` a repo and do an interactive rebase with several commits
3) Press RET on any of the commits

Perhaps this is an intended safety measure in this case, but I find it more annoying than anything. However, I usually like the save prompt in other situations.

Could `git-rebase-todo` be exempt from ` magit-save-repository-buffers` at least in the case of `git-rebase-show-commit`?

`Magit 20160902.1451, Git 2.9.3, Emacs 25.1.3, gnu/linux`

kyleam:
```
The git-rebase-todo buffer is frequently in a modified state, but there
is no need to save it before showing a commit.  It will be taken care of
by with-editor-return when the user exits through with-editor-finish.

The unsaved buffer should almost always be the git-rebase-todo because
the initial 'git rebase' call will fail if there are unstaged changes.
The user could unwisely go and modify some other tracked buffers after
starting a rebase, and this could put things into a bad state [1], but
this is unrelated to whether the user happened to call a command to show
a commit.

Closes #2770.

[1] And, when the user has done this, I don't think there is really any
    good action to take to fix the situation.  It's bad whether or not
    the buffer is saved.  The user should just never start an
    interactive rebase and then change buffers of tracked files before
    triggering the rebase with with-editor-finish.
```

### DIFF
--- a/lisp/git-rebase.el
+++ b/lisp/git-rebase.el
@@ -128,7 +128,7 @@
     (define-key map (kbd "q")    'undefined)
     (define-key map [remap undo] 'git-rebase-undo)
     (define-key map (kbd "RET") 'git-rebase-show-commit)
-    (define-key map (kbd "SPC") 'magit-diff-show-or-scroll-up)
+    (define-key map (kbd "SPC") 'git-rebase-show-or-scroll-up)
     (define-key map (kbd "x")   'git-rebase-exec)
     (define-key map (kbd "c")   'git-rebase-pick)
     (define-key map (kbd "r")   'git-rebase-reword)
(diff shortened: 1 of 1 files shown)


lisp/git-rebase.el
```diff
@@ -351,15 +351,26 @@ Like `undo' but works in read-only buffers."
   (let ((inhibit-read-only t))
     (undo arg)))
 
+(defun git-rebase--show-commit (&optional display)
```
tarsius: Why `display`, regardless of the value of this argument the buffer is displayed? Just `scroll` seems better.

kyleam: Right, that's not a good name.  I was trying to make the distinction
that, with the optional argument, the buffer is *just* displayed,
rather than displayed and then switched to.  `scroll` is better given
that `display` doesn't actually make the distiction I was trying to
and because `scroll` is in the name of `magit-diff-show-or-scroll-up`.

kyleam: Updated to use `scroll` in 7da77ee.

PR #2775, opened 2016-09-17
Redefine popup when magit-define-popup re-eval'd

vermiculus:
`defvar` will not change the value of a defined symbol, so we should
undefine it when `magit-define-popup` is re-evaluated.

`setq` could also have been used, but this would not behave correctly
with respect to the variable's reported default value.

### DIFF
--- a/lisp/magit-popup.el
+++ b/lisp/magit-popup.el
@@ -546,6 +546,7 @@ usually specified in that order):
        (defun ,name (&optional arg) ,doc
          (interactive "P")
          (magit-invoke-popup ',name ,mode arg))
+       (makunbound ',name)
        (defvar ,name
          (list :variable ',opt ,@args))
        (magit-define-popup-keys-deferred ',name)


tarsius:
This is not necessary, you can just `C-M-x` (`eval-defun`). With your patch applied `eval-buffer` would remove all customization done using `magit-popup-define-*`, which is undesirable when you change many parts of the file, but not every popup in it, and want to load all those modifications.

vermiculus:
Have you tested that? I don't believe it works – that's why I opened the PR.

Try evaluating both of these forms with `eval-defun`:

```elisp
(magit-define-popup test "test popup" 'fake-group
  :actions '((?t "test" find-file)))

(magit-define-popup test "test popup" 'fake-group
  :actions '((?t "test two" find-file)))
```

tarsius:
> Try evaluating both of these forms with `eval-defun`:

Works for me. What Emacs version are you using?

vermiculus:
GNU Emacs 25.0.92.1 (x86_64-apple-darwin13.4.0, NS appkit-1265.21 Version 10.9.5 (Build 13F1603)) of 2016-03-02

npostavs:
Works for me as well in Emacs versions 24.4, 24.5, and 25.1. I have a memory of this not working once though.

tarsius:
> GNU Emacs 25.0.92.1

25.1 just came out, give it a try! ;-)

vermiculus:
Oh, I'll update to 25-1 – that might fix it. I'm working on a prerelease.

tarsius:
Hehe

vermiculus:
How about that – updating worked 👍

christophejunke:
I am using emacs 25.2.2, apparently `eval-defun` does *not* descend into `progn` forms:

    (progn (defvar foo 1))

Evaluating the following modified form with `eval-defun` does not update the binding:

    (progn (defvar foo 2))
    foo => 1

I am using a custom macro which expands into a `progn` and expands into a call to `magit-define-popup`, which explains why I could not fix my problem with `eval-defun`. For now, I am using `unintern` to unbind the global variable.

PR #2782, opened 2016-09-20
Replaced obsolete function string-as-multibyte

iladin:
Saw the following error in my 'macs and wanted to get rid of it.

magit-git.el:623:8:Warning: ‘string-as-multibyte’ is an obsolete function (as
    of 25.2); use ‘decode-coding-string’.

See the build error here
https://travis-ci.org/magit/magit#L207


iladin:
Hmm I should look into this more before submitting a PR.

PR #2787, opened 2016-09-20
Update docs for magit-stash-drop keybinding

jschwab:
The keybinding assigned to magit-stash-drop (in magit-stash.el) is z k,
not z d.


tarsius:
Thanks.

I have regenerated `magit.texi` and amended your commit with that.

PR #2789, opened 2016-09-21
Add INSIDE_EMACS to magit-git-environment

gdobbins:
This follows core Emacs behavior. For example M-x shell, term and epa
all export similar env vars. This will enable functionality such as the
pinentry package distributed with Emacs 25.1 being used for the
passphrase for git commit signing with gpg.


tarsius:
Thanks. Applied with some minor changes: ce93d3d5f9d86548b6eccf5b550d4b2e8c162bfe.

PR #2791, opened 2016-09-22
magit-diff-update-hunk-region: New method for v25

### LINKED ISSUE #2758: Selecting long lines in a diff hunk in Emacs 25 highlights extra lines
1. First make some changes to a file so its diff contains lines longer than the width of the Emacs window.
2. From `emacs -Q` load magit and enter the status buffer for the repo.
3. View the diff and go the end of the line (past the window edge).
4. `C-SPC` to select the line.

Notice how it's normal in Emacs 24.5 but in 25.1 the lines above and below the line are highlighted as well, pushing the selected line and the one below it down 1/2 lines respectively.

PS: I noticed how if you have consecutive lines that go past the window edge and you try to use the mouse to click and drag a region, the selection will jump right to the bottom of the lines instead of going line by line as would happen when using the keyboard.

`Magit 20160902.1451, Git 2.9.3, Emacs 25.1.1, gnu/linux`

npostavs:
Fixes #2758. Refer also to #2293. I squashed everything into one commit, and added some more comments, otherwise the code is unchanged.

```
The trick of using overlays containing a 1 pixel high newline to render
magit's region lines causes various glitches in the display
engine (overlays containing newlines tend to cause trouble).

When running in Emacs 25, instead render the top line as an :overline
face, and the bottom as an :underline on the first and last line of the
region, respectively.  While extending the lines past the text until the
edge of the screen does cause some complications (and messes with
temporary-goal-column movement), these are less serious than those
caused by the original method.
```

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -2089,15 +2089,18 @@ are highlighted."
 (defvar magit-diff-unmarked-lines-keep-foreground t)
 
 (defun magit-diff-update-hunk-region (section)
-  (when (and (eq (magit-diff-scope section t) 'region)
-             (not (and (if (version< emacs-version "25.1")
-                           (eq this-command 'mouse-drag-region)
-                         (or (eq last-command 'mouse-drag-region)
-                             ;; When another window was previously
-                             ;; selected then the last-command is
-                             ;; a byte-code function.
-                             (byte-code-function-p last-command)))
-                       (eq (region-end) (region-beginning)))))
+  (when (eq (magit-diff-scope section t) 'region)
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks a lot, the comments helped a lot and for the most part survived my subsequent changes :-)

tarsius:
I've deleted the older iterations locally and here. But just in case we end up having to look up something, here are the tips f1d007febcc4df2141a6d905a475fb2d3671af1c, 314f7162c70c2c06d278811778a183b6a6f44167, ef9198773953f79efec962f3f00191e607858689.

PR #2804, opened 2016-09-27
magit-diff-visit-file: take narrowing into account

maio:
In order to jump to correct location, we need to widen narrowed buffer
before moving cursor.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -1026,10 +1026,12 @@ which, as the name suggests always visits the actual file."
       (magit-display-file-buffer buffer)
       (with-current-buffer buffer
         (when line
-          (goto-char (point-min))
-          (forward-line (1- line))
-          (when col
-            (move-to-column col)))
+          (save-restriction
+            (widen)
+            (goto-char (point-min))
+            (forward-line (1- line))
+            (when col
+              (move-to-column col))))
         (when unmerged-p
           (smerge-start-session))
         (run-hooks 'magit-diff-visit-file-hook)))))


tarsius:
Thanks! This helps a lot when the location we jump to is within the narrowed part of the buffer.

We still have to decide what to do when jumping outside the narrowing. Probably we should just `widen` in that case. I have added this to my todo list.

----

It appears that you have created this commit using Github and that their commit message interface still has a bug that causes commit messages to not end with a newline character as they are supposed to. I have "reworded" the message to contain that newline (`c a` and then `C-c C-c` without actually making any change).

maio:
@tarsius good point. I could implement that if you wish.

tarsius:
What do you propose should happen? Just remove the narrowing? I suppose in most cases that would be the appropriate thing to do.

maio:
Yes I would start with that. Maybe later we could make it configurable (e.g. warn user, use indirect buffer, ...) if someone finds it obtrusive, but I hope that no one will.

BTW there may be same problem with code folding? (I never used it in Emacs so I don't know).

tarsius:
> Yes I would start with that.

Do that please.

> but I hope that no one will.

Yes. Me too ;-)

> BTW there may be same problem with code folding?

Yes, and I use that a lot and it's quite annoying. I'll tackle that eventually, but its a bit more work than what we are doing here. I would for example in `outline-minor-mode` or `org-mode` be able to say "fold everything, except those that parts that have been modified (also show the headers of the appropriate ancestors)". Jumping to such a modified place and hiding the rest, is just a special case of, the this bigger feature.

PR #2806, opened 2016-09-30
Fix magit-clone (and others?) while TRAMPing

timvisher:
`magit-clone` would fail when called in a tramp directory. This was
specifically because the filename as passed to the the git process was
being expanded to include the magic portion of the name instead of just
the local portion.

All we need to do is test if the file is remote and get its local part
if it is.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -611,7 +611,9 @@ Sorted from longest to shortest CYGWIN name."
             (cl-assoc filename magit-cygwin-mount-points
                       :test (lambda (f cyg) (string-prefix-p cyg f))))
       (concat win (substring filename (length cyg)))
-    filename))
+    (if (file-remote-p filename)
+        (tramp-file-name-localname (tramp-dissect-file-name filename))
+      filename)))
 
 (defun magit-convert-git-filename (filename)
   (-if-let ((cyg . win)


lisp/magit-git.el
```diff
@@ -611,7 +611,9 @@ Sorted from longest to shortest CYGWIN name."
             (cl-assoc filename magit-cygwin-mount-points
                       :test (lambda (f cyg) (string-prefix-p cyg f))))
       (concat win (substring filename (length cyg)))
-    filename))
+    (if (file-remote-p filename)
```
tarsius: This also returns `t` if another file-handler is used.

tarsius:
> Fix magit-clone (and others?)

That's too sloppy. `magit-init` benefits from this, everything that uses `magit-with-temp-file` breaks. Instead we need a separate function `strip-remote-part-if-it-is-there`, and `magit-init` and `magit-clone` should call that explicitly. You should also investigate whether Tramp already provides such a function, and whether it would make more sense to call that function or `magit-convert-git-filename` first.

lisp/magit-git.el
tarsius:
```diff
-    filename))
+    (or (file-remote-p filename 'localname) filename)))
```

would *probably* be better.

tarsius:
Okay, figured out why this confused me so much. You are editing `magit-EXPAND-git-file-name` to fix `magit-clone`, which uses `magit-CONVERT-git-filename`.

tarsius:
But the whole thing is a bit confusing. I think @npostavs and I have to look into this. The proposed fix is not appropriate.

PR #2813, opened 2016-10-06
Properly decode git utf-8 output on Windows

tarsius:
All git commands which print file names and similar encoding in utf-8 on
Windows.  The encoding of commit messages is controlled by
i18n.logOutputEncoding, but it's easier to set that to UTF-8 than try to
distinguish between commands that respect and those that don't.

magit-log-output-coding-system should probably be obsoleted/removed.

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -45,14 +45,17 @@
 
 ;;; Options
 
-(defcustom magit-log-output-coding-system 'utf-8
-  "Default coding system for receiving log output from Git.
+(defcustom magit-git-output-coding-system
+  (and (eq system-type 'windows-nt) 'utf-8)
+  "Coding system for receiving output from Git.
 
-Should be consistent with the Git config value `i18n.logOutputEncoding'."
-  :package-version '(magit . "2.8.0")
+If non-nil, the Git config value `i18n.logOutputEncoding' should
+be set via `magit-git-global-arguments' to value consistent with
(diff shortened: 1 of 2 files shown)


lisp/magit-process.el
```diff
@@ -700,12 +719,13 @@ Return the matched string suffixed with \": \", if needed."
             (t                             (concat prompt ": "))))))
 
 (defun magit--process-coding-system ()
-  (if magit-process-ensure-unix-line-ending
-      (cons (coding-system-change-eol-conversion
-             (car default-process-coding-system) 'unix)
-            (coding-system-change-eol-conversion
-             (cdr default-process-coding-system) 'unix))
-      default-process-coding-system))
+  (let ((fro (or magit-git-output-coding-system
```
tarsius: "fro"?

tarsius:
> magit-log-output-coding-system should probably be obsoleted/removed.

I think making it an alias would make the most sense.

tarsius:
I am all in favor of this change. Do you need anything else from me?

lisp/magit-process.el
npostavs: I was feeling poetic ;) https://en.wiktionary.org/wiki/to_and_fro
Maybe too obfuscated?

tarsius: Well, I wouldn't want to discourage that!

Issue #2812, opened 2016-10-05
feature: diff improvements in Magit Refs buffer


ryantm:
Hello.

# Background
I posted about this on Stack Overflow, because I thought maybe I was missing a feature of magit, but it seems like not.

http://stackoverflow.com/questions/39863277/in-magit-how-do-i-see-a-diff-what-a-branch-adds-to-my-current-branch/39870727#39870727

> In the Magit Refs buffer, you can press tab on a branch (the branch I want to merge in) to see what commits would be added if this branch were merged into the currently checked out branch, and you can press enter on the commits to see their changes. Is there a way to see all those changes unified together?
>I don't think the magit diff dwim feature is what I want, because it shows all the changes that are on the current branch but not on the branch I want to merge in. I usually don't care about seeing those changes, because I am mostly interested in what the branch I am merging will change, not everything else that changed.

@tarsius replied

>Yes, "diff dwim" doesn't really do what I mean here either. That should be improved, open a feature request please.
>Meanwhile you can use <kbd>d r</kbd> and then type the range (at least with completion). Or you could use <kbd>d d</kbd> and then flip the revisions (<kbd>D f</kbd>) as well as switch the range style (<kbd>D r</kbd>). Together this changes `feature..master` to `master...feature`.

# The feature

When in the Magit Refs buffer it would be nice to have a command (maybe <kbd>d d</kbd>) that opens a diff between the checked out branch and the branch at point that does not include the changes on the checked out branch since the branches diverged. 

A shell script that generates the diff I am talking about is:

```
#!/usr/bin/env sh

BRANCH_AT_POINT="branch/at/point/name"
CURRENT_BRANCH=`git branch | grep \* | cut -d ' ' -f2`
MERGE_BASE=`git merge-base $CURRENT_BRANCH $BRANCH_AT_POINT`

git diff $BRANCH_AT_POINT --not $MERGE_BASE
```

Please let me know if you need me to clarify anything, and thanks for considering this feature!
Ryan

tarsius:
I don't see any reason why the old behavior should be favored. In fact I have a hard time imagining a situation in which it would be useful ;-)

So I made the minimal change necessary to make it behave as you are suggesting. That's actually useful in most cases.

For the record, the old behavior was introduced in 03dd71683ee0afa2326f04c0f7e286b4bb5e1900. Prior to that `magit-diff-dwim` did just fall back to `magit-diff` in most cases when there is a branch at point, i.e. no dwimming.

[...]

PR #2814, opened 2016-10-06
magit-diff-dwmin: on branch, show diff for MERGE-BASE...BRANCH

ryantm:
This shows all the changes that would be made if BRANCH were to be
merged into CURRENT. It can also be very quick for small changes being
merged into a master branch that has diverged from a feature branch from
further development.

Re: #2812 

@tarsius Thanks for your addressing of #2812. It showed me how to do what I was hoping for. I've modified the behavior of the diff-dwim again in this PR to match what I was asking for, but this version may be strange depending on people's workflows. For my workflow, the change in this PR would be better for dwim, but I'm not sure if that's true for everyone. 

If this doesn't seem like the right behavior for DWIM, I'd like to propose making a new diff function that does this, since it is extremely useful for my workflow. Perhaps we could call it "Diff merge-base" on <kbd>d m</kbd>.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -727,7 +727,8 @@ If no DWIM context is found, nil is returned."
                       (if (magit-anything-modified-p)
                           current
                         (cons 'commit current)))
-                  (format "%s...%s" current atpoint))))
+                  (let ((base (magit-git-string "merge-base" current atpoint)))
+                    (format "%s...%s" base atpoint)))))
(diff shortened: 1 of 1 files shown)


ryantm:
It could just be called "Diff merge"

npostavs:
Is this the same as `magit-merge-preview`? <kbd>m p</kbd>

ryantm:
@npostavs Yes! Thanks. I guess I'll close it.

PR #2815, opened 2016-10-08
let-bind `shell-file-name' around with-editor calls

### LINKED ISSUE #2794: [w32] Interactive rebasing doesn't work with non-cmd shell-file-name
For a long time (possibly forever) I had a problem using interactive rebasing from magit.  Today I had updated magit and tried to dig into it.

Disclaimer: It was hard to find it, and I'm not sure that I got to the right place, and it might be something with my own environment.  Also, I found it in `with-editor` and the readme file there said to post issues here...

The place where I tracked it to is the place where `with-editor` does this:

    (shell-quote-argument with-editor-emacsclient-executable)

This gets a path that starts with "c:/" and since I'm using cygwin, that returns

    "c\\:/Program\\ Files/Emacs/bin/emacsclient.exe"

and that eventually leads to a failure when running the command which produces this error message (the line number in the error is probably a bit off since I added some printouts to debug that too):

    /mingw64/libexec/git-core/git-rebase--interactive: line 262: c;C:UserselbarzilAppDataLocalProgramsGitProgram FilesEmacsbinemacsclient.exe: command not found
    Could not execute editor

That's as much as I found.  To see if I can hack around it I replaced the shell quoting thing with:

      (push (concat with-editor--envvar "=\"" with-editor-emacsclient-executable
                    (and ...same...)
                    "\"")
            process-environment)

and that made it work.  Is this some known problem?  Any ideas what could the problem be or how I can debug it further?

---

`magit-version` says:

    Magit 2.8.0-28-g0ad041c, Git 2.10.0.windows.1, Emacs 24.5.1, windows-nt

`with-editor-debug` says:

    with-editor: c:/eli/.emacs.d/elpa/with-editor-20160408.201/with-editor.el
    emacs: c:/Program Files/Emacs/bin/emacs (24.5.1)
    system:
      system-type: windows-nt
      system-configuration: i686-pc-mingw32
      system-configuration-options: --prefix=/c/usr --host=i686-pc-mingw32
    server:
      server-running-p: t
      server-process: #<process server>
      server-use-tcp: t
      server-name: server
      server-socket-dir: nil
        WARNING: not an accessible directory
      server-auth-dir: ~/.emacs.d/server/
        server
    magit-emacsclient-executable:
     value:   c:/Program Files/Emacs/bin/emacsclient.exe (24.5)
     default: c:/Program Files/Emacs/bin/emacsclient.exe (24.5)
     funcall: c:/Program Files/Emacs/bin/emacsclient.exe (24.5)
    path:
      $PATH: "...ellided..."
      exec-path: (...ellided...)
      with-editor-emacsclient-path:
        c:/Program Files/Emacs/bin (t)
          c:/Program Files/Emacs/bin/emacsclient.exe (24.5)
          c:/Program Files/Emacs/bin/emacsclientw.exe (nil)
        C:/bin (t)
        ...ellided...
        C:/Users/elbarzil/AppData/Local/Programs/Git/cmd (t)
        C:/Program Files/nodejs/ (t)
        C:/windows/system32 (t)
        C:/windows (t)
        C:/windows/System32/Wbem (t)
        ...ellided...
        c:/Program Files/Emacs/libexec/emacs/24.5/i686-pc-mingw32 (t)
        c:/cygwin/bin (t)

npostavs:
Fixes #2794.

https://github.com/magit/magit/issues/2794#issuecomment-252183157:
> If it's around every call, why not just move it inside?

The need to do this binding is git specific, so the decision can't be made by `with-editor`.

> `magit-with-editor` is probably the best we can do.

Yes.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -203,6 +203,13 @@ change the upstream and many which create new branches."
                value)))
        ,@body)))
 
+(defmacro magit-with-editor (&rest body)
+  "Like `with-editor' but let-bind some more variables."
+  (declare (indent 0) (debug (body)))
+  `(let ((magit-process-popup-time -1)
+     (with-editor "GIT_EDITOR"
+       ,@body)))
+
 (defun magit-process-git-arguments (args)
   "Prepare ARGS for a function that invokes Git.
 
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -207,6 +207,18 @@ change the upstream and many which create new branches."
   "Like `with-editor' but let-bind some more variables."
   (declare (indent 0) (debug (body)))
   `(let ((magit-process-popup-time -1)
+         ;; The user may have customized `shell-file-name' to
+         ;; something which results in `w32-shell-dos-semantics' nil
+         ;; (which changes the quoting style used by
+         ;; `shell-quote-argument'), but Git for Windows expects shell
+         ;; quoting in the dos style.
+         (shell-file-name (if (and (eq system-type 'windows-nt)
+                                   ;; If we have Cygwin mount points
+                                   ;; the git flavor is cygwin, so dos
+                                   ;; shell quoting is probably wrong.
+                                   (not magit-cygwin-mount-points))
+                              "cmdproxy"
+                            shell-file-name)))
      (with-editor "GIT_EDITOR"
        ,@body)))
 
--- a/lisp/magit-commit.el
+++ b/lisp/magit-commit.el
@@ -249,12 +249,11 @@ depending on the value of option `magit-commit-squash-confirm'."
           (unless edit
             (push "--no-edit" args))
           (if rebase
-              (with-editor "GIT_EDITOR"
-                (let ((magit-process-popup-time -1))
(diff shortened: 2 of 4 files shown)


lisp/magit-git.el
```diff
@@ -207,6 +207,17 @@ change the upstream and many which create new branches."
...
+         ;; something which results in `w32-shell-dos-semantics' nil
+         ;; (which changes the quoting style used by
+         ;; `shell-quote-argument'), but Git for Windows expects shell
+         ;; quoting in the dos style.
+         (shell-file-name (if (and (eq system-type 'windows-nt)
+                                   ;; If we have Cygwin mount points
+                                   ;; the git flavor is cygwin, so dos
+                                   ;; shell quoting is probably wrong.
+                                   (not magit-cygwin-mount-points))
+                              "cmdproxy" shell-file-name)))
```
tarsius: Maybe it's a tick, but I never ever do that, and when I see it I get nervous. I only put THEN and ELSE on the same line, iff the COND is also on that same line.

Otherwise looks exactly as I have hopped it would ;-)

PR #2818, opened 2016-10-10
magit-diff-visit-file: improve narrowing support

maio:
Try to preserve narrowing when possible. When it isn't possible widen
buffer.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -1068,7 +1068,7 @@ the buffer in another window."
                      ((derived-mode-p 'magit-status-mode)
                       (magit-rev-name "HEAD"))))
           (unmerged-p (magit-anything-unmerged-p file))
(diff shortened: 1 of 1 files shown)


tarsius:
There have been some other changes to that function recently and I have decided that it needs some refactoring. I have started with that on `wip/diff-visit`, but that isn't finished. There's a bug related to "visit last rev that had the removed line at point" for staged and unstaged changes, and I want to see if I can cleanup things further.

There is however a commit  (17b88c2f65ef430d8bc96ba490eb1b4a00fee629) which accomplishes the same as what you are doing here, but I cannot apply it just yet. I should get to that soon so I won't merge this as that would make the general refactoring harder. Thanks anyway!

tarsius:
Actually I can apply my version now; that doesn't really make the upcoming refactoring harder. I've just done so.

PR #2822, opened 2016-10-11
Fix backwards obsolete variable alias

kini:
Just saw a warning when installing magit and went poking around briefly. I can't vouch for this change being correct ― take this PR as a linter suggestion :)

CC @npostavs who authored d176dfc5b8f00053bf02485a9d461639f9e38c05.

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -867,8 +867,8 @@ as argument."
 
 ;;; magit-process.el ends soon
 
-(define-obsolete-variable-alias 'magit-git-output-coding-system
-  'magit-log-output-coding-system "Magit 2.9.0")
+(define-obsolete-variable-alias 'magit-log-output-coding-system
+  'magit-git-output-coding-system "Magit 2.9.0")
 
 (provide 'magit-process)
 ;; Local Variables:


tarsius:
Thanks!

> CC @npostavs who authored d176dfc.

Actually it was me you wrote that faulty part, sorry.

Please don't use Github to create commits because  that makes it very easy to create broken commit messages. Unless you explicitly end the message with a newline the recorded message won't end with a newline as it is supposed to. I have fixed that locally for this commit and pushed the amended commit (it just adds the newline at the end of the message).

kini:
Noted ― I'll make sure not to do that in the future.

Just to clarify, what exactly did you change about the commit? https://github.com/magit/magit/commit/a36cb1976845ad7be6d4fc50fff8e22e607e723f.patch and https://github.com/kini/magit/commit/8f5d80956994bc86438f7e8a04d96f16eef59a41.patch have identical md5sums if you remove their first line...

kini:
Aha, it's visible in the actual commit objects

```
[2] fs@erdos ~/src/magit $ git cat-file -p 8f5d80956994bc86438f7e8a04d96f16eef59a41 | md5sum
40d77f6ca7b0610ab4199e4bba8db9c0  -
[2] fs@erdos ~/src/magit $ git cat-file -p a36cb1976845ad7be6d4fc50fff8e22e607e723f | md5sum
d6f36376407fd67a0745efc3420b36cc  -
```

Yikes, haha.

PR #2825, opened 2016-10-12
Only turn (n)linum-mode off in global-(n)linum-mode

fice-t:
There's no need to turn it off simply if `linum-mode` is bound (which is always, as `linum-mode` is autoloaded).


tarsius:
Thanks!

PR #2829, opened 2016-10-13
Add --full-history option in magit-log

Yuki-Inoue:
Tested on my local.

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -295,6 +295,7 @@ the upstream isn't ahead of the current branch) show."
                (?s "Show diffstats"          "--stat")
                (?h "Show header"             "++header")
                (?D "Simplify by decoration"  "--simplify-by-decoration")
+               (?F "Full History"            "--full-history")
                (?f "Follow renames when showing single-file log" "--follow"))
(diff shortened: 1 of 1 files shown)


tarsius:
Please see https://github.com/magit/magit/wiki/Additional-proposed-infix-arguments-and-suffix-commands.

fice-t:
@tarsius 
Should popup argument/command suggestions still be done in the issue tracker? Or should all suggestions just go straight to that page? 

If the latter, should there be a separate section on that page for "suggested arguments/commands that have not yet been rejected by the maintainers"?

tarsius:
@fice-t no they can be brought up here first.

@Yuki-Inoue I didn't mean to be rude, sorry for skipping a proper response. This is a useful argument in some cases but so are many of the other arguments that aren't featured in the popup. I will eventually go over the possible candidates again, and maybe this one will make the cut. But this likely also requires changes to the popup as a whole and even the implementation underlying it - there are just to many arguments to present them all unconditionally.

Yuki-Inoue:
@tarsius It's OK, I understand.  Just proposal that putting a label to categorise those PRs ToBeReconsidered might help..?

Issue #1581, opened 2014-10-28
Refining hunks can be slow because smerge-refine-subst is inefficient


yyr:
This issue was originally titled *args-out-of-range and other errors with latest next branch.* but we then went on mainly discussing the performance issue which was also mentioned.

It was then changed to *Highlighting of diffs sometimes takes a long time* and later to what you see above.

----

Press n on magit-status buffer I get this backtrace.

```
Debugger entered--Lisp error: (args-out-of-range [cl-struct-magit-section branch "master" #<marker (moves after insertion) at 1 in *magit: ~/.emacs.d/*> #<marker at 58 in *magit: ~/.emacs.d/*> #<marker at 144 in *magit: ~/.emacs.d/*> nil nil nil nil [cl-struct-magit-section status nil #<marker at 1 in *magit: ~/.emacs.d/*> nil nil nil nil nil nil nil (#1)] ([cl-struct-magit-section branch "origin/master" #<marker (moves after insertion) at 58 in *magit: ~/.emacs.d/*> nil #<marker at 122 in *magit: ~/.emacs.d/*> nil nil nil nil #1 nil] [cl-struct-magit-section tag "v0.4" #<marker (moves after insertion) at 122 in *magit: ~/.emacs.d/*> nil #<marker at 143 in *magit: ~/.emacs.d/*> nil nil nil nil #1 nil])] 12)
  magit-section-forward()
  call-interactively(magit-section-forward nil nil)
  command-execute(magit-section-forward)
```

There are other problems I see, may or may not be related. this is in my from *messages* buffer

magit-section-ident: Wrong type argument: arrayp, ([cl-struct-magit-section branch "master" #<marker (moves after insertion) at 1 in *magit: ~/.emacs.d/*> #<marker at 58 in *magit: ~/.emacs.d/*> #<marker at 144 in *magit: ~/.emacs.d/*> nil nil nil nil ...])

Another one also from *messages* buffer 

```
[...]

Issue #2666, opened 2016-05-28
Hide trailing carriage returns in commit messages?


npostavs:
There seems to be a github bug that inserts carriage returns in commit messages (except for 1st and last lines) when creating a commit through the web interface.

`git show` seems to hide the carriage returns, so magit should too, right? (the carriage returns can be seen by `git cat-file -p`).

tarsius:
Yes, Magit should do that. I wasn't aware Github messes this up, so while I have seen some such messages, I though this was a very rare occurance, which is why I haven't done anything about it yet.

PR #2834, opened 2016-10-23
Inhibit fsync when refining hunks

alphapapa:
I noticed that Emacs would sometimes thrash the disk very hard when I saved a file to disk.  By running `pkill -SIGUSR2 -f emacs` I saw that it was happening in magit's hunk-refining code.  I followed the calls and tracked it down to `smerge-refine-chopup-region`.  The problem seems to be that it writes many small files, and when Emacs is running in interactive mode, `write-region` calls `fsync`.  This results in many calls to `fsync`, which results in much disk thrashing and an unresponsive Emacs.

When `magit-diff-refine-hunk` is set to `all`, and the status buffer shows many changed hunks, this is a big problem, causing Emacs to thrash the disk for many seconds at a time, even when the status buffer is buried.

However, there seems to be a simple fix for this problem: Set `write-region-inhibit-fsync` in `magit-diff-update-hunk-refinement`, which causes the calls to `write-region` to happen without `fsync`.  This way those many small writes never touch the disk, the thrashing is eliminated, and the diff refinement seems instanteous.  It's effectively a one-line change that solves the whole problem.  :)

I searched the bug tracker and found #1581; I guess this basically fixes that issue too.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -2235,7 +2235,9 @@ are highlighted."
              (goto-char (magit-section-start section))
              ;; `diff-refine-hunk' does not handle combined diffs.
              (unless (looking-at "@@@")
-               (diff-refine-hunk))))
+               ;; Avoid fsyncing many small temp files
+               (let ((write-region-inhibit-fsync t))
+                 (diff-refine-hunk)))))
           ((or `(nil t ,_) `(t t nil))
            (setf (magit-section-refined section) nil)
            (remove-overlays (magit-section-start section)


tarsius:
Nice find. Thanks!

Please move the let-binding directly around `(diff-refine-hunk)` and move the comment to a separate line. Not sure you really have to mention `smerge-refine-chopup-region`, something like "This creates many temporary files, not fsyncing speeds that up.".

alphapapa:
Good point, I should have confined the let to that part.  I adjusted the comment as well.  Does this look okay?  Thanks.

tarsius:
@npostavs @kyleam It's now possible to push to the branch in the fork that was used to create a pull request. Unless the user unticked "Allow edits from maintainers". If we are allowed to push, then we see something like "Add more commits by pushing to the hunk-refinement-inhibit-fsync branch on alphapapa/magit" on the pull request.

In this case I have used this to rebase before merging. I think we should make use of this feature to avoid closing issues and then adding a comment like "I have cherry-picked your commit". When doing the latter, the metadata suggests the pull request was rejected.

tarsius:
@alphapapa Thanks ;-)

npostavs:
> I think we should make use of this feature to avoid closing issues and then adding a comment like "I have cherry-picked your commit". When doing the latter, the metadata suggests the pull request was rejected.

There's a newish option to "Squash and merge"[(1)][1] which has the same as effect as cherry-pick for a single-commit pull-request, but shows the pull request as "merged" rather than "closed". And also a "Rebase and merge", though I haven't tried that one out yet.

[1]: https://github.com/blog/2141-squash-your-commits

tarsius:
> it seems "Squash and merge" is disabled in the magit repo settings, on purpose?

Yeah, until Github stops defaulting to corrupted commit messages I want to look at commits locally.

npostavs:
> Yeah, until Github stops defaulting to corrupted commit messages

You mean the carriage returns (mentioned in #2666)?

alphapapa:
@monnier This is a bit of a blast from the past.  :)  What brought this to your attention?

> on my systems (and I expect that holds for most GNU/Linux systems) this is a filesystem using tmpfs

I don't know if that's true for most systems.  What distro are you using?  It's not the case on my Ubuntu 14.04 systems, nor on my Debian Stretch systems.  (I don't think I would want it to be, because I wouldn't want a process writing large temporary files to use up gigs of memory (e.g. burning a DVD and writing the ISO image), and I like being able to write files to `/tmp` and reboot and have them persist for a period of time (configurable somewhere, I forget where on Debian and Ubuntu)).

Now I think it is true that `/run` is using tmpfs on most Linux distros nowadays.  I remember reading about that transition on LWN a few years ago.

> More generally, given the expected semantics of /tmp it should ideally always be the case that fsync is pointless there.

I'm not familiar with that idea.  I don't think there is e.g. a `nofsync` mount option.  It may be the case on some distros that `/tmp` is cleared on every reboot, in which case that semantic would make sense, but I'm not aware of that being universal or standard.  It's certainly not in POSIX, is it?

> So, do you happen to know why in our case fsync generates so much disk traffic?

I'm not an expert on kernel semantics or syscalls, etc.  AFAIK, if `temporary-file-directory` is backed by a device filesystem (as opposed to something like tmpfs), `fsync` will cause the file (or should I say, the dirty pages in the cache) to be written to the device.

What do you think?

monnier:
According to the [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard), `/tmp` has the following semantics:

>  Temporary files (see also `/var/tmp`). Often not preserved between system reboots, and may be severely size restricted. 

So it  is a prime candidate for tmpfs (solves the kind of problems you've experienced once and forall), hence its name.  I assume its use is widespread, but I don't actually know it for a fact, indeed.
IIUC `/run` is quite different because it's not writable by normal users.

alphapapa:
> So it is a prime candidate for tmpfs (solves the kind of problems you've experienced once and forall)

Sorry, I don't understand what you mean about it solving the problems I mentioned, i.e. a place to write large temporary files to disk rather than RAM, and a place to write files that are eventually, automatically cleaned up, but not necessarily on the next reboot.

monnier:
By "problems you've experienced" I meant the performance problems in this bug-report.  As for your other needs, the FHS suggests `/var/tmp` as the place for temp files that aren't cleaned at the next reboot.

PR #2835, opened 2016-10-23
Update Bisect documentation keybindings

fice-t:


kyleam:
Thanks for catching those.

(By the way, assuming you made the texi changes in this commit
manually, you can instead just edit the Org file.  We can regenerate
the texi with ox-texinfo+.el and amend your commit.)

fice-t:
Ah, I just did `dired-do-find-regexp-and-replace` on both files since it was a tiny change. I'll keep that in mind if I edit the file again, thanks.

PR #2838, opened 2016-10-24
Displace obsolete decoding function

basil-conto:
##### Motivation
- As of `Emacs 25.2`, `string-as-multibyte` is obsolete.
##### Changelog
- Displace `string-as-multibyte` with the recommended `decode-coding-string` using the internal `utf-8-emacs` coding system.
##### Caveat
- I do not know whether `utf-8-emacs` is the correct coding system to use here; it was merely suggested as equivalent in the documentation for `string-as-multibyte`.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -56,6 +56,18 @@
   (list (format "INSIDE_EMACS=%s,magit" emacs-version))
   "Prepended to `process-environment' while running git.")
 
+(defcustom magit-git-output-coding-system
+  (and (eq system-type 'windows-nt) 'utf-8)
+  "Coding system for receiving output from Git.
+
+If non-nil, the Git config value `i18n.logOutputEncoding' should
+be set via `magit-git-global-arguments' to value consistent with
+this."
+  :package-version '(magit . "2.9.0")
+  :group 'magit-process
+  :type '(choice (coding-system :tag "Coding system to decode Git output")
+                 (const :tag "Use system default" nil)))
+
 (defcustom magit-git-executable
   ;; Git might be installed in a different location on a remote, so
   ;; it is better not to use the full path to the executable, except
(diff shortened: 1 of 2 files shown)


npostavs:
> I do not know whether utf-8-emacs is the correct coding system to use here; it was merely suggested as equivalent in the documentation for string-as-multibyte.

I think it should be `(or magit-git-output-coding-system (car default-process-coding-system))` (which will usually be utf8, so you won't see any difference, probably).

tarsius:
Does this also work with older Emacsen going back to 24.4?

basil-conto:
@npostavs Thank you for correcting that; I was not aware of `default-process-coding-system`. The problem now is that `magit-git-output-coding-system` is defined and used exclusively in `magit-process.el`. This package, in turn, `require`s `magit-git.el`, which is where `magit-decode-git-path` is defined. Would it make sense to define `magit-git-output-coding-system` in `magit-git.el` instead?

@tarsius I do not currently possess the ability to test this thoroughly across various platforms. Nevertheless, `decode-coding-string` has been around since at least `Emacs 23` and trivial examples like the following seem to work without complaint on both GNU/Linux and Windows NT:

```elisp
(decode-coding-string (read) (car default-process-coding-system) t)
```

tarsius:
I understand the sentiment, I usually also address such warning in the packages that I use. But we have to be sure this really does not cause any issues and since work on `25.2` just began, there is no hurry. Once someone looks into this more, it would be nice if s/he could document the steps required to verify that this is steps, so that someone else can easily retrace them.

basil-conto:
That is fair. I am interested in exploring these internals more but I expect slow progress on my part over the coming weeks.

I would keep this PR open as a reminder/invitation as long as that is in line with project policy, but I understand if it is preferred to demote/close it in the meantime.

tarsius:
We can leave this open.

tarsius:
I have cherry-picked this with small changes.

* 513aeb2f7607f70bbaeb6dfc3cd9ad33d2547bc9 drops a let-binding
* 9b218a39cc527f01859003117357bd0afc882909 doc-string was extended

basil-conto:
Thank you and sorry for the confusion and not getting back to this.

PR #2840, opened 2016-10-26
Document improved hunk-refining performance in release notes

alphapapa:
Documenting #2834 in release notes.


kyleam:
Thanks.  Please use \`' quotes rather than \`` for consistency.

I'm ok with the text as is, though I think I'd prefer something a
little less detailed, perhaps as short as

```
* Hunk refinement is now faster.  Users who previously set
  `magit-diff-refine-hunk' to nil due to performance issues may find
  that it now works well even when set to `all'.
```

alphapapa:
Oops, missed that quote.  I'm used to `smartparens` inserting the right pair for me automatically, but not in text files.  :)  I used the text you preferred.

PR #2842, opened 2016-10-28
Don’t raise errors when initializing ‘magit-need-cygwin-noglob’

phst:
When magit-process.el is loaded in an environment where Git can’t be
executed (e.g. in a container or when Git isn’t installed), loading
fails.  Work around this by catching ‘file-error’ and turning it into a
warning.


npostavs:
Thanks, I've also added a commit to avoid running this check except on Windows (that's the only place we actually care about the result anyway).

PR #2847, opened 2016-11-01
magit-log-wash-rev: set author to nil for stash lines

kyleam:
```
magit-log-stash-re matches the space after the hash as the author to fit
in with the other regexps used by magit-bind-match-strings in
magit-log-wash-rev.  This, however, results in a non-nil author argument
to magit-format-log-margin.  Because magit-stashes-mode derives from
magit-reflog-mode and uses its date-only margin specifications, a
non-nil author argument pushes the date outside of the margin width in
magit-stashes-mode.

Make magit-log-wash-rev set author to nil when washing stash lines so
that magit-format-log-margin receives the expected argument of nil for
author.

See #2846.
```

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -919,7 +919,10 @@ Do not add this to a hook variable."
       (magit-delete-line)
       (magit-insert-section section (commit hash)
         (pcase style
-          (`stash      (setf (magit-section-type section) 'stash))
+          (`stash      (setf (magit-section-type section) 'stash)
+                       ;; Replace blank space with nil for
+                       ;; `magit-format-log-margin'.
+                       (setq author nil))
           (`module     (setf (magit-section-type section) 'module-commit))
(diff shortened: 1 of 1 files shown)


lisp/magit-log.el
```diff
@@ -919,7 +919,10 @@ Do not add this to a hook variable."
       (magit-delete-line)
       (magit-insert-section section (commit hash)
         (pcase style
-          (`stash      (setf (magit-section-type section) 'stash))
+          (`stash      (setf (magit-section-type section) 'stash)
+                       ;; Replace blank space with nil for
+                       ;; `magit-format-log-margin'.
+                       (setq author nil))
```
zenspider: backquote? doesn't make sense in this context, does it? (I know you're just following the pattern already laid out here... but still)

npostavs: It does make sense, `pcase` in Emacs 24 requires backquote here (in Emacs 25, you can use straight quote instead)

PR #2853, opened 2016-11-09
Use nbutlast instead of butlast

fice-t:
It seems safe to use `nbutlast` in these instances.

### DIFF
--- a/lisp/magit-blame.el
+++ b/lisp/magit-blame.el
@@ -313,7 +313,7 @@ This is intended for debugging purposes.")
         (unless (string-match-p "\n\\'" string)
           (process-put process 'chunk chunk)
           (process-put process 'partial-line (car (last lines)))
-          (setq lines (butlast lines)))
+          (setq lines (nbutlast lines)))
         (dolist (line lines)
           (cond
            ((equal line ""))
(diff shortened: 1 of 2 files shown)


npostavs:
I agree they're safe now, but is it worth switching? It could lead to bugs when changing this code in the future.

fice-t:
I think it's worth it in `magit-blame-process-filter` as it handles lists of 100+ a good few times each `magit-blame`.

I just grepped for `butlast` and changed the other two invocations of it, so if there's worry over it then I don't think those really matter.

On the other hand, based on a quick check it doesn't seem like magit shies away from destructive functions all that much.

kyleam:
@fice-t Do you see a noticeable speedup?  Have you tried any timing?
I wouldn't expect this to make a noticeable difference because I'd
guess that the process call swamps other factors.

fice-t:
@kyleam It seems any small effect I noticed was placebo. On larger files changing it to `nbutlast` manages to shave off about 200KB (<1%) of memory usage, but that's indeed not much compared to the rest of `magit-blame-process-filter`. I don't see a reason to avoid `nbutlast` here, but it looks like it doesn't really matter either way.

> I'd guess that the process call swamps other factors.

It looks that way in terms of CPU. `magit-blame-make-overlay` accounts for 90% of memory usage, which is unsurprising. `generate-new-buffer` (from `with-temp-buffer`) takes up a surprising 20% though.

tarsius:
So what do we do about this? `butlast` sounds simpler than `nbutlast`, no "what's `n`?".

npostavs:
@tarsius I think we've spent enough time discussing this minor thing, it's best if you just make an executive decision now, whatever feels right.

tarsius:
Oki

PR #2854, opened 2016-11-09
display 'k' rather than 'C-k' for rebase-kill-line

marcowahl:
reverse the lines defining key-bindings for git-rebase-kill-line.

this is a pragmatic change to display 'k' rather than 'C-k' for git-rebase-kill-line in the help-text at interactive rebasings.

### DIFF
--- a/lisp/git-rebase.el
+++ b/lisp/git-rebase.el
@@ -160,6 +160,7 @@
 
 (put 'git-rebase-reword :advertised-binding "r")
 (put 'git-rebase-move-line-up :advertised-binding (kbd "M-p"))
+(put 'git-rebase-kill-line :advertised-binding "k")
 
 (easy-menu-define git-rebase-mode-menu git-rebase-mode-map
   "Git-Rebase mode menu"


tarsius:
Please use `(put 'git-rebase-kill-line :advertised-binding "k")` for this.

marcowahl:
Okay, I can do that.  Thanks for the advice.

Actually I thought that the user might be a (very) little happier if he sees 'k' instead of 'C-k'.

tarsius:
> Actually I thought that the user might be a (very) little happier if he sees 'k' instead of 'C-k'.

That's the purpose of the change I suggested too. Doesn't it work?

marcowahl:
Yes your change is fine. I evaluated the code and the k appeared in the rebase menu.

My point is that _every_ user should see the information about 'k' by default.  I.e. 'k' should be the advertised binding of git-rebase-kill-line by default.

tarsius:
Aha. I meant "please edit your commit to use this approach instead", not "please do this locally" ;-)

marcowahl:
Ahh! Looks like the communication finally worked. ;)

Please consider the brand new edit. (Which in summary results solely in the addition of the line you suggested.)

Bye!

kyleam:
Thanks.

PR #2860, opened 2016-11-19
note SPC at rebase. distinguish SPC, RET

marcowahl:
Hi guys!

This is another detail suggestion concerning the documentation of keys at rebase.

Recent: SPC is missing.

New:
1. SPC occurs in the "Commands" list.
2. SPC gets the text of (the recent) RET.
3. RET gets the hint that point moves away.

Best regards!

    Marco

### DIFF
--- a/lisp/git-rebase.el
+++ b/lisp/git-rebase.el
@@ -56,7 +56,9 @@
 ;;   x        Add a script to be run with the commit at point
 ;;            being checked out.
 ;;
-;;   RET      Show the commit at point in another buffer.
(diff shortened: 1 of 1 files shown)


tarsius:
> RET gets the hint that point moves away.

Saying that "[the buffer's] window is selected" seems more precise (since every buffer has a point). I'll squash that in a bit and then merge the result if you have no objections.

marcowahl:
No objections. And I agree that speaking of selecting the window is more precise.

PR #2864, opened 2016-11-20
"make clean" should do "rm *.elc"

YorkZ:
`make clean` should do `rm *.elc` instead of `rm $(ELCS)` to ensure removing all
the .elc files, including those obsolete residue ones. Obsolete residue .elc
files are usually not included in `$(ELCS)` because `$(ELCS)` is computed from
`$(ELS)` by the following substitution rule:

  `ELCS = $(ELS:.el=.elc)`

while the obsolete source files are not in `$(ELS)`.

A real life example is the residue file `lisp/with-editor.elc`. Apparently,
`with-editor.el` used to be part of Magit, but was moved out into a dependency
package. The old "make clean" wasn't able to delete the file `with-editor.elc`
for the reason explained above. This caused the `make` command always used the residue file `with-editor.elc` instead of the new one in the `with-editor`
package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,7 +105,7 @@ emacs-Q: clean-lisp
 
 clean: clean-lisp clean-docs clean-archives
 	@printf "Cleaning...\n"
-	@$(RM) $(ELCS) $(ELGS) # temporary cleanup kludge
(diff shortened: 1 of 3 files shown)


tarsius:
Thanks!

PR #2865, opened 2016-11-20
Raise not just window but frame also

s9gf4ult:
Usefull for multiframe configrations, overriding `magit-display-buffer-function` does not work as expected.


tarsius:
This seems safe, but I'll leave this unmerged for a while in case someone knows why it is not.

tarsius:
I've merged that now. Thanks!

PR #2867, opened 2016-11-21
Sort tags according to semver

Fuco1:
I think sorting the tags as versions beats the lexicographic default sorting.  Git can by default handle things like `-RC` or `-alpha` tags even.

The only problem is that on git version 2.5.0 and less `--sort` and `-n` does not work together.  I can't find what the minimal git version is, and I don't know the version when it started working (I updated straight to 2.10.2)

Alternatively we could add a switch to configure this.

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -1184,7 +1184,7 @@ reference, but it is not checked out."
 
 (defun magit-insert-tags ()
   "Insert sections showing all tags."
-  (-when-let (tags (magit-git-lines "tag" "-l" "-n"))
+  (-when-let (tags (magit-git-lines "tag" "-l" "-n" "--sort=v:refname"))
     (magit-insert-section (tags)
       (magit-insert-heading "Tags:")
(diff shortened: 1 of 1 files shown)


tarsius:
Ah nice, I forgot about that. I have just set this locally using `tag.sort`. Until we have figured out how to offer this to users in Magit and how to deal with older Git versions, I recommend you do that too.

> Alternatively we could add a switch [in the `magit-refs-popup`] to configure this.

It would probably be better to use the variable instead (similar to how this is done in the `magit-branch-config-popup`) and to also add `versionsort.prereleaseSuffix`.

kyleam:
> I can't find what the minimal git version is, and I don't know the
> version when it started working (I updated straight to 2.10.2)

I believe it was git's b7cc53e (tag.c: use 'ref-filter' APIs, 2015-09-11),
so version 2.7.0.

tarsius:
I think we should go with the variable variant and that we should delay doing so until I have gotten around to improve the popup implementation. So I am closing this.

PR #2868, opened 2016-11-22
Improve consistency and safety

tarsius:
This improves the consistency and safety of some commands mostly be changing the default behavior of certain commands. So some things, including key bindings, change and that's always a bit risky so I would appreciate it if you, @kyleam and @npostavs, could have another look at this.

I intend to create a release this week and would like to get this in first. So please review soonish. Thanks!


npostavs:
Looks good, I pushed a few minor fixes for the docs.

> I recently saw some Magit screecasts and noticed that even some users
> who enjoy Magit enough to create a screencast about it do not hide the
> section with common commands.

I have to confess that I don't bother hiding the common commands even though I enjoy Magit enough to maintain it :blush:

lisp/magit.el
```diff
@@ -2874,14 +2928,32 @@ Currently this only adds the following key bindings.
...
+                                                 :fun 'magit-section-toggle)
+                         (make-magit-popup-event :key 'return
+                                                 :fun 'magit-visit-thing))))
+  (let ((map (copy-keymap (current-local-map))))
+    (use-local-map map)
+    (define-key map "\t"     'magit-invoke-popup-action)
+    (define-key map "\r"     'magit-invoke-popup-action)
+    (define-key map [tab]    'magit-invoke-popup-action)
+    (define-key map [return] 'magit-invoke-popup-action)))
+
```
kyleam: What's the point of distinguishing between the keys and the ASCII
control characters here rather than letting the keys be automatically
translated as usual?

I think for TAB this is just redundant, but for RET, it leads to an
error when `magit-visit-thing` is called because `key-binding` returns
nil for `[return]`.  Removing `[return]` above (or, much uglier,
adding an explicit `[return]` for all the different maps with visit
thing functions) fixes this.

Documentation/RelNotes/2.9.0.txt
```diff
@@ -1,6 +1,75 @@
...
 ==================================
 
+Breaking changes since v2.8.0
+-----------------------------
+
+* For historic reasons "RET" behaved differently in `magit-refs-mode'
+  than anywhere else.  "RET" is one of Magit's most essential key
+  bindings and users quickly learn that regardless of what is actually
+  at point, it always shows more information about that thing in
+  another buffer.  It can be very confusing if references buffers
```
kyleam: I know "references buffers" is used elsewhere, but the double plural
noun is hard for me to parse.  In these situations, the first noun is
often made singular (i.e., "reference buffers"), but in this context
that leads to confusion with buffers that show a single reference.
Maybe it could be rephrased to talk in terms of a single buffer ("a
references buffer") or just `magit-refs-mode' buffers.  Dunno.

Documentation/RelNotes/2.9.0.txt
```diff
@@ -4,15 +4,42 @@ Magit v2.9.0 Release Notes (draft)
...
+
+  Furthermore the option `magit-visit-ref-create' was removed.  If you
+  previously had that set to `t', then you can achieve the same using:
+
+    (add-to-list 'magit-visit-ref-behavior 'create-branch)
+
+  However we recommend that you instead press "b c RET RET", just like
+  you always had to do in other buffers.
+
+* The option `magit-branch-popup-show-variables' now defaults to `nil',
```
kyleam: Should the changes below be squashed to
"magit-branch-popup-show-variables: default to nil" commit?

Documentation/RelNotes/2.9.0.txt
```diff
@@ -21,26 +48,27 @@ Breaking changes since v2.8.0
 * The option `magit-popup-show-common-commands' now defaults to `nil'.
 
   This was done to make popups less noisy but increases the odds that
```
kyleam: Squash below to "magit-popup-show-common-commands: default to nil"?

Documentation/RelNotes/2.9.0.txt
```diff
@@ -21,26 +48,27 @@ Breaking changes since v2.8.0
...
-  also mention the key that can be used to show those common commands
+  some users will never become aware of the common commands.
+
+  To counter that the usage instructions displayed in the echo area
+  now also mention the key that can be used to show these commands
   temporarily.
 
   You can easily revert this change using:
 
     (setq magit-popup-show-common-commands t)
```
kyleam: Squash below to "magit-popup-use-prefix-argument: default to default"?

Documentation/magit.org
```diff
@@ -2522,38 +2522,78 @@ should also change the others to keep things aligned.  The following
...
+  ~magit-visit-ref~.
+
+  "RET" is one of Magit's most essential keys and at least by default
+  it should behave consistently across all of Magit, especially
+  because users quickly learn that it does something very harmless; it
+  shows more information about the thing at point in another buffer.
+
+  However "RET" used to behave differently in ~magit-refs-mode~ buffers,
+  doing surprising things, some of which cannot really be described as
+  "visit this thing".  If you have grown accustomed to such
```
kyleam: I'd prefer shortening "If you have grown ... to get it back" to "If
you've grown accustomed this behavior, you can restore it by adding
one or more of the below symbols to the value of this option."

lisp/magit.el
```diff
@@ -201,16 +201,72 @@ in the current buffer using the command `magit-toggle-margin'."
...
+the current situation, then the symbols that follow do not affect
+the outcome.
+
+`focus-on-ref'
+
+  With a prefix argument update the buffer to show commit counts
+  and lists of cherry commits relative to the reference at point
+  instead of relative to the current buffer or HEAD.
+
+  Instead of adding this symbol, consider pressing \"y RET\".
```
kyleam: I'm confused.  Doesn't "y" just give the `magit-show-refs-popup`
(which defaults to `magit-show-refs-head`)? Should this be this be
"C-u y o RET" ?

kyleam:
I think these changes are a good idea.  I've added a few fixup commits
and comments.

fice-t:
I'm a bit confused about `magit-popup-use-prefix-argument`. If `default` invokes the "popup's default action", then why does `C-u y` show the popup instead of what I presume to be the default action (`magit-show-refs-head`)?

Speaking of which, why does one have to use `C-u y` to bring up the refs popup anyway? I think it'd be more consistent for `y` to show `"*magit-show-refs-popup*"` by default.

kyleam:
> If default invokes the "popup's default action", then why does C-u y
> show the popup instead of what I presume to be the default action
> (magit-show-refs-head)?

`magit-popup-use-prefix-argument` can be overridden on a per-popup
basis by the `magit-define-popup`'s `:use-prefix` keyword.

> Speaking of which, why does one have to use C-u y to bring up the
> refs popup anyway?

Convenience.  This works on the assumption that the default action,
without flag tweaking, is by far the most common action used.

fice-t:
Ah, I see. Thanks.

> Convenience. This works on the assumption that the default action,
without flag tweaking, is by far the most common action used.

True, but `y y` isn't much of a burden. The refs popup is the only one like this, right? I just find the inconsistency a bit odd.

lisp/magit.el
tarsius: I'll have to look into this again, but basically I've done this because with my configuration `C-i` aka `TAB` is bound to `previous-line` or similar across Emacs and ensuring that in this case turned out to be difficult after the fact.

Documentation/RelNotes/2.9.0.txt
tarsius: Went with "It can be very confusing if the references buffer breaks this convention."

Documentation/RelNotes/2.9.0.txt
tarsius: I don't think so, all of these entries come with quite a bit of option specific text.

Documentation/magit.org
tarsius: Done.

Documentation/RelNotes/2.9.0.txt
tarsius: Ah, I see what you mean now. Yes.

tarsius:
> The refs popup is the only one like this, right? I just find the inconsistency a bit odd.

Of course you have a strong argument here, considering the title of this pr is "Improve consistency and safety". That being said that really depends on what you compare it with. The refs buffer is like the status buffer in that it provides a grand overview. So it's consistent that both buffers are brought up without a popup (actually poping up).

Changing this is something that we can consider, but if we do, then it doesn't have to happen in this pr.

kyleam:
> The refs popup is the only one like this, right?

I believe it's currently the only one.  The blame popup used to as
well.

> I just find the inconsistency a bit odd.

I think that's fair.  I view `:use-prefix 'popup` as saying "it is OK
if users think of this soley as a non-popup command" because using
other actions or modifying the flags isn't expected to be done much.

Anyway, I'm not sure where I stand on it, but feel free to open this
as a new issue.

PR #2869, opened 2016-11-24
doc fix: correct git commit style convention name

basil-conto:
The docstring refers to `overlong-summary-line`, whereas the symbol used in the file is called `overlong-first-line`.

Although the latter is consistent with `non-empty-second-line` in its use of line numbers, I think "summary line" indeed makes more sense.


kyleam:
Thanks!

PR #2875, opened 2016-11-28
magit-emacs-Q-command: new command, for debugging

### LINKED ISSUE #2856: Provide a function for launching “barebones Emacs” with Magit
Right now, people who use Magit through Elpa is advised to test any
issues, bugs etc. by opening a “barebones” Emacs instance in order to
not let the specifics of ones setup interfere with reproduction of the
issue. These are the current instructions in `./ISSUES_TEMPLATE` on
this:

```
If you installed Magit from Melpa, then run commands similar to the ones below, but use tab completion to replace the various Ns with the correct versions:

  $ cd ~/.emacs.d/elpa/magit-N
  $ emacs -Q --debug-init --eval '(setq debug-on-error t)' -L ../dash-N -L ../git-commit-N -L ../magit-popup-N -L ../with-editor-N -L . -l magit
```

While this is already very helpful, it can be error-prone since you have
to replace all the `N`s with the specific suffix (version) that you
have in your own `.emacs.d`. While you’re in the middle of shuffling
strings around, trying to tab-complete on paths in the shell, etc. you
can forget to put some parts of it together. You might for example forget to
include one of the libraries.

What would be super-convenient would be if there was some Magit function
that searches all the relevant paths in your Elpa directory and tries to
complete these paths, and then launches the above command. In case that
is harder than I imagine that it is, or perhaps a bit ambigious (I don’t
know how Elpa works and if these paths can deviate from the general
pattern, if there can be several different versions, etc.), the function
can choose to be conservative and report that it wasn’t able to
automatically “deduce” the command in question, and that the programmer
should fill in the blank themselves. In either case, it seems that a
function like this would be super-helpful whenever Emacs/Magit is able
to deduce all the relevant Elpa paths (which I *imagine* is the common
case).

Alternatively, the function could just fill in the command and copy the
string to the kill ring, for the programmer to paste into their
terminal.

Context
=======

This came up on <https://gitter.im/magit/magit> today.

npostavs:
Refer: https://github.com/magit/magit/issues/2856#issuecomment-261773690

> It might be better to show the actual command, possibly prefixed with instructions.

I did that. Not sure if showing the command is a good idea, it's a bit long.

Fixes #2856.

### DIFF
--- a/Documentation/RelNotes/2.9.0.txt
+++ b/Documentation/RelNotes/2.9.0.txt
@@ -139,6 +139,10 @@ Changes since v2.8.0
   a regular file whose contents describe the location of the real git
   directory.
 
+* Added command `magit-emacs-Q-command' to make troubleshooting and
+  bug reporting easier for users not installing from the git
+  repository. #2856
+
 * The command `magit-branch-spinoff' learned to remove only a subset
   of commits from the previously current branch, instead of resetting
   it to its upstream, when the region selects commits reachable from
--- a/ISSUE_TEMPLATE
+++ b/ISSUE_TEMPLATE
@@ -43,10 +43,11 @@ If you run Magit from its Git repository, then you can do so using:
   $ cd /path/to/magit
   $ make emacs-Q
 
-If you installed Magit from Melpa, then run commands similar to the ones below, but use tab completion to replace the various Ns with the correct versions:
+Alternatively, run `M-x magit-emacs-Q-command RET` to save a shell command to the `kill-ring` and the system's clip-board, which you can then copy into a shell to run.
+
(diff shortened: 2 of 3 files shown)


tarsius:
>> It might be better to show the actual command, possibly prefixed with instructions.

> I did that. Not sure if showing the command is a good idea, it's a bit long.

On second thought I agree.

Please update all the relevant parts of the documentation.

npostavs:
> Please also update all the relevant parts of the documentation.

I updated `ISSUE_TEMPLATE` and added an entry in the release notes, is that what you had in mind?

tarsius:
I'v kept the old instructions in the documentation, just in case.

Issue #2881, opened 2016-12-02
magit-wip-after-apply-mode lighter typo


darkfeline:
Lighter variable has a typo.

https://github.com/magit/magit/blob/master/lisp/magit-wip.el#L128

PR #2882, opened 2016-12-02
Fix magit-wip-after-apply-mode lighter typo

darkfeline:
https://github.com/magit/magit/issues/2881


kyleam:
Thanks!  I edited the commit message to remove the link to #2881
because that issue doesn't add any new information.

Issue #2792, opened 2016-09-23
magit-log: ability to change commit timestamp format


razzmatazz:
In Magit log buffer commits (one per line) are shown with relative date. This is nice for recent commits, but makes little sense for older changes. In particular, this makes it annoyingly hard to relate commit dates to specific date, e.g. to find answer to question "which change might have broken the release on 3rd of April?".

Is it possible to somehow change the way commit dates are shown? I only found magit-duration-spec, but that's not exactly what I need.
- The text above is from http://emacs.stackexchange.com/questions/14483/magit-log-dates where it was suggested for the submitter to create a feature request for this. I could not find on magit issue tracker on github here so I went on to create it myself.

tarsius:
I've started work on this but it makes some ugly code even uglier and the doc-string of `magit-log-margin-spec` even more so. Too many implementation details are leaking into that variable, making it rather hard to customize it properly.

The `%ar` and `%cr` formats are not supported, for that the existing elisp implementation has to be used. `%ad` and `%cd` are also not supported, meaning that the date part of the margin has to be at least 25 characters wide. Together with the name part the total margin width should not be much less than 42 characters wide, which doesn't leave much room for the commit hash, graph and message. So I am wondering whether this would actually end up being used much.

I will have to investigate whether the use of `%ad` together with `--date=` allows using shorter date strings, which don't lose any information except maybe the timezone. The timezone cannot just be removed without the date/time being adjusted accordingly to UTC or local time. I don't know if Git can do that or whether we have to translate (which would be rather nasty).

I substituted an elisp implementation for Git's relative date formatting because we want something with a predictable and short width, e.g. "13 months". But Git would say "1 year and 1 month" ago in that case, which also is to wide. I am however wondering whether it would not have been better to translate Git's output into our preferred format. In that case we would have to check whether Git ever prints more than two time units.

The main reason Magit uses relative dates is that someone a long time contributed it that way. I have massively refactored that code by now, but I have kept the relative dates, because users were already used to that. Also supporting absolute dates isn't as trivial as it might seem, and I just never got around to it.

Anyway the implementation is at https://github.com/magit/magit/compare/jb/log-margin-format?expand=1, but it is unlikely that this or something similar will be merged anytime soon. I will work on this again eventually, but cannot say when.

tarsius:
[...]

PR #2885, opened 2016-12-04
Generalize margin and add commands to change its appearance

tarsius:
This is the second part. The first part happened directly on `master`, mostly in 186c3d6ccbca78386cb0faa69164b3396bb38ae6 and in response to #2792.

* Generalize margin support as described in 3881bbb2540e07c6f75ef3c675b38bca023f2649.
* Add commands to toggle margin details and cycle the margin style.

This changes the form of margin options again, but better do that now after five days and before the release than later when more than a handful of users have already adapted.

### DIFF
--- a/Documentation/RelNotes/2.9.0.txt
+++ b/Documentation/RelNotes/2.9.0.txt
@@ -85,9 +85,9 @@ Breaking changes since v2.8.0
   optionally in some other buffers that contain logs is now more
   flexible and easier to customize and to change on the fly.
 
-  The popup that "L" is bound to now features two commands that
-  for changing the appearance of the margin: `magit-toggle-margin'
-  and `magit-toggle-margin-details'.
+  The popup that "L" is bound to now features three commands that
+  for changing the appearance of the margin: `magit-toggle-margin',
+  `magit-toggle-margin-details' and `magit-cycle-margin-style'.
 
   The binding for `magit-toggle-margin' has changed from "L t" to the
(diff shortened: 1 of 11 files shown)


Documentation/RelNotes/2.9.0.txt
```diff
@@ -83,7 +83,17 @@ Breaking changes since v2.8.0
 
 * The margin that by default is displayed in pure log buffers and
   optionally in some other buffers that contain logs is now more
-  flexible and easier to customize.
+  flexible and easier to customize and to change on the fly.
+
+  The popup that "L" is bound to now features three commands that
```
kyleam: s/that//  (not adding as fixup to 093065b because it will result in
conflict with fdceddf)

lisp/magit-margin.el
```diff
@@ -0,0 +1,234 @@
...
+    (user-error "Magit margin isn't supported in this buffer"))
+  (setcar magit-buffer-margin (not (magit-buffer-margin-p)))
+  (magit-set-buffer-margin))
+
+(defun magit-cycle-margin-style ()
+  "Cycle style used for the Magit margin."
+  (interactive)
+  (unless (magit-margin-option)
+    (user-error "Magit margin isn't supported in this buffer"))
+  ;; This is only suitable for commit margins (there are not others).
```
kyleam: Maybe "This is only suitable for commit margins, which is currently
the only type of margin."?

PR #2888, opened 2016-12-06
Discard highlight-overlays

jockej:
Set `magit-section-highlight-overlays' to nil in addition to mapping
delete-overlay over it to remove all references to those overlays.
Otherwise `magit-section-highlight-overlays' grows unbounded, since `delete-overlay' only causes overlays to be unlinked form the buffer, they don't actually cease to exist. When we remove all references they should (hopefully) be garbage collected.


tarsius:
Thanks!

PR #2893, opened 2016-12-09
Use absolute path for magit-emacs-Q-command load file

fice-t:
This matches the absolute paths for library directories.

### DIFF
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -484,7 +484,8 @@ ACTION is a member of option `magit-slow-confirm'."
                               (file-name-directory (locate-library lib)))
                             '("magit" "magit-popup" "with-editor"
                               "git-commit" "dash"))))
-                "-l" "magit")
+                ;; Avoid Emacs bug#16406 by using full path.
+                "-l" ,(file-name-sans-extension (locate-library "magit")))
               " ")))
     (message "Uncustomized Magit command saved to kill-ring, %s"
              "please run it in a terminal.")


npostavs:
But setting up the library directories should make specifying the path loading `magit` redundant, no?

fice-t:
It seems that when using `-l", it tries to load from the current directory first. The reason I made this patch is when I execute the command from `~`, then I get:

```elisp
Debugger entered--Lisp error: (file-error "Cannot open load file" "Is a directory" "/home/alex/magit")
  load("/home/alex/magit" nil t)
  command-line-1(("--eval" "(setq debug-on-error t)" "-L" "/home/alex/.emacs.d/elpa/magit-20161208.1727/" "-L" "/home/alex/.emacs.d/elpa/magit-popup-20161207.613/" "-L" "/home/alex/.emacs.d/elpa/with-editor-20161201.925/" "-L" "/home/alex/.emacs.d/elpa/git-commit-20161206.1703/" "-L" "/home/alex/.emacs.d/elpa/dash-20161105.557/" "-l" "magit"))
  command-line()
  normal-top-level()
```

npostavs:
Ah, [Emacs bug#16406](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=16406). I fixed that recently, but that's only for Emacs 26, so we should indeed make `magit-emacs-Q-command` work for earlier versions, even if the user happens to execute just above the magit directory (it's supposed to be foolproof after all).

fice-t:
Good to know, thanks for fixing it upstream.

PR #2896, opened 2016-12-09
ISSUE_TEMPLATE: fill paragraphs to 70 columns

LemmingAvalanche:
Fill paragraphs instead of using one-line-per-paragraph.

The precedence for this change is other documentation files in the `./`
and `./Documentation` directories which use paragraph filling instead of
one-line-per-paragraph.  The precedence for using 70 columns is that
these files seem to stick to using 70 characters as a maximum value
pretty consistently.

### DIFF
--- a/ISSUE_TEMPLATE
+++ b/ISSUE_TEMPLATE
@@ -2,7 +2,9 @@
 DO NOT use this issue tracker for support requests
 =================================================================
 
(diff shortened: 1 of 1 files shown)


tarsius:
This file is one of the very few places where I use one-line-per-paragraph. And there is a reason for that - it is primarily being looked at [here](https://github.com/magit/magit/issues/new). I would like as much information to be displayed above the fold, and that requires longer lines. We could fill paragraphs at around 100 characters, but then that would look bad for users who have increased their browsers font size or view the file in an Emacs window with a reasonable width.

PR #2903, opened 2016-12-15
Fix typo in magit-diff-buffer-file-locked docstring

wbolster:


tarsius:
Thanks. I have applied your commit after fixing the commit message. It did not end with a newline as it is supposed to. Unfortunately creating a commit on github by default creates such a commit message. You have to explicitly add a newline at the end.

PR #2911, opened 2016-12-31
Improve module commands

tarsius:
@npostavs This is still very much work-in-progress and completely untested and undocumented, but gives a pretty good idea of I intend to do. This succeeds your earlier drafts; do you agree this is the way forward?

### DIFF
--- a/lisp/magit-submodule.el
+++ b/lisp/magit-submodule.el
@@ -102,14 +102,17 @@ an alist that supports the keys `:right-align' and `:pad-right'."
 (magit-define-popup magit-submodule-popup
   "Popup console for submodule commands."
   :man-page "git-submodule"
-  :actions  '((?a "Add"    magit-submodule-add)
-              (?b "Setup"  magit-submodule-setup)
-              (?i "Init"   magit-submodule-init)
-              (?u "Update" magit-submodule-update)
-              (?s "Sync"   magit-submodule-sync)
-              (?d "Deinit" magit-submodule-deinit)
-              (?f "Fetch"  magit-fetch-modules)
-              (?l "List"   magit-list-submodules)))
+  :actions
+  '((?a "Add"         magit-submodule-add)
+    (?r "Register"    magit-submodule-register)
+    (?p "Populate"    magit-submodule-populate)
+    (?u "Update"      magit-submodule-update)
+    (?s "Synchronize" magit-submodule-synchronize)
+    (?d "Unpopulate"  magit-submodule-unpopulate)
+    nil
+    (?l "List all modules"  magit-list-submodules)
+    (?f "Fetch all modules" magit-fetch-modules))
+  :max-action-columns 1)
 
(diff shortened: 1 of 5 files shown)


tarsius:
An earlier discussion can be found in #2674.

tarsius:
@npostavs I am going to delete `np/submodules` and `np/submodules-v2` in a few days, so if you want to hold on to something in there you should make sure you have those branches locally.

npostavs:
Hi, sorry I keep losing steam on this, I don't actually use submodules often enough to stay motivated I guess.

> This succeeds your earlier drafts; do you agree this is the way forward?

If I'm reading this right, the main change is move the submodule update method from a prompt to a popup switch. That looks like a good idea.

Putting all those actions in the submodule popup's top level looks a bit overwhelming to me though.

tarsius:
> Putting all those actions in the submodule popup's top level looks a bit overwhelming to me though.

Right. That's the main reason I have been keeping it back ;-)

tarsius:
I think this is a dead end. And unfortunately my ideas for alternative approaches are rather vague at the moment.

npostavs:
I'm in the process of removing the submodules from yasnippet, and it seems there is no way to do this from magit. The answers in https://stackoverflow.com/questions/1260748/how-do-i-remove-a-submodule make it seem complicated, though possibly simpler in more recent git versions.

fbergroth:
What about putting all the `magit-submodule-*-all` under another popup, say, `magit-submodule-foreach-popup`. The same lowercase keys can be used as their corresponding "one module" variant. Then binding a key, say, `F` (as in "Foreach module...") from `magit-submodule-popup` to `magit-submodule-foreach-popup`.

tarsius:
Thanks for the suggestion. No comment for now - but I intend to look into this in about a week.

lisp/magit-submodule.el
```diff
@@ -158,41 +181,67 @@ it is nil, then PATH also becomes the name."
...
 ;;;###autoload
-(defun magit-submodule-init ()
-  "Register all remaining modules."
-  (interactive)
+(defun magit-submodule-register (modules)
+  "Register MODULES."
+  ;; This command and the underlying "git submodule init" do NOT
+  ;; "initialize" modules.  They merely "register" modules in the
+  ;; super-projects $GIT_DIR/config file, the purpose of which is to
+  ;; allow users to change such values before actually initializing
```
npostavs: Wouldn't it be better to have this explanation in the docstring, rather than hiding it away in source code comments?

tarsius: I plan to come back to this eventually, but considering for how long I have been dragging this out I am pretty happy I finally merged something. I didn't want to further delay this by polishing documentation for who knows how long, so this is why I have put that in a comment for the time being. 😜

npostavs: Hah, fair enough. :)

PR #2912, opened 2016-12-31
Make Shift-selection less surprising

tarsius:
Normally `C-S-n`, sets the mark and then moves down one line, resulting in one line being selected by the region. The same thing happens in Magit (because we don't handle this specifically yet), but because Magit's "selection mechanism" is only *based* on the region, but actually spans a wider area, this keeps surprising users.

The selection begins at the beginning of the line on which the beginning of the region is and ends at the end of the line on which the end of the region is. The effect of this is that a single `C-S-n` selects *two* lines. The region is still the same, but the region is not visualized and so it looks like `C-S-n` works differently in Magit buffers.

For multi-section selections I think we should stick to that. Such a selection always contains at least two sections. Or we might want to start with a region that isn't a valid selection and therefore is visualized as such.

For hunk-internal regions I think we should change the behavior, so that it does something different and thereby looks like it does *not* do something different ;-) Here the first `C-S-n` should just set the mark but *not* move to the next line.

The current implementation on this branch just always does forgo the movement of the first `C-S-n`.

### DIFF
--- a/Documentation/RelNotes/2.11.1.txt
+++ b/Documentation/RelNotes/2.11.1.txt
@@ -4,6 +4,20 @@ Magit v2.11.1 Release Notes (unreleased)
 Changes since v2.11.0
 ---------------------
 
+* Added new commands `magit-previous-line' and `magit-next-line' as
+  substitutes for `previous-line' and `next-line'.  Magit's selection
+  mechanism is based on the region but selects an area that is larger
+  than the region.  This causes shift selection to select two lines on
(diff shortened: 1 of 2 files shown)


kyleam:
I see the purpose of special casing this in internal hunks, but agree
it shouldn't be used elsewhere.  It bugs me a bit that C-S-n would
behave differently than C-SPC C-n because I view them as equivalent,
but I suppose that's kind of the point.

PR #2914, opened 2016-12-31
In the manual, add dummy subsection, wrapping initial section content

tarsius:
It's nice to be able to collapse the text in section that comes before its subsections when working on the org source. But we don't want these dummy sections to appear in the exported files. `ox-extra` implements support for that, but unfortunately it appears to have a bug (maybe a regression when used with org v9), sections now contain the subsection menu twice.

I have contacted the author, @aecay, by email (because he isn't very active on github and might not see this otherwise).

### DIFF
--- a/Documentation/magit-popup.org
+++ b/Documentation/magit-popup.org
@@ -121,6 +121,7 @@ in packages that are not related to Magit.  But keep in mind that it
 will be deprecated eventually.
 
 * Usage
+** _ :ignore:
 
 Every popup buffers created with a prefix command contains a section
 named "Actions" listing the available suffix commands.  Most buffers
@@ -386,6 +387,7 @@ To emphasize the default action by making it bold use this:
 #+END_SRC
 
 * Defining Prefix and Suffix Commands
+** _ :ignore:
 
 If you write an extension for Magit then you should use this library
 now and later when ~transient~ is released port to that.
@@ -625,6 +627,8 @@ Except when redefining the former, you should not use these directly.
   arguments that don't match any of the regexps are returned, or ~:only~
   which doesn't change the behavior.
 
+* _ :ignore:
+
 #  LocalWords:  ARG DESC KEY's LocalWords Magit OPTVAL POPUP PREPEND
 #  LocalWords:  Popup SHORTNAME args desc magit manpage popup popup's
 #  LocalWords:  popups pre prepend
@@ -633,4 +637,5 @@ Except when redefining the former, you should not use these directly.
 # eval: (require 'magit-utils nil t)
 # eval: (require 'org-man     nil t)
 # eval: (require 'ox-texinfo+ nil t)
+# eval: (and (require 'ox-extra nil t) (ox-extras-activate '(ignore-headlines)))
 # End:
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -123,6 +123,7 @@ porcelain commands: ~add~, ~am~, ~bisect~, ~blame~, ~branch~, ~checkout~, ~cherr
 commands are implemented on top of Git plumbing commands.
 
 * Installation
+** _ :ignore:
 
 Magit can be installed using Emacs' package manager or manually from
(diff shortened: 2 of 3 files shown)


aecay:
When org parses a buffer into a lisp data structure, each `headline` contains (only) a `section`, which in turn contains the contents of that headline/section.  My preliminary analysis is that this issue arises because the `ox-extra` code leaves a stray `section` when suppressing a headline -- usually this causes no problems, but the texinfo exporter uses `section` as the trigger for menu insertion.

I'll be traveling for most of January, and I am also not very active in org development lately.  I will try to find time to fix this issue, but regretfully I cannot promise when that will be.

tarsius:
> I'll be traveling for most of January, and I am also not very active in org development lately. I will try to find time to fix this issue, but regretfully I cannot promise when that will be.

No problem at all. I am doing some end-of-year cleaning which in this case meant documenting this issue more visibly than in the commit you can see above. But I can wait for this.

Or maybe, if you already have an idea of where this would have to be addressed, you can point me to the function that I would have to adjust to come up with a, potentially temporary, fix.

tarsius:
Have fun traveling!

tarsius:
@aecay Could you please have a look at this now?

kyleam:
As @aecay notes, it has to do with having multiple `section` elements.

`ox-texinfo` uses sections to generate the menu.
`org-texinfo--normalize-headlines` ensures that each headline has a
section, and `org-texinfo-section` kicks off the menu generation for
headlines associated with the section that it's processing.

To ignore headlines, `org-export-ignore-headlines` transforms the tree
from something like

```emacs-lisp
(headline
 (:raw-value "a" ...)
 (section      ;; <----- single section
  (:begin ...)
  (paragraph
   (:begin ...)))
 (headline
  (:raw-value "_" :begin ... :tags ("ignore"))
  (section
   (:begin ...)
   (paragraph
    (:begin ...)
    #("ignore para\n"))))
 (headline
  (:raw-value "bb" ...)
  (section
   (:begin ...)
   (paragraph
    (:begin 384 ...)
    #("bb para\n" ...)))))
```

to

```emacs-lisp
(headline
 (:raw-value "a" ...)
 (section      ;; <----- original section
  (:begin ...)
  (paragraph
   (:begin ...)))
 (section      ;; <----- ignored headline's section
  (:begin ...)
  (paragraph
   (:begin ...)
   #("ignore para\n")))
 (headline
  (:raw-value "bb" ...)
  (section
   (:begin ...)
   (paragraph
    (:begin ...)
    #("bb para\n" ...)))))
```

Each of the sections in the second snippet triggers
`org-texinfo-section` to add a menu for the "a" headline.

As I understand it, a given headline should contain only one section
(spanning to the next headline), so I think `org-texinfo-section` is
correct in assuming that it should be able to generate the menu from
the section's parent.

That means that `org-export-ignore-headlines` should probably make
sure that its manipulations result in a single section.  I couldn't
see an easy way to modify the current code to insert into the correct
place initially, so another approach would be to merge the sections
after the fact (using advice for testing):

```emacs-lisp
(defun org-extra--merge-sections (data _backend info)
  (org-element-map data 'headline
    (lambda (hl)
      (let ((sections
             (cl-loop
              for el in (org-element-map (org-element-contents hl)
                            '(headline section) #'identity info)
              until (eq (org-element-type el) 'headline)
              collect el)))
        (when (and sections
                   (> (length sections) 1))
          (apply #'org-element-adopt-elements
                 (car sections)
                 (cl-mapcan (lambda (s) (org-element-contents s))
                            (cdr sections)))
          (mapc #'org-element-extract-element (cdr sections)))))
    info))

(advice-add
   'org-export-ignore-headlines :after
   #'org-extra--merge-sections
   '((name . "org-export-ignore-headlines--merge-sections")))
```

And that seems to work.

kyleam:
I think the `make texi` target needs to be updated as well.

tarsius:
Thanks! I've done that now.

PR #2918, opened 2017-01-05
Autoload ido-enter-magit-status

alezost:
Do you mind to make `ido-enter-magit-status` available right away?


kyleam:
Don't see why it shouldn't be.  Thanks.

alezost:
Thanks!

PR #2919, opened 2017-01-05
magit-push-popup: add --follow-tags switch

rosston:
`--follow-tags` has been in git since 1.8.3 (https://github.com/git/git/commit/c2aba15), and seems like a useful option. No problem if it doesn't make sense to add it here, but I thought it would be nice.

The `?t` switch seems mnemonic to me, but I'm open to it being anything.

I didn't add any release notes because I'm guessing this wouldn't go out in the next patch release (2.10.1).

Any feedback is welcome!

### DIFF
--- a/lisp/magit-remote.el
+++ b/lisp/magit-remote.el
@@ -424,6 +424,7 @@ removed after restarting Emacs."
               (?F "Force"            "--force")
               (?h "Disable hooks"    "--no-verify")
               (?d "Dry run"          "--dry-run")
+              (?t "Follow tags"      "--follow-tags")
               ,@(and (not magit-push-current-set-remote-if-missing)
                      '((?u "Set upstream"  "--set-upstream"))))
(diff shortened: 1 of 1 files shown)


kyleam:
Thanks, I've added this suggestion to the [wiki] for now.  I'm not
opposed to including this in the popup by default, though I wonder if
users that want this behavior should just set push.followTags in their
config.

[wiki]: https://github.com/magit/magit/wiki/Additional-proposed-infix-arguments-and-suffix-commands#add-argument---follow-tags

rosston:
Fair enough! `push.followTags` does seem like a fine alternative.

Thank you to all involved with magit for making/keeping it so extensible (e.g., `magit-define-popup-switch`).

tarsius:
I am closing this then.

For the grand picture on how I intend to address such requests for additional arguments in the future, see #2941.

PR #2946, opened 2017-01-09
Fix git-commit to honor 'enable-local-variables'

alezost:
Hello, I hope the commit message is clear what I mean: I work on a
project that has `.dir-locals.el` file with unsafe variables, so every
time I make a commit, I get a standard message about local variables
that "may not be safe", etc.

This happens because `normal-mode` sets `enable-local-variables` to `t`
when it is called without arguments.

I called `normal-mode` with `t` argument since any non-nil value suits,
although you can guess from the docstring that `find-file` should
probably be a string, so using `t` is like violating the documented
thing; is that OK?

### DIFF
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -407,7 +407,7 @@ already using it, then you probably shouldn't start doing so."
                                                (regexp-quote buffer-file-name)
                                                "\\'")
                                        git-commit-major-mode))))
-      (normal-mode)))
+      (normal-mode t)))
   (setq with-editor-show-usage nil)
   (with-editor-mode 1)
   (add-hook 'with-editor-finish-query-functions


tarsius:
From a quick reading of the doc-strings of `normal-mode` and `enable-local-variables` I get the impression that the current behavior is actually what you want: an attempt is made to enable all local variables, which can lead to the user being prompted:

from `normal-mode`:

> if you run `normal-mode' [...] from Lisp without specifying the optional argument FIND-FILE; in that case, this function acts as if `enable-local-variables' were t.

from `enable-local-variables`:

> A value of t means file local variables specifications are obeyed if all the specified variable values are safe; if any values are not safe, Emacs queries you, once, whether to set them all.

When queried, the user can choose to never again be queried about certain unsafe settings.

What am I missing?

alezost:
I set `enable-local-variables` to `:safe` and I don't want to be queried
at all!  Never!  :smile:

> When queried, the user can choose to never again be queried about certain unsafe settings.

I don't see this option there: there are only `y`/`n` to apply/ignore temporarily and `!` to apply and mark as safe permanently.  There is no choice to mark as unsafe permanently.  But even if it were, it doesn't matter.  A user like me just wants to set `enable-local-variables` to `:safe` or `nil` and not be bothered about local variables anymore.

tarsius:
So I was wrong about the options you have when being prompted. But again, does your proposed patch really accomplish your goal? My understanding of the quoted documentation is that it simply reinforces the current behavior, making it more explicit. Isn't `(normal-mode enable-local-variables)` what your really want?

npostavs:
> although you can guess from the docstring that `find-file` should
> probably be a string

I don't think that's true, certainly the code doesn't require it to be a string. Using `t` should be fine.

> Isn't (normal-mode enable-local-variables) what your really want?

No, because:

```elisp
(defun normal-mode (&optional find-file)
  ...
  (let ((enable-local-variables (or (not find-file) enable-local-variables)))
    ...
```

alezost:
> So I was wrong about the options you have when being prompted. But again, does your proposed patch really accomplish your goal?

Yes, it does.

> My understanding of the quoted documentation is that it simply reinforces the current behavior, making it more explicit. Isn't (normal-mode enable-local-variables) what your really want?

No, I just want to make git-commit honor the user setting of `enable-local-variables`.  Currently it is simply ignored and always assumed to be `t`.

tarsius:
Ah, I get it now, and think this should me merge. @npostavs what do you think?

npostavs:
Yes, it looks good to me.

tarsius:
Tadaaa! And thanks for explaining it to me - I had my head elsewhere.

alezost:
Thanks a lot!  I'm happy now :-)

PR #2967, opened 2017-01-19
Fix magit-version if magit.el is a symlink

raxod502:
I've written a package manager to replace `package.el`, and one of the
things it does differently is that instead of copying a package's
files to build it, it symlinks them. That is, the `build/magit` folder
is added to the `load-path`, and `build/magit/magit.el` is a symlink
to `repos/magit/lisp/magit.el`, where `repos/magit` is the Git repository
for `magit`.

Unfortunately, `magit-version` doesn't know how to work with
symlinks—that is, it expects the `.git` directory to live in
`build/magit`, when in fact it lives in `repos/magit`. Would there be
any unforeseen consequences to making `magit-version` resolve
symlinks?

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -877,14 +877,16 @@ Git, and Emacs in the echo area."
         debug)
     (unless (and toplib
                  (equal (file-name-nondirectory toplib) "magit.el"))
-      (setq toplib (locate-library "magit.el")))
+      (setq toplib (locate-library "magit.el"))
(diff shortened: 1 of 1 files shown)


tarsius:
> I've written a package manager to replace `package.el`,

Welcome to the club! So have I: https://emacsair.me/2016/05/17/assimilate-emacs-packages-as-git-submodules/.

> Would there be any unforeseen consequences

Probably not(tm). I have rebased you commit onto `master`.

raxod502:
Thanks!

PR #2975, opened 2017-01-28
Kill Ediff Registry buffer

xiongtx:
The [Ediff Registry](https://www.gnu.org/software/emacs/manual/html_node/ediff/Registry-of-Ediff-Sessions.html) buffer is the only remaining Ediff-related buffer that Magit does not clean up.

### DIFF
--- a/lisp/magit-ediff.el
+++ b/lisp/magit-ediff.el
@@ -486,6 +486,7 @@ stash that were staged."
     (ediff-kill-buffer-carefully ediff-error-buffer)
     (ediff-kill-buffer-carefully ediff-msg-buffer)
     (ediff-kill-buffer-carefully ediff-debug-buffer)
+    (ediff-kill-buffer-carefully ediff-registry-buffer)
     (when (boundp 'ediff-patch-diagnostics)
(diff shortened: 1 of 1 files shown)


tarsius:
This seems to simplistic. What if there actually *are* other active connections?

xiongtx:
Hmm...I suppose, although I don't think it's too likely. Ediffing *without* Magit, not quiting, then ediffing *with* Magic?

tarsius:
I would say, yes, that is possible.

Counter question: what's the downside of not deleting this buffer?

xiongtx:
:man_shrugging: Commands should clean up after themselves. Leaving the buffer open when it doesn't need to be violates that.

tarsius:
That's a good reason why someone(tm) should eventually look into this. (Or now, but I don't have the time to do it now.)

But it is not a good enough reason to break the registry for people who use Ediff with and without Magit.

tarsius:
Are you interested in performing the necessary research, coding a solution, and documenting it so that I can retrace your steps? If so, please quickly note so here. I am closing this pr because I think it is better to needlessly keep a buffer around than it is to break a feature for some users.

PR #2977, opened 2017-01-31
Allow cherry-pick of merge commits

bennorth:
The git command `git cherry-pick` allows you to cherry-pick a merge commit as long as you tell it which parent you want it to compute the diff with respect to.  E.g., `git cherry-pick --mainline=1 COMMIT`.  Under `magit`, all cherry-picks of merges are forbidden.  This PR sketches a fix for this:

The `--mainline` option should be an 'option' not a 'switch', as it requires an argument.

Even with that fix, magit asserts that cherry-picks must be of single-parent commits.  I'm not sure of the best way to address this.  The PR includes changes which only do the `magit-assert-one-parent` if the user has _not_ specified `--mainline`, but:
* I'm not sure how well this plays with a cherry-pick of multiple commits in a region.
* Maybe it would be better to leave out the `magit-assert-one-parent` altogether, and rely on git itself to raise an error if required.  Slightly worse user experience (you have to hit `$` to see what went wrong), but will definitely be correct compared to trying to replicate git's logic.

Does something along these lines look reasonable?

Thanks!


tarsius:
> Under magit, all cherry-picks of merges are forbidden.

I wouldn't say it is forbidden - it just doesn't work due to the two bugs you discovered. I think these bugs should be fixed in a single commit with a summary such as "Fix two bugs that prevented cherry-picking with --mainline".

lisp/magit-sequence.el
```diff
@@ -134,6 +134,9 @@ This discards all changes made since the sequence started."
             (magit-read-other-branch-or-commit prompt))
         (magit-cherry-pick-arguments)))
 
+(defun magit-cherry-pick-args-include-mainline (args)
+  (member t (mapcar '(lambda (a) (string-prefix-p "--mainline" a)) args)))
```
tarsius: This can be written as `(--any (string-prefix-p "--mainline" it) args)`, which is so short and obvious we don't need to add a named function to wrap it.

lisp/magit-sequence.el
```diff
@@ -134,17 +134,21 @@ This discards all changes made since the sequence started."
...
 Prompt for a commit, defaulting to the commit at point.  If
 the region selects multiple commits, then pick all of them,
 without prompting."
   (interactive (magit-cherry-pick-read-args "Cherry-pick"))
-  (magit-assert-one-parent (car (if (listp commit)
-                                    commit
-                                  (split-string commit "\\.\\.")))
-                           "cherry-pick")
+  (unless (magit-cherry-pick-args-include-mainline args)
+    (magit-assert-one-parent (car (if (listp commit)
```
tarsius: Instead of adding with `unless` we could also replace `magit-assert-one-parent` with `magit-cherry-pick-assert-parent` (suggestions for a better name?), which return t if there is a single parent or the `--mainline` argument is specified.

tarsius: Hm. No lets go with the `unless` and deal with this later.

lisp/magit-sequence.el
tarsius: Also never quote `lambda`.

tarsius:
Lets fix the bugs essentially as you suggested. I will open a new issue about doing this instead: Remove the `--mainline` argument and instead ask "Apply changes from Nth parent: " (defaulting to second, RET to confirm C-g to decline) when appropriate.

bennorth:
Thanks for detailed comments.  I'll re-do as suggested.  One query:

Is there a reason for the difference between the original single-parent assertion on
* plain `commit` in `magit-cherry-apply` and
* `(car (if (listp commit) commit (split-string commit "\\.\\.")))` in `magit-cherry-pick`?

bennorth:
New version created as #2978.

tarsius:
> Is there a reason for the difference

`magit-cherry-apply` should probably do the same. Moving the whole thing into `magit-cherry-pick-read-args` might be a good idea. The argument name and doc-strings should also be made more accurate.

PR #2986, opened 2017-02-04
Abbreviate default directory for buffer listings

### LINKED ISSUE #2974: abbreviate list-buffers-directory
There are two instances of `list-buffers-directory` in `magit-mode.el`. It would be nice if `default-directory` was changed to `(abbreviate-file-name default-directory)` so space is not wasted in the buffer listing telling the user where their home directory is.

holomorph:
Fixes #2974


kyleam:
Thanks

PR #2988, opened 2017-02-06
Record the working tree for separated git directories

### LINKED ISSUE #2955: magit-diff-while-commiting shows empty when the repo was made with --separate-git-dir
When editing a commit message at "COMMIT_EDITMSG" buffer, it is expected the staged changes are shown in "*magit-diff: foo" buffer, however, the "*magit-diff: foo" buffer says "(empty)" and the "foo" (directory name) is outside the git repository.

This happens the all the time as long as the repository has been made with --separate-git-dir option.
When I manually executed the function "magit-toplevel" at "COMMIT_EDITMSG" buffer, it returned wrong directory (one level higher directory than the correct git top level directory).

This is also reproduced at the pure emacs instance of "magit-emacs-Q-command".

Magit 20170106.1903, Git 2.11.0, Emacs 25.1.1, darwin
GNU Emacs 25.1.1 (x86_64-apple-darwin16.1.0, Carbon Version 157 AppKit 1504.6) of 2016-11-03
macOS Sierra version 10.12.2

kyleam:
I'm not sure if this is the direction we should go, but it's the best
I've come up with so far.  If we do want to go with someting along
these lines, I should think more carefully about the last paragraph of
the commit message.

#2955

-----------------

```
magit-toplevel needs to determine the working tree from inside .git/
when a buffer is visiting COMMIT_EDITMSG or various other files.  To
do this, it uses the following logic:

 1) file is in .git/modules/<module>/: set working tree to the output
    of "git rev-parse --show-toplevel"

 2) file is in .git/worktrees/<wtree>/: set working tree to the path
    in .git/worktrees/<wtree>/gitdir, minus the trailing "/.git"

 3) file is in .git: set working tree to the parent directory

This, however, fails when a repository was set up by passing
--separate-git-dir to "git init" or "git clone".  (It follows step 3,
returning an unrelated parent directory.)  In Git version 2.8.4 and
lower, core.worktree was set when --separate-git-dir was used, so "git
rev-parse --show-toplevel" (step 1) would also work for separated git
directories.  However, it turns out that Git was not intentionally
setting core.worktree here intentional [*].

The most visible consequence of magit-toplevel failing to return the
working tree for separated gitdirs was that an empty diff buffer was
displayed while committing (issue #2955).  9e0e2a24 (While committing,
diff from inside gitdir if necessary, 2017-01-17) worked around this
by introducing a defvar that could be let-bound to instruct
magit-toplevel to return the gitdir instead.  This resulted in the
diff buffer correctly showing staged changes because "git diff
--cached" works fine from the gitdir, but it broke
magit-diff-visit-file in these buffers (issue #2981) because the
default directory was the git directory rather than the top-level of
the working tree.  This approach also didn't consider other cases
where magit-toplevel would fail inside a separated gitdir, such as
running git-rebase-show-commit in a git-rebase-todo buffer.

Instead, let's record the worktree -> gitdir mapping of separated
repositories before the git call in magit-run-git-with-editor.  When
magit-toplevel is called from a gitdir file (COMMIT_EDITMSG,
MERGE_MSG, git-rebase-todo, ...), it can look for a working tree
associated with the current git directory.  If one isn't found, it can
take the parent directory as the working tree, as usual.  This comes
at a price of a magit-toplevel and magit-git-dir call during editing
commands, but it seems unlikely that any solution could avoid these
calls.

This should cover all cases where Magit throws users into a buffer
inside the git directory.  If this is true, it should be unlikely that
magit-toplevel uses an invalid mapping.  A user would have to call a
command that involves magit-run-git-with-editor and then, before
finishing that process, go to another working tree that points to the
same git directory and again call a command that involves
magit-run-git-with-editor.  And the user would have had to set up
these working trees outside of the "git worktree" mechanism because
magit-toplevel handles "git worktree" directories fine.

Fixes #2955.
Re: #2981.

[*] https://public-inbox.org/git/87h94d8cwi.fsf@kyleam.com/T/#u
```

### DIFF
--- a/Documentation/RelNotes/2.10.2.txt
+++ b/Documentation/RelNotes/2.10.2.txt
@@ -12,3 +12,6 @@ Fixes since v2.10.1
 * magit-abbrev-length mishandled two edge cases, returning 0 when
   called within an empty repository or outside of a repository.
   077740f5
+
+* A bug fix in the last release broke the visit-file functionality in
+  diff buffers displayed while committing.  #2988
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -395,6 +395,21 @@ absolute path is returned."
         (setq it (file-name-as-directory (magit-expand-git-file-name it)))
         (if path (expand-file-name (convert-standard-filename path) it) it)))))
 
+(defvar magit--separated-gitdirs nil)
+
+(defun magit--record-separated-gitdir ()
+  (let ((topdir (magit-toplevel))
+        (gitdir (magit-git-dir)))
+    ;; We want to delete the entry for `topdir' here, rather than within
+    ;; (unless ...), in case a `--separate-git-dir' repository was switched to
+    ;; the standard structure (i.e., "topdir/.git/").
+    (setq magit--separated-gitdirs (cl-delete topdir
+                                            magit--separated-gitdirs
+                                            :key #'car :test #'equal))
+    (unless (equal (file-name-as-directory (expand-file-name ".git" topdir))
+                   gitdir)
+      (push (cons topdir gitdir) magit--separated-gitdirs))))
+
 (defun magit-toplevel (&optional directory)
   "Return the absolute path to the toplevel of the current repository.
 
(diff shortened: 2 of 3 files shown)


tarsius:
> but it's the best I've come up with so far

I think that's what we'll have to do. The only alternative I can think of is to teach `with-editor` to pass along some additional piece of information, in the case of Magit the top-level. But that would break user customizations to the editor scripts, and risks introducing other bugs too. So I would rather add the kludge you are proposing.

kyleam:
> The only alternative I can think of is to teach with-editor to pass
> along some additional piece of information [...]

Yeah, the only other approach I considered was to have
`magit-toplevel` check that, when it steps out into the parent
directory of "<parent>/.git", the return value `magit-git-dir` there
matches the gitdir we expect.  It's not guarenteed that this would
have been the working tree, but a non-matching value is a good
indication that the user set up the repo with `--separate-git-dir`.
However, I think the main downsides of this approach are that

1) we then need to tell the user "it looks like you used
   --separate-git-dir consider; setting core.worktree", then abort and
   clean up,

2) it adds more git calls than the approach in this PR because it does
   the check for every `magit-toplevel` call that is in a gitdir,
   rather than storing info once per `magit-run-git-with-editor` call.

3) users of `--separate-git-dir` repos need manually to intervene before
   being able to perform actions that involve editing a file inside
   the gitdir.

> So I would rather add the kludge you are proposing.

OK, I'll rebase and push again to this PR.

PR #2991, opened 2017-02-09
Add new function `git-rebase-noop` to add a noop action at point

fabacino:
If you do a rebase and drop all the commits in the file, git will do nothing, because the file is considered empty (everything is commented). If you really do want to remove all the commits, you have to insert a `noop` action which prevents the file from being empty. It has no other purpose.

This commit lets the user add such a `noop` action. Currently, you have to switch to a different mode (e.g. text-mode), disable the read only mode, add `noop` manually and switch back.

### DIFF
--- a/Documentation/RelNotes/2.10.2.txt
+++ b/Documentation/RelNotes/2.10.2.txt
@@ -1,6 +1,13 @@
 Magit v2.10.2 Release Notes (unreleased)
 ========================================
 
+Changes since v2.10.1
+---------------------
+
+* The new command `git-rebase-noop' adds a noop action during a
+  rebase.  It can be used to make git perform a rebase even if none
+  of the commits are selected.  #2991
+
 Fixes since v2.10.1
 -------------------
 
--- a/lisp/git-rebase.el
+++ b/lisp/git-rebase.el
@@ -55,6 +55,7 @@
 ;;   f        Like "s" but don't also edit the commit message.
 ;;   x        Add a script to be run with the commit at point
 ;;            being checked out.
+;;   z        Add noop action at point.
 ;;
 ;;   SPC      Show the commit at point in another buffer.
 ;;   RET      Show the commit at point in another buffer and
@@ -162,6 +163,7 @@
     (define-key map (kbd "s") 'git-rebase-squash)
     (define-key map (kbd "x") 'git-rebase-exec)
     (define-key map (kbd "y") 'git-rebase-insert)
+    (define-key map (kbd "z") 'git-rebase-noop)
     (define-key map (kbd "SPC")     'git-rebase-show-or-scroll-up)
     (define-key map (kbd "DEL")     'git-rebase-show-or-scroll-down)
     (define-key map (kbd "C-x C-t") 'git-rebase-move-line-up)
@@ -207,7 +209,8 @@
     (git-rebase-show-commit
      . "show the commit at point in another buffer and select its window")
     (undo                         . "undo last change")
-    (git-rebase-kill-line         . "drop the commit at point")))
+    (git-rebase-kill-line         . "drop the commit at point")
+    (git-rebase-noop              . "add noop action at point")))
 
 ;;; Commands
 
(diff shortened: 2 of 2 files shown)


tarsius:
Thanks! Could you please also add an entry to the release notes. In a new section "Changes since v2.10.1". See older release notes for how to word your entry.

kyleam:
Interesting.  I wasn't aware that a noop line was accepted.  Is there
a situation where you prefer this to just a hard reset, or is this
just useful in cases when you've started to interactively rebase and
then realize you should have done a hard reset?

I'd prefer that you replace your current commit message body with the
first paragraph of the PR, which explains why this is useful instead
describing the behavior of the `git-rebase-noop` command and matching
its docstring.

Also, since the git-rebase manpage doesn't mention noop, it might make
sense to also explain why this command is useful in
`git-rebase-noops`'s docstring.

lisp/git-rebase.el
```diff
@@ -373,6 +376,28 @@ remove the command on the current line, if any."
...
+  "Add noop action at point.
+
+If the current line already contains a a noop action, leave it
+unchanged. If there is a commented noop action present, remove
+the comment. Otherwise add a new noop action. With a prefix
+argument insert a new noop action regardless what is already
+present on the current line.
+The extra space at the end of the action is only there to make
+it consistent with the other actions (action argument). This
+keeps the regexp `git-rebase-line' from getting complicated."
```
kyleam: I think "The extra space..." should be moved to a code comment.

lisp/git-rebase.el
```diff
@@ -373,6 +376,28 @@ remove the command on the current line, if any."
...
+  (interactive "P")
+  (goto-char (line-beginning-position))
+  (let ((noop-string "noop \n"))
+    (if (or arg
```
kyleam: s/if/when/

fabacino:
Thank you both for your input. I have updated the code accordingly.

@kyleam I guess it's just part of my workflow. I do rebasing all the time but hardly any hard resets. I find it more efficient to just start rebasing and dropping the commits I don't need than checking beforehand whether or not to use a hard reset. The time for checking is usually there anyway, but if I check beforehand and find out that I need some of the commits, I have to start again with the rebase. On the other hand, if I start rebasing and find out that none of the commits are left, I can just add the noop action and get the same as a hard reset.

Documentation/RelNotes/2.10.2.txt
```diff
@@ -1,6 +1,13 @@
 Magit v2.10.2 Release Notes (unreleased)
 ========================================
 
+Changes since v2.10.1
+---------------------
+
+* The new function `git-rebase-noop' adds a noop action during a
+  rebase. It can be used to make git perform a rebase even if none
```
kyleam: nitpicks: s/function/command/, use double space to end sentence for consistency

lisp/git-rebase.el
```diff
@@ -373,6 +376,32 @@ remove the command on the current line, if any."
...
+no commits are selected.  Without the noop action present, git
+would see an empty file and therefore do nothing."
+  (interactive "P")
+  (goto-char (line-beginning-position))
+  ;; The extra space at the end is only there to make the action
+  ;; consistent with the others (action argument). This keeps
+  ;; the regexp `git-rebase-line' from getting complicated.
+  (let ((noop-string "noop \n"))
+    (when (or arg
+            (not (looking-at noop-string)))
```
kyleam: minor: indentation is off.  Also,  I'm fine with two lines, but I think `(not ...)` on same line would read fine.

kyleam:
> On the other hand, if I start rebasing and find out that none of the
> commits are left, I can just add the noop action and get the same as
> a hard reset.

Yep, I end up with an empty rebase list every so often and then have
to follow up with a hard reset.  It'll be nice to have this command.

tarsius:
Would you mind telling us your full name, so that you can be properly attributed in https://github.com/magit/magit/blob/master/Documentation/AUTHORS.md?

fabacino:
Not at all, please check my (updated) profile.

PR #2992, opened 2017-02-10
Add `&' on a file to open it with dired-do-async-shell-command

DamienCassou:
Emacs is not good at opening every possible kind of file. For those rare cases, it's always good to be able to launch an external command.

### DIFF
--- a/Documentation/RelNotes/2.10.3.txt
+++ b/Documentation/RelNotes/2.10.3.txt
@@ -4,5 +4,9 @@ Magit v2.10.3 Release Notes (unreleased)
 Changes since v2.10.2
 ---------------------
 
+* The new command `magit-do-async-shell-command' opens file at point
+  with `dired-do-async-shell-command'.  "&" is now bound to this
+  command.  #2992
+
 Fixes since v2.10.2
 -------------------
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -1432,6 +1432,7 @@ is set in `magit-mode-setup'."
     (define-key map "C" 'magit-commit-add-log)
     (define-key map "s" 'magit-stage)
     (define-key map "u" 'magit-unstage)
+    (define-key map "&" 'magit-do-async-shell-command)
     map)
   "Keymap for `file' sections.")
 
--- a/lisp/magit-extras.el
+++ b/lisp/magit-extras.el
@@ -130,6 +130,17 @@ is no file at point then instead visit `default-directory'."
                                         file))
                              (concat default-directory "/."))))
 
+;;;###autoload
+(defun magit-do-async-shell-command (file)
+  "Open FILE with `dired-do-async-shell-command'.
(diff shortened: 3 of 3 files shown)


tarsius:
Also this isn't the right place to define this command.

DamienCassou:
> Also this isn't the right place to define this command.

could you please tell me where is the right place?

tarsius:
I'm haven't checked yet. Please fix the other issue first. And just leave the definition where it is for the time being. Or look around a bit, I'll have to do the same.

tarsius:
Probably `magit-extras.el`.

DamienCassou:
Thanks for your feedback. However, I can't find a place where it does not work. In the status buffer, it works for a file in the `Untracked files`, `Unstaged changes` and `Staged changes` sections. In Magit Rev mode, it works for both the commit summary lines and each patch header. Can you please tell me what I can do?

I moved the change to `magit-extras`.

tarsius:
I didn't remember correctly what `magit-apply--get-selection` does.

Do you think it would make sense for this command to be able to act on multiple files at once?

tarsius:
Why do you want to use `dired-do-async-shell-command`? Maybe this should be more similar to `magit-shell-command`. It could even be a specialized variant that places the file(s) after the cursor when reading the command in the minibuffer.

tarsius:
Anyway, this is a clean implementation:

```lisp
(defun magit-do-async-shell-command (files)
  "Open the file at point with `dired-do-async-shell-command'."
  (interactive (list (or (--when-let (magit-file-at-point) 
                           (list it))
                         (magit-read-files "Act on files: " nil))))
  (dired-do-async-shell-command
   (dired-read-shell-command "& on %s: " current-prefix-arg files)
   nil files))
```

DamienCassou:
Your version doesn't work as `magit-read-files` returns a comma-separated string and not a list. Here is a version based on yours that inline parts of `magit-read-files`.

DamienCassou:
Should we improve it to handle a region of files then?

tarsius:
> Should we improve it to handle a region of files then?

Yeah, it doesn't make much sense for the fallback to support multiple files, when the normal case only supports a single file.

If you yourself don't have any need for this command to support multiple files, then lets not do it. If you agree, then please rename the `files` argument to `file` and make the corresponding changes. We can still implement multi-file support later when/if someone actually requests it.

Finally please also add an entry to the release notes (see older notes for how to word it) and move the definition right after `magit-dired-jump`.

DamienCassou:
What do you think about this one?

tarsius:
The `--when-let` is redundant now.

tarsius:
Thanks!

tarsius:
Quickly noticed an issue after pushing, immediately force pushed HEAD~, and then pushed a fix commit.

DamienCassou:
Thanks @tarsius.

PR #2997, opened 2017-02-18
Add defcustom to toggle fetching when adding a remote

spwhitton:
I have a use-case for this, and others probably do too.  Thanks!

### DIFF
--- a/lisp/magit-remote.el
+++ b/lisp/magit-remote.el
@@ -104,6 +104,9 @@ variable isn't already set."
 (magit-define-popup magit-remote-popup
   "Popup console for remote commands."
   :man-page "git-remote"
+  :default-arguments '("-f")
+  :switches '("Switches for add"
+              (?f "Fetch after add" "-f"))
   :actions  '((?a "Add"     magit-remote-add)
               (?r "Rename"  magit-remote-rename)
               (?k "Remove"  magit-remote-remove)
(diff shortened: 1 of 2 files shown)


spwhitton:
Thank you for your feedback.

For my use case, adding an additional flag is not worthwhile.  So I
won't be implementing your suggestions.

(To be specific: on my git server repos are created when something is
pushed to them, so I get a failed fetch error when first adding the
remote.  But that is only mildly annoying and not worth typing the two
extra keystrokes required to disable the fetch.)

-- 
Sean Whitton

kyleam:
The idea was that it could also support the same behavior you propose.
If `-f` is added as a switch and enabled by default to keep the
current behavior, it gives you the option to toggle it per call.

But you could achieve the same effect as your patch's
`(setq magit-remote-add-fetch nil)` by `(setq magit-remote-arguments nil)`,
with the advantage that you could toggle `-f` for any given call.

kyleam:
IOW, `-f` would be on by default (like `magit-remote-add-fetch` is t
by default in your patch), but you could customize
`magit-remote-arguments` so that `-f` is off by default.

spwhitton:
Thank you for your clarificatory reply!  I will try to take another look
at this soon.

-- 
Sean Whitton

spwhitton:
Hello @kyleam.  I've rebased my branch and implemented your design.

kyleam:
Thanks. I've pushed a couple fixup commits to your branch.  Aside from
those changes, this looks good to me.

I'll wait a couple of days for others to comment and then merge if
there are no objections.

lisp/magit-remote.el
```diff
@@ -100,10 +100,20 @@ variable isn't already set."
                  (string :tag "set if named")
                  (const  :tag "don't set")))
 
+(defcustom magit-remote-arguments '("-f")
```
kyleam: `magit-define-popup` already takes care of setting up this defcustom.

spwhitton:
Hello @kyleam.  I know that `magit-define-popup` creates the variable, but I noticed that other magit commands add a defcustom so that they can add the help text.  So I thought I should follow those.

kyleam:
> but I noticed that other magit commands add a defcustom so that they
> can add the help text

I believe the ones you're noticing are actually popups that don't use
`magit-define-popup` (diff/log/commit).  Most popups do not have a
separate defcustom for the arguments.  (You can see this by running
`customize-apropos` with "magit arguments".)

lisp/magit-remote.el
```diff
@@ -125,10 +128,10 @@ variable isn't already set."
         (`(,(pred stringp) ,_) t)
         ((or `(ask ,_) `(ask-if-unset nil))
          (y-or-n-p (format "Set `remote.pushDefault' to \"%s\"? " remote))))
-      (progn (magit-call-git "remote" "add" "-f" remote url)
+      (progn (magit-call-git "remote" "add" (magit-remote-arguments) remote url)
              (setf (magit-get "remote.pushDefault") remote)
```
kyleam: Looking at this again... most commands that process popup arguments
handle the `magit-*-arguments` call in their `interactive` form.

spwhitton:
@kyleam,

`magit-remote-arguments` are now handled in the `interactive` for `magit-remote-add`.

I've autosquashed -- since you used fixup commits, I assumed that you are okay with losing attribution for your changelog entry.

Hopefully we are now done, so thank you once again for your feedback.

kyleam:
> I've autosquashed -- since you used fixup commits, I assumed that
> you are okay with losing attribution for your changelog entry.

Yep, you assumed correctly.

I'll let this sit for another day before merging in case anyone else
has comments on the latest version.

Thanks for putting this together.

tarsius:
Looks good to me.

Issue #3004, opened 2017-02-26
Git shell command completion in git command.


thierryvolpiatto:
Would be great to have git shell command completion in `magit-git-command`.
One way to achieve this is using `read-shell-command` in `magit-read-shell-command` with bash-completion package installed (works fine here), but perhaps you have a better idea or something is already existing ?

```elisp
(defun magit-read-shell-command (prompt &optional root)
  (let ((dir (if (or root current-prefix-arg)
                 (or (magit-toplevel)
                     (user-error "Not inside a Git repository"))
               default-directory)))
    (list
     (replace-regexp-in-string
      "\\`git " ""
      (read-shell-command
       (format (concat prompt ": ") (abbreviate-file-name dir))
       "git " 'magit-git-command-history))
     dir)))
```

Thanks.

tarsius:
I have to admit, I never use this or related commands, exactly because they don't offer any completion yet. But I had the intention to look into this for a while now.

I've tried out our change and installed `bash-completion`, and these seems to work very well. But I'll have to look into this a bit more before applying it. Some things we have to look into:

- Do we still need the `eshell-parse-arguments`?
- Also, it always felt weird that we are using something from `eshell` here, isn't there something equivalent that isn't intended for `eshell`?
- We could remove the git-specific variants and just keep `magit-shell-command{,-topdir}` . Those commands could then insert the "git " initial input, and users who want to run another command could just delete that.
- Since this does depend on a third-party package, we will have to communicate to users that they might want to install that.
- Or should we simply depend on it? Most likely not.
- Alternatives to `bash-completion.el` exist, e.g. [`pcmpl-git`](https://github.com/leoliu/pcmpl-git-el).
- There's even a package whose sole purpose is to read git commands in the minibuffer: [`git-command`](https://github.com/10sr/git-command-el). Unfortunately it doesn't send output to the magit process buffer  though.

So I do think the change you are proposing is a big step in the right decision, but suspect that we will go with another approach eventually.

I would apply your change anyway now, if it did not require inserting "git " as user-editable initial input. I actually tend to like that (because as mentioned above that would allow us to turn four commands into two) but this is something that is likely to change again once we settle for another approach.

Could you please look into make it unnecessary to insert "git "?

thierryvolpiatto:
Jonas Bernoulli <notifications@github.com> writes:

> I have to admit, I never use this or related commands, exactly because
> they don't offer any completion yet. But I had the intention to look
> into this for a while now.
>
[...]

Issue #2951, opened 2017-01-11
Teach magit-git-command to complete arguments


tarsius:
Personally I almost never use `magit-git-command` and friends because they offer no completion, and instead I just drop to a terminal.  But this command is often recommended when someone asks how something can be done in Magit and it can't, as in "you can always use ! to enter arbitrary git commands".  Well yes, but this shouldn't be less nice in Magit than it is in a shell.  (Also we probably should provide the feature in question in some other way, but that's not the point here.)

The built-in `pcomplete` package provides something similar to "bash-completion", but like with that someone has to actually compile a "list" of possible completions.

The `pcmpl-git` package does that. It provides the function `pcomplete/git`: "Completion rules for the `git` command.".  It also contains code to automatically compile the lists from the asciidoc sources of the git man-pages.

The `git-command` package uses `pcmpl-git` to implement a command similar to `magit-git-command`, but it doesn't send the output to a Magit buffer.  Never-the-less we should also luck at this package, it might contain some insights.

tarsius:
Also look at the discussion in #3004.

tarsius:
> Teach magit-git-command to complete arguments

Actually, since there recently have been some attempts to improve `magit-git-command`, let me rephrase that: "do not provide magit-specific but inferior alternatives to `async-shell-command`.

For the time being this can rephrased as "do not try to incrementally improve `magit-git-command` et al., it is a lost cause.

The magit-specific variants have some features that the generic ones do not. Because of that we cannot just remove the former, some users will favor those features over those available in the generic variants. Also I don't think those features are useless, just significantly less important. They are:

1. Magit's status buffer is refreshed when the process returns.
2. Output goes into the "magit process" buffer.
3. Some variants save the user from having to type "git ".

If someone wants to tackle before I get around to it, then s/he has to produce something that is just as capable as `shell-command` and that preserves these three features.

---

Here are some other discussions about the subject. They do contain useful discussions. #3004, #3005,  #3007, #3041.

npostavs:
So basically we just want to use a shell for `magit-git-command` as well?

npostavs:
As in: https://github.com/magit/magit/compare/np/cmd-shell
(this drops `magit-git-global-arguments`, probably they should be kept somehow)

npostavs:
Looks good to me. I'm a bit curious now as to what your original idea was though :)

tarsius:
[...]

PR #3005, opened 2017-02-27
Cmd line completion (#3004)

thierryvolpiatto:
#3004

Now only the git commands (`magit-git-command` and `magit-git-command-topdir`) use "git " as initial input.
The usage of `eshell-parse-arguments` have been removed.
Completion works with both bash-completion and pcml-git.
This is a first shot where we can discuss around.
Some thoughts:
I have no problems entering manually "git" at prompt, after all it is only three characters + the space, as I said I found myself often entering "git" when using `magit-git-command` and failing.
I have no problems also using `C-u` for `magit-shell-command` and `magit-git-command` so perhaps the topdir commands could be removed.
Git help shell commands are not working, perhaps they could be handled specially for emacs, calling man or info with emacs functions ?

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -600,9 +600,10 @@ Run Git in the top-level directory of the current repository.
 With a prefix argument run the command in the root of the current
 repository, otherwise in `default-directory'."
   (interactive (magit-read-shell-command "Shell command (pwd: %s)"))
-  (setq args (split-string args))
   (let ((default-directory directory))
-    (apply #'magit-start-process (car args) nil (cdr args)))
+    (start-file-process-shell-command
+     "magit-shell-command"
+     (magit-process-buffer 'nodisplay) args))
   (magit-process-buffer))
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
I see one problem now with `magit-shell-command` where you use `(apply #'magit-start-process (car args) nil (cdr args))`.
This will not work for redirections, pipe, for in in foo etc...
We need here to use start-(file)-process-shell-command.

thierryvolpiatto:
Working fine with `start-process-shell-command`, but we loose all the specific magit stuff `magit-start-process` is providing, probably this should be treated in another PR.

thierryvolpiatto:
Using `start-process-shell-command` in `magit-start-process` seems to work very well.

```diff
@@ -493,3 +493,3 @@ Magit status buffer."
-             (apply #'start-file-process
-                    (file-name-nondirectory program)
-                    process-buf program args))))
+             (start-file-process-shell-command
+              (file-name-nondirectory program)
+              process-buf (mapconcat 'identity (cons program args) " ")))))

```

tarsius:
It will be a while until I can get back to this.

thierryvolpiatto:

> It will be a while until I can get back to this.

No urgency ;-)

-- 
Thierry

tarsius:
Great!

tarsius:
Thanks everyone.

tarsius:
Uh wait, this is the pull request.

tarsius:
Closing again. Please see #2951.

PR #3007, opened 2017-02-28
Use start-file-process-shell-command in magit-start-process.

thierryvolpiatto:
* lisp/magit-process.el (magit-start-process): Do it.

Redirections, pipes and generally complex commands are not supported in `magit-shell-command` because `magit-start-process` use `(apply #'start-process args)` which assume that `(car args)` is a command where the rest of args are applied, for example the cmd line "echo foo > test.txt" will never redirect "foo" to the file test.txt but instead echo "foo > test.txt" as a string.
This PR is using `start-file-process-shell-command` which fix it.

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -490,9 +490,9 @@ Magit status buffer."
                  (process-environment (append (magit-cygwin-env-vars)
                                               process-environment))
                  (default-process-coding-system (magit--process-coding-system)))
-             (apply #'start-file-process
-                    (file-name-nondirectory program)
-                    process-buf program args))))
+             (start-file-process-shell-command
+              (file-name-nondirectory program)
+              process-buf (mapconcat 'identity (cons program args) " ")))))
     (with-editor-set-process-filter process #'magit-process-filter)
     (set-process-sentinel process #'magit-process-sentinel)
     (set-process-buffer   process process-buf)


tarsius:
> Redirections, pipes and generally complex commands are not supported in `magit-shell-command` because `magit-start-process` ...

Which means that `magit-start-process` is unsuitable for `magit-start-process. It is however suitable for all other callers but after your change it no longer is. This change has at least to negative consequences: (1) where previously one process had to be started, two are required now, and (2) there's now a shell involved and we would have to start quoting arguments accordingly all over the place.

So you need to create a modified *copy* of `magit-start-process` instead. Do it directly in #3004, because this likely will be a temporary solution only. Eventually something has to be done about the resulting code duplication.

npostavs:
> So you need to create a modified copy of `magit-start-process`

I think it should be okay to make a `magit-start-shell-process` wrapper which does something along the lines of 
```elisp
(apply #'magit-start-process shell-file-name INPUT
       shell-command-switch (mapconcat #'shell-quote-argument ARGS " "))
```

thierryvolpiatto:

>     Redirections, pipes and generally complex commands are not supported in magit-shell-command because magit-start-process ...
>
> Which means that magit-start-process is unsuitable for
> `magit-start-process. It is however suitable for all other callers but
> after your change it no longer is. This change has at least to
> negative consequences: (1) where previously one process had to be
> started, two are required now, and (2) there's now a shell involved
> and we would have to start quoting arguments accordingly all over the
> place.
>
> So you need to create a modified copy of magit-start-process
> instead. Do it directly in #3004, because this likely will be a
> temporary solution only. Eventually something has to be done about the
> resulting code duplication.

Sorry but I have not the time actually to look at this, feel free to
modify the PR or delete it.

Thanks.

--
Thierry

tarsius:
I think I'll go with letting it just sit there for a while ;-)

thierryvolpiatto:

> I think I'll go with letting it just sit there for a while ;-)

Just for your info I use this since some time now without experimenting
any problems, both in git commands and regular shell commands with pipe,
redirections, loops etc... And with completion everywhere :-)

-- 
Thierry

npostavs:
> Just for your info I use this since some time now without experimenting any problems,

But you probably restrict yourself to sensible file names (i.e., without spaces or other strange characters).

thierryvolpiatto:
Noam Postavsky <notifications@github.com> writes:

>     Just for your info I use this since some time now without experimenting any problems,
>    
> But you probably restrict yourself to sensible file names (i.e.,
> without spaces or other strange characters).

You can't handle that even with (mapconcat 'shell-quote-argument ...)
you must quote spaces in filenames yourself.
otherwise accentued character are handled properly and it is the same
with "strange characters" you have to quote them yourself.
Try with shell or eshell a command on a filename with spaces not quoted,
you will see it is not handled.
Otherwise what can be done is returning by completion the filename
already quoted, I was wrongly removing the quote in helm completion, I
have now reverted the change.


-- 
Thierry

thierryvolpiatto:
capture d ecran_2017-03-30_07-05-18
capture d ecran_2017-03-30_07-05-57
capture d ecran_2017-03-30_07-06-23
capture d ecran_2017-03-30_07-08-07

thierryvolpiatto:
And here passing filename non-quoted to ls in rxvt:
capture d ecran_2017-03-30_07-35-59

npostavs:
> You can't handle that even with (mapconcat 'shell-quote-argument ...) you must quote spaces in filenames yourself.

How about `magit-blame` in such a file?

thierryvolpiatto:
Noam Postavsky <notifications@github.com> writes:

>     You can't handle that even with (mapconcat 'shell-quote-argument ...) you must quote spaces in filenames yourself.
>    
> How about magit-blame in such a file?

Sorry I don't understand.

-- 
Thierry

npostavs:
>> How about magit-blame in such a file?
>
> Sorry I don't understand.

I mean, with your changes applied, add and commit a file named `file with spaces` to your repo, open it and run `M-x magit-blame`. I think it would fail.

thierryvolpiatto:
Noam Postavsky <notifications@github.com> writes:

>         How about magit-blame in such a file?
>
>     Sorry I don't understand.
>
> I mean, with your changes applied, add and commit a file named file
> with spaces to your repo, open it and run M-x magit-blame. I think it
> would fail.

It fails indeed because IMO the caller should quote arguments where
needed.
Using shell-quote-argument afterward will fail in such cases because the
list of args will be something like ("command" "~/file" "with"
"spaces.txt") instead of ("command" "~/file\\ with\\ spaces.txt").

--
Thierry

thierryvolpiatto:

>     Redirections, pipes and generally complex commands are not supported in magit-shell-command because magit-start-process ...
>
> Which means that magit-start-process is unsuitable for
> `magit-start-process. It is however suitable for all other callers but
> after your change it no longer is. This change has at least to
> negative consequences: (1) where previously one process had to be
> started, two are required now, and (2) there's now a shell involved
> and we would have to start quoting arguments accordingly all over the
> place.

One option is adding a new argument `shell-command` to
`magit-start-process` and use this argument only for shell/git-command,
with a nil value `magit-start-process` would behave like before.

This would avoid duplicating code and allow using
`start-process-shell-command` for shell commands and `start-process` as
before everywhere else.
See:
https://github.com/thierryvolpiatto/magit/commit/f471d36516b17e5b75d8c3c5ca169426c2e22315

--
Thierry

tarsius:
> This would avoid duplicating code

What about @npostavs [suggestion](https://github.com/magit/magit/pull/3007#issuecomment-290089108), which would additionally avoid changing function signatures?

thierryvolpiatto:

>     This would avoid duplicating code
>    
> What about @npostavs suggestion, which would additionally avoid changing function signatures?

As explained in previous posts I don't think this would work.

-- 
Thierry

npostavs:
> As explained in previous posts I don't think this would work.

Ah, because I just posted a vague outline of what to do. See #3041 for actual working code.

thierryvolpiatto:
Noam Postavsky <notifications@github.com> writes:

>     As explained in previous posts I don't think this would work.
>    
> Ah, because I just posted a vague outline of what to do. See #3041 for actual working code.

Ah, yes passing directly cmd looks a very good idea.

-- 
Thierry

thierryvolpiatto:
Noam Postavsky <notifications@github.com> writes:

>     As explained in previous posts I don't think this would work.
>    
> Ah, because I just posted a vague outline of what to do. See #3041 for actual working code.

But now it is not working with this:

  2 /bin/bash -c ls ~/tmp/test/dir\ with\ spaces
ls: cannot access /home/thierry/tmp/test/dir with spaces: No such file or directory

-- 
Thierry

thierryvolpiatto:
No that's just a stupid error of myself, so yes #3041 seems to be the best alternative and I confirm it is working fine for shell commands AFAICT.

Issue #2920, opened 2017-01-06
magit-branch-spinoff should support a single commit


tarsius:
It already supports spinning off fewer than all unpushed commits by using the region to mark multiple commits. It is not possible to select a single commit (or any section) using this selection mechanism and so it is not possible to spin off a single commit yet.

This command should be taught to spin off just the `HEAD` if point is on it in the list of unpushed commits, but not when it is on e.g. if it is on the `Head: ...` line at the top of the buffer.

There are complications to think about.

kyleam:
Hmm, but there is a docstring/code mismatch [1], right? If the code
followed the region behavior currently described in
magit-branch-spinoff's docstring, wouldn't that make it possible to
select one commit?  In the case of a region spanning two commits, the
original branch would be reset to the bottom commit and only the top
commit would be in the spinoff branch.

[1] https://groups.google.com/d/msg/magit/GzX3EKK6sj4/fj3oYCgBEgAJ

tarsius:
Yes, there is a docstring/code mismatch.

tarsius:
And the code is who is right.

tarsius:
I have fixed the documentation now.

kyleam:
> I have fixed the documentation now.

Thanks.  I would have liked to see some explanation here or in the
commit message of why that was the chosen way.  I had a slight
preference for making the code match the docstring.

kyleam:
Ah, sorry for the noise.  I see that there is some discussion at #3013.  Going there to read...

tarsius:
Please discuss here instead of in #2920, quoting bits from there as necessary.

tarsius:
Cooking in #3234.

PR #3013, opened 2017-03-02
Fix magit-branch-spinoff on selected commits

darkfeline:
This brings magit-branch-spinoff’s behavior in line with its
docstring and clarifies the dosctring.

https://github.com/magit/magit/issues/2920

### DIFF
--- a/lisp/magit-branch.el
+++ b/lisp/magit-branch.el
@@ -290,9 +290,9 @@ of the new branch, instead of the starting-point itself.
 
 If optional FROM is non-nil, then the source branch is reset to
 that commit, instead of to the last commit it shares with its
-upstream.  Interactively, FROM is non-nil, when the region
-selects some commits, and among those commits, FROM it is the
-commit that is the fewest commits ahead of the source branch.
+upstream.  Interactively when the region selects some commits,
+FROM is the selected commit that is the fewest commits ahead of
+the source branch.
 
 The commit at the other end of the selection actually does not
 matter, all commits between FROM and `HEAD' are moved to the new
@@ -320,7 +320,7 @@ from the source branch's upstream, then an error is raised."
         (when (and tracked
                    (setq base
                          (if from
-                             (concat from "^")
+                             from
                            (magit-git-string "merge-base" current tracked)))
                    (not (magit-rev-eq base current)))
           (magit-call-git "update-ref" "-m"


tarsius:
This is not an appropriate fix. When there are three commits a,b,c and I select b,c then I expect those to be spun off, not just c. Before your change the former happens, afterwards the latter.

For git to do something to a range of commits you have to tell it "do something with the commits *following* X (exclusive) and up to Z (inclusive)". But in the Magit ui we instead use " do something with Y and Z and everything in between". So the Y has to be translated to X, and the easiest way to do that is to use call it `Y~`.

`magit-branch-spinoff` is not unable to spin off a single commit because of a bug, but because the mechanism used to select the commits was not designed to be able to select a single commit. This means that another mechanism has to be used instead or in addition. Instead breaking the convention that selecting >2 neighboring items causes the invoked command to act on all of those, is not an appropriate solution.

Other commands that act on either a single item or alternatively on multiple items, usually act on the item at point if there is no select. At present I am not sure why I decided that this would not  be appropriate here. Maybe I was wrong.

I will correct the doc-string.

kyleam:
> This is not an appropriate fix. When there are three commits X,Y,Z
> and I select Y,Z then I expect those to be spun off, not just
> Z.

You expect that, but not someone who has read the docstring, or even
you at the time you wrote the docstring :)

But the point is that if a user selects something with the region,
they'll assume it is included in the operation.  Fair enough.

> Other commands that act on either a single item or alternatively on
> multiple items, usually act on the item at point if there is no
> select. At present I am not sure why I decided that this would not
> be appropriate here. Maybe I was wrong.

I had the thought that this should act like an interactive rebase and
just take the base as `commit-at-point^`, but I suppose that would
lead to too many false positives where the user happened to have point
on a commit but really wanted the old branch to be set back to the
last commit it shares with upstream.  So having the user select a
region makes their intention clear, but it introduces the corner case
where you can select one commit.

kyleam:
> it introduces the corner case where you can select one commit.

s/can/can't/

tarsius:
I think we should go with something involving `magit-log-select`, similar to how selecting a fixup target works. I'll try to get back to this in a week or so.

darkfeline:
That sounds like a good idea, thanks.

tarsius:
Since I am not going to merge this, I am closing it in favor of #2920.

PR #3016, opened 2017-03-06
magit-remote-add-set-remote.pushDefault: doc correction

duianto:
Added "that" between "provided" and "the name of".
Removed the duplicate words "the name of".

The corrections were made in the
magit-remote-add-set-remote.pushDefault
variables docstring and in it's magit.org documentation.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -4600,10 +4600,9 @@ Also see [[man:git-remote]]
   remote.
 
   If ~ask~, then always ask.  If ~ask-if-unset~, then ask, but only if the
-  variable isn't set already.  If ~nil~, then don't ever set.  If the
-  value is a string, then set without asking, provided the name of the
-  name of the added remote is equal to that string and the variable
-  isn't already set.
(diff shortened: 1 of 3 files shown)


kyleam:
Thanks for spotting this.  Do you mind updating the commit to use a
valid email?  Also, I think the commit message is superfluous here and
can be dropped.

duianto:
The message was removed, and i think the email also got changed. Where can i check here on Github if the email was changed?

kyleam:
> i think the email also got changed. Where can i check here on Github
> if the email was changed?

I fetched the updated ref for the PR; it's changed.  (I'm not sure if
this can be easily checked on GitHub.)

One last thing: if you don't mind giving your full name, we'll use
this to list you as a contributor in [AUTHORS.md].

[AUTHORS.md]: https://raw.githubusercontent.com/magit/magit/master/Documentation/AUTHORS.md

duianto:
Thanks for checking that it had changed.
My full name is: Duianto Vebotci

kyleam:
Applied.  Thanks again.

PR #3021, opened 2017-03-09
Update doc-strings: HEAD changed to `HEAD'

duianto:


kyleam:
[Lazily using a block to get plain text.]

```

I agree it'd be nice to be consistent, but I don't favor any
particular form too strongly.  Visually I prefer HEAD to `HEAD', but
it makes sense to avoid it for the reason 2ab4f4ee8 gives (i.e., avoid
confusion with parameters, which are uppercase and unquoted).  I may
prefer 'HEAD' to `HEAD', probably because HEAD isn't anything that
Emacs can actually link to.  But again, I think consistency is a good
goal, but don't care much beyond that.

With respect to "the" HEAD, without the "the" sounds better to me.
Sloppily grepping the Git documentation, it looks like the the-less
form is more common.  "the HEAD" does occur, but "the" seems to be
used most often when there is an additional noun, like "the HEAD
commit" or "the HEAD file".


```

kyleam:
> probably because HEAD isn't anything that Emacs can actually link to

which is a weak justification for my preference because `' are
frequently used in docstrings around symbols that aren't linkable

tarsius:
```
Could you please also look at the manual? In the org source we should
use ~HEAD~ for cases where we would use `HEAD' in doc-strings. You don't
need to re-generate the texinfo file, I can do that and then amend to
your commit.
```

duianto:
I'm guessing that HEAD should be left unchanged in SRC blocks like this:
https://github.com/magit/magit/blob/master/Documentation/magit.org#resolving-conflicts
```
#+BEGIN_SRC undefined
  <<<<<<< HEAD
  Take the blue pill.
  =======
  Take the red pill.
  >>>>>>> feature
#+END_SRC
```

There are 3 instances where HEAD is written as "HEAD":
>Normally you would edit one version and then tell Smerge to keep only that version. Use C-c ^ m (smerge-keep-mine) to keep the “HEAD” version or C-c ^ o (smerge-keep-other) to keep the version that follows “|||||||”.

>When the HEAD is detached then “HEAD” is in place of <branchref>.

>When snapshots are created while the HEAD is detached then “HEAD” is used in place of <branchref>.

should they be left unchanged?

In the RelNotes for example this line:
https://github.com/magit/magit/blob/master/Documentation/RelNotes/0.8.1.txt#L17
> * `magit-tag' learned to place the tag on something other than HEAD.

should HEAD be written as `HEAD' or ``~HEAD~``?

tarsius:
> It isn't optimal since one have to copy and paste from there, but it's a lot less work.

Check out https://github.com/alpha22jp/atomic-chrome.

> I'm guessing that HEAD should be left unchanged in SRC blocks

Yes.

> should they be left unchanged?

The first should be changed. The others not.

> should HEAD be written as `HEAD' or ~HEAD~?

Since these aren't org files we shouldn't use ~HEAD~. In the future we might start using org here, and then we should start using ~HEAD~ though. But until then I prefer to stick with just HEAD; even if it should turn out that some of these talk about the argument HEAD, while others talk about the concept.  On the other hand if HEAD has a prefix, as in "origin/HEAD", then we should use the string quoting.

So it seems to boil down to that one instance of HEAD that has to be changed in the manual.

tarsius:
Amended and applied. Thanks!

PR #3024, opened 2017-03-12
git-commit-setup: improve check before inserting newline

10ne1:
git-commit-setup always inserts a hardcoded newline after the
beginning of a new commit message and this is very annoying when
committing with --signoff because you get by default 3 empty lines
before the "Signed-off-by:" and you have to manually remove that
hardcoded line to conform to general git log style.

This happens because `git commit` also inserts a newline (only if
--signoff is used) so in that case a redundant newline is created
(thanks tarsius for figuring this out!).

So improve the check before inserting the newline to not insert if
--signoff is used, but keep old behaviour when the flag is absent.

Signed-off-by: Ioan-Adrian Ratiu <adi@adirat.com>

### DIFF
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -426,8 +426,7 @@ already using it, then you probably shouldn't start doing so."
     (setq save-place nil))
   (save-excursion
     (goto-char (point-min))
-    (when (= (line-beginning-position)
-             (line-end-position))
+    (when (looking-at "\\`\\(\\'\\|\n[^\n]\\)")
       (open-line 1)))
   (run-hooks 'git-commit-setup-hook)
   (set-buffer-modified-p nil))


kyleam:
> git-commit-setup always inserts a hardcoded newline after the
> beginning of a new commit message and this is very annoying when
> committing with --signoff because you get by default 3 empty lines
> above the "Signed-off-by:" line and you have to manually remove that
> hardcoded line to conform to general git log style.

Thanks, that seems worth avoiding.

> I can't see a good reason to keep hardcoding that pesky line and
> force magit users to especially advice git-commit-setup to remove it,
> so let's please just remove it by default.

You should be more thorough here, I think.  You don't see any good
reason, but there is no discussion of why fb66888f (git-commit-mode:
insert a newline after commit summary line, 2013-09-14) introduced
these lines in the first place and whether removing those lines now
will make that issue pop back up.  What problem was fb66888f trying to
solve?

Unfortunately, that commit suffers from a common problem of relying on
GitHub issue references rather than providing a self-contained
description of the problem in the commit message.  I think that was
from a time when git-commit.el was in a separate repo, so I don't know
if there is a way to get that information.  Perhaps it would become
obvious with a bit of testing and experimentation.

npostavs:
I think the issue is here https://github.com/magit/git-modes/issues/57 and there are some comments on the commit here: https://github.com/praet/git-modes/commit/1393d0fa9fbab1c5ed1f86131367098808eedd3f#comments

tarsius:
The empty line is inserted because it looks better (and apparently other reasons, some of which seem to be no longer relevant). `git commit` does the same, but only if `--signoff` (and possibly other arguments) is used. In that case two empty lines are currently inserted.

```lisp
    (when (looking-at "\\`\\(\\'\\|\n[^\n]\\)")
      (open-line 1))
```

seems to address that.

10ne1:
Thank you for the help everyone, I've updated the commit + desc.

kyleam:
@npostavs Yep, thanks.

The new commit looks fine to me.

tarsius:
Cherry-picked. Thanks!

Issue #2925, opened 2017-01-06
Branches should be listed before tags when reading a ref/rev


tarsius:
When reading a revision from the user and offering references as completion candidates (i.e when reading "a branch or something else"), then branches should be listed before tags. This is especially important when using a completion framework that shows a subset of completion candidates upfront, such as ivy and helm.

In most cases the user will want to select a branch, and while it is nice that tags are offered as candidates they should not hide the branches as one might not always remember the exact names of those.

It has gotten more urgent with 7300024190bcdafe71dc0527d1f35ea4dab8cb20, now <kbd>l o</kbd> throws us into a list of tags.

Issue #2872, opened 2016-11-25
Sort branch names by commit date


mgeisler:
I often find myself switching between the same few branch names and recently I started having two branches with the same prefix: `release-2.0` and `release-2.1`. I use magit-ido-completing-read, and with the shared prefix, I end up typing `rel21` again and again.

I would therefore like to see `--sort=-committerdate` included in magit-list-refnames as @kyleam kindly suggested:
```diff
diff --git a/lisp/magit-git.el b/lisp/magit-git.el
index af9134c..7af3975 100644
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1012,6 +1012,7 @@ (defun magit-list-unmerged-to-upstream-branches ()

 (defun magit-list-refnames (&rest args)
   (magit-git-lines "for-each-ref" "--format=%(refname:short)"
+                   "--sort=-committerdate"
                    (or args magit-list-refs-namespaces)))

 (defun magit-list-branch-names ()
```

In my testing, this makes it much faster to select the "right" branch -- which in my case means the most recently active branch.

tarsius:
That is already possible by <kbd>C-u y = s</kbd> `c` `<kbd>RET</kbd>. However the next time you want to list references you have to do it all over. I noticed that while working on #2894 and am going to fix it while working on that.

kyleam:
As far as I understand, this isn't about the magit-refs-mode.  It's
about changing the order of the refs in completion prompts.

tarsius:
Ups ;-)

kyleam:
@mgeisler As of 9c003a936, setting `magit-list-refs-sortby` to
"-creatordate" should provide the behavior you want.

mgeisler:
@kyleam I just tested it and it works, thanks a lot! :-)

alexanderkjeldaas:
This should be the default IMO - it's such an efficiency gain.

PR #3029, opened 2017-03-17
Order of ref completions

kyleam:
This PR makes the following changes:

  * Allow the sorting key of `magit-list-refs` and friends to be
    customized.

  * Prevent the default Emacs completion from alphabetically sorting
    the collection items before displaying them to the user.  There's
    still a remaining issue with the `crm` completion functions and
    global `helm-mode` or global `ivy-mode` that this PR doesn't
    address.

Please see the commit messages.

Re: #2872, #2925

### DIFF
--- a/Documentation/RelNotes/2.10.4.txt
+++ b/Documentation/RelNotes/2.10.4.txt
@@ -10,6 +10,9 @@ Changes since v2.10.3
   behavior is still to fetch after adding a remote (i.e., the switch
   is enabled by default).  #2997
 
+* Added option `magit-list-refs-sortby' to allow more control over the
+  order of refs in prompts.  #2872
+
 Fixes since v2.10.3
 -------------------
 
--- a/Documentation/RelNotes/2.10.4.txt
+++ b/Documentation/RelNotes/2.10.4.txt
@@ -13,6 +13,12 @@ Changes since v2.10.3
 * Added option `magit-list-refs-sortby' to allow more control over the
   order of refs in prompts.  #2872
 
+* The Magit wrappers around the default Emacs completion functions now
+  override the default behavior of alphabetically sorting choices when
+  displaying them in the "*Completions*" buffer.  In repositories with
+  many release tags, the new behavior prevents completion prompts from
+  being dominated by version tags instead of branch names.  #2925
+
 Fixes since v2.10.3
 -------------------
 
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -1191,6 +1191,14 @@ More frameworks will be supported in the future.
   comes with a supposed drop-in replacement ~ido-completing-read~, but
   that has too many deficits to serve our needs.
 
+- User Option: magit-list-refs-sortby
+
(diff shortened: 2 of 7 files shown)


tarsius:
Yes, lets do that. Any luck with crm yet?

kyleam:
> Any luck with crm yet?

No, haven't looked into the {helm,avy}-mode/crm interaction further.
I'll try to revisit it soon.

tarsius:
There's no hurry. I just worried that my thumbs up might have been a little to subtle. ;-)

kyleam:
> haven't looked into the {helm,avy}-mode/crm interaction further.

Doh, s/avy/ivy/

kyleam:
> I just worried that my thumbs up might have been a little to subtle. ;-)

heh, and you'd be right.  I did not notice it :)

kyleam:
OK, the updated series

  * deals with a bug in `completion-pcm--all-completions`

    This function returns the collection in reverse order, which is
    why `display-sort-function` had to be set to `nreverse` in some
    cases to preserve the collection.  But this then led to the
    completion being in the wrong order in other cases (particulary
    when partial completion was used).

    There is already an Emacs [bug report] with a fix for this, as
    @npostavs pointed out to me when I opened a [new report] for it.

  * solves the `crm` sorting for `ivy-mode` issue because that was
    actually due to the `completion-pcm--all-completions` issue above.

  * makes `magit-push-refspecs` also preserve the collection order

That leaves the issue with `helm-mode` and `crm` completion.  The
reason that the sorting isn't preserved is because `helm-mode` sets
`completion-in-region-function` to `helm--completion-in-region`, which
sorts the entries with the function `helm-generic-sort-fn`.  I'll open
a helm issue and see if it'd be possible to introduce a variable that
disables sorting.  (Well, `helm-completion-in-region-fuzzy-match` is
already a variable that will disable sorting when it's non-nil, but it
does more than just turn off sorting.)

[bug report]: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=24676#11

[new report]: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=26313#5

kyleam:
> That leaves the issue with helm-mode and crm completion.

Fixed with c00279fe.  That relies on a new option in Helm, which means
that only users with a recent version will get the intended sorting
behavior.  But given that the current `helm-mode`/`crm` has been
around for a long time (rather than being a regression introduced by
this PR), I think that's fine.

tarsius:
You obviously put a lot of thought into this, so please go ahead and merge it when you feel it is ready.

Two little things I noticed:
* "was added added in" in 8a9c84495aa8514615abe69bc72c035151d5ec1c commit message.
* Long relnotes line added in 323f89a5bf0a403783088282c7ba592e8d4e1389.

kyleam:
> Two little things I noticed: [...]

Thanks for spotting those.  Will fix.

> please go ahead and merge it when you feel it is ready.

OK, I'll plan to merge it tomorrow if I don't run into any issues
while running using this branch locally.

PR #3033, opened 2017-03-26
Fix magit-version if magit.el is a symlink, part ii

raxod502:
I didn't quite catch all the possible cases in the control flow when I submitted #2967, so I recently ran into another problem with symlinks. This pull request makes it so that symlinks are *definitely* resolved properly by `magit-version`.

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -877,8 +877,8 @@ Git, and Emacs in the echo area."
         debug)
     (unless (and toplib
                  (equal (file-name-nondirectory toplib) "magit.el"))
-      (setq toplib (locate-library "magit.el"))
-      (setq toplib (and toplib (file-chase-links toplib))))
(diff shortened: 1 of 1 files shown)


tarsius:
Could you please give an example of a situation where this change is necessary.

raxod502:
`magit-version` fails when

1. `magit.el` is a symlink, and
2. has not been byte-compiled (so that there is no `magit.elc` in the same directory as the symlink), and
3. is being loaded for the first time, or the user is currently viewing `magit.el`.

Then the condition in the `unless` is triggered, and `file-chase-links` is not invoked.

Even without this use case, the previous logic (which I wrote, oops) is just wrong. It doesn't make any sense to only resolve symlinks sometimes.

tarsius:
I've cherry-picked this. Thanks.

PR #3040, opened 2017-04-01
Allow commit and amend to use custom commands

jasonm23:
The motivation for this change is to allow magit users to configure the commit commands when committing and amending a commit. (defaulting of course to "commit")

When using commit wrappers such as git-duet, it's necessary to use git duet-commit for committing, and a simple commit for amending.

By adding these custom variables, we provide the ability to use aliases and wrappers for committing and amending.

### DIFF
--- a/lisp/magit-commit.el
+++ b/lisp/magit-commit.el
@@ -74,6 +74,20 @@
   :group 'magit-commands
   :type 'boolean)
 
+(defcustom magit-commit-comands-commit "commit"
+  "The git command used for commit.  This option exists to
+  allow support for git aliases and extensions such as git duet."
+  :group 'magit-commands
+  :safe 'stringp
+  :type 'stringp)
+
+(defcustom magit-commit-comands-commit-for-amend "commit"
+  "The git command used for commit --amend.  This option exists to
+  allow support for git aliases and extensions such as git duet."
+  :group 'magit-commands
+  :safe 'stringp
+  :type 'stringp)
+
 (defcustom magit-commit-squash-confirm t
   "Whether the commit targeted by squash and fixup has to be confirmed.
 When non-nil then the commit at point (if any) is used as default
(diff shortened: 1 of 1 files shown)


tarsius:
This looks innocent enough on first sight, but after looking at [`git-duet`](https://github.com/git-duet/git-duet)'s readme, I am worried that merging this would result in lots of noise solely for its benefit. After all `git-duet` provides variations of more than just `commit`.

I prefer your approach in https://gist.github.com/jasonm23/6326d278d99fe82f252a90d9b9ad61db.

But down the road that could lead to a lots of commands being copied from Magit. So you might also want to consider implementing this completely in elisp instead of wrapping `git duet-COMMAND`. You could use around advises to bind the variables and `magit-commit-setup-hook` to insert the signoff line.

jasonm23:
Hi Jonas, thanks for taking a look.

FWIW My original plan was to create a minor mode which read the git config instead. Only offering a duet commit option when the appropriate config keys were available.  

Regarding the Gist solution.  I am fairly happy to use this myself.  If people​ comment and request a proper Magit extension for it. I'd be happy to rework that implementation.

I'm not sure how advise can allow me to override a string literal for `"commit"` unless I go and catch and map all incoming commands? 

Can you help me understand your suggested solution a bit more?

jasonm23:
In other words how would the user flow change (if at all)?

kyleam:
> I'm not sure how advise can allow me to override a string literal
> for "commit" unless I go and catch and map all incoming commands?

I might be misunderstanding Jonas's comment, but I thought his
suggestion was to implement the `git duet-commit` functionality in
elisp.  So, you wouldn't need to change the "commit" string literal;
you'd need to manipulate the commiter name/email as well as the sign
off.

jasonm23:
@tarsius @kyleam Thanks, that's what I'd originally intended to do.

Just checking in case I'd missed something.

kyleam:
To point to an example that might be helpful: `magit-commit-reword`
overrides `GIT_COMMITTER_DATE`.  Similarly, you could put advice
around `magit-commit` and `magit-commit-amend` to override
`GIT_COMMITTER_NAME` and `GIT_COMMITTER_EMAIL`.

jasonm23:
@kyleam thank you Kyle.

tarsius:
@jasonm23 So you are not lobbying for this to be merged, and and I can close it, right?

jasonm23:
@tarsius while I don't share your reservations about merging this, there are a number of ways I can achieve my goal.

I think it might be useful to avoid string literals for git commands, I don't think this is the patch to achieve that cleanly.

Feel free to close it.

Thanks again for your attention on this.

PR #3043, opened 2017-04-03
Use a shell in magit-git-command et al

tarsius:
Re #2951.


npostavs:
https://github.com/magit/magit/issues/2951#issuecomment-291023248
> (this drops `magit-git-global-arguments`, probably they should be kept somehow)

Sorry, I forgot about this when looking at your updates. Would missing the `--no-pager` argument be a problem for us?

tarsius:
Just pushed a fix for that.

PR #3044, opened 2017-04-04
Improvements to magit-version

npostavs:
From recent stackexchange questions, I think our `magit-version` command is not sufficiently foolproof for a troubleshooting tool.

http://emacs.stackexchange.com/questions/31913/how-to-direct-m-m-xs-output-to-the-current-buffer-rather-than-to-the-ec
http://emacs.stackexchange.com/questions/31868/magit-version-post-installation-does-not-look-as-in-the-manual

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -312,9 +312,12 @@ add a section in the respective process buffer."
                       (setq msg (with-temp-buffer
                                   (insert-file-contents log)
                                   (goto-char (point-max))
-                                  (and (re-search-backward
-                                        magit-process-error-message-re nil t)
-                                       (match-string 1))))
+                                  (cond
+                                   ((functionp magit-git-debug)
+                                    (funcall magit-git-debug (buffer-string)))
+                                   ((re-search-backward
+                                     magit-process-error-message-re nil t)
+                                    (match-string 1)))))
                       (let ((magit-git-debug nil))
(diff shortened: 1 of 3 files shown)


tarsius:
I was thinking the same.

In addition to this, we might also want to create something like `with-editor-debug`, but for `git` instead of `emacsclient`.

lisp/magit-git.el
```diff
@@ -387,6 +390,34 @@ call function WASHER with ARGS as its sole argument."
     (if raw it (and (string-match "^\\([0-9]+\\.[0-9]+\\.[0-9]+\\)" it)
                     (match-string 1 it)))))
 
+(defun magit-git-debug ()
```
npostavs: > In addition to this, we might also want to create something like `with-editor-debug`, but for `git` instead of `emacsclient`.

Like this?

tarsius:
> Like this?

Yes, but I would name it `magit-debug-git-exectuable` and move the definition somewhere near that of `magit-version`. Also I think that the value of `magit-git-exectuable` should be formatted using `%S` instead of `%s`, and if the value is `"git"`, then the value of `(exectuable-find magit-git-executable)` should also be displayed.

npostavs:
Makes sense, updated.

tarsius:
Please also add documentation.

npostavs:
I added a bit, do we need more?

PR #3046, opened 2017-04-06
Test with more Emacs versions

npostavs:
I have some Emacs binaries for Travis at https://github.com/npostavs/emacs-travis/releases

They can run in the non-sudo environment, but unfortunately, magit's tramp test requires sudo to run. Possibly we could change it to use ssh instead (I've checked that sshd is installed on Travis), though I'm not sure if it's worth bothering with that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,30 +1,23 @@
+# We need sudo for the tramp test.
+sudo: true
 language: generic
 env:
   global:
     - CURL="curl -fsSkL --retry 9 --retry-delay 9"
   matrix:
-    # - EMACS=emacs24 ## This is 24.3 which is too old to run magit
-    - EMACS=emacs-snapshot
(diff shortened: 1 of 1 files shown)


tarsius:
Great!

> though I'm not sure if it's worth bothering with that.

Seem okay to me that we have to enable `sudo`.

npostavs:
Supposedly the `sudo`-less environment is faster, although the `fix.CVE-2015-7547` thing that runs every time probably drowns out the gains anyway.

npostavs:
> Possibly we could change it to use ssh instead (I've checked that sshd is installed on Travis), 

Actually I worked out a simpler way, we can just override the tramp method to avoid calling `sudo`: 20ba0d6145e447c5a62fdb50bbcae16d0923b1e3

PR #3051, opened 2017-04-15
log: use NUL to delimit refs

### LINKED ISSUE #3028: Parentheses in a branch name, breaks that commits log entry
#### What behavior you expected
The magit log shouldn't look any different when there are parentheses in a branch name, or if one should avoid using parentheses, then there probably should be a warning if one tries to use them.

#### What behavior you observed
The log entry looks broken, for a commit with a branch name that has parentheses in it.
The hash, author and date columns are empty. Everything is written in the middle column that usually just shows the branch names, remote branch names and the commits subject line. 

The middle column also shows `(refs/heads/` before the branch name. And there's an empty commit line below it. 

Here's a screenshot:

emacs_2017-03-16_12-43-17

When the parentheses are removed from the branch name, then the log entry looks normal again:

emacs_2017-03-16_13-02-42

#### How we can reproduce the issue.
1. Start Emacs (with magit installed).
2. Open a repository
3. Rename an existing branch or create a new branch with parentheses in the name.
4. Open the magit-log buffer.

#### System info
Magit 20170312.616, Git 2.11.1.windows.1, Emacs 25.1.1, windows-nt

kyleam:
Please see commit messages.

Fixes #3028.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1261,8 +1261,15 @@ Return a list of two integers: (A>B B>A)."
                   'face (if (equal ref head) 'magit-branch-current face)))))
 
 (defun magit-format-ref-labels (string)
+  ;; To support Git <2.2.0, we remove the surrounding parentheses here
+  ;; rather than specifying that STRING should be generated with Git's
+  ;; "%D" placeholder.
+  (setq string (->> string
+                    (replace-regexp-in-string "\\`\\s-*(" "")
+                    (replace-regexp-in-string ")\\s-*\\'" "")))
   (save-match-data
-    (let ((regexp "\\(, \\|tag: \\| -> \\|[()]\\)") head names)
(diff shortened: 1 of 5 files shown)


lisp/magit-git.el
```diff
@@ -1262,7 +1262,14 @@ Return a list of two integers: (A>B B>A)."
 
 (defun magit-format-ref-labels (string)
   (save-match-data
-    (let ((regexp "\\(, \\|tag: \\| -> \\|[()]\\)") head names)
+    ;; To support Git <2.2.0, we remove the surrounding parentheses
+    ;; here rather than specifying that STRING should be generated
+    ;; with Git's "%D" placeholder.
+    (let ((str (->> string
```
tarsius: I would prefer if you did just `(setq string ...)` here. Otherwise it is not clear whether the old value of `string` is preserved because it is still needed for something.

kyleam: Good point.

Issue #3060, opened 2017-05-03
Signature Removes magit-log Commit Messages


base698:
(1) If show signature is enabled in .gitconfig you can't see commit messages in magit-status for signed commits.  
  (2) the commit log message missing and instead shows 
```
gpg: Signature made Wed 03 May 2017 08:05:16 AM PDT
```
  (3) set .gitconfig showSignature = true and open magit status with a signed commit.

  M-x magit-version RET

```
Magit 20170425.638, Git 2.11.0, Emacs 25.1.1, gnu/linux
```

kyleam:
Thanks for a report.

It looks like `log.showSignature=true` can lead to issues anywhere
`git log` or `git show` is called because it will prepend the
`--show-signature` output to any `--format` specification.  There
isn't a "don't show signature" flag to override this, although we
could add "-c" "log.showSignature=false" to each of these calls.  But
I think the best approach would be to just add "-c"
"log.showSignature=false" to `magit-git-global-arguments`.

PR #3061, opened 2017-05-03
fix show signature

base698:
Fix commit messages which are over written with signature, when
using `log.signatureEnabled=true`. #3061


kyleam:
Given [my comment] in #3060, I'm of course fine with this change.
However, I don't find the commit message sufficient.  If I were to
look at this patch without the context of #3060, I wouldn't understand
why this is needed.  Also, could you add an entry to
Documentation/RelNotes/2.11.0.txt?

With those changes, I'll merge this in a couple days unless there are
objections.

Thanks.

[my comment]: https://github.com/magit/magit/issues/3060#issuecomment-298986087

base698:
Made said changes.

base698:
:+1:

PR #3062, opened 2017-05-04
magit-show-commit: when in a log buffer, use log's file restriction

### LINKED ISSUE #3056: Support restricting diff/commit view to current file
I like the default behaviour of `git-log-files` from the `git-emacs` package better than `magit-log-buffer-file`. If I press `d` to see the diff between two commits in the resulting list from `git-log-files` the diff is restricted to the file in question. In contrast if I press enter in the `magit-log-buffer-file` list, I get to see the whole commit, which is more often than not, not what I need.

The semantic of the diff in `git-log-files` is a little different since it calculates an actual diff between  two commits, whereas `magit-log-buffer-file` shows the actual commit. But git also supports to restrict the diff of a commit to certain files, so it should be easy to support that while keeping current semantic. One way this could work in the UI is to provide a toggle switch in the log view to specify if commit views opened from there should be restricted by file (since the log is already filtered by the file in question, it's clear in this context which file is meant).

I think this is the only feature left from `git-emacs` which I use regularly, so if magit would implement that I could simplify my configuration.

`git-log-files` also supports restricting the log and diff to multiple files, but the most common usecase for me is currently viewing the log and the commits of the current buffer, so that might not be considered part of this request.

kyleam:
```
All calls to magit-show-commit take the file restriction from
magit-diff-arguments.  But when a call is made from a log buffer, the
user is viewing the revision in the context of the log, so the log's
file restriction is more relevant than the diff's [*].

Adjust magit-show-commit calls to respect the log's file restriction
if the current buffer is a log buffer.  Leave calls to
magit-show-commit in git-rebase.el and magit-blame.el unchanged
because these aren't designed to be executed from a log buffer.

Closes #3056

[*] This patch assumes that all users would prefer the log's file
    restriction over the diff's in this situation, but of course we
    can add an option to revert to the previous behavior if this turns
    out not to be the case.
```

Re: #3056

### DIFF
--- a/Documentation/RelNotes/2.11.0.txt
+++ b/Documentation/RelNotes/2.11.0.txt
@@ -46,6 +46,9 @@ Changes since v2.10.3
 
 * Added `magit-file-checkout' to `magit-reset-popup'.  #3052
 
+* When a revision is shown from a log buffer, the revision buffer now
+  inherits the file restriction of the log buffer.  #3056
+
 Fixes since v2.10.3
 -------------------
 
--- a/Documentation/RelNotes/2.11.0.txt
+++ b/Documentation/RelNotes/2.11.0.txt
@@ -49,6 +49,8 @@ Changes since v2.10.3
 * When a revision is shown from a log buffer, the revision buffer now
   inherits the file restriction of the log buffer.  #3056
 
+* Add new command `magit-revision-toggle-file-filter'.  #3062
+
 Fixes since v2.10.3
 -------------------
 
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -2473,6 +2473,15 @@ without having to using one of the diff popups.
   the committer image, accordingly.  Either the car or the cdr may
   be nil.
 
+- Key: M-x magit-revision-toggle-file-filter, magit-revision-toggle-file-filter
+
+  This command toggles the file restriction of the current revision
+  buffer.  It is particularly useful when you display a revision from
+  a log buffer that is restricted to a file or files.  In this case,
+  the revision buffer is also restricted to these files, and this
+  command allows you to quickly switch between viewing all the changes
+  in the commit and the restricted subset.
+
 ** Ediffing
 
 This section describes how to enter Ediff from Magit buffers.  For
--- a/Documentation/magit.texi
+++ b/Documentation/magit.texi
@@ -3379,6 +3379,19 @@ the committer image, accordingly.  Either the car or the cdr may
 be nil.
 @end defopt
 
+@table @asis
+@kindex M-x magit-revision-toggle-file-filter
+@cindex magit-revision-toggle-file-filter
+@item @kbd{M-x magit-revision-toggle-file-filter} @tie{}@tie{}@tie{}@tie{}(@code{magit-revision-toggle-file-filter})
+
+This command toggles the file restriction of the current revision
+buffer.  It is particularly useful when you display a revision from
+a log buffer that is restricted to a file or files.  In this case,
+the revision buffer is also restricted to these files, and this
+command allows you to quickly switch between viewing all the changes
+in the commit and the restricted subset.
+@end table
+
 @node Ediffing
 @section Ediffing
 
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -1938,6 +1938,24 @@ or a ref which is not a branch, then it inserts nothing."
                (if magit-revision-use-gravatar-kludge slice2 slice1)
                (if magit-revision-use-gravatar-kludge slice1 slice2)))))))
 
+(defvar-local magit-revision-files nil)
+
+(defun magit-revision-toggle-file-filter ()
+  "Toggle the file restriction of the current revision buffer."
+  (interactive)
+  (with-current-buffer (or (and (derived-mode-p 'magit-revision-mode)
+                                (current-buffer))
+                           (magit-mode-get-buffer 'magit-revision-mode)
+                           (user-error "No revision buffer found"))
+    (let ((files (nth 3 magit-refresh-args)))
+      (unless (or magit-revision-files files)
+        (user-error "No file filter to toggle"))
+      (setf (nth 3 magit-refresh-args) (if (not files)
+                                           magit-revision-files
+                                         (setq magit-revision-files files)
+                                         nil))
+      (magit-refresh))))
+
 ;;; Diff Sections
 
 (defvar magit-unstaged-section-map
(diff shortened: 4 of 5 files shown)


tarsius:
I think we should also add a command to toggle between using the log buffer's file filter in the revision buffer or not. That command would have to remember when the user explicitly chose not to apply the log filter, probably using a buffer-local variable in the revision buffer and `magit-show-commit--arguments` should respect that.

Also `magit-show-commit--arguments` when called in `magit-log-maybe-update-revision-buffer-1` assumes that we are still in the log buffer. It's highly likely that we are, but we could also record the log buffer in addition to the rev buffer, and make the former current before calling `magit-show-commit--arguments`.

Only marginally related, but "Commit abcdef in file foobar" doesn't make much sense. What about "Commit abcdef limited to file foobar" instead?

kyleam:
Thanks for taking a look.

> I think we should also add a command to toggle between using the log
> buffer's file filter in the revision buffer or not. That command would
> have to remember when the user explicitly chose not to apply the log
> filter, probably using a buffer-local variable in the revision buffer
> and magit-show-commit--arguments should respect that.

Hmm, I think @bitti [expressed] a similar desire for a toggle.  I'll
add a commit for that.

I think the main reason I'm hesitant about it is because I don't see
how this file restriction case differs from other cases already in
Magit, and these don't have a toggle.

[expressed]: https://github.com/magit/magit/issues/3056#issue-225168898

> Also magit-show-commit--arguments when called in
> magit-log-maybe-update-revision-buffer-1 assumes that we are still in
> the log buffer. It's highly likely that we are, but we could also
> record the log buffer in addition to the rev buffer, and make the
> former current before calling magit-show-commit--arguments.

Good point.

> Only marginally related, but "Commit abcdef in file foobar" doesn't
> make much sense. What about "Commit abcdef limited to file foobar"
> instead?

Yep, that's better.

bitti:
I don't really need the toggle, I only suggested it in the hope it might increase the probability of this being implemented, because it shouldn't break anyones workflow. I would be perfectly fine if there is just a simple way to get from the filtered commit to a full commit view.

According to the YAGNI principle you could leave out the toggle and just see how many people complain and if it's actually needed. (It's not the first time that @tarsius changed behaviour in order to increase consistency.)

tarsius:
> I think the main reason I'm hesitant about it is because I don't see how this file restriction case differs from other cases already in Magit, and these don't have a toggle.

I assume you are mainly talking about other settings that can be toggled by using the appropriate "change arguments" popup.

For some of those there already exist additional toggle (or other) commands that can be used from outside popup. For example `magit-diff-{less,more}-context`.  And other commands like `magit-diff-flip-revs` could theoretically be bound outside of the popup, but currently are available only from within the popup.  Maybe we should only do the latter for the file-filter toggle command too.

But more importantly, the main difference between the proposed file-filter toggle command and other settings, that don't have such a dedicated command, is that the log-derived file-filter isn't something the user explicitly requested, but some dwim behavior.  That's why I don't think it is inconsistent to have a command to change the file-filter, while not having one to change `--diff-algorithm` for example.

Also I think it is a reasonable work-flow to limit a log to a certain file and only wanting to see the diff for that file *initially*. And to then view the whole commit for context if necessary, before moving to another commit, which again should be limited to just the file.

Whether all users appreciate this behavior is to be seen, and if not, then we'll have to add an option to disable it completely.  Also there are some edge cases we might or might not have to address explicitly, for example what should be done when the revision buffer is already visible and unfiltered at the time the log buffer is being filtered?  We can figure out these things as we go, but I think the toggle command should be part of the initial version.

kyleam:
>> I think the main reason I'm hesitant about it is because I don't
>> see how this file restriction case differs from other cases already
>> in Magit, and these don't have a toggle.
>
> I assume you are mainly talking about other settings that can be
> toggled by using the appropriate "change arguments" popup.

Sorry for the confusion.  I am only talking about the file-restriction
cases.

For example:

  * When generating a new log, the file restriction is taken from
    `magit-log-arguments`.

  * When generating a new diff, the file restriction is taken from
    `magit-diff-arguments`.

  * When generating a new revision buffer, the file restriction is
    taken from `magit-diff-arguments`.

    The only change this patch makes is that it says "when we're in a
    log buffer `magit-log-arguments` is the more meaninfgul
    restriction"

So it isn't that we are now considering a file restriction when we
weren't before.  My point was that none of the above situations have a
toggle.  Here we're only changing a detail about the last item.

tarsius:
In the case of `log`, `diff`, and - for the time being - `revision` buffers, the user explicitly requested the file restriction at some point - for that particular buffer. It might be that s/he doesn't want that restriction anymore. There isn't much we can do about this - we remember an old, explicit setting and cannot know at what time it would be better to reset to the default setting.

So we either decide to never remember settings (I think your personal configuration contains an option to allow that) or the user has to explicitly discard the no longer appropriate setting at some point.

But when consulting the log buffer to determine the file restriction for the revision buffer, the user did not actually explicitly request that this be done. We are currently operating under the assumption that users will find this useful, but if that turns out to be wrong for some users, then fixing that involves adding an option to disable that behavior completely. The command that I am suggesting serves a different purpose.

The "inherit the file restriction from log" toggle command does not remove a buffer-local setting that is no longer useful, instead it switches between two alternative views of the same commit that may both be useful. The ability to quickly switch between them make these views more useful. Switching to one view doesn't mean that the user doesn't intend to use the previous view anymore for some time.

The restricted view allows seeing the changes we are really interested in without first having to first navigate there and potentially being distracted by the surrounding noise; and switching to the unrestricted view allows seeing additional context when it turns out that is actually necessary. Being able to do this using a simple command instead of having to do it in a way that involves a popup and having to confirm the default choice at a prompt makes that much more pleasant to do.

The other restrictions don't need such a toggle command because for those restrictions one usually does not want to go back an forth - instead one wants to drop a setting that is no longer suitable.

I hope this makes it a bit clearer what I am talking about. If not, then I will try another day, possibly in form of an implementation.

kyleam:
> I think your personal configuration contains an option to allow that

Yeah, I suppose it's no secret that I [don't like] sticky arguments :)

[don't like]: https://github.com/magit/magit/pull/2146

> But when consulting the log buffer to determine the file restriction
> for the revision buffer, the user did not actually explicitly
> request that this be done.

Perhaps that's true for most users, but it's not for me.  Consider two
actions that limit the file restriction in the revision buffer.

  * Diffing some changes with a file restriction.  Some time later
    showing a revision (say from the log, but anywhere would do).

  * Logging a buffer file and then showing a revision from the log.

Though neither of these is an explicit request to restrict the files,
the second situation is much closer to an explicit request to my mind.

> I hope this makes it a bit clearer what I am talking about.

It does, especially the part about why you view the toggle as better
than going through the popup with `D=f`.  And I see why it's useful
regardless of the user's opinion on the two situations I gave above.

Thanks.

PR #3064, opened 2017-05-05
missing comma

0fiscalentropy:


0fiscalentropy:
@kyleam I've created a new [PR](https://github.com/magit/magit/pull/3065) with a couple of more cases, one commit per file. Is it better if I squash all the commits into one and call it something like "add missing commas to various doc-strings"?

kyleam:
> I've created a new PR with a couple of more cases, one commit per
> file.

Thanks!  For future reference, we generally prefer if you just force
push to the PR.

> Is it better if I squash all the commits into one and call it
> something like "add missing commas to various doc-strings"?

Yes, that's my preference in this case.

PR #3065, opened 2017-05-06
add some missing commas

0fiscalentropy:


lisp/magit-mode.el
```diff
@@ -230,11 +230,11 @@ improve performance."
 (defcustom magit-save-repository-buffers t
   "Whether to save file-visiting buffers when appropriate.
 
-If non-nil then all modified file-visiting buffers belonging
-to the current repository may be saved before running Magit
-commands and before creating or refreshing Magit buffers.
-If `dontask' then this is done without user intervention, for
-any other non-nil value the user has to confirm each save.
+If non-nil, then all modified file-visiting buffers belonging to
```
kyleam: Right, despite being very short, I suppose "If non-nil" is still a
dependent clause where the subject and verb are implied.  Also,
grepping Emacs's sources suggests most instances use a comma after
this.

lisp/magit-mode.el
```diff
@@ -230,11 +230,11 @@ improve performance."
...
+the current repository may be saved before running Magit commands
+and before creating or refreshing Magit buffers.  If `dontask'
```
kyleam: Add one here too, then?

lisp/magit-section.el
```diff
@@ -1032,9 +1032,9 @@ either `hide' or nil."
 (defun magit-section-siblings (section &optional direction)
   "Return a list of the sibling sections of SECTION.
 
-If optional DIRECTION is `prev' then return siblings that come
-before SECTION, if it is `next' then return siblings that come
-after SECTION.  For all other values return all siblings
+If optional DIRECTION is `prev', then return siblings that come
+before SECTION, if it is `next', then return siblings that come
```
kyleam:
```
While we're here, I'd be fine making the s/, if it is `next'/.  If it is `next'/ correction.
```

lisp/magit-section.el
```diff
@@ -1032,9 +1032,9 @@ either `hide' or nil."
...
+after SECTION.  For all other values, return all siblings
```
kyleam: OK by me, but I think this one is a prepositional phrase (rather than
a dependent clause), and the comma is more a matter of preference.

PR #3070, opened 2017-05-10
git-commit-filename-regexp: autoload

basil-conto:
This will allow users to register `git-commit-mode` with `auto-mode-alist` without having to enable `global-git-commit-mode` and thus eagerly load the whole `git-commit.el` package.

The motivation for this change was [this Q&A](https://emacs.stackexchange.com/q/32676/15748). If the proposed addition of an autoload cookie is not acceptable/desirable, I would be grateful for any suggestions addressing the points in the linked question.

### DIFF
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -369,6 +369,7 @@ already using it, then you probably shouldn't start doing so."
 
 ;;; Hooks
 
+;;;###autoload
 (defconst git-commit-filename-regexp "/\\(\
 \\(\\(COMMIT\\|NOTES\\|PULLREQ\\|TAG\\)_EDIT\\|MERGE_\\|\\)MSG\
 \\|BRANCH_DESCRIPTION\\)\\'")
@@ -390,6 +391,7 @@ already using it, then you probably shouldn't start doing so."
        (string-match-p git-commit-filename-regexp buffer-file-name)
        (git-commit-setup)))
 
+;;;###autoload
 (defun git-commit-setup ()
   ;; cygwin git will pass a cygwin path (/cygdrive/c/foo/.git/...),
   ;; try to handle this in window-nt Emacs.


kyleam:
I'm OK with this.  But since you want to do

    (add-to-list 'auto-mode-alist
                 (cons git-commit-filename-regexp #'git-commit-mode))

doesn't `git-commit-mode` need to be autoloaded as well?

basil-conto:
Good point; `use-package` has spoiled me...

basil-conto:
After taking a closer look at the source I'm no longer convinced this is a good approach. Am I correct in understanding that calling `git-commit-mode` directly skips calling `git-commit-setup`? If so, that would explain why I had to add `with-editor` hooks to kill `git-commit` buffers on cancel.

In other words, I would still like autoloaded access to `git-commit-filename-regexp` so that I can defer loading `git-commit.el` one way or another, but `git-commit-mode` does not seem like a valid entry point to the package.

basil-conto:
If anything else were to be autoloaded, it would have to be `git-commit-setup`, it being the main worker, right?

I have tried adding a function which calls `global-git-commit-mode` to either `find-file-hook` or `auto-mode-alist` before removing itself from these, but this doesn't seem to work reliably. In the `find-file-hook` case I have to also run `git-commit-setup` because it's already late in the `after-find-file` sequence. In both cases `find-file-hook` doesn't end up containing `git-commit-setup`.

What seems to work perfectly is adding `git-commit-setup` to `auto-mode-alist`. Do you see anything wrong with this approach? If you are in favour of autoloading `git-commit-setup` I will update this pull request; though it's not as important as the autoload cookie for `git-commit-filename-regexp` because users can autoload functions themselves.

kyleam:
> git-commit-mode does not seem like a valid entry point to the package

No, it doesn't.  (Sorry, I just commented on your snippet from Emacs
StackExchange without giving it much more thought.)

> What seems to work perfectly is adding git-commit-setup to
> auto-mode-alist. Do you see anything wrong with this approach?

Yeah, I think `git-commit-setup` is the entry point you're looking
for.  Should you do something like
`(lambda nil (or (featurep 'git-commit) (git-commit-setup)))`
so that your `auto-mode-alist` entry doesn't run `git-commit-setup`
after `global-git-commit-mode` adds `git-commit-setup-check-buffer` to
`find-file-hook`?

basil-conto:
I do not think that will be necessary. The `auto-mode-alist`-abusing approach achieves the same effect as `global-git-commit-mode`, but better lends itself to autoloading, so there is no need to enable the latter. I feel a bit dirty doing this, but faster init times are good for my heart. :)

kyleam:
OK, so you're also explicilty disabling `global-git-commit-mode` then?

kyleam:
In any case, I see no issue with adding the autoload cookie to
`git-commit-filename-regexp` and `git-commit-setup` if it's helpful.

basil-conto:
> OK, so you're also explicilty disabling `global-git-commit-mode` then?

I realise that the autoloaded mode variable defaults to `t`, but this does not actually have any effect until its corresponding mode function is called, right? Or is that the case when using `custom-set-variables` (which I do not)?

> I see no issue with adding autoload cookies to `git-commit-filename-regexp` and `git-commit-setup`

I'll force-push this PR with the latter then.

kyleam:
> but this does not actually have any effect until its corresponding mode function is called, right?

I think it has an effect without a call to the mode function, through
the `initializer` call the ends up happening in `define-minor-mode`.

Without loading Magit, I tested

    (add-to-list 'auto-mode-alist
                 (cons git-commit-filename-regexp #'git-commit-setup))

After visiting a file named COMMIT_EDITMSG, I see
`git-commit-setup-check-buffer` in `find-file-hook`.

basil-conto:
@tarsius Sorry about that omission. I can only promise to follow this process hereafter.

@kyleam Thanks for investigating; I should indeed be disabling the global mode then. Thanks also for merging this and sorry for turning such a small change into a support issue.

tarsius:
No problem. At some point these comments will be imported as git notes.

Issue #3022, opened 2017-03-10
Consider adding imenu support


DamienCassou:
It would be nice to be able to use imenu in magit buffers:

- in magit-log, imenu could let the user go to a commit rapidly by hash or description
- in magit-rev, imenu could let the user go to a file rapidly by filename
- in magit status, imenu could let the user go to a section rapidly

tarsius:
That would be nice, and implementing it might actually help me use imenu more (in other contexts). But it's not a high priority.

tarsius:
Is there anything else you want to add, or can we close this?

DamienCassou:
We could add it to `magit-rebase-mode` too, I forgot it. What do you think?

tarsius:
Feel free to do that if/when you have time to do it. Eventually someone will ask for it, so why not add it sooner rather than later.

kyleam:
> We could add it to git-rebase-mode too

A few other candidates:

  * `magit-refs-mode`

  * `magit-diff-mode` and `magit-stash-mode`

    I think it'd make sense to define a single set of imenu functions
    for diff/revision/stash buffers, setting these functions in
    `magit-diff-mode` rather than `magit-revision-mode`.
    `magit-revision-mode` and `magit-stash-mode` would then inherit
    these from `magit-diff-mode`.

  * `magit-reflog-mode` and `magit-stashes-mode`

     Right now, these inherit the log's imenu functions, which don't
     return useful results.  If the regexp of the log's imenu
     functions aren't easily extended to work with these, I think
     these mdoes should have their own imenu functions or should at
     least set the imenu functions to nil.

PR #3072, opened 2017-05-14
Add support for imenu

DamienCassou:
This is a step forward for issue #3022.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -1756,7 +1756,11 @@ Staging and applying changes is documented in info node
 
 \\{magit-revision-mode-map}"
   :group 'magit-revision
-  (hack-dir-local-variables-non-file-buffer))
+  (hack-dir-local-variables-non-file-buffer)
+  (setq imenu-prev-index-position-function
+        #'magit-revision-imenu-prev-index-position-function)
+  (setq imenu-extract-index-name-function
+        #'magit-revision-imenu-extract-index-name-function))
 
 (defun magit-revision-refresh-buffer (rev __const _args files)
   (setq header-line-format
@@ -2449,6 +2453,24 @@ https://github.com/magit/magit/pull/2293 for more details)."
       (diff-fixup-modifs (point-min) (point-max))
       (setq patch (buffer-string)))
     patch))
+
+;;;; Imenu Support
+
+(defun magit-revision-imenu-prev-index-position-function ()
+  "Move point to previous file line in current buffer.
+This function is used as a value for
+`imenu-prev-index-position-function'."
+  (magit-section--backward-find
+   (lambda ()
+     (and (equal (magit-section-type (magit-current-section)) 'file)
+          (equal (magit-section-type (magit-section-parent (magit-current-section))) 'commitbuf)))))
+
+(defun magit-revision-imenu-extract-index-name-function ()
+  "Return imenu name for line at point.
+This function is used as a value for
+`imenu-extract-index-name-function'.  Point should be at the
+beginning of the line."
+  (buffer-substring-no-properties (line-beginning-position) (line-end-position)))
 
 (provide 'magit-diff)
 ;;; magit-diff.el ends here
--- a/lisp/magit-section.el
+++ b/lisp/magit-section.el
@@ -1113,6 +1113,25 @@ If optional SECTION is nil, use the current section."
                 (magit-section-position-in-heading-p section (region-end))))
        t))
 
+(defun magit-section--backward-protected ()
+  "Move to the beginning of the current or the previous visible section.
+Same as `magit-section-backward' but do not raise error and
+return non-nil if there is a section, nil if not."
+  (condition-case nil
+      (progn
+        (magit-section-backward)
+        t)
+    (user-error nil)))
+
+(defun magit-section--backward-find (predicate)
+  "Move to the first previous section satisfying PREDICATE.
+PREDICATE does not take any parameter and should not move
+point."
+  (let (not-end)
+    (while (and (setq not-end (magit-section--backward-protected))
+                (not (funcall predicate))))
+    not-end))
+
 (defun magit-wash-sequence (function)
   "Repeatedly call FUNCTION until it returns nil or eob is reached.
 FUNCTION has to move point forward or return nil."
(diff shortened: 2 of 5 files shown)


lisp/magit-section.el
```diff
@@ -230,6 +230,27 @@ the beginning of the current section."
                  (forward-line -1)
                  (magit-section-goto (magit-current-section))))))))))
 
+(defun magit-section-backward-protected ()
```
tarsius: Since these two functions are used in all three commits/buffers, please add them in a separate commit. It should mention that this is done for `imenu` support.

lisp/magit-section.el
```diff
@@ -230,6 +230,27 @@ the beginning of the current section."
...
+    (user-error nil)))
+
+(defun magit-section-backward-find (predicate)
+  "Move to the first previous section satisfying PREDICATE.
+PREDICATE does not take any parameter and should not move
+point."
+  (let (not-end)
+    (while (and (setq not-end (magit-section-backward-protected))
+                (not (funcall predicate)))
+      t)
```
tarsius: Is this `t` necessary?

lisp/magit-section.el
```diff
@@ -230,6 +230,27 @@ the beginning of the current section."
...
                  (magit-section-goto (magit-current-section))))))))))
 
+(defun magit-section-backward-protected ()
+  "Move to the beginning of the current or the previous visible section.
+Same as `magit-section-backward' but do not raise error and
+return non-nil if there is a section, nil if not."
+  (interactive)
+  (condition-case nil
+      (progn
+        (magit-section-backward)
```
tarsius: That's actually intended for interactive use. But I think simply binding `magit-section-movement-hook` would do the trick.

lisp/magit-section.el
```diff
@@ -230,6 +230,27 @@ the beginning of the current section."
                  (forward-line -1)
                  (magit-section-goto (magit-current-section))))))))))
 
+(defun magit-section-backward-protected ()
+  "Move to the beginning of the current or the previous visible section.
+Same as `magit-section-backward' but do not raise error and
+return non-nil if there is a section, nil if not."
+  (interactive)
```
tarsius: I don't think this is useful interactively.

lisp/magit-status.el
```diff
@@ -282,6 +285,37 @@ Type \\[magit-commit-popup] to create a commit.
...
+    (while (magit-section-backward-find
+            (lambda ()
+              (let* ((section (magit-current-section))
+                     (type (magit-section-type section))
+                     (parent (magit-section-parent section))
+                     (parent-type (magit-section-type parent)))
+                (and (-contains? entry-types type)
+                     (-contains? menu-types parent-type)))))
+      (let* ((section (magit-current-section))
+             (type (magit-section-type section))
```
tarsius: This isn't used.

lisp/magit-section.el
tarsius: Also once this is removed, both of these two functions are not commands, so it is not really appropriate to place them in the section "Commands > Movement". There are some other non-command functions in this section already, but those are used to implement the commands, while these two functions make use of a command.

So I think it would be better to place them in the "Utilities", right before `magit-wash-sequence`. Also not a perfect location, maybe we can think of something better later on.

I would like to avoid these functions becoming part of a public api, so please use the `magit-section--` prefix.

lisp/magit-diff.el
```diff
@@ -1754,7 +1754,27 @@ Staging and applying changes is documented in info node
...
 \\{magit-revision-mode-map}"
   :group 'magit-revision
-  (hack-dir-local-variables-non-file-buffer))
+  (hack-dir-local-variables-non-file-buffer)
+  (setq imenu-prev-index-position-function
+        #'magit-revision-imenu-prev-index-position-function)
+  (setq imenu-extract-index-name-function
+        #'magit-revision-imenu-extract-index-name-function))
+
+(defun magit-revision-imenu-prev-index-position-function ()
```
tarsius: This, and all the other `*-imenu-*` functions you are adding are not a core part of the respective mode. Since I try to place "details" later in the file and/or relevant section, you should do the same for those functions.

Please move these functions to sections named "Imenu Support" sections at the very end of the respective files. I am also considering whether it wouldn't be better to place all those six functions, and possibly even the `magit-section...` functions you are adding to a new file `magit-imenu.el`.

lisp/magit-status.el
DamienCassou: removed

lisp/magit-diff.el
DamienCassou: > move these functions to sections named "Imenu Support" sections at the very end of the respective files

done

DamienCassou: should I use a `--`-based function name to hide these functions from the public api too?

tarsius:
Please add an entry in the release notes, something like "* Add imenu support in log, revision, and status buffers.  #3022". You can just add that in the last commit.

tarsius:
> I should be able to find some time to lead the work on that now.

That's really something I should do myself, because I think that beside @kyleam and @npostavs, I am the only one who knows about all the issues (including accidental complexity, and "to hard to implement" features) that should be fixed by the new implementation. To my surprise I have not yet opened an issue that at least summarizes some of my thoughts on the matter. I might do that in a few days.

Also expect some news about my plans for Magit in about two weeks. Until then I don't really want to discuss this. (Also 2016 and the first half of 2017 have been slow, but I hope this will change soon.)

tarsius:
@kyleam I am probably gonna merge this after the release note entry has been added. Any objections?

kyleam:
> Any objections?

Haven't been following too closely, but looks good to me.  Thanks for
working on this, @DamienCassou!

DamienCassou:
> Please add an entry in the release notes

done

> That's really something I should do myself

good for me. I would love to be able to use this in other packages.

lisp/magit-section.el
tarsius: > can you please help me understand how binding `binding magit-section-movement-hook` would help me get rid of the `condition-case`?

It doesn't. But it avoids things from happening that are only supposed to be happening when interactively moving to a section.

tarsius: I have amended to the commit.

tarsius:
Merged with a small change to the first commit.

Thanks!

PR #3074, opened 2017-05-15
magit-use-sticky-arguments: new option

kyleam:
[resurrection and reroll of #2146]


tarsius:
Could you please split this into two commits, the first one only adding `magit-*-get-buffer-args` in order to remove code duplication but without changing the behavior. Also please put `(nth {1,2} magit-refresh-args)` on separate lines.

kyleam:
> Could you please split this into two commits [...]

Yes, that's better.  Updated.

Issue #3054, opened 2017-04-26
How to show the log of selected region?


stardiviner:
Is there a way to show the log of the selected region related commits log?

tarsius:
Yes, there definitely has to be a way to do this easily.

RenWenshan:
@stardiviner with a region selected, pressing `M-x magit-log-buffer-file`, this will show the region related commits only.

stardiviner:
Thanks

PR #3075, opened 2017-05-15
magit-log-buffer-file: fix `-L` arg when region is active

jguenther:
Previously, the line numbers passed to `git log`'s `-L` arg were calculated by subtracting 1 from
`line-number-at-pos` for `region-beginning` and `region-end`, which was incorrect in some cases.

e.g. calling `mark-whole-buffer` in a file with a single line with a trailing newline would call
`git log -L0,1:file`, instead of the expected `git log -L1,1:file`.

Additionally, if the active region does not include a trailing newline, the end of the range passed
to `-L` would also be off by 1. (e.g. in a 1-line file with no trailing newline, it would call `git
log -L0,0:file`)

This fix changes `magit-log-buffer-file` to calculate `-L` as a range starting at `the line number
at the start of the region`, and ending at `the line number at start + total lines in the region -
1`.

This appears to fix the problem, at least in all cases I've tried.

I can add regression tests if necessary--I wanted to make sure this is an OK approach first.

I'd also love if `magit-log-refresh-buffer` would update the header line with the line range passed to `-L`, similar to how limiting the log to a list of files updates the header to ` touching <files>`. I think it should work OK to check `args` for `(string-match-p "^-L")`, parse the value if present, and update the header appropriately, but is there a better way?

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -634,11 +634,17 @@ completion candidates."
   "Show log for the blob or file visited in the current buffer.
 With a prefix argument or when `--follow' is part of
 `magit-log-arguments', then follow renames."
-  (interactive (if (region-active-p)
-                   (list current-prefix-arg
-                         (1- (line-number-at-pos (region-beginning)))
-                         (1- (line-number-at-pos (region-end))))
-                 (list current-prefix-arg)))
+  (interactive
+   (cons current-prefix-arg
+         (and (region-active-p)
+              (list (line-number-at-pos (region-beginning))
+                    (line-number-at-pos
+                     (let ((end (region-end)))
+                       (if (char-after end)
+                           end
+                         ;; Ensure that we don't get the line number
+                         ;; of a trailing newline.
+                         (1- end))))))))
   (require 'magit)
   (-if-let (file (magit-file-relative-name))
       (magit-mode-setup-internal
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -637,14 +637,16 @@ With a prefix argument or when `--follow' is part of
   (interactive
    (cons current-prefix-arg
          (and (region-active-p)
-              (list (line-number-at-pos (region-beginning))
-                    (line-number-at-pos
-                     (let ((end (region-end)))
-                       (if (char-after end)
-                           end
-                         ;; Ensure that we don't get the line number
-                         ;; of a trailing newline.
-                         (1- end))))))))
+              (save-restriction
+                (widen)
+                (list (line-number-at-pos (region-beginning))
+                      (line-number-at-pos
+                       (let ((end (region-end)))
+                         (if (char-after end)
+                             end
+                           ;; Ensure that we don't get the line number
+                           ;; of a trailing newline.
+                           (1- end)))))))))
   (require 'magit)
   (-if-let (file (magit-file-relative-name))
       (magit-mode-setup-internal
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -806,7 +806,10 @@ Type \\[magit-reset] to reset `HEAD' to the commit at point.
         (propertize
          (concat " Commits in " (mapconcat 'identity revs  " ")
                  (and files (concat " touching "
-                                    (mapconcat 'identity files " "))))
+                                    (mapconcat 'identity files " ")))
+                 (--some (and (string-prefix-p "-L" it)
+                              (concat " " it))
+                         args))
          'face 'magit-header-line))
   (unless (= (length files) 1)
     (setq args (remove "--follow" args)))
--- a/Documentation/RelNotes/2.11.0.txt
+++ b/Documentation/RelNotes/2.11.0.txt
@@ -70,6 +70,9 @@ Changes since v2.10.3
 * The commands `magit-subtree-add', `magit-subtree-pull', and
   `magit-subtree-push' now offer more completion candidates.  #3081
 
+* The log buffer now shows the line range restriction (i.e. `-L'
+  argument) in its header.  #3075
+
 Fixes since v2.10.3
 -------------------
 
--- a/Documentation/RelNotes/2.11.0.txt
+++ b/Documentation/RelNotes/2.11.0.txt
@@ -101,6 +101,9 @@ Fixes since v2.10.3
 * The command `magit-show-commit' mistakenly displayed a tag's commit
   when point was on a branch with the same name.  #3098
 
+* The command `magit-log-buffer-file' miscalculated line numbers from
+  the region.  #3075
+
 This release also contains the fixes described in the v2.10.{1,2,3}
 release notes, as well as other minor improvements, bug fixes, typo
 fixes, and documentation fixes.


lisp/magit-log.el
```diff
@@ -633,9 +633,10 @@ completion candidates."
 With a prefix argument or when `--follow' is part of
 `magit-log-arguments', then follow renames."
   (interactive (if (region-active-p)
-                   (list current-prefix-arg
-                         (1- (line-number-at-pos (region-beginning)))
-                         (1- (line-number-at-pos (region-end))))
+                   (let* ((first-line (line-number-at-pos (region-beginning)))
+                          (total-lines (count-lines (region-beginning) (region-end)))
+                          (last-line (1- (+ first-line total-lines))))
+                     (list current-prefix-arg first-line last-line))
```
kyleam: Or

```emacs-lisp
(list current-prefix-arg
      (line-number-at-pos (region-beginning))
      (line-number-at-pos (region-end)))
```

?

kyleam: In addition to any off-by-one errors, the line numbers will be
incorrect if a buffer is narrowed.  It'd be good if your proposed
change could address this issue as well.

jguenther: That doesn't work correctly if the region contains a trailing newline (e.g. results in `-L1,2:file` if `file` is a single line + trailing newline)

jguenther: Ah good point. I will take a look at that

jguenther:
@kyleam I believe I've addressed your comments. It should work with narrowed buffers now and I cleaned up the commit msg.

Any thoughts on the last paragraph of the PR description? I might take a look at that tonight if that seems reasonable.

lisp/magit-log.el
```diff
@@ -633,9 +633,12 @@ completion candidates."
 With a prefix argument or when `--follow' is part of
 `magit-log-arguments', then follow renames."
   (interactive (if (region-active-p)
-                   (list current-prefix-arg
-                         (1- (line-number-at-pos (region-beginning)))
-                         (1- (line-number-at-pos (region-end))))
+                   (save-restriction
+                     (widen)
```
kyleam: Thanks for accounting for narrowed buffers.  Please mention this fix
in the commit message, or even better, place it as a separate commit
before the boundary calculation commit.

lisp/magit-log.el
```diff
@@ -633,9 +633,12 @@ completion candidates."
...
+                     (let* ((first-line (line-number-at-pos (region-beginning)))
+                            (total-lines (count-lines (region-beginning) (region-end)))
+                            (last-line (1- (+ first-line total-lines))))
+                       (list current-prefix-arg first-line last-line)))
```
kyleam: I think this changes the intended behavior of the region selection.
If point is at the beginning of a non-empty line, it will no longer be
included.

How about this?

```diff
diff --git a/lisp/magit-log.el b/lisp/magit-log.el
index f8462a0d..4f235be8 100644
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -635,10 +635,15 @@ (defun magit-log-buffer-file (&optional follow beg end)
   (interactive (if (region-active-p)
                    (save-restriction
                      (widen)
-                     (let* ((first-line (line-number-at-pos (region-beginning)))
-                            (total-lines (count-lines (region-beginning) (region-end)))
-                            (last-line (1- (+ first-line total-lines))))
-                       (list current-prefix-arg first-line last-line)))
+                     (list current-prefix-arg
+                           (line-number-at-pos (region-beginning))
+                           (line-number-at-pos
+                            (let ((end (region-end)))
+                              (if (char-after end)
+                                  end
+                                ;; We're at the end of the buffer and
+                                ;; may be on a trailing newline.
+                                (1- end))))))
                  (list current-prefix-arg)))
   (require 'magit)
   (-if-let (file (magit-file-relative-name))

```

kyleam:
> Any thoughts on the last paragraph of the PR description? I might
> take a look at that tonight if that seems reasonable.

I do think the `-L` argument would be nice to indicate in the header,
but based on f06edb21 (magit-log-{popup,buffer-file}: add `-L'
argument, 2015-09-30), I wonder if @tarsius wants wait until he
revisits this functionality.

tarsius:
Please go ahead adding this information to the header. While I probably meant things like this when I wrote f06edb21d303e22e999c490b4fc2f23d3e1220fc, 2016 has been a slow year and I see no reason to delay this. This change is also fairly independent of the more complicated part, which I would still like to do in one go - essentially #3054.

lisp/magit-log.el
jguenther: What do you mean? I can't replicate that problem.

Do you mean a situation like this?
image

If so, point is on line 2, but the selected region is from the beginning of line 1 up to and including the trailing newline of line 1. If I set the mark at the bol of line 1, and then `next-line`, I wouldn't expect to include line 2 in the call to `git log`, no?

kyleam: > What do you mean? I can't replicate that problem.

Given what you say next, it seems that you _can_ replicate it; you
just don't agree that it's a problem :)

Based on the code added in f06edb21, I'm pretty sure that the
intention was to say "if a point or mark is on a line, include the
line".  If a commit is aimed at correcting the off-by-one boundary
calculation, it shouldn't change this behavior (or, at the very least,
the commit message should explicitly discuss this change in behavior).

But I don't find the "any line touched" behavior unreasonable.  A
region has character resolution and the `-L` argument has line
resolution, so you have to define some point where a line is included.
This is the same issue that Magit deals with when a region is used to
select commits or lines of a hunk, and it does so by _extending_ the
effective region to the end of the line.

In Magit buffers, we make this more obvious by highlighting the
effective region.  We can't use the same approach here because we're
in a non-Magit buffer and we don't know what command will be called at
the time the user is defining the region.  We could flash the
effective region once we're in magit-log-buffer-file's interactive
form, but I'm not sure it's worth the effort.

jguenther: > But I don't find the "any line touched" behavior unreasonable. A
region has character resolution and the -L argument has line
resolution, so you have to define some point where a line is included.
This is the same issue that Magit deals with when a region is used to
select commits or lines of a hunk, and it does so by extending the
effective region to the end of the line.

That's a good point. I had thought that e.g. `mark-whole-buffer` or `mark-defun` would include the trailing newline but it actually doesn't.

> We could flash the effective region once we're in magit-log-buffer-file's interactive form, but I'm not sure it's worth the effort.

I think a good compromise would be to show the effective region in the `magit-log-header`. Flashing the effective region would be really nifty though. 

I'll take another look at this branch tonight. Thanks for the feedback!

jguenther: @kyleam Could you take another look when you get a chance? I made your suggested change and also updated the `magit-log` header.

I tested with the edge cases I could think of and it seems to work as expected, and the `-L` header info seems to be OK with both `magit-log-buffer-file` as well as `magit-log-popup` (including if `=F` and `=L` were both specified).

kyleam: > Could you take another look when you get a chance?

Thanks for the update.  I've pushed a few fixup commits (tip at
9ff31c46c).  See the commit message bodies for more information.

> including if =F and =L were both specified

I'm guessing you meant `=f`.  As I discuss in 65cfc1069, pathspecs and
-L are actually incompatible.

Please let me know if you have any issues with my suggested changes.

jguenther:
@kyleam your fixup commits make the magit-log header not show the `-L` arg if the `=f` *and* `=L` args are used in the magit-log popup.

Are your changes because I didn't use your preferred style, or did it not work correctly?

kyleam:
> Are your changes because I didn't use your preferred style, or did it not work correctly?

I'm confused by your question.  I've already mentioned why I think
considering the pathspec and the line range argument together is
incorrect in my [last comment] to you and in the commit message of
65cfc1069 (first bullet and footnote).

[last comment]: https://github.com/magit/magit/pull/3075#discussion_r119989879

jguenther:
@kyleam ahh. My apologies, somehow I missed that comment.

kyleam:
> FYI this works just fine for me on git 2.13

Interesting.  Can you provide more details?  Here is what I see with
2.13.

With a pathspec but not file given to -L:

    $ git log -L45,50 -- lisp/magit-bisect.el
    fatal: -L argument not 'start,end:file' or ':funcname:file': 45,50


If I path a pathspec that matches the file given to -L,

    $ git log -L45,50:lisp/magit-bisect.el -- lisp/magit-bisect.el


the output is the same as with `git log -L45,50:lisp/magit-bisect.el`,
so the pathspec is useless but doesn't hurt.

With any other file for the pathspec, such as

    $ git log -L45,50:lisp/magit-bisect.el -- lisp/magit-imenu.el

the output is empty.


Since -L was add in Git's 12da1d1f6 (Implement line-history search
(git log -L), 2013-03-28), its documentation has said "You may not
give any pathspec limiters".

kyleam:
> I will clean this PR up when I get a chance

If "clean" just refers to squashing down the fixup commits, I can take
care of that when merging.

kyleam:
I've rebased the branch, squashing in the fixup commits.  I'll merge
tomorrow if there are no further comments.

jguenther:
> If I path a pathspec that matches the file given to -L,

@kyleam that's the situation I was talking about. `git log -Lx,y:file -- file` gives the same output as `git log -Lx,y:file` both from the command line as well as in magit, which is something that does come up in my workflow. e.g. I look at the log for a file, then I try narrowing it down by using `-L`.

The problem is that currently this *does* show the expected log contents for the line range specified, but the log header doesn't show the `-L` arg (only the pathspec arg).

So either the pathspec arg should be removed/ignored if `-L` is specified, or we just pass along the requested args to git and show the results, and ensure that the header shows those args so it's obvious if the log is unexpectedly empty. I'd prefer the latter.

> With any other file for the pathspec, such as `$ git log -L45,50:lisp/magit-bisect.el -- lisp/magit-imenu.el` the output is empty.

This is similar to what happens if I use a pathspec of a file that has never existed in the repo: I get an empty log and not an error from magit, and I can see the pathspec in the header so I know why the log is empty. IMO this should also be the case if I specify both a pathspec and an `-L` arg for a different file:  it should give the same output as if I'd done it from the command line.

Thoughts?

> If "clean" just refers to squashing down the fixup commits, I can take care of that when merging.

:+1:

kyleam:
> that's the situation I was talking about. git log -Lx,y:file -- file gives the same output as git log -Lx,y:file both from the command line as well as in magit, which is something that does come up in my workflow. e.g. I look at the log for a file, then I try narrowing it down by using -L.

OK, thanks for clarifying.  Because Git explicitly documents -L as
incompatible with a pathspec, I think this behavior is unspecified and
perhaps unintentional.

> The problem is that currently in this PR this does show the expected log contents for the line range specified, but the log header doesn't show the -L arg (only the pathspec arg).

OK, fair enough.  It's fine for me to say "well, they're incompatible,
so you shouldn't be passing both", but since it does work if the
pathspec and -L file match, we should display -L.  We could move the
-L check above the `files` check, but then we're still not being too
helpful when a blank log shows for the mismatching -L/pathspec case.

> So either the pathspec arg should be removed/ignored if -L is specified, or we just pass along the requested args to git and show the results, and ensure that the header shows those args so it's obvious if the log is unexpectedly empty. I'd prefer the latter.

I was originally thinking we should signal an error, but because the
matching -L/pathspec case works, I suppose we probably shouldn't be
stricter than Git itself is.

> and I can see the pathspec in the header so I know why the log is empty.

I don't believe your original implementation (snippet below) made
things much clearer.  If -L and a pathspec was given, it would display
only the pathspec file, so it wasn't obvious that there was a
mismatch.

```emacs-lisp
(concat " Commits in " (mapconcat 'identity revs " ")
        (and files (concat " touching "
                           (mapconcat 'identity files " ")))
        (--when-let
            (--first (string-match "^-L\\([0-9]+\\),\\([0-9]+\\):\\(.+\\)"
                                   it)
                     args)
          ;; Show "trace line evolution" (-L) value in header
          (concat (if (not files)
                      (concat " touching " (match-string 3 it)))
                  (format ":%s-%s"
                          (match-string 1 it)
                          (match-string 2 it)))))
```

I've pushed an update that just independently shows the files and the
*unprocessed* -L argument.  I don't think there's much gain in being
clever with the formatting.

Thanks for the pushback.

jguenther:
> I don't believe your original implementation (snippet below) made
things much clearer. If -L and a pathspec was given, it would display
only the pathspec file, so it wasn't obvious that there was a
mismatch.

Agreed.

> I've pushed an update that just independently shows the files and the
unprocessed -L argument. I don't think there's much gain in being
clever with the formatting.

I will take a look, thanks!

SteveJobzniak:
I wonder if this bug is also why "Magit stage selected lines" (selecting a range of lines (rather than hunks or files) and then staging just those lines) often produced corrupt commits with diffs that literally broke the target-files.

I began using Magit in March, and did lots of work in March. But when I noticed that Magit produced corrupt commits when staging only certain lines, I stopped using that method.

Does anyone know if this pull request fixed that bug? Or perhaps I was the only person staging line-by-line and nobody else knows about the bug yet. ;-)

npostavs:
> I wonder if this bug is also why "Magit stage selected lines" (selecting a range of lines (rather than hunks or files) and then staging just those lines) often produced corrupt commits with diffs that literally broke the target-files.

Hmm, I know that sometimes line staging doesn't work, e.g., if you need to stage 2 non-consecutive lines from a single hunk, the second probably won't apply. But I've never seen nor heard of it causing corruption. I suggest opening a new issue with all the details.

kyleam:
> Does anyone know if this pull request fixed that bug?

The merged PR fixes the line calculation for the
`magit-log-buffer-file` command.  I don't understand why you think
that's related to the issue you're describing.

SteveJobzniak:
@npostavs @kyleam I just searched all release notes for all versions that came out since dec 2016. For words like "region", "file", "lines", "stage", "hunk", etc.

It's not mentioned there, so I guess the bug remains then. I was sick so I didn't have energy to open a ticket about it.

Anyway, the bug is simple: Select a range of lines (within a hunk) and press `s` to stage them. Do that a few times for different lines inside the file at various places, and then commit. The resulting git commit will sometimes be corrupt and will destroy the target-file once committed...

Should I open a ticket?

npostavs:
> Should I open a ticket?

Yes please, at worst we'll mark it "needinfo" if you can't identify the exact conditions required. Spacemacs & evil-mode should be (hopefully) irrelevant, but I suspect the file+diff being staged may well be important. If you could post those it would be very helpful.

SteveJobzniak:
Alright, here is a ticket #3182! 👍 Sorry to have asked in this unrelated PR. I thought the line-offset bug was related.

PR #3076, opened 2017-05-17
magit-abort-dwim: New function to abort current operation

DamienCassou:


tarsius:
Please

* Add a release note entry.
* No newline after `cond`, I don't expect another case to appear before the current first case, so avoiding a slightly less trivial future diff, is a lost cause.
* Align the condition BODYs. That's just how I like it in such trivial cases.

tarsius:
* Mention the "operations" this may abort in the doc-string, in order. (How sure are you this is the optimal order?)
* Mention this in the manual.
  * You just have to edit `magit.org`, I can regenerate `magit.texi` for you.
  * Do so probably at the bottom of "Merging"...
  * ... so you should also place it after `magit-merge-abort` in the source.

kyleam:
The byte-compiler isn't happy:

```
In end of data:
magit.el:1085:1:Warning: the following functions are not known to be defined:
    magit-rebase-in-progress-p, magit-rebase-abort, magit-am-in-progress-p,
    magit-am-abort, magit-sequencer-in-progress-p, magit-sequencer-abort,
    magit-bisect-in-progress-p, magit-bisect-reset
```


Perhaps this should be added it magit-extras.el instead?

tarsius:
> Perhaps this should be added to magit-extras.el instead?

Yes, I think that would be best. Probably in a new section "Miscellaneous" at the bottom.

DamienCassou:
I would first like to thank you for the speed at which you review pull requests. I appreciate it a lot because it means I can focus on a task for a few days and then forget it forever. Thanks.

> Add a release note entry

done

> Remove the newline after `cond`

done

> Align the condition BODYs

done

> Mention the "operations" this may abort in the doc-string

done

> How sure are you this is the optimal order?

I have no clue. What kind of ordering conflict/problem do you have in mind? Every `*-in-progress-p` function checks for the existence of a file. Is it possible that 2 files exist at the same time?

> Mention this in the manual

I didn't yet touch the manual as it covers very few of the `magit-extras` functions. Also, I didn't find a valid place to add to now that you changed your mind regarding location of the function (from _"after `magit-merge-abort`"_ to _"end of `magit-extras`"_). Please advise.

> Perhaps this should be added to magit-extras.el instead?

done

DamienCassou:
What do you think regarding binding a key to this function? Any suggestion?

tarsius:
>> How sure are you this is the optimal order?

> I have no clue. What kind of ordering conflict/problem do you have in mind? Every `*-in-progress-p` function checks for the existence of a file. Is it possible that 2 files exist at the same time?

Rebase uses cherry-pick internally, I believe. So the rebase case should come before the cherry-pick case, which it does. Also I am not sure this is really required, but better play it save.

Please just start each operation once and the abort using this command. If that works, which I expect it does, then that's good enough for me.

>> Mention this in the manual

> I didn't yet touch the manual as it covers very few of the `magit-extras` functions. Also, I didn't find a valid place to add to now that you changed your mind regarding location of the function (from _"after `magit-merge-abort`"_ to _"end of `magit-extras`"_). Please advise.

Okay, let's leave it unmentioned for now.

(We should probably add "Miscellaneous > Assorted Commands" at some point, and then mention all those "extra" commands in there.)

> What do you think regarding binding a key to this function? Any suggestion?

Can't really think of anything appropriate. Also since I don't really expect all that many users to use this command, I don't really want to use up any of the few remaining "good" ones. So I would suggest that we leave it for users who do want to use it, to find a key that suits them.

tarsius:
I've cherry-picked this. Thanks!

PR #3080, opened 2017-05-18
magit-clone: improve suggested directory name

### LINKED ISSUE #3079: Improve target directory name guessing in magit-clone
The `magit-clone` command could do a better job when name of the repo to clone contains dots. For example, if I invoke `magit-clone` with `https://github.com/mrkkrp/mrkkrp.github.io.git`, it thinks that the directory where to clone the repo should be called `io`, while the correct behavior would be to infer `mrkkrp.github.io` and propose that.

kyleam:
```
Extract the directory name by taking the tail of the url up to the
first slash or colon, not the first slash or period.

This makes magit-clone's default directory name match the behavior of
"git clone <url>" in two places where it did not:

  1. https://[...]/foo.bar.git => "foo.bar" (instead of "bar")

  2. host.xz:foo/.git => "foo" (instead of "git")

Fixes #3079.
```

### DIFF
--- a/Documentation/RelNotes/2.11.0.txt
+++ b/Documentation/RelNotes/2.11.0.txt
@@ -64,6 +64,9 @@ Changes since v2.10.3
   Depending on the context, this will abort a merge, a rebase, a
   patch application, a cherry-pick, a revert, or a bisect.  #3017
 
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks!

PR #3081, opened 2017-05-20
Use completion in magit-subtree.el

justbur:
* lisp/magit-subtree.el (magit-subtree-add): Complete repo and branch
(magit-subtree-pull): Complete repo and branch
(magit-subtree-push): Complete repo

This could probably be taken farther, but this was as far as I was comfortable with.

### DIFF
--- a/lisp/magit-subtree.el
+++ b/lisp/magit-subtree.el
@@ -90,10 +90,12 @@
 ;;;###autoload
 (defun magit-subtree-add (prefix repository ref args)
   "Add REF from REPOSITORY as a new subtree at PREFIX."
-  (interactive (list (magit-subtree-prefix "Add subtree")
-                     (magit-read-string-ns "Repository")
-                     (magit-read-string-ns "Ref")
-                     (magit-subtree-args)))
+  (interactive
+   (cons (magit-subtree-prefix "Add subtree")
+         (let ((remote (magit-read-remote-or-url "From repository")))
+           (list remote
+                 (magit-read-refspec "Ref" remote)
+                 (magit-subtree-args)))))
   (magit-git-subtree "add" prefix args repository ref))
 
 ;;;###autoload
@@ -115,17 +117,19 @@
 ;;;###autoload
 (defun magit-subtree-pull (prefix repository ref args)
   "Pull REF from REPOSITORY into the PREFIX subtree."
-  (interactive (list (magit-subtree-prefix "Pull into subtree")
-                     (magit-read-string-ns "From repository")
-                     (magit-read-string-ns "Ref")
-                     (magit-subtree-args)))
+  (interactive
+   (cons (magit-subtree-prefix "Pull into subtree")
+         (let ((remote (magit-read-remote-or-url "From repository")))
+           (list remote
+                 (magit-read-refspec "Ref" remote)
+                 (magit-subtree-args)))))
   (magit-git-subtree "pull" prefix args repository ref))
 
 ;;;###autoload
 (defun magit-subtree-push (prefix repository ref args)
   "Extract the history of the subtree PREFIX and push it to REF on REPOSITORY."
   (interactive (list (magit-subtree-prefix "Push subtree")
-                     (magit-read-string-ns "To repository")
+                     (magit-read-remote-or-url "To repository")
                      (magit-read-string-ns "To reference")
                      (magit-subtree-args)))
   (magit-git-subtree "push" prefix args repository ref))
--- a/lisp/magit-subtree.el
+++ b/lisp/magit-subtree.el
@@ -88,13 +88,13 @@
   (magit-run-git-async "subtree" subcmd (concat "--prefix=" prefix) args))
 
 ;;;###autoload
-(defun magit-subtree-add (prefix repository commit args)
-  "Add COMMIT from REPOSITORY as a new subtree at PREFIX."
+(defun magit-subtree-add (prefix repository ref args)
+  "Add REF from REPOSITORY as a new subtree at PREFIX."
   (interactive (list (magit-subtree-prefix "Add subtree")
                      (magit-read-string-ns "Repository")
-                     (magit-read-string-ns "Commit")
+                     (magit-read-string-ns "Ref")
                      (magit-subtree-args)))
-  (magit-git-subtree "add" prefix args repository commit))
+  (magit-git-subtree "add" prefix args repository ref))
 
 ;;;###autoload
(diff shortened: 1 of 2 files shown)


lisp/magit-subtree.el
```diff
@@ -90,10 +90,12 @@
...
   "Add COMMIT from REPOSITORY as a new subtree at PREFIX."
-  (interactive (list (magit-subtree-prefix "Add subtree")
-                     (magit-read-string-ns "Repository")
-                     (magit-read-string-ns "Commit")
-                     (magit-subtree-args)))
+  (interactive (let ((prefix (magit-subtree-prefix "Add subtree"))
+                     (remote (magit-read-remote-or-url "From Repository")))
+                 (list prefix
+                       remote
+                       (magit-read-remote-branch "Commit" remote)
```
kyleam: Two thoughts:

  * `magit-read-remote-branch` lists local refs in
    `refs/remotes/<remote>`.  If a url is given for the remote, this
    works ok because `magit-list-remote-branch-names` returns nil
    (i.e., it doesn't error if `<remote>` doesn't exist).

    Because this is working with local refs, this list isn't
    necessarily an up-to-date representation of what is available on
    the remote.

    It might be better to use `magit-read-refspec` here, which will
    reach out to a remote (and work for urls too).  Perhaps having an
    up-to-date list isn't worth the extra time it takes to connect,
    but I think this consideration should at least be mentioned in the
    commit message.

  * I don't think "Commit" is a good prompt, though I know you are
    just inheriting the decision to use "commit" as the argument.

    I'd be in favor of a patch (ideally preceding this one) that
    changes the argument to "ref".  This is consistent with the name
    used in "man git-subtree", and, more importantly, is a better
    representation of what `git subtree <repo> <arg>` expects for
    `<arg>`.  This form in general works with a ref and will fail when
    given a hash unless that boject happens to already be in the local
    repository.

lisp/magit-subtree.el
```diff
@@ -90,10 +90,12 @@
 ;;;###autoload
 (defun magit-subtree-add (prefix repository commit args)
   "Add COMMIT from REPOSITORY as a new subtree at PREFIX."
-  (interactive (list (magit-subtree-prefix "Add subtree")
-                     (magit-read-string-ns "Repository")
-                     (magit-read-string-ns "Commit")
-                     (magit-subtree-args)))
+  (interactive (let ((prefix (magit-subtree-prefix "Add subtree"))
```
justbur: I wanted to preserve the same order for the prompts. Prefix then repo then commit.

lisp/magit-subtree.el
justbur: I thought about `magit-read-refspec`. I'm in favor of that too. I also don't think "Commit" is a great prompt. 

In general, I was trying to push the prompts in this file in the direction of being more helpful. I think they can be made better even. I just wasn't sure if there was any logic to how they were currently written that I wasn't understanding.

justbur:
Hi @kyleam, thanks for the comments. I change the commit message when the other details get worked out.

lisp/magit-subtree.el
npostavs: > I wanted to preserve the same order for the prompts.

If you're thinking that the evaluation order of function parameters is unspecified like in Scheme; Elisp is not scheme: they're evaluated from left to right. (If that's not what you meant, I can't make any sense of your comment)

justbur: @npostavs maybe you're thinking of a better way to do this that I don't see, but I was referring to the fact that if I just did this
```elisp
(let ((remote (magit-read-remote-or-url "From Repository")))
  (list (magit-subtree-prefix "Add subtree")
        remote
        (magit-read-remote-branch "Commit" remote)))
```
Then afterthis change the order of the prompts for the user would change from Prefix -> Repo -> Commit to Repo -> Prefix -> Commit

justbur:
@kyleam I just switched to `magit-read-refspec`, changed the prompt and reworded the commit message.

lisp/magit-subtree.el
npostavs: Ah, the Github display confused me, I didn't notice the added passing of `remote` to `magit-read-refspec`. Let-binding as you have done makes sense.

kyleam: > I wanted to preserve the same order for the prompts.

Doh, makes sense.

lisp/magit-subtree.el
```diff
@@ -90,10 +90,12 @@
...
+                     (remote (magit-read-remote-or-url "From Repository")))
+                 (list prefix
+                       remote
+                       (magit-read-refspec "Refspec" remote)
```
kyleam: Despite this function's name, I'd prefer the prompt be "Ref" because
"ref" is what is used in git-subtree's manpage and because I don't
think a `<src>:<dest>` refspec mapping makes sense in the context of
`git subtree`.

justbur: No problem. Just made the change

kyleam:
The most recent update (8780f35f) looks good to me.  Thanks.

It'd be good to mention this in the release notes, but I can add that.

justbur:
Thanks again

tarsius:
Personally I would avoid the let binding using

```lisp
  (interactive
   (cons (magit-subtree-prefix "Pull into subtree")
         (let ((remote (magit-read-remote-or-url "From repository")))
           (list remote
                 (magit-read-refspec "Ref" remote)
                 (magit-subtree-args)))))
```

That makes it more obvious that while `remote` is used more than once, the value read with `magit-subtree-prefix` is not.

justbur:
How about? 
```elisp
(interactive
 (let (remote)
   (list (magit-subtree-prefix "Pull into subtree")
         (setq remote (magit-read-remote-or-url "From repository"))
         (magit-read-refspec "Ref" remote)
         (magit-subtree-args))))
```
Saves 18 chars and keeps the arguments together visually. :-) 

Honestly, I don't really care. I'll write it how you want me to.

kyleam:
I'm mostly indifferent with regard to the current form and @tarsius's
suggestion.  I vote against the last suggestion :)

Let's go with @tarius's cons form.

@tarsius  Are you OK with the PR otherwise?

tarsius:
> Honestly, I don't really care. I'll write it how you want me to.

All three variants are reasonable. But I too think the last variant is the least desirable.

tarsius:
> Are you OK with the PR otherwise?

Yes.

justbur:
Updated

PR #3082, opened 2017-05-21
magit-diff-visit-file: Make sure point movement affects correct window

jbms:
The existing implementation does not seem to properly move the point
if the buffer is displayed in another window.

By explicitly selecting the window returned by
magit-display-file-buffer, we ensure the correct point is moved.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -1101,28 +1101,30 @@ the buffer in another window."
                    (or (get-file-buffer file)
                        (find-file-noselect file)))))
       (magit-display-file-buffer buf)
-      (with-current-buffer buf
-        (when line
-          (setq line
-                (cond ((eq rev 'staged)
-                       (apply 'magit-diff-visit--offset file nil line))
-                      ((and force-worktree
-                            (stringp rev))
-                       (apply 'magit-diff-visit--offset file rev line))
-                      (t
-                       (apply '+ line))))
-          (let ((pos (save-restriction
-                       (widen)
-                       (goto-char (point-min))
-                       (forward-line (1- line))
-                       (move-to-column col)
-                       (point))))
-            (unless (<= (point-min) pos (point-max))
-              (widen)
-              (goto-char pos))))
-        (when (magit-anything-unmerged-p file)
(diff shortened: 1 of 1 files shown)


lisp/magit-diff.el
```diff
@@ -1101,28 +1101,30 @@ the buffer in another window."
...
-                       (forward-line (1- line))
-                       (move-to-column col)
-                       (point))))
-            (unless (<= (point-min) pos (point-max))
-              (widen)
-              (goto-char pos))))
-        (when (magit-anything-unmerged-p file)
-          (smerge-start-session))
-        (run-hooks 'magit-diff-visit-file-hook)))))
+      (let ((win (magit-display-file-buffer buf)))
```
kyleam: Did you forget to remove the other `magit-display-file-buffer` call
above?

Anyway, `magit-display-file-buffer-function` is not documented as
returning a window---and the two built-in ones do not---so this will
lead to an error for most users.  (I'm assuming you tested this with
`magit-display-file-buffer-function` set to `display-buffer`, which
does return a window.)

I think something along the lines of

```emacs-lisp
(with-selected-window (or (get-buffer-window buf 'visible)
                          (error "File buffer is not visible"))
  (when line ...))
```

should work.

tarsius:
> The existing implementation does not seem to properly move the point if the buffer is displayed in another window.

I was not able to reproduce this. I tried with the file-visiting buffer being visible in another window of the same frame and also with it being visible in another frame. It is possible that we have to through an `with-selected-window` in there for some edge cases, but first we need reproducible steps so that we can inspect those edge cases.

tarsius:
I have applied the change suggested by @kyleam.

jbms:
I'll clean this up and provide a way to reproduce.  The duplicate call to magit-display-file-buffer was introduced when I copied my changes over in order to prepare the commit.  I do indeed have magit-display-file-buffer-function set to display-buffer.

tarsius:
Thanks, but I just pushed the change suggested by @kyleam a minute before you said that ;-) Please try it out (I already tested with `display-buffer` and `display-buffer-other-frame` though).

PR #3084, opened 2017-05-23
More imenu support

DamienCassou:
* Implement [@kyleam's suggestions](https://github.com/magit/magit/issues/3022#issuecomment-302590319) on current imenu support.
* Also add support in `git-rebase-mode`

### DIFF
--- a/lisp/magit-imenu.el
+++ b/lisp/magit-imenu.el
@@ -195,6 +195,28 @@ beginning of the line."
             (elt entry 0)
             (elt entry (1- (length entry))))))
 
+;;;; Process mode
+
+;;;###autoload
+(defun magit-imenu--process-prev-index-position-function ()
+  "Move point to previous process in magit-process buffer.
+This function is used as a value for
+`imenu-prev-index-position-function'."
+  (magit-section--backward-find
+   (lambda ()
+     (eq
+      (magit-section-type (magit-current-section))
+      'process))))
+
+;;;###autoload
+(defun magit-imenu--process-extract-index-name-function ()
+  "Return imenu name for line at point.
+This function is used as a value for
+`imenu-extract-index-name-function'.  Point should be at the
+beginning of the line."
+  (buffer-substring-no-properties (line-beginning-position)
+                                  (line-end-position)))
+
 ;;;; Rebase mode
 
 ;;;###autoload
--- a/lisp/magit-imenu.el
+++ b/lisp/magit-imenu.el
@@ -163,6 +163,27 @@ This function is used as a value for
 beginning of the line."
   (elt (tabulated-list-get-entry) 0))
 
+;;;; Repolist mode
+
+;;;###autoload
+(defun magit-imenu--repolist-prev-index-position-function ()
+  "Move point to previous line in magit-repolist buffer.
+This function is used as a value for
+`imenu-prev-index-position-function'."
+  (unless (bobp)
+    (forward-line -1)))
+
+;;;###autoload
+(defun magit-imenu--repolist-extract-index-name-function ()
+  "Return imenu name for line at point.
+This function is used as a value for
+`imenu-extract-index-name-function'.  Point should be at the
+beginning of the line."
+  (let ((entry (tabulated-list-get-entry)))
+    (format "%s (%s)"
+            (elt entry 0)
+            (elt entry (1- (length entry))))))
+
 ;;;; Rebase mode
 
 ;;;###autoload
--- a/lisp/magit-imenu.el
+++ b/lisp/magit-imenu.el
@@ -145,6 +145,17 @@ This function is used as a value for
    '(branch commit tag)
    '(local remote tags)))
 
+;;;; Cherry mode
+
+;;;###autoload
+(defun magit-imenu--cherry-create-index-function ()
+  "Return an alist of all imenu entries in current buffer.
+This function is used as a value for
+`imenu-create-index-function'."
+  (magit-imenu--index-function
+   '(commit)
+   '(cherries)))
+
 ;;;; Submodule list mode
 
 ;;;###autoload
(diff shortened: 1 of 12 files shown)


lisp/magit-refs.el
```diff
@@ -573,5 +574,15 @@ line is inserted at all."
...
 
+;;;; Imenu Support
+
+(defun magit-refs-imenu-create-index-function ()
+  "Return an alist of all imenu entries in current buffer.
+This function is used as a value for
+`imenu-create-index-function'."
+  (magit-section--imenu-index-function
+   '(branch commit tag)
+   '(local remote tags)))
```
kyleam: Nice to see how little code was needed for that.

lisp/magit-log.el
```diff
@@ -1532,15 +1532,16 @@ all others with \"-\"."
 This function is used as a value for
 `imenu-prev-index-position-function'."
   (magit-section--backward-find
-   (lambda () (eq (magit-section-type (magit-current-section)) 'commit))))
+   (lambda () (-contains? '(commit stash)
+                     (magit-section-type (magit-current-section))))))
```
kyleam: indentation

lisp/magit-section.el
```diff
@@ -1134,6 +1134,42 @@ point."
                 (not (funcall predicate))))
     found))
 
+(defun magit-section--imenu-index-function (entry-types menu-types)
+  "Return an alist of imenu entries in current buffer.
```
DamienCassou: I'm ok with that. I don't have any opinion. Please confirm you'd prefer me to do it this way.

DamienCassou:
I think I took care of all your feedback @kyleam.

lisp/magit-log.el
```diff
@@ -1532,15 +1532,17 @@ all others with \"-\"."
 This function is used as a value for
 `imenu-prev-index-position-function'."
   (magit-section--backward-find
-   (lambda () (eq (magit-section-type (magit-current-section)) 'commit))))
+   (lambda () (-contains?
+          '(commit stash)
+          (magit-section-type (magit-current-section))))))
```
kyleam: Sorry for not giving more info with my "indentation" [comment].  I
just meant that, at least for me, Emacs would indent your original
change as

```emacs-lisp
(lambda () (-contains? '(commit stash)
                       (magit-section-type (magit-current-section))))

```

not

```emacs-lisp
   (lambda () (-contains? '(commit stash)
                     (magit-section-type (magit-current-section))))))

```

[comment]: https://github.com/magit/magit/pull/3084#discussion_r118134116

lisp/magit-section.el
kyleam: I lean towards a separate file, but let's see what @tarsius thinks.

lisp/magit-log.el
DamienCassou: fixed. I just realised that `prettify-symbols-mode` affects indentation: the five-characters difference in your two snippets is the five-characters difference between `lambda` and `λ`.

lisp/magit-section.el
tarsius: Let's use a separate file then.

lisp/magit-log.el
tarsius: I don't know where @DamienCassou comment on that went, but here is my reply to that:

For that reason I either place all of the BODY on the same line as `lambda`, or if that is not feasible, then no part of it.

Also please use `-contains-p` instead of the `-contains?` alias.

kyleam: > I don't know where @DamienCassou comment on that went

https://github.com/magit/magit/pull/3084#discussion_r118425814

DamienCassou:
I think I took care of all your comments. Please double-check everything because it's my first new file in the project and I'm not quite sure about the setup.

lisp/magit-imenu.el
```diff
@@ -0,0 +1,188 @@
...
+
+;;; Commentary:
+
+;; Emacs' major modes can facilitate navigation in their buffers by
+;; supporting Imenu.  In such major modes, launching Imenu (M-x imenu)
+;; makes Emacs display a list of items (e.g., function definitions in
+;; a programming major mode).  Selecting an item from this list moves
+;; point to this item.
+
+;; magit-imenu.el adds Imenu support to every major mode in Magit
```
kyleam: Without grepping the source for modes, I can think of one that isn't
supported: magit-cherry-mode :)

tarsius:
I have added fixup commits and pushed them to your branch. You will have to review that in Magit because Github's log are completely messed up after one has rewritten history. (They are sorted by author date (not even committer date!) instead of according to the dag. Very silly default.)

For some of your commits there are two fixup commits. In that case the first deals with minor issues, if I remember correctly, only lines that are two long.

The second or only fixup commit deals with:

> Please double-check everything because it's my first new file in the project and I'm not quite sure about the setup.

We shouldn't add anonymous functions to mode hooks to add the imenu functions to the appropriate hooks. I have reverted this to add the functions directly in the mode functions and have arranged for the imenu functions to be autoloaded.

Seems to work fine, though I am a bit surprised the byte-compiler does not complain about the `#'functions-unknown-at-this-point`. I guess that's because these functions are not used *as functions* at the time they are being added to the hooks. But I only checked with `25.1`. Please make sure that in older and newer Emacsen the byte-compiler also does not complain about that - else remove the `#`.

DamienCassou:
> Without grepping the source for modes, I can think of one that isn't supported: magit-cherry-mode :)

thanks for the idea. I added cherry-mode, repolist-mode, submodule-list-mode and process mode. I think we have them all know :-).

> I have added fixup commits and pushed them to your branch

thank you very much. I fixed them up and did the same for the new modes. 

> make sure that in older and newer Emacsen the byte-compiler also does not complain about that

I tried with both Emacs master and 24.5. No problem detected.

kyleam:
@tarsius I'll plan to merge this tomorrow night (est) unless you have more comments.

DamienCassou:
Thanks for the fixups @kyleam. I squashed them in.

tarsius:
No objections from me.

Issue #3085, opened 2017-05-24
git-rebase-mode does not recognize default commented lines


darkfeline:
Lines that are commented out by default in the rebase file (for example for empty commits) have a space:

```
# pick 561728c053b2be014520eeb06530c0eb5f4bfe3d test
pick 38c1690 test

# Rebase df81a1a..38c1690 onto df81a1a (1 command)
#
# Commands:
# C-c C-c  tell Git to make it happen
# C-c C-k  tell Git that you changed your mind, i.e. abort
# p        move point to previous line
# n        move point to next line
# M-p      move the commit at point up
# M-n      move the commit at point down
# SPC      show the commit at point in another buffer
# RET      show the commit at point in another buffer and select its window
# C-_      undo last change
# k        drop the commit at point
# z        add noop action at point
# c        pick = use commit
# r, w     reword = use commit, but edit the commit message
# e, m     edit = use commit, but stop for amending
# s        squash = use commit, but meld into previous commit
# f        fixup = like "squash", but discard this commit's log message
# x        exec = run command (the rest of the line) using shell
#
# These lines can be re-ordered; they are executed from top to bottom.
#
# If you remove a line here THAT COMMIT WILL BE LOST.
#
# However, if you remove everything, the rebase will be aborted.
#
# Note that empty commits are commented out
```

`git-rebase-mode` does not handle these lines correctly because of the space after the `#`.

I fixed this locally by hacking `git-rebase-mode`:

```
(setq git-rebase-line
          (concat "^\\(\\(" (regexp-quote comment-start) " *\\)?"
                  "\\(?:[fprse]\\|pick\\|reword\\|edit\\|squash\\|fixup\\|exec\\|noop\\)\\) "
                  "\\(?:\\([^ \n]+\\) \\(.*\\)\\)?"))
```

In addition to this fix, it might be more consistent to make `git-rebase-kill-line` also insert the extra space.

PR #3086, opened 2017-05-24
Recognize commented lines with spaces during rebase

darkfeline:
Lines that are commented out by default in the rebase file (for
example for empty commits) have a space after the #.  Currently,
git-rebase-mode does not recognize these lines.

For consistency we also insert a space after git-rebase-kill-line.

https://github.com/magit/magit/issues/3085


tarsius:
Thanks for the report and the fix!

tarsius:
I've now addes such an entry.

darkfeline:
Thanks, I didn't have time yesterday.

tarsius:
No problem. I would have waited, but was in merging mood 😆

Thanks again!

Issue #3088, opened 2017-05-24
Magit fails to guess what I want to checkout


mrkkrp:
Previously this worked:

1. Open commit log by pressing <kbd>l l</kbd>.
2. Go for example one commit back (on `master`).
3. Type <kbd>b b</kbd> to checkout that commit.
4. Magit should propose something like `master~1`.

Currently it proposes only tag and branch names, but can't guess `master~1`. I liked that feature! I actually learned that syntax through Magit :-)

tarsius:
Hm, still works as before for me, but there was another report about such a regression just today on gitter.

What completion framework do you use? Can you reproduce this following the instructions at https://magit.vc/manual/magit/Debugging-Tools.html?

mrkkrp:
I'm using this: https://github.com/abo-abo/swiper. I think the package that does autocompletion is called `ivy`.

This is my current Magit config:

https://github.com/mrkkrp/dot-emacs/blob/master/mk/mk-magit.el#L26-L38

tarsius:
I use `ivy` too, but I kept the default value of `magit-completing-read-function` and instead simply turned on `ivy-mode`. If you have enabled `ivy-mode` too, then please try resetting `magit-completing-read-function` to see if that fixes the issue.

Note that its doc-string mentions that when using `helm` one should just use the default value. The same is the case for `ivy`, and I'll have to update the doc-string accordingly (but I'll have to first check whether setting the variable works when `ivy-mode` is *not* enabled).

mrkkrp:
I can confirm that if I keep the default value of `magit-completing-read-function` it works.

mrkkrp:
Looks like `magit-push-always-verify` can be dropped as well...

tarsius:
I've opened #3089 to address confusions about `magit-completing-read-function`.

PR #3089, opened 2017-05-24
Don't add default to prompt for helm and ivy completion; and avoid completion misconfiguration

tarsius:
I think we should merge this now (after verifying that the helm change is appropriate). 

But in the long run we should probably phase out the `magit-completion-read-function` option, or at least move the kludges that are currently duplicated in all the magit specific functions into `magit-completing-read`, so that the upstream completion functions can be used directly instead of our wrappers. But that requires a transition plan and testing, and I don't have time for that right now. If someone else wants to do it, please go ahead. Otherwise we should merge this soon.

Re #3088.

### DIFF
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -320,7 +320,11 @@ that this wrapper makes the following changes:
 The use of another completing function and/or wrapper obviously
 results in additional differences."
   (let ((reply (funcall magit-completing-read-function
-                        (concat prompt ": ") collection predicate
+                        (concat prompt ": ")
+                        (if (and def (not (member def collection)))
+                            (cons def collection)
+                          collection)
+                        predicate
                         require-match initial-input hist def)))
     (if (string= reply "")
         (if require-match


tarsius:
Seems this is a pretty common misconfiguration. I'll try to push a kludge soon.

kyleam:
I use `helm-mode`, but even with

    (setq magit-completing-read-function 'helm--completing-read-default)

helm seems to work fine.  I don't think 2e55f04 is needed.

kyleam:
I can reproduce the `ivy-completing-read` issue from #3088 with the
latest version of swiper.  (I was previously testing with swiper's
ce3c0ec, which doesn't have the issue.)  But 1fa8955 doesn't seem to
fix that issue for me.

kyleam:
> or at least move the kludges that are currently duplicated in all
> the magit specific functions into magit-completing-read, so that the
> upstream completion functions can be used directly instead of our
> wrappers

I'm fairly sure that when I was testing #3029,
`helm--completing-read-default`, `ivy-completing-read`, and
`magit-ido-completing-read` did not need the reverse-order kludge, and
my quick testing now suggests they still don't.

kyleam:
I've bisected this to ivy's 365008f (ivy.el: Don't add :preselect to
collection, 2017-05-23).  Based on my (incomplete) understanding of
the [swiper issue], I think one solution would be to make sure that we
preprend the default value to the collection when the default value
isn't already there.

I've tested this approach for `magit-read-other-branch-or-commit`, and
it seems to work.

```diff
diff --git a/lisp/magit-git.el b/lisp/magit-git.el
index aecb505c..daa64ff0 100644
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1556,7 +1556,9 @@ (defun magit-read-other-branch-or-commit
                       (and (not (equal current exclude)) current)
                       secondary-default
                       (magit-get-previous-branch))))
-    (or (magit-completing-read prompt (delete exclude (magit-list-refnames))
+    (or (magit-completing-read prompt
+                               (cons default
+                                     (delete exclude (magit-list-refnames)))
                                nil nil nil 'magit-revision-history default)
         (user-error "Nothing selected"))))

```

[swiper issue]: https://github.com/abo-abo/swiper/issues/1017

lisp/magit-utils.el
```diff
@@ -339,13 +352,13 @@ results in additional differences."
   "Magit wrapper for standard `completing-read' function."
   (cl-letf (((symbol-function 'completion-pcm--all-completions)
              #'magit-completion-pcm--all-completions))
-    (completing-read (magit-prompt-with-default prompt def)
+    (completing-read (if (or helm-mode ivy-mode)
```
kyleam: Use `bound-and-true-p` instead?  The `defvar` approach leads to an
error:

```emacs-lisp
(defvar tarsius)
tarsius ;; (void-variable tarsius)
```

tarsius: Silly me.

tarsius:
> I don't think 2e55f04 is needed.

Okay, dropped it.

tarsius:
> I can reproduce the ivy-completing-read issue from #3088

Turns out I was not actually using the latest version due to a misconfigured upstream in my local clone. I can reproduce this too now.

tarsius:
> Based on my (incomplete) understanding of the corresponding swiper issue, I think one solution would be to prepend the default value to the collection when the default value isn't already there.

I agree. Also see my comment at https://github.com/abo-abo/swiper/issues/1017#issuecomment-304091866.

---

I have applied the first commit to `master` and will open a new pr with a new fix.

tarsius:
> I can reproduce the `ivy-completing-read` issue from #3088 with the
> latest version of swiper.  (I was previously testing with swiper's
> ce3c0ec, which doesn't have the issue.)  But 1fa8955 doesn't seem to
> fix that issue for me.

Yes, that seems to be unnecessary. @mrkkrp was probably mistaken in that this makes any difference.

> I'm fairly sure that when I was testing #3029,
> helm--completing-read-default, ivy-completing-read, and
> magit-ido-completing-read did not need the reverse-order kludge, and
> my quick testing now suggests they still don't.

Great, so we don't have to do anything here.

----

At least in the short run, #3092 should be sufficient.

PR #3091, opened 2017-05-25
Avoid connecting to unrelated hosts when saving buffers

tarsius:


kyleam:
Nice.  Thanks, @jdtsmith and @tarsius

PR #3092, opened 2017-05-25
magit-completing-read: add default to choices if necessary

tarsius:
Until now, we did not bother adding the default choice to the list of
all choices, because we assumed that doing so was not necessary.

But it turned out that while it is possible to use such a default by
simply confirming it, `completing-read` does not actually added to the
list of all choices.  It therefore won't be shown when the user uses
TAB to list the choices and manually typing the default and then
exiting with RET also won't work.

Additionally `ivy-completing-read` was recently changed to completely
ignore a default that isn't a member of the choices, making a fix more
pressing.

This commit simply prepends the default to the list of choices, if it
is not a `member` yet.  Doing this is much simpler than adjusting each
and every caller of `magit-completing-read`, but it may in some cases
be weird for the default to be listed first, when the other choices
are sorted alphabetically.

### DIFF
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -320,7 +320,11 @@ that this wrapper makes the following changes:
 The use of another completing function and/or wrapper obviously
 results in additional differences."
   (let ((reply (funcall magit-completing-read-function
-                        (concat prompt ": ") collection predicate
+                        (concat prompt ": ")
+                        (if (and def (not (member def collection)))
+                            (cons def collection)
+                          collection)
+                        predicate
                         require-match initial-input hist def)))
     (if (string= reply "")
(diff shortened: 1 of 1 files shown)


kyleam:
Looks good to me, and nicely explained.

In 5ebc0cb's message:

> does not actually added to

s/added/add the default to/ ?

> manually typing the default and then exiting with RET also won't
> work

I've tested a few cases with the default completing-read, and I'm able
to manually enter the default.

tarsius:
> I've tested a few cases with the default completing-read, and I'm able to manually enter the default.

I have clarified the commit message. Only if REQUIRE-MATCH has to be non-nil, is it impossible to exit after typing the DEFAULT.

Issue #3098, opened 2017-06-08
magit-show-commit show the wrong commitid when branche name == tag name 


jixiuf:
(2) what behavior you observed
![snip20170608_2](https://user-images.githubusercontent.com/461743/26914798-c9cd605a-4c55-11e7-8fff-1b47768d5f38.png)

  (3) and how we can reproduce the issue.

it can be reproduced by `M-x magit-emacs-Q-command RET`

kyleam:
`magit-branch-or-commit-at-point` (and utlimately `magit-get-shortname`) should be returning a qualified refname here (i.e., "heads/v1").

PR #3101, opened 2017-06-09
Improve handling of ambiguous refnames

kyleam:
Please see commit messages.

Re: #3098

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -1339,7 +1339,11 @@ commit or stash at point, then prompt for a commit."
           (user-error "No commit on this line"))))
      (t
       (magit-section-case
-        ((commit branch)
+        (branch
+         (setq rev (magit-ref-maybe-qualify (magit-section-value it)))
+         (setq cmd 'magit-show-commit)
+         (setq buf (magit-mode-get-buffer 'magit-revision-mode)))
+        (commit
          (setq rev (magit-section-value it))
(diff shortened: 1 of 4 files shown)


lisp/magit-diff.el
```diff
@@ -1339,7 +1339,11 @@ commit or stash at point, then prompt for a commit."
           (user-error "No commit on this line"))))
      (t
       (magit-section-case
-        ((commit branch)
+        (branch
+         (setq rev (magit-ref-maybe-qualify (magit-section-value it))
+               cmd 'magit-show-commit
+               buf (magit-mode-get-buffer 'magit-revision-mode)))
```
tarsius: Please use three `setq`s. Stefan says that the ability to set multiple variables is only supported for backward compatibility.

kyleam: That multi-symbol style is used throughout
`magit-diff-show-or-scroll`.  It'd be better to change all the
single-symbol setq's in a separate step, no?

kyleam: > It'd be better to change all the single-symbol setq's in a separate step

The branch [km/multi-setq] should take care of all the multi-symbol
setq's in Magit, if that's the direction you want to go.

[km/multi-setq]: https://github.com/magit/magit/compare/km/multi-setq

kyleam: Done.  Thanks for the review.

PR #3103, opened 2017-06-13
Fix donation link in CONTRIBUTING.md

ebpa:
As much as I think everyone should see the hilarious Black Book lulz killer gif, the donation link currently 404's :-)


tarsius:
Thanks! I have applied the commit after fixing the issues that resulted from writing a commit message on github.

PR #3112, opened 2017-06-25
git-rebase-line regexp did not support spaces before pick commands

Peaker:
* This causes it to reject lines like "# pick ..." generated by
  magit-rebase when using git-rebase-kill-line


tarsius:
Thanks!

Issue #1639, opened 2014-11-25
Save commit buffer with bookmark


tuhdo:
It would be nice if I can quickly bookmark a `*magit-commit*` buffer and open it later, similar to how Dired and Bookmark work with each other. `orgit` is fantastic, but it could be improved along with adding this feature:
- We should be able to bookmark/link not only commit buffer, but log buffer as well. So, when we bookmark/link a log buffer, it saves the command that could be used to recreate that log buffer again.
- After the above is done, it is desirable if point is restored according to when the buffer is saved..

tarsius:
Like `orgit` this should be a separate package. I have created a separate (so far empty) repository for this (except for the TODO issue https://github.com/magit/magit-bookmark/issues/1).

yurikhan:
What happened to that repository? It says 404 now.

Anyway, I have implemented basic support for `magit-status` bookmarks. (Stores just the top-level directory and not the expand/collapse state of sections. It might make sense to save state of top-level sections (Untracked, Unstaged, Staged, Stashes, Unmerged, Unpulled) but everything below is way too volatile for a long-lived bookmark.

The code is 28 significant lines of Elisp, and I expect near-linear growth in the number of supported modes. @tarsius, are you sure you don’t want that in core?

tarsius:
We recently added `magit-imenu.el` to this repository. If you write `magit-bookmarks.el`, then we could add that too. I am not really familiar with `imenu` or `bookmarks`, but I suspect there might be some code that could be reused.

tarsius:
I eventually deleted the `magit-bookmarks` repository because it always was only a stub. I do not have a leftover copy of that stub.

yurikhan:
Okay, then I’m collecting feature requests/ideas as to which kinds of Magit buffers should be supported, in addition to “`magit-status` of a specific repository”.

I’m interpreting @tuhdo’s original comment as asking for bookmarks to “a specific commit in a specific repository” and, I guess, “log of a specific repository with specific options, starting with the current HEAD, a specific set of refs or all refs”.

PR #3113, opened 2017-06-25
Add bookmark support [#1639]

yurikhan:
Okay, here goes. In my limited testing, smoke does not come out.

Of all the major modes defined in Magit:

* `magit-status-mode`, `magit-refs-mode`, `magit-log-mode`, `magit-reflog-mode`, `magit-stashes-mode`, `magit-cherry-mode`, `magit-diff-mode`, `magit-revision-mode`, `magit-stash-mode`, and `magit-submodule-list-mode` are fully supported.
* `magit-log-select-mode` is not supported specifically but will inherit `magit-log-mode` treatment. Should I explicitly disallow that?
* `magit-repolist-mode` is not supported because I don’t see much sense in using repolist with bookmarks but feel free to convince me otherwise.
* `magit-merge-preview-mode` is supported but if the current branch is changed after the bookmark is created then activating the bookmark will preview a different merge than was bookmarked. That is because `magit-merge-preview` does not let me specify the target branch. If you are bothered by that, let’s work out the details separately.
* `git-rebase-mode` and `magit-popup-mode` are not supported because they are transient UI rather than content.
* `magit-process-mode` is not supported because none of its content is persistent and the buffer itself is always recreated and easily accessible from any other Magit buffer.

Theory of operation:

* When the user requests a bookmark for the current buffer and it’s in one of the supported modes, our `*-make-record` function will be called.
* It will create a bookmark record using the default function, set its `filename` property to the root of the current repository, `handler` to the corresponding `*-jump` function, and various `magit-*` properties to parts of `magit-refresh-args`.
* When the bookmark is activated, it is passed to its recorded handler function, which feeds the saved `magit-*` properties into the relevant Magit command and restores the position in buffer.
* The handler functions are marked autoload so that they pull us in when the user activates a bookmark even if neither `magit-bookmark.el` nor `magit.el` have been loaded yet.
* Otherwise, we get loaded after both `magit.el` and `bookmark.el` are loaded, which is the necessary condition for the user to start creating bookmarks.

### DIFF
--- a/lisp/magit-bookmark.el
+++ b/lisp/magit-bookmark.el
@@ -1,0 +1,364 @@
+;;; magit-bookmark.el --- bookmark support for Magit  -*- lexical-binding: t -*-
+
+;; Copyright (C) 2010-2017  The Magit Project Contributors
+;;
+;; You should have received a copy of the AUTHORS.md file which
+;; lists all contributors.  If not, see http://magit.vc/authors.
+
+;; Author: Yuri Khan <yuri.v.khan@gmail.com>
+;; Maintainer: Jonas Bernoulli <jonas@bernoul.li>
+
+;; Magit is free software; you can redistribute it and/or modify it
+;; under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 3, or (at your option)
+;; any later version.
+;;
+;; Magit is distributed in the hope that it will be useful, but WITHOUT
+;; ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+;; or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+;; License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with Magit.  If not, see http://www.gnu.org/licenses.
+
+;;; Commentary:
+
+;; Support for bookmarks for most Magit buffers.
+
+;;; Code:
+
+(require 'magit)
+(require 'bookmark)
+
+;;; Supporting primitives
+
+(defun magit-bookmark--jump (bookmark fn &rest args)
+  "Handle a Magit BOOKMARK.
+
+This function will:
+
+1. Bind `default-directory' to the repository root directory
+   stored in the `filename' bookmark property.
+2. Invoke the function FN with ARGS as arguments.  This needs to
+   restore the buffer.
+3. Restore the expanded/collapsed status of top level sections
+   and the point position."
+  (declare (indent 2))
+  (let* ((default-directory (bookmark-get-filename bookmark)))
+    (if default-directory
+        (apply fn args)
+      (signal 'bookmark-error-no-filename (list 'stringp default-directory)))
+    (when (derived-mode-p 'magit-mode)
+      (-when-let (hidden-sections (bookmark-prop-get bookmark
+                                                     'magit-hidden-sections))
+        (--each (magit-section-children magit-root-section)
+          (if (member (cons (magit-section-type it) (magit-section-value it))
+                      hidden-sections)
+              (magit-section-hide it)
+            (magit-section-show it)))))
+    (--when-let (bookmark-get-position bookmark)
+      (goto-char it))
+    (--when-let (bookmark-get-front-context-string bookmark)
+      (when (search-forward it (point-max) t)
+        (goto-char (match-beginning 0))))
+    (--when-let (bookmark-get-rear-context-string bookmark)
+      (when (search-backward it (point-min) t)
+        (goto-char (match-end 0))))
+    nil))
+
+(defun magit-bookmark--make-record (mode handler &optional make-props)
+  "Create a Magit bookmark.
+
+MODE specifies the expected major mode of current buffer.
+
+HANDLER should be a function that will be used to restore this
+buffer.
+
+MAKE-PROPS should be either nil or a function that will be called
+with `magit-refresh-args' as the argument list, and may return an
+alist whose every element has the form (PROP . VALUE) and
+specifies additional properties to store in the bookmark."
+  (declare (indent 1))
+  (unless (eq major-mode mode)
+    (user-error "Not in a %s buffer" mode))
+  (let ((bookmark (bookmark-make-record-default 'no-file)))
+    (bookmark-prop-set bookmark 'handler handler)
+    (bookmark-set-filename bookmark (magit-toplevel))
+    (when (derived-mode-p 'magit-mode)
+      (bookmark-prop-set
+       bookmark 'magit-hidden-sections
+       (--map (cons (magit-section-type it) (magit-section-value it))
+              (-filter #'magit-section-hidden
+                       (magit-section-children magit-root-section)))))
+    (when make-props
+      (pcase-dolist (`(,prop . ,value) (apply make-props magit-refresh-args))
+        (bookmark-prop-set bookmark prop value)))
+    bookmark))
+
+;;; Status
+
+;;;###autoload
+(defun magit-bookmark--status-jump (bookmark)
+  "Handle a Magit status BOOKMARK."
+  (magit-bookmark--jump bookmark
+      (lambda () (magit-status-internal default-directory))))
+
+;;;###autoload
+(defun magit-bookmark--status-make-record ()
+  "Create a Magit status bookmark."
+  (magit-bookmark--make-record 'magit-status-mode
+    #'magit-bookmark--status-jump))
+
+;;; Refs
+
+;;;###autoload
+(defun magit-bookmark--refs-jump (bookmark)
+  "Handle a Magit refs BOOKMARK."
+  (magit-bookmark--jump bookmark #'magit-show-refs
+    (bookmark-prop-get bookmark 'magit-refs)
+    (bookmark-prop-get bookmark 'magit-args)))
+
+;;;###autoload
+(defun magit-bookmark--refs-make-record ()
+  "Create a Magit refs bookmark."
+  (magit-bookmark--make-record 'magit-refs-mode
+    #'magit-bookmark--refs-jump
+    (lambda (refs args)
+      `((magit-refs . ,refs)
+        (magit-args . ,args)))))
+
+;;; Log
+
+;;;###autoload
+(defun magit-bookmark--log-jump (bookmark)
+  "Handle a Magit log BOOKMARK."
+  (magit-bookmark--jump bookmark #'magit-log
+    (bookmark-prop-get bookmark 'magit-revs)
+    (bookmark-prop-get bookmark 'magit-args)
+    (bookmark-prop-get bookmark 'magit-files)))
+
+(defun magit-bookmark--log-make-name (buffer-name revs _args files)
+  "Generate the default name for a log bookmark."
+  (concat
+   buffer-name " " (mapconcat #'identity revs " ")
+   (and files
+        (concat " touching " (mapconcat #'identity files " ")))))
+
+;;;###autoload
+(defun magit-bookmark--log-make-record ()
+  "Create a Magit log bookmark."
+  (magit-bookmark--make-record 'magit-log-mode
+    #'magit-bookmark--log-jump
+    (lambda (revs args files)
+      `((defaults    . (,(magit-bookmark--log-make-name
+                          (buffer-name) revs args files)))
+        (magit-revs  . ,revs)
+        (magit-args  . ,args)
+        (magit-files . ,files)))))
+
+;;; Reflog
+
+;;;###autoload
+(defun magit-bookmark--reflog-jump (bookmark)
+  "Handle a Magit reflog BOOKMARK."
+  (magit-bookmark--jump bookmark
+      (lambda ()
+        (let ((magit-reflog-arguments (bookmark-prop-get bookmark 'magit-args)))
+          (magit-reflog (bookmark-prop-get bookmark 'magit-ref))))))
+
+(defun magit-bookmark--reflog-make-name (buffer-name ref)
+  "Generate the default name for a reflog bookmark."
+  (concat buffer-name " " ref))
+
+;;;###autoload
+(defun magit-bookmark--reflog-make-record ()
+  "Create a Magit reflog bookmark."
+  (magit-bookmark--make-record 'magit-reflog-mode
+    #'magit-bookmark--reflog-jump
+    (lambda (ref args)
+      `((defaults   . (,(magit-bookmark--reflog-make-name (buffer-name) ref)))
+        (magit-ref  . ,ref)
+        (magit-args . ,args)))))
+
+;;; Stashes
+
+;;;###autoload
+(defun magit-bookmark--stashes-jump (bookmark)
+  "Handle a Magit stash list BOOKMARK."
+  (magit-bookmark--jump bookmark #'magit-stash-list))
+
+;;;###autoload
+(defun magit-bookmark--stashes-make-record ()
+  "Create a Magit stash list bookmark."
+  (magit-bookmark--make-record 'magit-stashes-mode
+    #'magit-bookmark--stashes-jump))
+
+;;; Cherry
+
+;;;###autoload
+(defun magit-bookmark--cherry-jump (bookmark)
+  "Handle a Magit cherry BOOKMARK."
+  (magit-bookmark--jump bookmark #'magit-cherry
+    (bookmark-prop-get bookmark 'magit-head)
+    (bookmark-prop-get bookmark 'magit-upstream)))
+
+(defun magit-bookmark--cherry-make-name (buffer-name head upstream)
+  "Generate the default name for a cherry bookmark."
+  (concat buffer-name " " head " upstream " upstream))
+
+;;;###autoload
+(defun magit-bookmark--cherry-make-record ()
+  "Create a Magit cherry bookmark."
+  (magit-bookmark--make-record 'magit-cherry-mode
+    #'magit-bookmark--cherry-jump
+    (lambda (upstream head)
+      `((defaults       . (,(magit-bookmark--cherry-make-name
+                             (buffer-name) head upstream)))
+        (magit-head     . ,head)
+        (magit-upstream . ,upstream)))))
+
+;;; Diff
+
+;;;###autoload
+(defun magit-bookmark--diff-jump (bookmark)
+  "Handle a Magit diff BOOKMARK."
+  (magit-bookmark--jump bookmark #'magit-diff-setup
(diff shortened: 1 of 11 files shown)


tarsius:
I am sorry I didn't have a change to check this out yet. I will try to do so soonish.

yurikhan:
No pressure.

lisp/magit-bookmark.el
```diff
@@ -0,0 +1,461 @@
...
+
+BODY needs to invoke the correct Magit command to restore the
+buffer.  After that, top level section status and point position
+will be restored."
+  (declare (indent 1))
+  `(let ((default-directory (bookmark-get-filename ,bookmark))
+         (hidden-sections (bookmark-prop-get ,bookmark 'magit-hidden-sections))
+         (forward-str (bookmark-get-front-context-string ,bookmark))
+         (behind-str (bookmark-get-rear-context-string ,bookmark))
+         (place (bookmark-get-position ,bookmark)))
```
kyleam: Though it doesn't seem like it matters for any of the current callers,
I'd only evaluate `bookmark` once, storing the value with an
uninterned symbol.

Aside from `default-directory`, I think all the variables in the let
form should be protected with an outer `cl-gensym` variable.  And some
of these variables could be eliminated by using `--when-let`.

lisp/magit-bookmark.el
```diff
@@ -0,0 +1,461 @@
...
+     (if default-directory
+         ,@body
+       (signal 'bookmark-error-no-filename (list 'stringp default-directory)))
+     (when (derived-mode-p 'magit-moge)
```
kyleam: code typo: moge

lisp/magit-bookmark.el
```diff
@@ -0,0 +1,461 @@
...
+       (--each (magit-section-children magit-root-section)
+         (if (member `(,(magit-section-type it) . ,(magit-section-value it))
```
kyleam: nit: I prefer `cons` here.

lisp/magit-bookmark.el
```diff
@@ -0,0 +1,461 @@
...
+specifies additional properties to store in the bookmark."
+  (unless (derived-mode-p mode)
+    (error "Not in a %s buffer" mode))
+  (let ((bookmark (bookmark-make-record-default 'no-file)))
+    (bookmark-prop-set bookmark 'handler handler)
+    (bookmark-set-filename bookmark (magit-toplevel))
+    (when (derived-mode-p 'magit-mode)
+      (bookmark-prop-set
+       bookmark 'magit-hidden-sections
+       (--map `(,(magit-section-type it) . ,(magit-section-value it))
```
kyleam: same comment as above about `cons`

lisp/magit-bookmark.el
```diff
@@ -0,0 +1,461 @@
...
+  "Create a Magit status bookmark."
+  (magit-bookmark--make-record
+   'magit-status-mode #'magit-bookmark--status-jump))
+
+(defun magit-bookmark--status-setup ()
+  "Enable bookmark support in a Magit status buffer."
+  (setq-local bookmark-make-record-function
+              #'magit-bookmark--status-make-record))
+
+(add-hook 'magit-status-mode-hook 'magit-bookmark--status-setup)
```
kyleam: For consistency with magit-imenu.el, `bookmark-make-record-function`
would be set to an autoloaded function in `magit-status-mode`'s
definition, a setup that was discussed somewhat in #3084.  (The same
would be true for all the modes below.)  Though to me it seems cleaner
to add a (named) function like you do here, but perhaps I'm
overlooking some downside of the add-hook approach.

lisp/magit-bookmark.el
```diff
@@ -0,0 +1,461 @@
...
+     rev)
+    ((and `(,rev1 ,rev2)
+          (guard (/= ?^ (aref rev1 0)))
+          (guard (= ?^ (aref rev2 0))))
+     (concat (substring rev2 1) ".." rev1))
+    ((and `(,rev1 ,rev2 ,rev3)
+          (guard (/= ?^ (aref rev1 0)))
+          (guard (/= ?^ (aref rev2 0)))
+          (guard (= ?^ (aref rev3 0))))
+     (ignore rev3)
```
kyleam: I was surprised that the byte-compiler complains here since `rev3` is
used in the guard.

lisp/magit-bookmark.el
```diff
@@ -0,0 +1,461 @@
...
+;;;###autoload
+(defun magit-bookmark--diff-jump (bookmark)
+  "Handle a Magit diff bookmark."
+  (magit-bookmark--jump bookmark
+    (magit-diff-setup (bookmark-prop-get bookmark 'magit-rev-or-range)
+                      (bookmark-prop-get bookmark 'magit-const)
+                      (bookmark-prop-get bookmark 'magit-args)
+                      (bookmark-prop-get bookmark 'magit-files))))
+
+(defun magit-bookmark--resolve (rev-or-range)
```
kyleam: Hmm, quickly look at git-rev-parse's output on the command line, it
seems like this should work fine.  I wonder if Git is consistent about
its output order across different input/Git versions.  But, in the
worse case, this hands back the range as is, so that should be OK.

Also, the differences in Git's handling of diff ranges vs ranges
everywhere else made me pause, but it's irrelevant because rev-parse
is just used to determine the hash and then reconstruct the range.

lisp/magit-bookmark.el
```diff
@@ -0,0 +1,461 @@
...
+          (guard (/= ?^ (aref rev1 0)))
+          (guard (/= ?^ (aref rev2 0)))
+          (guard (= ?^ (aref rev3 0))))
+     (ignore rev3)
+     (concat rev1 "..." rev2))
+    (_
+     rev-or-range)))
+
+(defun magit-bookmark--diff-make-name (buffer-name
+                                          rev-or-range const _args files)
```
kyleam: I prefer the entire argument list on a new line.

lisp/magit-bookmark.el
kyleam: And one such downside is that the hooks would need to be autoloaded
(wrapped in a `(eval-after-load 'magit ...`).

The

```
(eval-after-load 'bookmark
  '(require 'magit-bookmark))
```

in magit.el isn't sufficient because the value of
`bookmark-make-record-function` won't be set at the time of the first
call.  For example, if you create a status buffer and then run
`bookmark-set`, the `magit-status-mode-hook` has already been called,
so `bookmark-make-record-function` will be
`bookmark-make-record-default` at that point, leading to a "Buffer not
visiting a file or directory" error.

lisp/magit-bookmark.el
```diff
@@ -0,0 +1,461 @@
...
+    (magit-diff-setup (bookmark-prop-get bookmark 'magit-rev-or-range)
+                      (bookmark-prop-get bookmark 'magit-const)
+                      (bookmark-prop-get bookmark 'magit-args)
+                      (bookmark-prop-get bookmark 'magit-files))))
+
+(defun magit-bookmark--resolve (rev-or-range)
+  "Return REV-OR-RANGE with ref names resolved to commit hashes."
+  (pcase (magit-git-lines "rev-parse"
+                          (magit-abbrev-arg "short")
+                          rev-or-range)
```
kyleam: I'd rather not rely on the `--short` flag here because Git gives the
same non-zero exit status that it would give with `--verify`:

```
$ git rev-parse --short master^..master
d85547a4
^524ee09d
fatal: Needed a single revision
```

The individual revs below could instead be shortened with
`magit-rev-abbrev`.  It's one more git call, but I don't think that's
problematic in the context of storing a bookmark.

yurikhan:
>> `magit-log-select-mode` will inherit
>> `magit-log-mode` treatment. Should I explicitly disallow that?
> Unless it raises an error, I don't think it's necessary.

In fact it does. The structure of `magit-refresh-args` is different: `magit-log-mode` stores `(revs args files)` where `revs` is a list, while `magit-log-select-mode` stores `(rev args)` where `rev` is a single string. The Liskov Substitutability Principle does not hold.

For now, I will compare `major-mode` for exact match rather than derivation in `magit-bookmark--make-record`. This way, derived modes will not by default inherit bookmark behavior from their base.

> Is magit-merge-preview-mode worth supporting?

What made me make an effort to support it was the fact that it derived from `magit-diff-mode`, and I knew the generic diff bookmarking was going to break on `magit-merge-preview-mode`.

With inheritance out of the picture, I don’t see value in supporting `magit-merge-preview-mode` until there are actual user requests for that.

lisp/magit-bookmark.el
yurikhan: Right. Those differences were a source of worry for me as well.

Should I make it clear that this function is specifically for diff ranges?

lisp/magit-bookmark.el
yurikhan: A bit of investigation here: [#27545](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=27545)

yurikhan:
@kyleam, I have incorporated your suggestions.

Please let me know when the branch is ready for merging, I will squash it into one commit.

lisp/magit-bookmark.el
kyleam: > Should I make it clear that this function is specifically for diff ranges?

I'm OK with it as-is because it can handle either range.  Mostly just
thinking out loud.

lisp/magit-bookmark.el
kyleam: Thanks.  Looks like a known (and tricky-to-fix) annoyance then.

lisp/magit-bookmark.el
```diff
@@ -0,0 +1,366 @@
...
+     `((defaults . (,(magit-bookmark--stash-make-name
+                      (buffer-name)
+                      (magit-rev-abbrev magit-buffer-revision-hash)
+                      args files)))
+       (magit-stash . ,magit-buffer-revision-hash)
+       (magit-args . ,args)
+       (magit-files . ,files)
+       (magit-hidden-sections
+        . ,(--map `(,(magit-section-type it)
+                    . ,(replace-regexp-in-string (rx-to-string stash)
```
tarsius: Please don't use `rx-to-string`.

lisp/magit-bookmark.el
```diff
@@ -0,0 +1,367 @@
...
+
+BODY needs to invoke the correct Magit command to restore the
+buffer.  After that, top level section status and point position
+will be restored."
+  (declare (indent 1))
+  (let ((bookmark-var (make-symbol "bookmark")))
+    `(let* ((,bookmark-var ,bookmark)
+            (default-directory (bookmark-get-filename ,bookmark-var)))
+       (if default-directory
+           ,@body
```
tarsius: I think you could turn this macro into a function if you replaced `,@body` with `(apply fn args)`.

lisp/magit-bookmark.el
```diff
@@ -0,0 +1,367 @@
...
+MODE specifies the expected major mode of current buffer.
+
+HANDLER should be a function that will be used to restore this
+buffer.
+
+MAKE-PROPS should be either nil or a function that will be called
+with `magit-refresh-args' as the argument list, and may return an
+alist whose every element has the form (PROP . VALUE) and
+specifies additional properties to store in the bookmark."
+  (unless (eq major-mode mode)
```
tarsius: You can use `(declare (indent 1))` for functions too. I think it would be nice to put the first argument on the same line as `magit-bookmark--make-record` and then reindent the rest.

lisp/magit-bookmark.el
```diff
@@ -0,0 +1,367 @@
...
+(defun magit-bookmark--log-make-record ()
+  "Create a Magit log bookmark."
+  (magit-bookmark--make-record
+   'magit-log-mode #'magit-bookmark--log-jump
+   (lambda (revs args files)
+     `((defaults . (,(magit-bookmark--log-make-name
+                      (buffer-name) revs args files)))
+       (magit-revs . ,revs)
+       (magit-args . ,args)
+       (magit-files . ,files)))))
```
tarsius: I think that aligning the dots would improve readability here and in other such `lambda`s.

lisp/magit-bookmark.el
```diff
@@ -0,0 +1,367 @@
...
+    ((and `(,rev1 ,rev2)
+          (guard (/= ?^ (aref rev1 0)))
+          (guard (= ?^ (aref rev2 0))))
+     (concat (magit-rev-abbrev (substring rev2 1))
+             ".."
+             (magit-rev-abbrev rev1)))
+    ((and `(,rev1 ,rev2 ,rev3)
+          (guard (/= ?^ (aref rev1 0)))
+          (guard (/= ?^ (aref rev2 0)))
+          (guard (= ?^ (aref rev3 0))))
```
tarsius: Here too some alignment would help. Add a space character next to the `=` in `(= ?^ ...)`. I would probably put it *before*, but if that's too gaudy for you then you can also put it after.

Generally speaking, when multiple subsequent lines are essentially the same, I try to align them, so that (1) the parts that are *not* the same stand out more and (2) so that the parts that *are* the same, are also more easily absorbed.

lisp/magit-bookmark.el
```diff
@@ -0,0 +1,367 @@
...
+    (magit-show-commit (bookmark-prop-get bookmark 'magit-rev)
+                       (bookmark-prop-get bookmark 'args)
+                       (bookmark-prop-get bookmark 'files))))
+
+(defun magit-bookmark--revision-make-name (buffer-name rev _args files)
+  "Generate a default name for a revision bookmark."
+  (let ((subject (magit-rev-format "%s" rev)))
+    (concat buffer-name " "
+            (magit-rev-abbrev rev)
+            (cond
```
tarsius: Do you expect other cases to be added later and want to keep future diffs as simple as possible? If not, then I would drop the newline here.

lisp/magit-bookmark.el
```diff
@@ -0,0 +1,367 @@
...
+             (files (concat " " (mapconcat #'identity files " ")))
+             (subject (concat " " subject))))))
```
tarsius: Since we are nitpicking about alignment anyway... here's another instance where I would consider doing that.

lisp/magit-bookmark.el
```diff
@@ -0,0 +1,367 @@
...
+
+;;; Code:
+
+(require 'magit)
+(require 'bookmark)
+
+;;; Supporting primitives
+
+(defmacro magit-bookmark--jump (bookmark &rest body)
+  "Handle a Magit bookmark.
```
tarsius: You are not mentioning the `BOOKMARK` argument in this doc-string, just the term `"bookmark"`. You could change this line to `Handle the Magit BOOKMARK`. Maybe for this function you want to *not* do that and instead talk about it more explicitly later in the doc-string, but at least for the `magit-bookmark--{*}-jump` functions below I would go with the "the" and "BOOKMARK".

lisp/magit-bookmark.el
yurikhan: Hmmm… Actually, I can. The handlers can retrieve the arguments from the bookmark first, and then call `magit-bookmark--jump`, passing the actual worker function and the arguments. `magit-bookmark--jump` will then call the worker function with those arguments.

The only wrinkles will be with `m-b--status-jump`, where the worker function needs the value of `default-directory` set by `m-b--jump`, and with `m-b--reflog-jump` which needs to bind a variable around the worker function call. These can be solved with lambdas.

lisp/magit-bookmark.el
yurikhan: The original buffer’s sections have the values of the form `stash{0}`, `stash{0}^2`, and possibly `stash{0}^3`. In the bookmark record, they need to be stored as `HASH`, `HASH^2`, and `HASH^3`, respectively, so that the bookmark still refers to the same stash if the stash stack changes between creation and activation of the bookmark.

Would the following be an acceptable way to resolve stash sections, instead of the above `replace-regexp-in-string` invocation?

```elisp
                    . ,(--if-let (nth 1 (split-string (magit-section-value it)
                                                      "\\^"))
                           (format "%s^%s" magit-buffer-revision-hash it)
                         magit-buffer-revision-hash))
```

lisp/magit-bookmark.el
yurikhan: `(declare (indent 2))` would be even better but if the maximum line length is 80 then `indent 1` it is.

lisp/magit-bookmark.el
yurikhan: Let me try it on.

```elisp
(defun magit-bookmark--status-jump (bookmark)
  "Handle the Magit status BOOKMARK."
```

Nope. To me, “the” here looks like there can be only one status bookmark.

The idea that we want to express here is “handle BOOKMARK, which is expected to be a Magit status bookmark”.

I can settle with “Handle a Magit {kind} BOOKMARK”, if that’s okay with you.

lisp/magit-bookmark.el
yurikhan: Okay, I squashed this and the other `cond` and aligned them.

lisp/magit-bookmark.el
kyleam: I think non-aligners (like me) understand why aligners align.  They
just don't agree that it's worth it :)

[though I do try to follow your preference when touching Magit code]

lisp/magit-bookmark.el
tarsius: That's okay with me.

lisp/magit-bookmark.el
tarsius: Can't you just replace `rx-to-string` with `regexp-quote`?

yurikhan: That will work, too.

What is the problem with `rx-to-string`, is it a backward compatibility constraint, or are there other considerations?

tarsius: Well, it requires `rx` to be loaded. Also I am not really a big fan of `rx`. I can already read regular regular expressions, and so `rx` just adds an extra layer that I have to understand (which usually forces me to look things up). It may have its place, but for little things I try to avoid it.

yurikhan: Okay, replaced with `regexp-quote`.

yurikhan:
I should not push code that does not pass byte compilation.
I should not push code that does not pass byte compilation.
I should not push code that does not pass byte compilation.
…

tarsius:
I squashed, rebased, and merged. But before pulling `master` and after rebasing again I forgot to push again before merging. So it looks like this wasn't merge, sorry!

And thanks a lot for you work!

yurikhan:
Thank you!

PR #3115, opened 2017-06-27
magit-{init,status}: compare toplevels by truename

basil-conto:
##### Changes

This allows user input, whether read interactively or from `magit-repository-directories`, involving symlinked files to be compared to the result of `magit-toplevel` with `find-file-visit-truename` enabled. Otherwise, `magit-status` in particular fails to show the status corresponding to a symlinked directory in `magit-repository-directories`, and instead asks the user to create a new repository.

##### Background

My personal need for this arises from configuring `magit-repository-directories` to list a single directory containing symlinks to all other repositories in my filesystem. This saves me from having to maintain the list in my `user-init-file`. I only noticed the `magit-status` issue after recently enabling `find-file-visit-truename`. Though I can't imagine how the introduction of `file-equal-p` might affect anything else, an alternative solution in my configuration would be to advise `magit-list-repos` to return truenames.

### DIFF
--- a/Documentation/RelNotes/2.11.0.txt
+++ b/Documentation/RelNotes/2.11.0.txt
@@ -117,6 +117,11 @@ Fixes since v2.10.3
   not into its own upstream branch, then the user was not offered to
   force the deletion, causing it to fail.  #3107
 
+* With `find-file-visit-truename' set to a non-nil value, the commands
+  `magit-status' and `magit-init' failed to detect when their
+  directory argument was a symlink that resolved to the output of
+  `magit-toplevel'.  #3115
+
 This release also contains the fixes described in the v2.10.{1,2,3}
 release notes, as well as other minor improvements, bug fixes, typo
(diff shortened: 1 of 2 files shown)


kyleam:
> Though I can't imagine how the introduction of file-equal-p might affect anything else

I can't can't come up with any potential issues either.  This change
looks like a clear improvement.

Should your subject `s/magit-status/magit-{init,status}/`?

basil-conto:
> Should your subject `s/magit-status/magit-{init,status}/`?

Done. I was being lazy in using the feature `magit-status` to represent both functions defined therein, but luckily the amended commit subject still fits within 50 columns for my compulsive peace of mind. :)

basil-conto:
FWIW, despite this being a seemingly isolated use case for truenames, I also grepped the rest of the project in search of similar `string-equal/string=` filename comparisons and didn't see anything obvious. No promises about `equal` though. :)

Issue #3117, opened 2017-07-05
Apple Git-81 doesn't recognize git-name-rev --exclude option. Breaks magit-rev-name


lambdageek:
Magit 20170702.858, Git 2.11.0 (Apple Git-81), Emacs 25.2.1, darwin

Expected behavior:
 When I'm looking at a `magit-log` buffer (`l b`) and point is on the name of a branch `b k` should default to the branch at point as the default to delete.

Actual behavior:
  Default branch to be deleted is the currently checked out branch.

(`b k` behaves as expected in a `magit-refs` buffer (`y`))

tarsius:
Magit is supposed to be have the way you want it to behave and I just verified that it does for me.

Are you using a third-party completion framework? (It's supposed to work with those too, but that might help narrow the issue down.)

tarsius:
Also try to reproduce this issue using [`magit-emacs-Q-command `](https://magit.vc/manual/magit/Debugging-Tools.html).

lambdageek:
Confirmed - does the wrong thing with `magit-emacs-Q-command`, too.

tarsius:
I just tried that too and it works for me. Though maybe I misunderstood "point is on the name of a branch". Do you mean "point is on a line on which later a branch name appears"? Or do you mean "Multiple branches point at the same commit, including the current branch. When point is on a branch name in a line in a log (not just anywhere in that line), then offer that command as default."?

lambdageek:
I've tried both.  Doesn't seem to work even when point is on a line where there's just a single branch which is different from the current branch.

I just realized there might be something else in play - I have `hub` installed - `brew install hub`

```
$ brew info hub
hub: stable 2.2.9 (bottled), devel 2.3.0-pre9, HEAD
Add GitHub support to git on the command-line
https://hub.github.com/
/usr/local/Cellar/hub/2.2.9 (10 files, 9MB) *
  Poured from bottle on 2017-06-06 at 13:45:17
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/hub.rb
```

lambdageek:
If I try `magit-checkout` (`b b`) with point on the name of a branch it offers to checkout the numeric commit id instead of the branch name, by the way.  Any git plumbing commands that you'd like me to try?

tarsius:
Try

```lisp
(list (magit-branch-at-point)
      (magit-get-previous-branch)
      (magit-get-current-branch))
```

[...]

Issue #3106, opened 2017-06-20
git rev-parse --symbolic-full-name doesn't behave as we expect


tarsius:
`git rev-parse --symbolic-full-name` only works on names of actual refs. It does not work on something like `NAME~1`. But the changes introduced in #3101 assume that the latter works too.

If you do this:

```shell
git init testibus
cd testibus
git commit --allow-empty -m "A"; A=$(git show --format=%H)
git commit --allow-empty -m "B"; B=$(git show --format=%H)
git commit --allow-empty -m "C"; C=$(git show --format=%H)
git update-ref refs/foo/refs/heads/master C
git reset --hard A
git commit --allow-empty -m "D"; D=$(git show --format=%H)
git log --oneline --graph --all
git name-rev --refs='refs/heads/*' $A
git name-rev --refs='refs/heads/*' $B
git name-rev --refs='refs/heads/*' $C
git name-rev --refs='refs/heads/*' $D
echo "A $(git rev-parse --symbolic-full-name $(git name-rev --name-only --refs='refs/heads/*' $A))"
echo "B $(git rev-parse --symbolic-full-name $(git name-rev --name-only --refs='refs/heads/*' $B))"
echo "C $(git rev-parse --symbolic-full-name $(git name-rev --name-only --refs='refs/heads/*' $C))"
echo "D $(git rev-parse --symbolic-full-name $(git name-rev --name-only --refs='refs/heads/*' $D))"
```

we get:

```
Initialized empty Git repository in /tmp/testibus/.git/
[master (root-commit) bbad412] A
[master aebdbd7] B
[master f6c44d9] C
HEAD is now at bbad412 A
[master 224554d] D
* f6c44d9 (tag: C, refs/foo/refs/heads/master) C
* aebdbd7 (tag: B) B
| * 224554d (HEAD -> master, tag: D) D
|/  
* bbad412 (tag: A) A
A master~1
B foo/refs/heads/master~1
C foo/refs/heads/master
D master
A 
B 
C refs/foo/refs/heads/master
D refs/heads/master
```

`magit-get-shortname` now returns:

* for `A` line: `heads/master-1` when `master-1` would have been unambiguous.
* for `B` line: `heads/refs/heads/master~1`, which is wrong and causes `magit-show-diff` to fail.

tarsius:
This might be an appropriate fix, at least in the short run. But I haven't tested it yet:

```diff
diff --git a/lisp/magit-git.el b/lisp/magit-git.el
[...]

PR #3119, opened 2017-07-07
magit-rev-name: treat 'refs/<subdir>/*' as a namespace

kyleam:
```
Every 'git-name-rev --refs=PATTERN' instance in Magit has the
intention of restricting the results to refs within a particular
namespace ('refs/heads/', 'refs/remotes/', and so on).  This approach
has false positives, though, because git-name-rev matches PATTERN
against any part of the ref name, not only the beginning.  magit-wip's
names are frequent offenders (e.g., 'refs/wip/wtree/refs/heads/').

9ab6d554 (magit-rev-name: anchor the ref pattern by default,
2017-06-20) eliminated these non-namespace matches by automatically
tacking on an '--exclude=*/PATTERN', but we reverted it in b8d55867
because the --exclude option wasn't added until Git v2.13.0.

Instead let's discard git-name-rev's choice if it includes the
namespace as a substring.  Because the substring match could result
from a namespace repeated in the name (e.g., a terribly named
'refs/heads/refs/heads/foo'), we only discard the name if the ref
doesn't exist in the namespace.

A downside of this approach is that there may be some other ref in the
namespace that could be used to describe the revision, but we don't
have access to it because git-name-rev returns only its top
choice.  (The --exclude approach didn't have this downside.)  In the
magit-wip context, though, this situation appears to be uncommon.
When some commit can be named either as 'master~N' or
'wip/wtree/refs/heads/master~M', N is usually (always?) smaller than M
due to the way that wip refs are structured, so Git will choose
'master~N'.
```

Re: #3106, #3117

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -803,9 +803,24 @@ string \"true\", otherwise return nil."
                   (magit-rev-parse "HEAD")))))
 
 (defun magit-rev-name (rev &optional pattern)
-  (magit-git-string "name-rev" "--name-only" "--no-undefined"
-                    (and pattern (concat "--refs=" pattern))
-                    rev))
+  "Return a symbolic name for REV.
+PATTERN is passed to the `--refs' flag of `git-name-rev' and can
+be used to limit the result to a matching ref.  When structured
+as \"refs/<subdir>/*\", PATTERN is taken as a namespace.  In this
+case, the name returned by `git-name-rev' is discarded if it
+corresponds to a ref outside of the namespace."
+  (--when-let (magit-git-string "name-rev" "--name-only" "--no-undefined"
+                                (and pattern (concat "--refs=" pattern))
+                                rev)
+    ;; We can't use name-rev's --exclude to filter out "*/PATTERN"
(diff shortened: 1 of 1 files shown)


kyleam:
> Maybe we should name the argument prefix or namespace instead of
> pattern to make it clear that this does not behave the same way as
> git name-rev --refs.

Yeah, we probably should make that clearer.  I'm a bit hesitant to use
`prefix` or `namespace` because, if the argument doesn't look like a
namespace, we don't touch the results.  Do you think adding a
docstring explaining the behavior would do?

kyleam:
I also considered

  * adding a separate function that would accept an optional namespace
    as input and then construct the pattern, though I couldn't come up
    with a good name.

  * limiting `magit-rev-name` to a namespace without adding a separate
    function.  But then if we ever needed the output of a
    `git-name-rev` call with a non-namespace pattern, we'd need to add
    support for it and would be back in the same spot as we are now.

tarsius:
Lets just go with the doc-string for now.

kyleam:
> Lets just go with the doc-string for now.

Updated.

tarsius:
Thanks!

PR #3132, opened 2017-07-23
Don't use log's file restriction with --follow

kyleam:
Please see commit messages.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -2489,9 +2489,15 @@ without having to using one of the diff popups.
 The diffs shown in the revision buffer may be automatically restricted
 to a subset of the changed files.  If the revision buffer is displayed
 from a log buffer, the revision buffer will share the same file
-restriction as that log buffer.  Otherwise, the file restriction is
-determined by the file restriction in the repository's diff buffer, if
-it exists, and the value of the option ~magit-use-sticky-arguments~.
+restriction as that log buffer.  Note, however, that the log's file
+restriction will be ignored when ~magit-log-arguments~ includes
+~--follow~.  In this case, the ~-u~ argument of the log popup can be
+used to show the file-restricted diffs inline.
+
+If the revision buffer is not displayed from a log buffer, the file
+restriction is determined by the file restriction in the repository's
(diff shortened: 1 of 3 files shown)


tarsius:
Thanks! That was bothering me too.

tarsius:
We should probably find a key for magit-revision-toggle-file-filter. Maybe `f` in `magit-diff-refresh-popup`?

tarsius:
... after generalizing it as `magit-diff-toggle-file-filter`.

kyleam:
> We should probably find a key for magit-revision-toggle-file-filter. Maybe f in magit-diff-refresh-popup?

Sure, `f` sounds good to me.

> ... after generalizing it as magit-diff-toggle-file-filter.

That also makes sense.  I'll look into it.

tarsius:
Off topic: Is `magit-imerge` ready for Melpa? (If so I can add it. That's easier because I don't have to go through the review process.)

tarsius:
Ups, `f` is taken. Maybe `F`.

kyleam:
> Ups, f is taken.

Not by an action though.

kyleam:
Yeah, I think `magit-imerge` is ready for MELPA.  I was holding off to
see if I got any feedback from imerge users, but that probably won't
happen until it's on MELPA :)

> If so I can add it. That's easier because I don't have to go through the review process

Thanks, that'd be great.  Just a heads up: I haven't run
`package-lint` on it.

tarsius:
Not in revision and status buffers, but in diff buffers it is. It flips revisions.

kyleam:
> Not in revision and status buffers, but in diff buffers it is. It flips revisions.

Doh, indeed.  I should know that :)

tarsius:
> Doh, indeed. I should know that :)

Don't feel bad. I think I was looking at the **log** popup before suggesting `f` :wink: 
Time for bed.

tarsius:
```
1 issue found:

9:39: error: Package magit is not installable.
```

Looks like I have to teach `package-lint` about `borg`.

tarsius:
Where nicely alternating between the two topics... I think next time I will send a mail :smile_cat:

Issue #2972, opened 2017-01-25
Implement competitive collaboration tools


tarsius:
Features that facilitate collaboration are one of Magit's weak spots.  That is somewhat embarrassing, considering that this is were many other Git tools shine.

The primary reason for this is that I at some point decided to delay implementing such features because I already had enough on my hands and had to draw the line somewhere.  It was easy to draw the line here because Magit didn't offer much here to begin with, and because Magit is primarily an interface to Git and Github et al. are not Git.  I also put a lot of hope into some of the packages that set out to fill the gap.

But now the time has come for fully integrated collaboration tools.  Here's a quick run-down on what I intend to do.

- Support both web-service based, as well as mailing-list based work-flows.
  - Focus on web-services, or more specifically Github first.
    - *Which is why I am mostly talking about Github below - the competitors are implied.*
  - Feature parity is a long time goal.
- Store all data locally, most of it using Git.
  - Data is only synced periodically with the "remote", just like `git fetch`.
  - Pull requests are stored as branches.
    - All branch commands are automatically available for pr's.  No effort, major win.
    - Set the pr's target as upstream.
    - Set the pr's source as push target (using `branch.pushRemote`).
      - This allows the upstream to push to the repository of the contributor.
        - Provided the latter has not disabled this Github feature.
        - This is very useful because it allows upstream to easily fix a trivial error in an otherwise perfect pr, without having to either wait for the contributor to do it, ignore it and merge anyway, push the amended commits to the target
      - Because of that the name of the branch is fixed.
        - If it is bad, then that's just what it is.  You can still "change" it at the time of merge.
        - Name conflicts are worse, `master` being the most likely one. Conflicts will likely cause some lose of functionality, more specifically there won't be a branch.pushRemote`
      - This requires adding a remote.  But we don't have to fetch all the other irrelevant branches too if we setup `remote.fetch` accordingly.
  - Automatically add issue number to merge commits.
    - Like Github does, but give some control over formatting.
    - Since we already use the contributor's branch name, we have to store the number somewhere else, e.g. in `branch.<name>.pr`.
  - Preserve previous pr versions.
    - Probably store past versions outside `refs/heads/`, `refs/pr/<pr>/<version>` maybe?
    - The Github UI does not make previous versions of a pr easily accessible :-(
    - Unfortunately this also appears to apply to the API :-[
      - Hopefully I am overlooking something ([pulls](https://developer.github.com/v3/pulls/),
[...]

PR #3134, opened 2017-07-24
Make it easier to work with outside contributors' pull-requests

tarsius:
The new command `magit-branch-pull-request` creates a local branch for the pull-request, adds a remote if necessary, and configures the branch to make working with pull-requests from outside collaborators just as easy as working with pull-requests whose branches have been pushed to the upstream repository.

`magit-branch-delete` has been taught to cleanup when deleting a "pull-request branch". That may involve deleting the "contributor's remote".

This is only a tiny step toward the goals described in #2972, but for people like me who review all pull-requests in Magit instead of by clicking around in a browser, it is a huge win.

Before I can merge this, I have to improve `ghub.el` somewhat. Most importantly it has to be made easier to create a new token but there are also some bugs.

### DIFF
--- a/lisp/magit-collab.el
+++ b/lisp/magit-collab.el
@@ -1,0 +1,113 @@
+;;; magit-collab.el --- collaboration tools       -*- lexical-binding: t -*-
+
+;; Copyright (C) 2010-2017  The Magit Project Contributors
+;;
+;; You should have received a copy of the AUTHORS.md file which
+;; lists all contributors.  If not, see http://magit.vc/authors.
+
+;; Author: Jonas Bernoulli <jonas@bernoul.li>
+;; Maintainer: Jonas Bernoulli <jonas@bernoul.li>
+
+;; Magit is free software; you can redistribute it and/or modify it
+;; under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 3, or (at your option)
+;; any later version.
+;;
+;; Magit is distributed in the hope that it will be useful, but WITHOUT
+;; ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+;; or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+;; License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with Magit.  If not, see http://www.gnu.org/licenses.
+
+;;; Commentary:
+
+;; This library implements various collaboration tools.  These tools
+;; are only early incarnation -- implementing collaboration tools is
+;; a top priority for future development.
+
+;; Currently these tools (including `magit-branch-pull-request', which
+;; is defined elsewhere) only support Github, but support for other
+;; Git forges as well as mailing list based collaboration is in
+;; planning.
+
+;;; Code:
+
+(require 'magit)
+(require 'ghub)
+
+;;; Variables
+
+(defvar magit-github-token-scopes '(repo)
+  "The Github API scopes needed by Magit.
+
+`repo' is the only required scope.  Without this scope none of
+Magit's features that use the API work.  Instead of this scope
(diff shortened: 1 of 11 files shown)


orivej:
Thank you for this feature! Previously I had to use a custom program to make it easy to check out GitHub PRs.

I have noticed two issues:
1. When using `magit-checkout-pull-request` for the first time, `ghub` asks for the username and password, acquires a token, saves it in `.authinfo`, and `magit-checkout-pull-request` performs its job, but when using it for the second time, `ghub` prompt to acquire the token again. Only after Emacs is restarted `magit-checkout-pull-request` works without a prompt, with a token from `.authinfo`.
2. After using `magit-checkout-pull-request` for one PR, it can not be used for another PR from the same contributor if he has already deleted his branch of the first PR: Magit tries `git fetch author` which fails with `fatal: Couldn't find remote ref refs/heads/branch1`, and Magit continues to issue further git commands, all of which fail. Maybe Magit should fetch only the necessary branch?

vermiculus:
For (1), I imagine a call to `auth-source-forget-all-cached` is in order.

orivej:
Thank you! I have confirmed that after `auth-source-forget-all-cached` Magit no longer prompts to create the token and uses the one from `.authinfo`.

vermiculus:
Interestingly, this *should* already be handled:
```elisp
(defun ghub--token (host username package &optional nocreate)
  (let ((user (ghub--ident username package)))
    (or (ghub--auth-source-get :secret :host host :user user)
        (progn
          ;; Auth-Source caches the information that there is no
          ;; value, but in our case that is a situation that needs
          ;; fixing so we want to keep trying by invalidating that
          ;; information.
          (auth-source-forget (list :host host :user user))
          (and (not nocreate)
               (ghub--confirm-create-token host username package))))))
```

tarsius:
I have tested this function, but maybe I later broke it again.

Try `(auth-source-forget (list :max 1 :host host :user user))`.

tarsius:
For (2), I ran into that too (and consider it a misfeature of git). But I forgot to write a command to fix this situation before I merged this.

orivej:
Indeed, this fixes the first issue:
```patch
@@ -452 +452 @@ (defun ghub--token (host username package &optional nocreate)
-          (auth-source-forget (list :host host :user user))
+          (auth-source-forget (list :max 1 :host host :user user))
```

vermiculus:
> I have tested this function, but maybe I later broke it again.
>
> Try `(auth-source-forget (list :max 1 :host host :user user))`.
>
> This might also differ between Emacs versions.

Any reason why specifying `:max 1` makes a difference?  Is it because of `(apply #'auth-source-search :max 1 spec)` in `ghub--auth-source-get`?

tarsius:
Yes, and the reason why I had that there is that I was still using 25.3, which did consider irrelevant keys, and now I am using 26.1, which doesn't. At least I assume that's how things played out. I have to verify this of course.

tarsius:
(2) is fixed now. The behavior did differ between 25.3 and 26.1 as I assumed above.

lisp/magit-collab.el
```diff
@@ -0,0 +1,129 @@
...
+;;;###autoload
+(defun magit-browse-pull-request (pr)
+  "Visit pull-request PR using `browse-url'.
+
+Currently this only supports Github, but that restriction will
+be lifted eventually to support other Git forges."
+  (interactive (list (magit-read-pull-request "Visit pull request")))
+  (browse-url (format "https://github.com/%s/pull/%s"
+                      (--> pr
+                           (cdr (assq 'base it))
```
npostavs: Just FYI, using `it` below the top level expression requires a recent dash version which includes the patch from magnars/dash.el#221. Not necessarily a problem, but as of now, there is no tagged version of dash with the fix, so it could affect releasing.

PR #3136, opened 2017-07-31
Fix typos in Magit popup manual

mrkkrp:
The document says that the package will be deprecated eventually, what should I use instead? Does current version of Magit use `magit-popup`?

### DIFF
--- a/Documentation/magit-popup.org
+++ b/Documentation/magit-popup.org
@@ -40,7 +40,7 @@ suffix commands.
 
 Invoking such a prefix command displays a popup buffer which lists the
 associated infix arguments and suffix commands.  In that buffer each
-argument is prefixes with the key sequence that can be used to toggle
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks! Merged after regenerating `magit-popup.texi`.

mrkkrp:
Ping. Sorry, but you did not answer the question. Does current version of Magit use this library? I wanted to create a package based on this interface, but I'm hesitant to do that if it'll be deprecated soon.

tarsius:
Yes, Magit still uses this. But I also still plan to move away from it eventually.

I would tone down the warning in the manual a bit. The code is a mess, but the interface is solid and it works reliably. Also I plan to keep it alive even after Magit has switched away. At that point I will only make bugfixes and minor improvements, but not develop any new features. (But that's pretty much how I maintain this now already anyway.)

PR #3141, opened 2017-08-02
Speedup inserting module sections

tarsius:
Please see the commit messages.

### DIFF
--- a/lisp/magit-section.el
+++ b/lisp/magit-section.el
@@ -1196,5 +1196,20 @@ again use `remove-hook'."
         (set hook value)
       (set-default hook value))))
 
+(defun magit-run-section-hook (hook)
+  "Run HOOK, warning about invalid entries."
+  (--if-let (-remove #'functionp (symbol-value hook))
+      (progn
+        (message "`%s' contains entries that are no longer valid.
+%s\nUsing standard value instead.  Please re-configure hook variable."
+                 hook
+                 (mapconcat (lambda (sym) (format "  `%s'" sym)) it "\n"))
+        (sit-for 5)
+        (defvar magit--hook-standard-value nil)
+        (let ((magit--hook-standard-value
+               (eval (car (get hook 'standard-value)))))
+          (run-hooks 'magit---hook-standard-value)))
+    (run-hooks hook)))
+
 (provide 'magit-section)
 ;;; magit-section.el ends here
--- a/Documentation/RelNotes/2.11.0.txt
+++ b/Documentation/RelNotes/2.11.0.txt
@@ -97,6 +97,10 @@ Changes since v2.10.3
   `magit-module-sections-hook' and `magit-module-sections-nested',
   to allow tuning the behavior of this function.  #3141
 
+* Renamed function `magit-insert-submodules' to
+  `magit-insert-modules-overview' for consistency with other
+  `magit-insert-modules-*' functions.  #3141
+
 Fixes since v2.10.3
 -------------------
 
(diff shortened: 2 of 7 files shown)


lisp/magit-section.el
```diff
@@ -1196,5 +1196,20 @@ again use `remove-hook'."
         (set hook value)
       (set-default hook value))))
 
+(defun magit-run-section-hook (hook)
+  "Run HOOK, warning about invalid entries."
+  (--if-let (--filter (not (functionp it)) (symbol-value hook))
```
kyleam: I prefer `(-remove #'functionp (symbol-value hook))`.

lisp/magit-section.el
```diff
@@ -1196,5 +1196,20 @@ again use `remove-hook'."
...
+      (progn
+        (message "`%s' contains entries that are no longer valid.
+%s\nUsing standard value instead.  Please re-configure hook variable."
+                 hook
+                 (mapconcat (lambda (sym) (format "  `%s'" sym)) it "\n"))
```
kyleam: Depending on the value of `text-quoting-style`, you can get
inconsistent quotes here.  Using `format-message` could fix that, but
this isn't available until Emacs 25.

Just a note, though, because I don't think it's worth adding a
compatibility kludge in this case.

lisp/magit-section.el
```diff
@@ -1196,5 +1196,20 @@ again use `remove-hook'."
...
+        (sit-for 5)
+        (defvar magit--hook-standard-value nil)
+        (let ((magit--hook-standard-value
+               (-filter #'functionp (eval (car (get hook 'standard-value))))))
```
kyleam: I'm not sure what this is guarding against.  If `standard-value` is
the value set in the defcustom, can't we be confident that we've set
it to an exisiting function?  Is the issue that some functions might
not be loaded yet?

lisp/magit-section.el
tarsius: I briefly forgot about that.

lisp/magit-section.el
tarsius: Let's just leave it as it is. I am wondering though what would be the proper way if we could assume Emacs 25?

kyleam: > I am wondering though what would be the proper way if we could assume Emacs 25

Nothing comes to mind aside from using `format-message`.

lisp/magit-section.el
tarsius: It's useful when implementing a new interter and having to switch branches to do something else. That might lead to a member that is invalid on one branch. But this isn't really useful for users, and should not happen to them, so this works around such issues silently.

tarsius: But, I should probably drop it. If there actually was a permanent issue, then this would hide it.

lisp/magit-section.el
tarsius: Err, sorry I missed that earlier.

lisp/magit-section.el
kyleam: I see

kyleam: > But I should probably drop this. If there actually was a permanent issue, then this would hide it.

Yep, I agree.

tarsius:
Thanks for the review!

kyleam:
Looks like the branch you merged dropped my fixup commits. (Sorry, I
should have left a comment explicitly mentioning them.)  I think most
of these are obviously correct changes, so I've pushed these to
jb/speedup-modules-fixup assuming you'll want to add a squashed commit
on top.

tarsius:
I though I squashed those yesterday... fixing.

tarsius:
Okay I have added a new commit with those changes now. Thanks for noticing this.

kyleam:
No problem.  Thanks for the quick fix.

PR #3155, opened 2017-08-29
magit-run-git{k,-gui*}: Use magit-process-file

npostavs:
```
Using call-process fails when the PATH doesn't contain the git-gui
executable (e.g., when Git for Windows was installed with only the
cmd/ directory on PATH).
```

I'm not entirely sure if the use of `call-process` vs `process-file` is just an oversight, or it was meant to avoid opening a gui on the remote (when using tramp). In the latter case we might need to add a `magit-call-process` function.


tarsius:
I think it was just an oversight.

PR #3156, opened 2017-09-06
magit-switch-to-magit-buffer: new function

NicholasTD07:
My magit workflow looks like this:
```
... -> do some work -> magit status -> diff -> commit -> bury the magit status buffer -> ...
(repeat)
```

When using `magit-status` with `magithub` installed, `magithub` fetches data every time when I run `magit-status`, even when I have a magit status buffer open and `magithub` has cached data. Because `magithub` fetches data synchronously, I'd have to stop and wait for `magithub` to finish fetching. You can see this can get quite annoying quickly.

So I thought, why not create a function that switch to the magit status buffer if it exists and create one if there's no such buffer. :)

PS.

Thanks for the great work you pour into `magit` and `magithub`. They are awesome!

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -33,6 +33,7 @@
 (require 'dash)
 
 (require 'magit-section)
+(require 'magit-status)
 (require 'magit-git)
 (require 'magit-popup)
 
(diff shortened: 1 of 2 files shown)


kyleam:
Thanks for the PR.  My feeling is that a function like this is better
suited for a user's personal configuration, or perhaps magithub if it
solves a common problem for magithub users.  (I've never used
magithub, so I'm not sure if there's a better way for you to solve
your issue.)

NicholasTD07:
@kyleam Yeah, I agree. I also realized there's a down side of this function: it does not refresh the status buffer if there's one. I will close this PR. :) Thanks for the comment.

PR #3159, opened 2017-09-07
Apply ansi colors on the output of git process

Fuco1:
This will make the `*magit-process*` buffer a little bit more readable in case git runs hooks which produce colored output, as well as the output of git itself (which I suppose you now run without colors).

It is still not ideal, for example we run eslint on prehook and it prints a lot of control characters like this

```
 ❯ Running tasks for *.js
   ⠙ build/jscs.sh
[?25l[1000D[1A[1000D[1A[1000D[1A[1000D[1A[1000D ↓ Running tasks for *.{php,phpt} [skipped]
```
(and about 30 lines of the same...), but at least the final report gets colorized and readable :)

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -31,6 +31,7 @@
 
 ;;; Code:
 
+(require 'ansi-color)
 (require 'cl-lib)
 (require 'dash)
 
@@ -817,6 +818,7 @@ as argument."
                                         'magit-process-ok
                                       'magit-process-ng)))
           (set-marker-insertion-type marker t))
+        (ansi-color-apply-on-region (point) (point-max))   
         (if (= (magit-section-end section)
                (+ (line-end-position) 2))
             (save-excursion


Fuco1:
Before: 

After:

lisp/magit-process.el
```diff
@@ -817,6 +818,7 @@ as argument."
                                         'magit-process-ok
                                       'magit-process-ng)))
           (set-marker-insertion-type marker t))
+        (ansi-color-apply-on-region (point) (point-max))
```
tarsius: Only the text between `magit-section-content` and `magit-section-end` needs to be updated.

Fuco1: Doesn't the log always append?  I assumed that when we are processing this we are adding new section to the log.  I can add the change of course, but I'm interested in how this works.

Fuco1:
I can add an opt-in switch if you want.  Though I think that performance impact is going to be very very negligible.  I use `ansi-color-apply-on-region` on outputs of huge shell commands/compile buffers and I've never noticed any difference.

tarsius:
> I can add an opt-in switch if you want. Though I think that performance impact is going to be very very negligible.

Let's do that but use a variable instead of an option. If it turns out that there's no noticeable impact then we can change the default or even remove it. If we use an option then there might be an expectation that we won't remove it again.

Fuco1:
I don't understand the difference between option and variable.

tarsius:
The idea is that in the short run we don't advertise the availability of this feature (by using an option) and instead just tell those users about the variable when they ask. Yes, that might be a little silly, but it would allow testing this for a while and to remove it again if necessary.

basil-conto:
> I don't understand the difference between option and variable

Just in case the confusion arises from terminology, rather than intent: "option" usually refers to a `defcustom` and "variable" to a `defvar`. The latter is, of course, inherently more internal and subject to implementation change than the user-visible former.

tarsius:
I have pushed something to `master`. For now you have to set `magit-process-finish-apply-ansi-colors` to `t`.

Fuco1:
@tarsius Thanks, I'll enable it and keep testing.  I've been using this patch for a month and so far I had no problems (and I run quite heavy hooks in terms of asci escapes).

tarsius:
Eventually want to propertize the text as it is coming in.

PR #3160, opened 2017-09-07
Improve the looks of the submodules overview section

stepnem:


tarsius:
> This solution has seemed to only care about the "vNUM" case, but the world of version strings is much more colourful. Tags often start with more letters (or words such as "release" or "debian"), not just one followed by numbers and punctuation, so the alignment wasn't perfect anyway.

At least for Emacs the current approach works very well. Out of the 67 packages that I have installed only do not either use the `v` prefix or no prefix at all.

> Rather than trying to handle one special case, align everything to the same column and be done with it.

If you care enough about this then please add a new option but keep the current behavior as the default. (Please also update the manual and release notes.)

tarsius:
> An abbreviated commit SHA-1 is better than nothing.

It makes the output look more regular - that's the only benefit I can see. That has the drawback that it is now *harder* to spot the modules that don't even have a single tag. Being able to spot those modules is useful to me because in some cases (e.g. inside `~/.emacs.d~) it serves as an implicit TODO list (in the case of `~/.emacs.d~, the untagged modules are things I have not polished enough to release them yet).

I want to keep the current behavior and also don't think that an option should be added.

tarsius:
> Compute the path column width according to the longest path, with a
cap at half the window width. [...]

This sounds like an improvement, but I haven't tried it out yet.

stepnem:
> At least for Emacs the current approach works very well. Out of the 67
> packages that I have installed only do not either use the `v` prefix
> or no prefix at all.

There's something missing in your statement (org? undo-tree? emacs
itself?), but the particular repos or their number doesn't really
matter. Hardcoding a special case like this seems just wrong to me.

>> Rather than trying to handle one special case, align everything to the same column and be done with it.
>
> If you care enough about this then please add a new option but keep
> the current behavior as the default. (Please also update the manual
> and release notes.)

Yeah, I don't know. If an option, instead of a boolean it might perhaps
be more "sophisticated", e.g. defining the prefixes to be handled
magically like this, but I continue to think it's not needed at all, so
I may as well just keep the trivial change locally, especially if I'm
going to do that with the other changes as well.

stepnem:
> It makes the output look more regular - that's the only benefit I can
> see. That has the drawback that it is now *harder* to spot the modules
> that don't even have a single tag. Being able to spot those modules is
> useful to me because in some cases (e.g. inside `~/.emacs.d~) it
> serves as an implicit TODO list (in the case of `~/.emacs.d~, the
> untagged modules are things I have not polished enough to release them
> yet).

Hm. To me the 3rd column in submodule overview is just "this is where
the repo is at", which seems a more obvious notion than what you
describe, but I might be missing something. I've been using it e.g. for
copying the tag/SHA when needed, without having to enter the
subrepository or resort to other means. And after all, the SHA seems
still readily distinguishable from a real tag to me, though perhaps not
as conspicuous as the blank (but could be made even more conspicuous by
some kind of highlighting).

tarsius:
> so I may as well just keep the trivial change locally

Please do that for a while. I'll get back to this, but it's a pretty low priority.

tarsius:
> And after all, the SHA seems still readily distinguishable from a real tag to me, though perhaps not as conspicuous as the blank (but could be made even more conspicuous by some kind of highlighting).

Please do that, probably using `(or ... (magit-rev-format "%h"))` instead of `--always`.

stepnem:
> Please do that, probably using `(or ... (magit-rev-format "%h"))`
> instead of `--always`.

I have now rebased the branch and edited the relevant commit to that
effect. The hashes are highlighted (or rather lowlighted, so to speak)
using the magit-hash face.

tarsius:
I have written out the *strict alignment* commit and replaced it with ceb5b0dc08cac055939c88df889a012f6e372b89.

> Yeah, I don't know. If an option, instead of a boolean it might perhaps be more "sophisticated", e.g. defining the prefixes to be handled magically like this,

Maybe yes, but lets stick with this simple (and undocumented) variable for now. If others complain about the weird alignment, then I will look into this again.

stepnem:
OK, thanks!

tarsius:
Thanks to you too!

PR #3167, opened 2017-09-12
Fix typo in key chord of git-commit-save-message

jeremymeng:
Looks like it should be `C-c, M-s`, not `C-s, M-s`.


tarsius:
Thanks!

I have amended your commit to also update the `texi` file and slightly change the commit message.

PR #3175, opened 2017-09-18
add 'd' as alias key for git-rebase-kill-line

pau4o:
In git-rebase-todo buffer I have:
# d, drop — drop the commit at point
Unfortunately trying to kill line with pressing 'd' shows 'd is undefined'

### DIFF
--- a/lisp/git-rebase.el
+++ b/lisp/git-rebase.el
@@ -152,6 +152,7 @@
            (define-key map (kbd "M-p") 'git-rebase-move-line-up)
            (define-key map (kbd "M-n") 'git-rebase-move-line-down)
            (define-key map (kbd   "c") 'git-rebase-pick)
+           (define-key map (kbd   "d") 'git-rebase-kill-line)
            (define-key map (kbd   "k") 'git-rebase-kill-line)
            (define-key map (kbd "C-k") 'git-rebase-kill-line)))
     (define-key map (kbd "e") 'git-rebase-edit)


kyleam:
> In git-rebase-todo buffer I have:
> d, drop — drop the commit at point

I'm confused.  Doesn't it say

    # k        drop the commit at point

by default?

If you want to bind `git-rebase-kill-line` to "d" as well,

    (define-key git-rebase-mode-map "d" 'git-rebase-kill-line)

should work fine.

If you don't want it on "k" (or if the help message showing "k" for
`git-rebase-kill-line` bothers you), you can also

    (define-key git-rebase-mode-map "k" nil)

tarsius:
Yes, please add this binding locally. `k` is usually used for "delete" (i.e. "kill") commands in Emacs, and I would like to reserve `d` for a diff related command.

pau4o:
First of all I do agree that `k` is better suited for this purpose, but please read my investigations below.

Now I'm confused too. When I start with `emacs --no-init-file` and load magit manually from my packages folder then it say:
```
# k        drop the commit at point
```
I have same hint when run `LANG=C emacs`

My default $LANG is `LANG=bg_BG.UTF-8` and if i run just `emacs` then given hint is:
```
# d, drop — прескачане на подаването
```
There are other differences between two language versions of hints in **git-rebase-todo** buffer.
I can give full texts from hints section from both versions.

It turns out that hints are locale dependant. So now questions are:

- is it desirable magit-rebase to be language dependant (inconsistent in other words)?
- is it possible to configure via magit to not use localised version?


Can you help me a bit? (I'm not so skilled with elisp to figure out answers by my self.)

kyleam:
@pau4o Thanks for those details.  That's what we were missing.

> is it desirable magit-rebase to be language dependant (inconsistent in other words)?

No, that's a bug in how the help message is generated.

> Can you help me a bit? (I'm not so skilled with elisp to figure out answers by my self.)

I'll have a look in the next couple days if someone doesn't get to it
first.

tarsius:
> is it desirable magit-rebase to be language dependant (inconsistent in other words)?

I don't think so but apparently `git` is doing that. Some of the key bindings are printed by `git` itself and I was not aware that the keys change in addition to the descriptions.

Since the rest of Magit is not localized, I think we should just force English.

kyleam:
> Some of the key bindings are printed by git itself and I was not aware that the keys change in addition to the descriptions depending on the language setting.

The keys don't change, just the descriptions.  For example, here's the
relevant translation entry in Git's po/bg.po:

```
"Команди:\n"
" p, pick — прилагане на подаването\n"
" r, reword — прилагане на подаването, но смяна на съобщението му\n"
" e, edit — прилагане на подаването и спиране при него за още промени\n"
...
```

What's happening in the case above is that
`git-rebase-mode-show-keybindings` is not processing the help message
at all because it fails to find "# Commands:".

The format of the commands ("p, pick", ...) is actually quite stable.
Although some commands have been added, the leading lines haven't
changed since interactive rebase was introduced.  Because of this, we
could fix this issue with something like

```diff
diff --git a/lisp/git-rebase.el b/lisp/git-rebase.el
index bead77c0..afb7034c 100644
--- a/lisp/git-rebase.el
+++ b/lisp/git-rebase.el
@@ -535,15 +535,17 @@ (defun git-rebase-mode-show-keybindings ()
       (goto-char (point-min))
       (when (and git-rebase-show-instructions
                  (re-search-forward
-                  (concat git-rebase-comment-re " Commands:\n")
+                  (concat git-rebase-comment-re " p, pick")
                   nil t))
+        (goto-char (line-beginning-position))
         (--each git-rebase-command-descriptions
           (insert (format "%s %-8s %s\n"
                           comment-start
                           (substitute-command-keys (format "\\[%s]" (car it)))
                           (cdr it))))
         (while (re-search-forward (concat git-rebase-comment-re
-                                          "\\(  ?\\)\\([^,],\\) \\([^ ]+\\) = ")
+                                          "\\(  ?\\)\\([^\n,],\\) "
+                                          "\\([^\n ]+\\) ")
                                   nil t)
           (let ((cmd (intern (concat "git-rebase-" (match-string 3)))))
             (if (not (fboundp cmd))
```

The does lead to a weird mix of Magit's help in English and Git's help
in $LANG, though.

Or, as @tarsius mentions, we could force all the rebase help text to
be English.  Perhaps

```diff
diff --git a/lisp/magit-sequence.el b/lisp/magit-sequence.el
index 91372687..e27731f7 100644
--- a/lisp/magit-sequence.el
+++ b/lisp/magit-sequence.el
@@ -404,6 +404,8 @@ (defun magit-rebase-interactive-1 (commit args message &optional editor)
       (let ((process-environment process-environment))
         (when editor
           (push (concat "GIT_SEQUENCE_EDITOR=" editor) process-environment))
+        (push "LANG=C" process-environment)
+        (push "LC_ALL=C" process-environment)
         (magit-run-git-sequencer "rebase" "-i" args
                                  (unless (member "--root" args) commit)))
     (magit-log-select
```

I think this is the lowest level where we can set these.  We could go
higher up in the process call chain, but I'm concerned we'd be
needlessly going against users' settings and deviating from the Git
help output that a user expects to see (e.g., in the process buffer).
Even at this level, I don't know if there are unintended side effects
or if this is a bad solution for other reasons.  (I've only tested it
on a GNU/Linux system.)

Thoughts?

tarsius:
> Thoughts?

Given your concerns about setting `LANG` etc., let's just go with `re " p, pick"`.

pau4o:
I updated magit after @kyleam closed #3175 
Now all work as expected.

Thank you all for good work.

PR #3183, opened 2017-10-01
Document HEAD distances shown in status header "Tags" line.

jimblandy:


tarsius:
Thanks!

Next time please edit the `*.org` file instead. You may also regenerate the `*.texi` from that, but that requires some additional tools, so you can skip that and I will just amend to your commit.

I have edited `magit.org` myself in this case, see f737bc32ede569a9748803247d1c147cd1d48bd5.

PR #3186, opened 2017-10-03
Add magit-remote-prune

thejan2009:
This is the only command I frequently use that doesn't seem to be already included in magit. 

As a side note, thanks for developing and maintaining this awesome package.


tarsius:
You can already prune from the fetching popup, but I guess it doesn't hurt to add this here also.

Please also add an entry to the release notes (after looking at some old ones to see how to word that).

thejan2009:
To be honest, I wouldn't have bothered if I knew/noticed this feature already existed.

I've updated the commit, but feel free to close this PR if the feature seems unnecessarily duplicated.

tarsius:
Thanks. Yeah, I think I'll leave it out.

PR #3190, opened 2017-10-07
Make highlighting keywords in brackets optional

bastianbeischer:
Highlighting keywords in brackets is a useful feature in case the convention in the project is to use such brackets to mark references to issues and PRs (for example on github).

However, in our project such brackets are often used for other purposes and I dislike the highlighting, so I would like to make it optional. As far as I know, customizing the magit-keyword face does not work, because it is used in two different places where the surrounding characters have differing faces so I wouldn't know how to set the magit-keyword face such that it resolves to face A (in magit-log) and face B (in magit-diff). If you have an idea how to do this the patch here wouldn't be required and I would appreciate hearing it.

I would have preferred to add only one defcustom, but I didn't find a good place to put it. I guess it would have to go into magit.el since it would be used both in magit-log.el and magit-diff.el. But there are only two other defcustoms in magit.el so it doesn't seem like the proper way to do it. Let me know if you have an idea for a solution with only one defcustom.

### DIFF
--- a/Documentation/RelNotes/2.11.1.txt
+++ b/Documentation/RelNotes/2.11.1.txt
@@ -28,6 +28,11 @@ Changes since v2.11.0
   substitute for `magit-checkout'.  Setup instructions can be found in
   the manual.  #3104
 
+* Added new customizable options `magit-log-highlight-keywords' and
+  `magit-diff-highlight-keywords' which control whether text inside
+  brackets is highlighted in magit-log and magit-diff buffers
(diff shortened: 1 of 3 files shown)


tarsius:
Thanks! Please mention the new options in the release notes.

bastianbeischer:
> Thanks! Please mention the new options in the release notes.
Sure, done!

tarsius:
Thanks!

PR #3191, opened 2017-10-08
Custom buffer lock values

vermiculus:
Needed for many of the features listed in vermiculus/magithub#72.

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -866,6 +866,8 @@ See also `magit-toggle-buffer-lock'.")
 (cl-defun magit-buffer-lock-value
     (&optional (mode major-mode)
                (args magit-refresh-args))
+  "Find an appropriate buffer lock value for MODE under ARGS.
+See also `magit-buffer-lock-functions'."
   (cl-case mode
     (magit-cherry-mode
      (-let [(upstream head) args]
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -840,6 +840,29 @@ latter is displayed in its place."
                                   major-mode value)))
       (user-error "Buffer has no value it could be locked to"))))
 
+(defvar magit-buffer-lock-functions nil
+  "Provide buffer-locking support for third-party modes.
+An alist of symbols to functions.
+
+The symbol must be the major-mode the locked buffer will have.
+
+The function must take a list of arguments and return a value
+that identifies the buffer (i.e., its 'lock value').  If the
+third-party mode is invoked as
+
+    (magit-mode-setup-internal #\\='my-mode \\='(1 2 3) t)
+
+the function will be invoked as
+
+    (apply lock-func \\='(1 2 3))
+
+if the cons (my-mode . lock-func) is in this list.
+
+This variable is intended for third-party extensions;
+`magit-buffer-lock-value' implements all built-in behavior.
+
+See also `magit-toggle-buffer-lock'.")
+
 (cl-defun magit-buffer-lock-value
     (&optional (mode major-mode)
                (args magit-refresh-args))
@@ -870,7 +893,10 @@ latter is displayed in its place."
     ((magit-reflog-mode   ; (ref ~args)
       magit-stash-mode    ; (stash _const _args _files)
       magit-stashes-mode) ; (ref)
-     (car args))))
+     (car args))
+    (t
+     (--when-let (cdr (entry (assq mode magit-buffer-lock-functions)))
+       (apply it args)))))
 
 (defun magit-mode-bury-buffer (&optional kill-buffer)
   "Bury the current buffer.
--- a/Documentation/RelNotes/2.12.0.txt
+++ b/Documentation/RelNotes/2.12.0.txt
@@ -47,6 +47,9 @@ Changes since v2.11.0
 * Kewords in brackets are now being highlighted in `magit-refs-mode'
   buffers like they already were in logs.  #3179
 
+* Added new variable `magit-buffer-lock-functions' to support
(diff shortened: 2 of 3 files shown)


lisp/magit-mode.el
```diff
@@ -837,9 +837,31 @@ latter is displayed in its place."
...
+
+The function must take a list of arguments and return a value
+that identifies the buffer (i.e., its 'lock value').  If the
+third-party mode is invoked as
+
+    \(my-mode 1 2 3\)
+
+the function will be invoked as
+
+    \(apply lock-func '\(1 2 3\)\)
```
basil-conto: The elisp [documentation tips](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html) don't make it clear whether it's an open-paren as the *first* character or the first *non-whitespace* character that is problematic, but the other parens surely don't need to be escaped.

npostavs: It's only parens in column 0, it's explained in here: https://www.gnu.org/software/emacs/manual/html_node/emacs/Left-Margin-Paren.html

vermiculus: I wasn't clear on this either – fixed, thanks!

lisp/magit-mode.el
```diff
@@ -837,9 +837,31 @@ latter is displayed in its place."
...
+
+if the cons (my-mode . lock-func) is in this list.
+
+See `magit-toggle-buffer-lock' and `magit-buffer-lock-value'.")
+
 (cl-defun magit-buffer-lock-value
     (&optional (mode major-mode)
                (args magit-refresh-args))
+  "Find an appropriate buffer lock value for MODE under ARGS.
+See also `magit-buffer-third-party-mode-lock-functions'. "
```
kyleam: Since the docstring mentions `magit-buffer-third-party-mode-lock-functions`,
I'd prefer the commits be reordered so that the docstring commit comes
after the commit that introduces the variable.

nitpick: There's a trailing space at the end of the docstring.

lisp/magit-mode.el
```diff
@@ -867,7 +889,11 @@ latter is displayed in its place."
     ((magit-reflog-mode   ; (ref ~args)
       magit-stash-mode    ; (stash _const _args _files)
       magit-stashes-mode) ; (ref)
-     (car args))))
+     (car args))
+    (t
+     (-when-let* ((entry (assq mode magit-buffer-third-party-mode-lock-functions))
+                  (func  (cdr entry)))
+       (apply func args)))))
```
kyleam: nitpick: You could drop the `func` binding here.  Perhaps

```emacs-lisp
(--when-let (cdr (assq mode magit-buffer-third-party-mode-lock-functions))
  (apply it args))
```

Documentation/magit.texi
```diff
@@ -7949,7 +7949,7 @@ packages, without having to depend on Magit.  These libraries are
 described in separate manuals, see @ref{Top,,,with-editor,} and
 @ref{Top,,,magit-popup,}.
 
-If you are trying to find an unused key that you can bound to a
+If you are trying to find an unused key that you can bind to a
 command provided by your own Magit extension, then checkout
```
kyleam: magit.texi is generated from magit.org, so please edit that instead.
If you don't have ox-texinfo+.el set up, you don't have to worry about
generating magit.texi.  We can do it on our end and amend your commit.

tarsius:
Instead of `magit-buffer-third-party-mode-lock-functions ` I would prefer `magit-buffer-lock-functions`. Just mention in the documentation that built-in behavior is implemented directly in `magit-buffer-lock`.

Documentation/magit.texi
vermiculus: Ah right, I forgot we started that! Will do.

lisp/magit-mode.el
vermiculus: Yup, I can do that. I think when I was writing it the first time, I wanted to distinguish between 'the key wasn't found' and 'the key had a nil value'.

Do you think this distinction would be useful? If so, I can signal an error or something. (This is assuming we'll never hit this code with an unhandled mode; this would be more of an assertion than anything else.)

lisp/magit-mode.el
vermiculus: I'll take care of both, good catch!

lisp/magit-mode.el
```diff
@@ -837,9 +837,31 @@ latter is displayed in its place."
...
+  "Provide buffer-locking support for third-party modes.
+An alist of symbols to functions.
+
+The symbol must be the major-mode the locked buffer will have.
+
+The function must take a list of arguments and return a value
+that identifies the buffer (i.e., its 'lock value').  If the
+third-party mode is invoked as
+
+    (magit-mode-setup-internal #'my-mode '(1 2 3) t)
```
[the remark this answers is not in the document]
vermiculus: Can you give an example of what it should look like?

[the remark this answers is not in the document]
kyleam: Sure.  See `magit-pull-and-fetch-popup`.

lisp/magit-mode.el
kyleam: > Do you think this distinction would be useful?

no, but I also don't see any harm in that distinction being made if you prefer

vermiculus: I changed to use your `--when-let` form. I'm not a fan of anaphoric forms myself, but I see it's a prevailing style in this codebase.

vermiculus:
Hopefully I've addressed everyone's comments; let me know if I missed something 😄

vermiculus:
Rebased onto `master`.

tarsius:
Thanks a lot!

PR #3192, opened 2017-10-08
Handle empty credential.cache value

raxod502:
It is an intended use case for Git to override previously specified `credential.cache` values by providing an empty string as a new value. Fix the definition of `magit-credential-cache-daemon-socket` so that it handles this case, instead of throwing an error.


raxod502:
Thanks. Fixed.

tarsius:
Thanks!

tarsius:
(It's merged after an additional rebase.)

PR #3193, opened 2017-10-09
Remove 2 trailing double-quotes in documentation

DamienCassou:


tarsius:
@DamienCassou you are not really removing two double-quotes but two incarnations of the same one ;-) Since I was touching the commit for demonstration purposes anyway, I have dropped the "2".

@kyleam @npostavs When regenerating the manual for an existing commit you now have to use `AMEND=t make texi`, which will update the version strings to `TAG-gPREVHASH+1`. If you used just `make texi` then `TAG-gHASH_FOR_REPLACED_COMMIT+1` would be used.

Of course for trivial changes like this you don't actually have to update the version strings at all.

I will likely make it possible to bump the version strings in just one of the manuals.

kyleam:
I can't recall: Do we track magit.texi for compatibility with MELPA?
Or because you don't want to assume a machine building the source has
ox-texinfo+?  Some other reason?

tarsius:
Primarily for Melpa, but `make` users not needing `ox-texinfo+` is another benefit.

kyleam:
Yeah, not pretty, but what can you do

tarsius:
Well in my function as a regular Melpa contributor my opinion is that there are much more important things to do than to support all the weird things upstreams do to generate their documentation :stuck_out_tongue_winking_eye:

PR #3194, opened 2017-10-09
Align header-line contents with the text area

fice-t:
A space is not a robust alignment; the left fringe may not be the same pixel
width as a space, and there may not even be a fringe present (e.g., terminal
Emacs).

In my current installation, the fringe width is one pixel less than the width of a space, and once you see that one pixel offset you just can't ignore it.

### DIFF
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -603,12 +603,21 @@ Unless optional argument KEEP-EMPTY-LINES is t, trim all empty lines."
 (defun magit-set-header-line-format (string)
   "Set the header-line using STRING.
 Propertize STRING with the `magit-header-line' face if no face is
-present, and make sure it aligns with the text area."
+present, and pad the left and right sides of STRING equally such
+that it will align with the text area."
   (let ((header-line
          (concat (propertize " "
                              'display
                              '(space :align-to 0))
-                 string)))
+                 string
+                 (propertize
+                  " "
+                  'display
+                  `(space :width (+ left-fringe
+                                    left-margin
+                                    ,@(and (eq (car (window-current-scroll-bars))
+                                               'left)
+                                           '(scroll-bar))))))))
     (setq header-line-format
           (if (text-property-not-all 0 (length header-line) 'face nil header-line)
               header-line
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -605,25 +605,42 @@ Unless optional argument KEEP-EMPTY-LINES is t, trim all empty lines."
 Propertize STRING with the `magit-header-line' face if no face is
 present, and pad the left and right sides of STRING equally such
 that it will align with the text area."
-  (let ((header-line
-         (concat (propertize " "
-                             'display
-                             '(space :align-to 0))
-                 string
-                 (propertize
-                  " "
-                  'display
-                  `(space :width (+ left-fringe
-                                    left-margin
-                                    ,@(and (eq (car (window-current-scroll-bars))
-                                               'left)
-                                           '(scroll-bar))))))))
+  (let* ((header-line
+          (concat (propertize " "
+                              'display
+                              '(space :align-to 0))
+                  string
+                  (propertize
+                   " "
+                   'display
+                   `(space :width (+ left-fringe
+                                     left-margin
+                                     ,@(and (eq (car (window-current-scroll-bars))
+                                                'left)
+                                            '(scroll-bar)))))))
+         (len (length header-line)))
     (setq header-line-format
-          (if (text-property-not-all 0 (length header-line) 'face nil header-line)
-              header-line
+          (if (text-property-not-all 0 len 'face nil header-line)
+              (let ((face (get-text-property 0 'face string)))
+                (when (and (atom face)
+                           (magit-face-property-all face string))
+                  (add-face-text-property 0 1 face nil header-line)
+                  (add-face-text-property (1- len) len face nil header-line))
+                header-line)
             (propertize header-line
                         'face
                         'magit-header-line)))))
+
+(defun magit-face-property-all (face string)
+  "Return non-nil if FACE is present in all of STRING."
+  (cl-loop for pos = 0 then (next-single-property-change pos 'face string)
+           unless pos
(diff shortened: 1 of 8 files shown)


fice-t:
I also thought that it would be nice if the right side of header lines were padded as well. Unfortunately I couldn't find an easy and robust solution, but I doubt people will be switching the scroll bar status in a magit buffer often.

I opened an [Emacs bug report](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=28771) partially about this, but until the feature is merged I think this is the best that can be done without adding a hook to refresh the header line.

@tarsius 
I missed a couple of header lines that aren't propertized in `magit-header-face`.

It looks like `magit-log-select`'s header-line should use `magit-propertize-header-line`, but `magit-stash-refresh-buffer`'s header-line shouldn't have its faces overridden. What do you think?

lisp/magit-utils.el
```diff
@@ -604,11 +604,20 @@ Unless optional argument KEEP-EMPTY-LINES is t, trim all empty lines."
...
+   (concat (propertize " "
+                       'display
+                       '(space :align-to 0))
+           string
+           (propertize
+            " "
+            'display
+            `(space :width (+ left-fringe
+                              left-margin
+                              ,@(when (eq (car (window-current-scroll-bars))
```
tarsius: Since this isn't about side-effects you should use`and` instead.

tarsius:
> It looks like `magit-log-select`'s header-line should use `magit-propertize-header-line`, but `magit-stash-refresh-buffer`'s header-line shouldn't have its faces overridden. What do you think?

I think it would be best if `magit-propertize-header-line` only set the face when no character in `STRING` already had a face. Similar to what `magit-insert-heading` does (maybe you can find a less hacky way of doing that).

fice-t:
>I think it would be best if magit-propertize-header-line only set the face when no character in STRING already had a face. 

So it's intentional for those two header lines to have different faces? I was thinking that it would be nice for them to be more uniform, though maybe you would like for them to stand out.

If that's the case, then I think at least `magit-log-select`'s header-line should have its own customizable face instead of just `bold`. It might also be nice to highlight the commands in the header-line (`C-c C-c` and `C-c C-k` for me).

> Similar to what magit-insert-heading does (maybe you can find a less hacky way of doing that).

What about adding an optional argument `preserve-face` that would avoid overwriting the face and setting it in the two above cases?

tarsius:
> So it's intentional for those two header lines to have different faces? I was thinking that it would be nice for them to be more uniform, though maybe you would like for them to stand out.

* The log-select header serves a purpose different from that of other headers. So it should look different, but it's not all that important what it looks like.i
* But the stash header demonstrates why I don't want headers to use the same face for all their text. The different parts of the string represent different pieces of information that when displayed in a buffer get different faces. I think more headers should do that. They don't do so yet because it is not that important but eventually I want to go into that direction.

> If that's the case, then I think at least `magit-log-select`'s header-line should have its own customizable face instead of just `bold`.

I think that is overkill, but add one if you wish.

> What about adding an optional argument `preserve-face` that would avoid overwriting the face and setting it in the two above cases?

In some cases calculating the value of the `STRING` argument takes a lot of code, so doing this would result in lonely trailing `t`s, much like the `SEPARATOR` of `mapconcat`. I think just automatically detecting whether the string doesn't have to be propertized because it already is propertized is just fine.

Instead of the proposed function we could also add a macro similar to `magit-insert-heading` (also containing the `(setq header-format ...)` part). It could be named `magit-set-header-format` and it would not necessarily have to be a macro.

I admit that this might be a bit too clever, but since we are already doing it for `magit-insert-heading` I think it is okay here too.

tarsius:
> Pad the right side of header lines

I don't actually understand what that is good for.

fice-t:
> I don't actually understand what that is good for.

Here's what it looks like with my current theme (doom-one) before and after this PR:

header-line-before

header-line-after

Notice how the alignment is fixed in the second picture (1st commit), and there is padding on the right with equal width to that of the left padding (2nd commit).

I think the right padding gives the header a nice bit of breathing room.

fice-t:
> Similar to what magit-insert-heading does (maybe you can find a less hacky way of doing that)

Yeah, `magit-insert-heading` should use `text-property-not-all` like so:

```elisp
(text-property-not-all 0 (length heading) 'face nil heading)
```

> But the stash header demonstrates why I don't want headers to use the same face for all their text. The different parts of the string represent different pieces of information that when displayed in a buffer get different faces. I think more headers should do that. They don't do so yet because it is not that important but eventually I want to go into that direction.

Sounds good, but I believe in that case those different parts of the string should get new faces so that the user can customize them to match his/her customization of the `header-line` face. I pushed another commit that does this for the keys in the log select header-line (I'll rebase at the end, of course).

>In some cases calculating the value of the STRING argument takes a lot of code, so doing this would result in lonely trailing ts, much like the SEPARATOR of mapconcat. I think just automatically detecting whether the string doesn't have to be propertized because it already is propertized is just fine.

I figured that there's not much of a point to do that calculation at run-time, but I suppose the check is fast enough. I pushed a commit to do that.

>Instead of the proposed function we could also add a macro similar to magit-insert-heading (also containing the (setq header-format ...) part). It could be named magit-set-header-format and it would not necessarily have to be a macro.

Yeah, that sounds good. To avoid confusion with headers in Magit, should it be called `magit-set-header-line-format`? I don't see a reason to make it a macro, though.

tarsius:
Everything sounds good! Please do that :wink:

fice-t:
Okay, I did that and rebased it.

fice-t:
@tarsius 
I added a (possibly temporary) 5th commit that propertizes the padding strings if a particular face (e.g., `magit-header-line-log-select`) is present in the entire string. Otherwise, the padding doesn't match with the rest of the header-line.

It currently fails though if there are multiple faces at the first character in the `header-line`. Perhaps a better way would be to have the relevant callers (right now, just `magit-log-select`) use `add-face-text-property` on the padding? It's getting late here, so I'll leave it at this for now...

tarsius:
Thanks!

PR #3201, opened 2017-10-12
Avoid crash in magit-debug-git-executable

vermiculus:
If `magit-git-executable` is an absolute path, the `unless` form will evaluate to nil, which crashes `insert`.

Wrap all the inserted forms with `concat` first, which handles nil more intuitively.


tarsius:
Thanks!

PR #3205, opened 2017-10-14
magit-insert-{local,remote}-branches: use for-each-ref

### LINKED ISSUE #3203: magit-show-refs-popup broken for Git <2.13
The magit-show-refs command when the git repository is accessed over tramp has become broken over the last few days. If I open a magit-status buffer over tramp, and then run `magit-show-refs-popup` from within that status buffer, the resulting `*magit-refs` buffer doesn't show the branches.

```elisp
(magit-status "/ssh:david.hull@cos6:src/ssrtb-server/")
```

Now:
```
Branches:

Origin (git@example.com:orange/ssrtb-server.git):

Tags (385)
  rpm-9_9_0-x86_64          tag 'rpm-9_9_0-x86_64' created by fw-revision-git
  rpm-9_99_0-x86_64         tag 'rpm-9_99_0-x86_64' created by fw-revision-git
```

Previously (as of around Tuesday):
```
Branches (7)
  apps-15392-device-id-throttle-priority WIP.
@ apps-15447-map-request-state Update for ox-request-context-6.0.0.  Version 9.172.0.
  expose-endpoint-latency-via-http-and-json ssrtb_wm_latency: New module to report latency stats over HTTP.
  jsons                     oxjson: Pre-convert object keys.
  master                    [origin/master] APPS-15451: Initialize token bucket to full capacity.  Version 9.171.2.
  resolicit-2.5-web-index   SOLICIT-133: Update web_index to call ssrtb_service:solicitBidders.
  set-per-bidder-worker-heap-size Set initial heap size when spawning per-bidder workers.

Origin (git@example.com:orange/ssrtb-server.git) (2)
  origin/buy-690-notify-error-reply BUY-413: Hide private fields for external users.
  origin/master             APPS-15451: Initialize token bucket to full capacity.  Version 9.171.2.

Tags (385)
  rpm-9_9_0-x86_64          tag 'rpm-9_9_0-x86_64' created by fw-revision-git
  rpm-9_99_0-x86_64         tag 'rpm-9_99_0-x86_64' created by fw-revision-git

```

Magit version: `Magit 20171011.1130, Git 2.7.4, Emacs 25.3.2, darwin`

kyleam:
```
9cac1f53 (magit-insert-{local,remote}-branches: use NUL separated
fields, 2017-10-11) breaks magit-show-refs because "git branch" does
not support the --format option in Git versions earlier than 2.13.

[WIP] We can *almost* use for-each-ref as a drop-in replacement.  The
only discrepancy I've noticed is that the local refs do not include an
entry for a detached head.  At first glance, I don't see an issue with
dropping this because the buffer is about listing refs.  Thoughts?
```

Fixes #3203.

### DIFF
--- a/lisp/magit-refs.el
+++ b/lisp/magit-refs.el
@@ -418,9 +418,10 @@ line is inserted at all."
   "Insert sections showing all local branches."
   (magit-insert-section (local nil)
     (magit-insert-heading "Branches:")
-    (dolist (line (magit-git-lines "branch" "--format=\
+    (dolist (line (magit-git-lines "for-each-ref" "--format=\
 %(HEAD)%00%(refname:short)%00%(objectname:short)%00%(subject)%00\
 %(upstream:short)%00%(upstream)%00%(upstream:track,nobracket)"
+                                   "refs/heads"
                                    (cadr magit-refresh-args)))
       (pcase-let ((`(,head ,branch ,hash ,message
                            ,upstream ,uref ,utrack)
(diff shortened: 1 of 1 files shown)


tarsius:
Well in that case we won't have a line beginning with `@`, but because the header-line still contains `HEAD`, I think that's okay.

tarsius:
I went ahead and merged this after updating the commit message.

kyleam:
Thanks!  In the future I'd appreciate if you give me a chance to
revisit a commit I've marked as WIP, if only to update WIP bits of the
commit message.  My turnaround usually isn't too slow :)

tarsius:
Okay! Sorry.

PR #3207, opened 2017-10-14
Fix documentation typos and grammar

danielmai:
Like 1f5482361c62dccd0c58eddb6cc72e2f8dcad936, but updating magit.org instead of magit.texi.

> - Update some typos, such as "to" to "too"
> - Update grammar where appropriate, such as commas and verb conjugation
> - Update to name spelling, such as Git and TRAMP

I have some questions related to the documentation style and formatting:

1. Is there a documentation style guide on the appropriate use of "Git" and "git"?  My thought is that "git" is strictly used for the `git` command line tool and "Git" otherwise.
2. Is there `fill-column` convention for magit.org? Paragraphs in the file are broken into several lines, but when I use `fill-paragraph` the formatting changes significantly even on unedited paragraphs.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -7,7 +7,7 @@
 #+TEXINFO_DIR_CATEGORY: Emacs
 #+TEXINFO_DIR_TITLE: Magit: (magit).
 #+TEXINFO_DIR_DESC: Using Git from Emacs with Magit.
-#+SUBTITLE: for version 2.11.0 (2.11.0-101-g83ac4f12+1)
+#+SUBTITLE: for version 2.11.0 (2.11.0-104-g80d9ca41+1)
 #+BIND: ox-texinfo+-before-export-hook ox-texinfo+-update-version-strings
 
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks!

> My thought is that "git" is strictly used for the git command line tool and "Git" otherwise.

Yes.

> Is there `fill-column` convention for magit.org? Paragraphs in the file are broken into several lines, but when I use `fill-paragraph` the formatting changes significantly even on unedited paragraphs.

On my system it seems to be 70 so maybe we should enforce that. On the other hand the value might have been something else since `fill-paragraph` often would like to make changes.

So I guess the convention for now is "when you edit something, preserve the `fill-column` that appears to have been in use when the surrounding paragraphs were written. I prefer if a few lines are a little to long instead of honoring `fill-column` to the letter. Also some of the documentation was copied to or from doc-strings, so sometimes the convention from one place was carried over to the other. In summary, use common sense. :wink:

PR #3208, opened 2017-10-15
Fix man page search

fice-t:
I noticed that searching the man pages didn't work very often, so I thought I would fix it. Unfortunately, I didn't expect that there would be so many edge cases, so the end result looks awful. Perhaps I should put more comments in?

There are still two cases that aren't handled, though. The first is the `--` argument, which is sometimes absent and sometimes present as `[--]`, and the options that start with `++` (e.g., those in `magit-log-popup`). What would you like to happen with those? Should the `++` options have their own help pages?

I also made it a `user-error` to try and bring up the man page for an invalid option/switch.

### DIFF
--- a/lisp/magit-popup.el
+++ b/lisp/magit-popup.el
@@ -956,7 +956,16 @@ and are defined in `magit-popup-mode-map' (which see)."
     (user-error "No man page associated with %s"
                 (magit-popup-get :man-page)))
   (when arg
-    (setq arg (magit-popup-event-arg arg)))
+    (setq arg (magit-popup-event-arg arg))
+    (when (string-prefix-p "--" arg)
+      ;; handle --[no-] options
+      (setq arg (if (string-prefix-p "--no-" arg)
+                    (concat "--"
+                            "\\[?no-\\]?"
+                            (substring arg 5))
+                  (concat "--"
+                          "\\(?:\\[no-\\]\\)?"
+                          (substring arg 2))))))
   (let ((winconf (current-window-configuration)) buffer)
     (pcase magit-popup-manpage-package
       (`woman (delete-other-windows)
--- a/lisp/magit-popup.el
+++ b/lisp/magit-popup.el
@@ -958,14 +958,17 @@ and are defined in `magit-popup-mode-map' (which see)."
   (when arg
     (setq arg (magit-popup-event-arg arg))
     (when (string-prefix-p "--" arg)
-      ;; handle --[no-] options
-      (setq arg (if (string-prefix-p "--no-" arg)
-                    (concat "--"
-                            "\\[?no-\\]?"
-                            (substring arg 5))
-                  (concat "--"
-                          "\\(?:\\[no-\\]\\)?"
-                          (substring arg 2))))))
+      ;; handle '--' option and the '--[no-]' shorthand
+      (setq arg (cond ((string= "-- " arg)
+                       "\\(?:\\[--\\] \\)?<[^[:space:]]+>\\.\\.\\.")
+                      ((string-prefix-p "--no-" arg)
+                       (concat "--"
+                               "\\[?no-\\]?"
+                               (substring arg 5)))
+                      (t
+                       (concat "--"
+                               "\\(?:\\[no-\\]\\)?"
+                               (substring arg 2)))))))
   (let ((winconf (current-window-configuration)) buffer)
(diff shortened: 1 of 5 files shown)


fice-t:
@tarsius During my testing I noticed that the man-page for `magit-notes-popup` is `"git-tag"` instead of `"git-note"`; is that intentional?

tarsius:
> the options that start with `++` (e.g., those in `magit-log-popup`). What would you like to happen with those?

How about adding a doc-string the the relevant function and then using `describe-function`?

> During my testing I noticed that the man-page for `magit-notes-popup` is `"git-tag"` instead of `"git-notes"`; is that intentional?

Sounds like a pasto.

tarsius:
> Perhaps I should put more comments in?

That's probably a good idea. At least at ten o'clock at night it's not immediately obvious to me what the various handled edge cases. So yes, please make it "late at night safe" :wink:

fice-t:
> How about adding a doc-string the the relevant function and then using describe-function?

What relevant functions? `magit-insert-log` doesn't appear to use them to call any functions.

I did everything else.

tarsius:
> What relevant functions? 

`++order` could be documented in `magit-log-select-order`, which is the function used to set the value of that option.

For switches like `++header` a pseudo setter function could be added. Quick testing indicates that `magit-popup` doesn't start treating such a switch as an option instead, though this might cause other issues.

fice-t:
I added the relevant code to search the docstrings. I think that it would be better if you made the docstrings/new procedures, though.

lisp/magit-popup.el
```diff
@@ -52,6 +52,7 @@
 (require 'cl-lib)
 (require 'dash)
 (require 'format-spec)
+(require 'subr-x) ; string-remove-suffix
```
tarsius: `(require 'subr-x)` aka. require stuff that we have to require explicitly even though we shouldn't have to. Or in other words, drop the comment :wink:

lisp/magit-popup.el
```diff
@@ -933,9 +934,17 @@ and are defined in `magit-popup-mode-map' (which see)."
                   (lookup-key (current-global-map) key))))
     (pcase def
       (`magit-invoke-popup-switch
-       (magit-popup-manpage man (magit-popup-lookup int :switches)))
+       (--if-let (magit-popup-lookup int :switches)
+           (if (string-prefix-p "++" (magit-popup-event-arg it))
+               (magit-popup-describe-function (magit-popup-event-fun it))
```
tarsius: Make sure we don't try to call `nil` as a function.

tarsius: Please also add documentation for `++order=` and `++header`. In the latter case by adding a function that does nothing* and is named `magit-log++header`.

(*) Or whatever it has to do to serve as a no-op.

lisp/magit-popup.el
npostavs: Should be `(eval-when-compile (require 'subr-x))` perhaps? (`string-remove-suffix` is a `defsubst`, so it's compiled out)

tarsius: We also require `subr-x` in two other libraries. If we can do that there too, then yes. Otherwise it won't make much of a difference, but of course doing it would still be nicer.

npostavs: The only defun in subr-x.el is/was* `read-multiple-choice`, and it seems we're not using it. So putting `eval-when-compile` should be fine everywhere. 

*That function is moved to its own file in Emacs 26.

tarsius: Okay let's do that then.

lisp/magit-popup.el
fice-t: @tarsius It looks like `magit-popup-convert-switches` strips the function during conversion. Should it be changed to preserve it?

tarsius:
> I think that it would be better if you made the docstrings/new procedures, though.

You have already done that by now... looks okay.

> It also seems that ++header isn't documented in the manual.

I'll we do that after merging your branch

> Should it be changed to preserve it?

It's a bit nasty - but that just means it fits in very well with the surrounding code :wink: Anyway I have done that now. (Luckily I will soon start writing the popup replacement.)

tarsius:
I think this is ready - anything you want to do before I merge?

fice-t:
No, I don't think so. The regex should be good enough at this point.

PR #3209, opened 2017-10-15
magit-branch-reset: use magit-local-branch-at-point

npostavs:
Using the non-local version was just a copy-pasta mistake, right?


tarsius:
One would think that resetting to the upstream or push-target would be a common use-case, so yes, this appears to be a pasto.

Issue #2950, opened 2017-01-10
Allow saving un-/staged changes to a patch file


tarsius:
I don't consider this to be an important feature, so it's okay if this uses some indirection, but it should be possible without having to resort to <kbd>!</kbd>.

<kbd>C-x C-w</kbd> in a buffer which was created with <kbd>d s</kbd> or <kbd>d u</kbd> should offer to save the respective diff to a patch file.

PR #3217, opened 2017-10-21
Add basic commands for saving and applying plain patch files

### LINKED ISSUE #2940: Support applying simple diff patch files
Magit 2.10.0-dirty, Git 2.6.6, Emacs 24.5.1, gnu/linux

On the command line one can `git apply foo.patch` where foo.patch is unadorned `git diff` output, without any of the additional patch file formatting that magit supports/requires.

It would be helpful if one could apply such unformatted patch files in magit.

Attempting to do so via the patch popup with `w w`causes a `git am` command, which fails with "Patch format detection failed."

Presumably magit would itself need to detect the file format (or lack thereof) in order to intelligently issue a `git apply` command instead of a `git am` command, but I think an `a` for `git apply` Action could usefully be added to that menu alongside "`w Apply patches    m Apply maildir`" (the `w` label would then need to be differentiated). That would seem like a least-effort approach to achieving the goal?

tarsius:
Closes #2940 and #2950.

### DIFF
--- a/Documentation/RelNotes/2.12.0.txt
+++ b/Documentation/RelNotes/2.12.0.txt
@@ -55,6 +55,9 @@ Changes since v2.11.0
 * The remote popup can now be used to set various remote releated
   variables.  #3165
 
+* Added new commands `magit-patch-apply' and `magit-patch-apply-popup'.
+  The latter is available from inside `magit-am-popup' ("w").  #2940
+
 Fixes since v2.11.0
 -------------------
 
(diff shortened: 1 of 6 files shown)


fice-t:
@tarsius 

`magit-patch-apply` raises an error when `(magit-file-at-point)` is nil, since `file-relative-name` expects a string.

Also, `default-directory` doesn't need to be specified in `file-relative-name`.

tarsius:
Fixed. Thanks.

PR #3219, opened 2017-10-21
Bind [tab] whenever "C-i" is bound

moyamo:
Same people have [tab] and "C-i" uncoupled so that binding to "C-i" will
not necessarily bind to the tab key.

Thus we should bind to <tab> whenever we bind to "C-i".

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -358,6 +358,7 @@ starts complicating other things, then it will be removed."
            (define-key map (kbd "C-m") 'magit-visit-thing)
            (define-key map (kbd "C-M-i") 'magit-dired-jump)
            (define-key map (kbd "C-i") 'magit-section-toggle)
+           (define-key map [tab]       'magit-section-toggle)
            (define-key map [C-tab]     'magit-section-cycle)
            (define-key map [M-tab]     'magit-section-cycle-diffs)
            ;; [backtab] is the most portable binding for Shift+Tab.
--- a/lisp/magit-popup.el
+++ b/lisp/magit-popup.el
@@ -188,6 +188,7 @@ before entering a popup has.
            (define-key map (kbd "DEL") 'backward-button)
            (define-key map (kbd "C-p") 'backward-button)
            (define-key map (kbd "C-i") 'forward-button)
+           (define-key map [tab]       'forward-button)
            (define-key map (kbd "C-n") 'forward-button)))
     map)
   "Keymap for `magit-popup-mode'.
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -550,6 +550,7 @@ defaulting to the tag at point.
            (define-key map [return] 'magit-invoke-popup-action))
           (t
            (define-key map (kbd "C-i") 'magit-invoke-popup-action)
+           (define-key map [tab]       'magit-invoke-popup-action)
            (define-key map (kbd "C-m") 'magit-invoke-popup-action)))
     map)
   "Keymap used by `magit-dispatch-popup'.")


npostavs:
If they want to uncouple those keys, why should we ignore that?

tarsius:
There are at least two possible reasons to decouple `[tab]` and `\C-i`.

1. To use `\C-i` for something else. For example I use `C-[jkil]` for basic navigation instead of `C-[bpnf]`.
2. To use `[tab]` for something else. For example to bind a more advance completion/toggling/dwim command while leaving the usual completion/toggling binding alone.

> If they want to uncouple those keys, why should we ignore that?

@moyamo appears to be in the first group. If one does this then one has to be ready for an uphill battle: one has to modify basically every keymap that binds `TAB`.

I think the best way to ease the pain would be to reverse the remapping. But that would require a lot of keymaps to be changes and we would never get to the point where each and every keymap in third-pary packages would have been adjusted. But at least once "completed", such a change should not cause any problems for those in the second group. At least conceptually and at least in this day and age changing the direction just makes more sense: most keyboards have a key labeled "Tab" and most users press that key (not  "i" while holding "Control") when they want to do "the completion thing".

Anyway, I don't really expect this to ever happen. I have no immediate plan of bringing this up on `emacs-devel` as this would no doubt result in an endless discussion. At the same time I also don't think that a user who wants to re-purpose `\C-i` should try to get this "fixed" in all packages that s/he uses. Such a change can have negative consequences for people in the second group and unless the consensus is to change the direction of the remapping across all of Emacs and beyond, we should therefore leave things as is.

moyamo:
If you look at the surrounding keybindings

````
(define-key map (kbd "C-i") 'magit-section-toggle)
(define-key map [C-tab]     'magit-section-cycle)
(define-key map [M-tab]     'magit-section-cycle-diffs)
````
You'll notice that `magit-section-cycle` and `magit-section-cycle-diffs` are bound to <kbd>tab</kbd>, but `magit-section-toggle` is bound to <kbd>C-i</kbd>. The reason it is bound to <kbd>C-i</kbd> is that in a terminal the <kbd>tab</kbd> key is indistinguishable from <kbd>C-i</kbd>. I assume that the intention of binding to <kbd>C-i</kbd> is actually to bind to <kbd>tab</kbd>.  However in the emacs GUI *<kbd>C-i</kbd> is not necessarily <kbd>tab</kbd>*.
.
So you really should bind to <kbd>tab</kbd> and <kbd>C-i</kbd> to ensure keybindings work across GUI emacs and terminal emacs.

**Comment by moyamo** [comment]
I just noticed that in terminal emacs <kbd>M-tab</kbd> binds to `magit-dired-jump` but in GUI emacs <kbd>M-tab</kbd> binds to `magit-section-cycle-diffs`.

**Comment by tarsius** [comment]
> However in the emacs GUI *<kbd>C-i</kbd> is not necessarily <kbd>tab</kbd>*.

I know. Otherwise I would not have been able to give two potential reasons why a use would make that so.

> So you really should bind to <kbd>tab</kbd> and <kbd>C-i</kbd> to ensure keybindings work across GUI emacs and terminal emacs.

You are kinda ignoring what I have said. In particular the part about there being another reason for decoupling beside yours (which is also mine) and that we have to think of those users also:

> At the same time I also don't think that a user who wants to re-purpose \C-i should try to get this "fixed" in all packages that s/he uses. Such a change can have negative consequences for people in the second group and unless the consensus is to change the direction of the remapping across all of Emacs and beyond, we should therefore leave things as is.

**Comment by npostavs** [comment]
> I think the best way to ease the pain would be to reverse the remapping. But that would require a lot of keymaps to be changes and we would never get to the point where each and every keymap in third-pary packages would have been adjusted.

I wonder if it could be possible to allow `(define-key map [(control) ?i] 'foo)` to bind <kbd>Ctrl+i</kbd> with higher priority than `(define-key map "\t" 'foo)`, the same way `(define-key map [tab] 'foo)` has higher priority for <kbd>tab</kbd>.

**Comment by tarsius** [comment]
As in introducing a new key binding syntax with special handling? (Or is `[(control) ?i]` valid now?)

By the way (depending on how you look at it of course) `[tab]` doesn't have higher priority than `\C-i`, `[tab]` is just remapped to `\C-i` if it doesn't have a binding of it's own.

I undo that locally using

```lisp
(defun kmu-remove-key (keymap key)
  "In KEYMAP, remove key sequence KEY.
Make the event KEY truely undefined in KEYMAP by removing the
respective element of KEYMAP (or a sub-keymap) as opposed to
merely setting its binding to nil.

There are several ways in which a key can be \"undefined\":

   (keymap (65 . undefined) ; A
           (66))            ; B

As far as key lookup is concerned A isn't undefined at all, it is
bound to the command `undefined' (which doesn't do anything but
make some noise).  This can be used to override lower-precedence
keymaps.

B's binding is nil which doesn't constitute a definition but does
take precedence over a default binding or a binding in the parent
keymap.  On the other hand, a binding of nil does _not_ override
lower-precedence keymaps; thus, if the local map gives a binding
of nil, Emacs uses the binding from the global map.

All other events are truly undefined in KEYMAP.

Note that in a full keymap all characters without modifiers are
always bound to something, the closest these events can get to
being undefined is being bound to nil like B above."
  (when (stringp key)
    (setq key (kmu-parse-key-description key t)))
  (define-key keymap key nil)
  (setq key (cl-mapcan (lambda (k)
                         (if (and (integerp k)
                                  (/= (logand k ?\M-\0) 0))
                             (list ?\e (- k ?\M-\0))
                           (list k)))
                       key))
  (if (= (length key) 1)
      (delete key keymap)
    (let* ((prefix (vconcat (butlast key)))
           (submap (lookup-key keymap prefix)))
      (when (and (not (eq submap 'ESC-prefix))
                 (= (length submap) 1))
        (delete (last key) submap)
        (kmu-remove-key keymap prefix)))))

(kmu-remove-key       function-key-map [(           tab)])
(kmu-remove-key local-function-key-map [(           tab)])
(kmu-remove-key       function-key-map [(meta       tab)])
(kmu-remove-key local-function-key-map [(meta       tab)])
(kmu-remove-key       function-key-map [(meta shift tab)])
(kmu-remove-key local-function-key-map [(meta shift tab)])
(kmu-remove-key       function-key-map [(     shift tab)])
(kmu-remove-key local-function-key-map [(     shift tab)])

(kmu-remove-key       function-key-map [(           return)])
(kmu-remove-key local-function-key-map [(           return)])
(kmu-remove-key       function-key-map [(meta       return)])
(kmu-remove-key local-function-key-map [(meta       return)])
(kmu-remove-key       function-key-map [(meta shift return)])
(kmu-remove-key local-function-key-map [(meta shift return)])
(kmu-remove-key       function-key-map [(     shift return)])
(kmu-remove-key local-function-key-map [(     shift return)])
```

tarsius:
Two random examples of keymaps I had to fix up for my own use:

```lisp
(kmu-define-keys message-mode-map message
  "C-i" > "tab" message-tab
  "C-a" > "C-u" message-beginning-of-line)

(kmu-define-keys mspools-mode-map mspools
  "C-m" > "return" mspools-visit-spool
  "p"   > "i"      previous-line
  "n"   > "k"      next-line)
```

After having tried many different approaches (contemplating asking everyone to make changes so I wouldn't have to, various graphical key binding interfaces, automated changes (= many false positives), patching every package I use...) I have finally settled on doing it manually (a select few packages I continue to patch).

tarsius:
Just mentioning in case there is any doubt that I have not though about this sufficiently :stuck_out_tongue_closed_eyes:

npostavs:
> As in introducing a new key binding syntax with special handling? (Or is [(control) ?i] valid now?)

Oh sorry, I got confused, I meant `[(control ?i)]` which currently means the same as `"\t"` or `[?\C-i]`.
I was experimenting with `[(control) ?i]` before, so I didn't actually notice the equivalence.

tarsius:
I don't remember the details but I think that redirecting two key bindings to each other doesn't work properly.

npostavs:
So I believe the current situation is this:

```
<tab>                    Ctrl+i
  |                       |    
  |                       |
  |                       |
  |                       |
  V                       V
[tab] ---translate----> [?\t]
```

But it should be changed to something along the lines of this:

```

<tab>                    Ctrl+i
  |                       |    
  |                       |
  |                       |
  |                       |
  V                       V
[tab]                   (vector (+ ?\C-0 ?i))
 \                        /
  \                      /
   \                    /
    \                  /
     \   translate    /
      \              /
       \            /
        `-> [?\t] <-

```

`(vector (+ ?\C-0 ?i))` is a provisional notation to indicate that we are encoding <kbd>Ctrl</kbd> with `2**26` (cf [(elisp) Keyboard Events](https://www.gnu.org/software/emacs/manual/html_node/elisp/Keyboard-Events.html)), and not `-2**6`. Not sure if taking `[(control ?i)]` for that would be a good idea.

moyamo:
If you are really against binding <kbd>\<tab></kbd> then I don't mind patching the keymaps in my own configuration. If I rebind C-i in the `override-global-map` keymap of `override-global-mode` minor mode. Then something like this

``` 
;; Make the <tab> key do whatever C-i is supposed to do.
(defun my/do-tab-without-override ()
  (interactive)
  (let* ((override-global-mode nil)
         (key (kbd "C-i"))
         (command (key-binding key)))
    (call-interactively command)))
(bind-key "<tab>" 'my/do-tab-without-override)
```

Will bind <kbd>\<tab></kbd> to whatever <kbd>C-i</kbd> was bound to.

tarsius:
@moyamo I have done similar things in the past. If it works for you, great! I eventually gave up on that because there actually were existing <kbd>C-i</kbd> bindings that I wanted to keep.

npostavs:
By the way, based on some [recent discussion in gnu-emacs-help][geh], a possible way of achieving this sort of decoupling is to use `input-decode-map`:

```elisp
(define-key input-decode-map "\t" [Ctrl+i])
(define-key function-key-map [Ctrl+i] "\t")
```

Now pressing <kbd>C-i</kbd> will look for bindings for `[Ctrl+i]` and if not found, look for bindings for `TAB`. Note that `Ctrl+i` is just an arbitrary symbol, you can use anything else instead.

[geh]: https://lists.gnu.org/archive/html/help-gnu-emacs/2019-06/msg00082.html

PR #3224, opened 2017-10-24
magit-blame-copy-hash: improve documentation

DamienCassou:


lisp/magit-blame.el
```diff
@@ -529,8 +529,8 @@ then also kill the buffer."
 (defun magit-blame-copy-hash ()
   "Save hash of the current chunk's commit to the kill ring.
 
-When the region is active, then save that to the `kill-ring',
-like `kill-ring-save' would."
+When the region is active, the command saves the region's content
+instead of the hash, like `kill-ring-save' would."
```
tarsius: Since doc-strings are supposed in imperative form, I think we should use this here:

> When the region is active, ~~the command saves~~ *then save* the region's content instead of the hash, like `kill-ring-save' would."

DamienCassou: done. I prefer the explicit form here so I kept it like that in the documentation. Do you want me to change it there too?

tarsius: No, in the documentation we should write like actual humans ;-)

Though in many cases I have just copied the doc-strings without adjusting the wording yet. Someday...

PR #3226, opened 2017-10-24
swap args 2 & 3 in magit-ediff-stage's call to magit-completing-read

dellison:
I was getting an error when I tried to use `magit-ediff-stage` where emacs would try to call the list of tracked files as a function, resulting in an error like this:

    Invalid function: (".gitignore" .....)

It seemed to me after a little investigation that this was happening because of `magit-ediff-stage` calling out to `magit-completing-read` with `(magit-tracked-files)` as the PREDICATE argument, when it should be COLLECTION (which was `nil`).

This pull request simply swaps those two arguments in that particular function call: now the call to `magit-completing-read` from `magit-ediff-stage` uses `(magit-tracked-files)` as the candidates, and `nil` as the predicate. An hour or two of casual use hasn't turned up any errors for me, so I thought I'd submit the change.

### DIFF
--- a/lisp/magit-ediff.el
+++ b/lisp/magit-ediff.el
@@ -159,8 +159,8 @@ conflicts, including those already resolved by Git, use
   "Stage and unstage changes to FILE using Ediff.
 FILE has to be relative to the top directory of the repository."
   (interactive
-   (list (magit-completing-read "Selectively stage file" nil
-                                (magit-tracked-files) nil nil nil
+   (list (magit-completing-read "Selectively stage file"
+                                (magit-tracked-files) nil nil nil nil
                                 (magit-current-file))))
   (magit-with-toplevel
     (let* ((conf (current-window-configuration))


tarsius:
Yep, those two arguments have to be switched.

Could you please change this to look like

```lisp
   (list (magit-completing-read "Selectively stage file"
                                (magit-tracked-files) nil nil nil nil
                                (magit-current-file))))
```

and while you are at it the commit message to

```
magit-ediff-stage: swap 2th and 3th args in call to magit-completing-read
```

Thanks.

dellison:
Sure, sounds good! Thanks for the feedback. :)

dellison:
Sorry about the extra merge commit, I think that happened because I amended the previous commit to be how you suggested after pushing. I hope that doesn't cause any problems for you!

kyleam:
@dellison Please just force push the amended commit.

Also, it looks like your updated commit missed @tarsius's suggestion
about how to break the lines.

dellison:
@kyleam Oh dang, I thought I'd gotten it. Thanks, I'll take another look.

dellison:
I couldn't quite be savvy enough to do that cleanly the way you suggested, @kyleam, so I just added another commit to address the line spacing issue. I'm awfully sorry about the mess! :( I guess I'm constantly learning how little I really know about git.

kyleam:
@dellison No problem, thanks for spotting the issue and addressing our
comments.  I'll squash the commits for you.

dellison:
@kyleam I appreciate it, thank you!

tarsius:
> I'll squash the commits for you.

I've just done that.

kyleam:
or @tarisus will take care of it :)

tarsius:
A commented just now
B commented just now

tarsius:
You must have been in a hurry - usually you don't mistype my nickname :wink:

dellison:
@tarsius @kyleam my thanks again to you both! I really appreciate all your work on magit.

PR #3232, opened 2017-10-29
Add option to decouple confirmation and completion

### LINKED ISSUE #2916: Add option to decouple confirmation and completion
Hi Jonas,

Other than removing `P P` and `F F` (which I restored with a local patch), I've been extremely happy with Magit v2.9; great work!

Two key bindings I'm wondering about, and why they behave differently than so many others:

- `k` almost everywhere prompts y/n to kill the thing or region at point.  On stashes, however, it prompts me with a list of all stashes, to select one.  Since my cursor is already on the stash I want to kill, why doesn't it just ask me y/n if I want to kill that stash?

- `E` on a merge conflicted file used to enter ediff to order to resolve the merge.  Now if I type `E E` on a `unmerged` line, it pops me into a 2-way diff.  I need to type `E m` to resolve a merge resolution, but instead of acting on the merge conflict line that I'm on, it prompts me with a list of all the merge conflicts in the project.

Request: Harmonize these bindings with the way that other aspects of Magit operate, so that `k` on a stash line will prompt to kill that stash, `E E` on a merge conflict line will resolve the merge, as would `E m`.

Thanks,
  John

### LINKED ISSUE #2926: Embrace ivy (and support other frameworks better too)
[`ivy`](https://github.com/abo-abo/swiper) is "a generic completion mechanism", which "aims to be more efficient, smaller, simpler, and smoother to use yet highly customizable" than the alternatives.

Coming from `ido` (which is complex and set in stone), that's exactly what I wished for. And it does deliver on that promise! With `ido` I had to install half a dozen packages that monkey-patch it to get something reasonably useful, and then I still had to work with the maintainers of those packages to fix a variety of issues.

With `ivy` things just work. Sure I had to customize some things and I am not quite done changing key bindings to my liking and such (and getting over muscle memory from `ido`), but it has been a smooth ride right from the start. I should have switched earlier and am not going back.

So I intend to eventually recommend on my blog that users switch to `ivy`, at least for Magit. But things should be improved a bit before I do that.

Magit's "completion framework wrapper" is a bit strange...

1. Should we really append ": " if that isn't there yet?
2. Should we really unconditionally add " (default: DEFAULT)"?
   * When using `ivy` (and potentially `helm`) we probably shouldn't since the default is initially highlighted anyway.
3. Maybe (probably not) we can remove the `magit-completing-read-function` option.
   * For `ivy` the correct function actually is `magit-builtin-completing-read`, a very thin wrapper around `completing-read` (which does nothing but implement the strange features mentioned above). Therefore enabling `ivy-mode` does the job already.
    * However `ivy`'s documentation claims that `(setq magit-completing-read-function 'ivy-completing-read)` is necessary. Maybe that is true for Magit v1.4.2, which is what `ivy`'s maintainer uses I believe (1). For v2 it is not, and the documentation should be updated accordingly in both places (though that is not urgent as this setting does no harm either).
   * For `ido` we probably need a wrapper because `ido-completing-read` is not actually a sufficient drop-in replacement for `completing-read`. `ido-completing-read+` from the package by the same name is a sufficient replacement though, and that's what we use.
    * `ido-completing-read+.el` contains code that prevents it from being used unless explicitly enabled using this option. It might be possible to remove that code and just rely on `ido-ubiquitous-mode` (from the same author). Not sure if that is desirable though.
   * For `helm` we wrap `helm--completing-read-default`. I don't really know whether that's the appropriate thing to do. *Edit: the doc-string says it is not necessary.* ~~I also don't know if `helm` contains could which actively prevents it from being used unless this is explicitly enabled using this option.~~

---

* (1) @abo-abo I vaguely remember that v2 broke some of your expectations and that you "have no need for the new features" in that release. For obvious reasons I think you too would benefit greatly from updating to v2 and if you share your issues with that, then I would be willing to give you some one-on-one help. (Though I have little time for that this month.)

### LINKED ISSUE #3220: Remove `A` keybinding for magit-stash-pop?
Since `A` is the keybinding for `magit-cherry-pick-popup`, one can accidentally pop a stash if the point happens to be on a stash when they want to cherry pick something.

`z p` should be good enough, and one is free to rebind `A` if they'd like. I think it's best to avoid conflicts like these if possible, especially if they modify the repository.

tarsius:
Closes #3220.
Closes #2926.
Closes #2916.

### DIFF
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -97,8 +97,8 @@ Only the following commands support this option:
     (magit-stash-branch       nil t)
     (magit-stash-branch-here  nil t)
     (magit-stash-format-patch nil t)
-    (magit-stash-drop         nil nil)
-    (magit-stash-pop          nil t))
+    (magit-stash-drop         nil ask)
+    (magit-stash-pop          nil ask))
   "When not to offer alternatives and ask for confirmation.
 
 Many commands by default ask the user to select from a list of
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -92,7 +92,13 @@ Only the following commands support this option:
   `magit-branch-rename'
   `magit-tag'")
 
-(defcustom magit-dwim-selection nil
+(defcustom magit-dwim-selection
+  '((magit-stash-apply        nil t)
+    (magit-stash-branch       nil t)
+    (magit-stash-branch-here  nil t)
+    (magit-stash-format-patch nil t)
+    (magit-stash-drop         nil nil)
+    (magit-stash-pop          nil t))
   "When not to offer alternatives and ask for confirmation.
 
 Many commands by default ask the user to select from a list of
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -74,7 +74,7 @@ or `helm--completing-read-default'."
                 (function-item helm--completing-read-default)
                 (function :tag "Other function")))
 
-(defcustom magit-no-confirm-default nil
+(defvar magit-no-confirm-default nil
   "A list of commands which should just use the default choice.
 
 Many commands let the user choose the target they act on offering
@@ -82,8 +82,7 @@ a sensible default as default choice.  If you think that that
 default is so sensible that it should always be used without even
 offering other choices, then add that command here.
 
-Commands have to explicitly support this option.  Currently only
-these commands do:
+Only the following commands support this option:
   `magit-branch'
   `magit-branch-and-checkout'
   `magit-branch-orphan'
@@ -91,16 +90,43 @@ these commands do:
     For these four commands `magit-branch-read-upstream-first'
     must be non-nil, or adding them here has no effect.
   `magit-branch-rename'
-  `magit-tag'"
-  :package-version '(magit . "2.9.0")
+  `magit-tag'")
+
+(defcustom magit-dwim-selection nil
+  "When not to offer alternatives and ask for confirmation.
+
+Many commands by default ask the user to select from a list of
+possible candidates.  They do so even when there is a thing at
+point that they can act on, which is then offered as the default.
+
+This option can be used to tell certain commands to use the thing
+at point instead of asking the user to select a candidate to act
+on, with or without confirmation.
+
+The value has the form ((COMMAND nil|PROMPT DEFAULT)...).
+
+- COMMAND is the command that should not prompt for a choice.
+  To have an effect, the command has to use the function
+  `magit-completing-read' or a utility function which in turn uses
+  that function.
+
+- If the command uses `magit-completing-read' multiple times, then
+  PROMPT can be used to only affect one of these uses.  PROMPT, if
+  non-nil, is a regular expression that is used to match against
+  the PROMPT argument passed to `magit-completing-read'.
+
+- DEFAULT specifies how to use the default.  If it is t, then
+  the DEFAULT argument passed to `magit-completing-read' is used
+  without confirmation.  If it is `ask', then the user is given
+  a chance to abort.  DEFAULT can also be nil, in which case the
+  entry has no effect."
+  :package-version '(magit . "2.12.0")
   :group 'magit-commands
-  :type '(list :convert-widget custom-hook-convert-widget)
-  :options '(magit-branch
-             magit-branch-and-checkout
-             magit-branch-orphan
-             magit-branch-rename
-             magit-worktree-branch
-             magit-tag))
+  :type '(repeat (list command
+                       (regexp :tag "Prompt regexp")
+                       (choice (const "Offer other choices" nil)
+                               (const "Require confirmation" ask)
+                               (const "Use default without confirmation" t)))))
 
 (defconst magit--confirm-actions
   '((const reverse)           (const discard)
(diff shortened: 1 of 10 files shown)


tarsius:
@kyleam I apologize for all the typos. I tried, I really tried. Also the new documentation could of course be a lot better, but it is a step in the right direction. But if you want to reword large parts of that, then you are more than welcome to do that, but no pressure :wink: (I would like to merge this rather soon, so a larger rewrite should probably be done separately.)

kyleam:
great.  I'll have a look by (my) tonight

kyleam:
"great" for this PR being ready for review in general, not a sarcastic
great for the typos it may or may not contain :)

lisp/magit-utils.el
```diff
@@ -300,45 +332,75 @@ and delay of your graphical environment or operating system."
 
 ;;; User Input
 
-(defun magit-completing-read
-    (prompt collection &optional predicate require-match initial-input hist def)
-  "Magit wrapper around `completing-read' or an alternative function.
+(defvar magit-completing-read--silent-default nil)
```
kyleam: The purpose of this variable wasn't obvious to me.  Perhaps it's worth
adding a comment.  AFAICT this variable exists to let functions
(currently just `magit-branch-read-args`) tweak their use of
non-`magit-completing-read` functions (currently just
`magit-read-string-ns`) to behave consistently with the
`magit-dwim-selection` customization.

lisp/magit-utils.el
```diff
@@ -74,33 +74,65 @@ or `helm--completing-read-default'."
...
+Many commands by default ask the user to select from a list of
+possible candidates.  They do so even when there is a thing at
+point that they can act on, which is then offered as the default.
+
+This option can be used to tell certain commands to use the thing
+at point instead of asking the user to select a candidate to act
+on, with or without confirmation.
+
+The value has the form ((COMMAND nil|PROMPT DEFAULT)...).
+
```
kyleam: Hmm `prompt` seems like it would be a bit fragile, but I can't think
of a better way.

I'd prefer to swap the order of `PROMPT` and `DEFAULT` so that the
more important and frequently used value is first.

lisp/magit-utils.el
```diff
@@ -300,45 +332,75 @@ and delay of your graphical environment or operating system."
...
-            (user-error "Nothing selected")
-          nil)
-      reply)))
+  uses `magit-prompt-with-completion', neither `ivy-mode' nor
+  `helm-mode' is enabled, and DEF or FALLBACK is non-nil, then
+  PROMPT is modified to end with \" (default DEF|FALLBACK): \"."
+  (setq magit-completing-read--silent-default nil)
+  (-if-let (dwim (and def
+                      (or (nth 2 (-first (pcase-lambda (`(,cmd ,re ,_))
+                                           (and (eq this-command cmd)
```
kyleam: Won't this fail when a command is called from the popup because
`this-command` will be `magit-invoke-popup-action`?

lisp/magit-utils.el
```diff
@@ -300,45 +332,76 @@ and delay of your graphical environment or operating system."
...
+- PROMPT is modified to end with \" (default DEF|FALLBACK): \"
+  provided that DEF or FALLBACK is non-nil, that neither
+  `ivy-mode' nor `helm-mode' is enabled, and that
+  `magit-completing-read-function' is set to its default value of
+  `magit-builtin-completing-read'."
+  (setq magit-completing-read--silent-default nil)
+  (-if-let (dwim (and def
+                      (or (nth 2 (-first (pcase-lambda (`(,cmd ,re ,_))
+                                           (and (eq this-command cmd)
+                                                (or (not re)
```
kyleam: GitHub is marking my [previous comment] as stale because of the
doc-string fixup that I pushed.

[previous comment]: https://github.com/magit/magit/pull/3232#discussion_r149277837

lisp/magit-utils.el
tarsius: I picked the current order so that all all of the "key" comes before the value - alist-style. Maybe it would be better to use `(((COMMAND nil|PROMPT) DEFAULT)...)` for forward compatibility, just in case we implement something less fragile than relying on the prompt.

lisp/magit-utils.el
tarsius: No, `magit-invoke-popup-action` explicitly sets `this-command` to the invoked action.

lisp/magit-utils.el
tarsius: I replied to that. Just mentioning that in case Github doesn't feel like telling you about it.

tarsius:
> I found reading the first patch a bit confusing because some of the changes refer to some yet-to-be-introduced things.

Do you feel its necessary that I spread that across the commits or would it be okay to move everything to a "update the documentation" commit or even leave it as is?

lisp/magit-utils.el
tarsius: I replied that. Just mentioning that in case Github doesn't have the curtacy

lisp/magit-utils.el
```diff
@@ -74,33 +74,65 @@ or `helm--completing-read-default'."
...
+
+The value has the form ((COMMAND nil|PROMPT DEFAULT)...).
+
+- COMMAND is the command that should not prompt for a choice.
+  To have an effect, the command has to use the function
+  `magit-completing-read' or a utility function which in turn uses
+  that function.
+
+- If the command uses `magit-completing-read' multiple times, then
+  PROMPT can be used to only affect on of these uses.  PROMPT, if
```
npostavs: s/on of/one of

Also, I think that sentence could use some rephrasing, though I can't quite come up with a good replacement at the moment.

kyleam:
> Do you feel its necessary that I spread that across the commits or would it be okay to move everything to a "update the documentation" commit or even leave it as is?

My preference would be for the documentation changes to come at the
end, but it's also not very important so I'm OK with you not wasting
time on it and leaving it as is.

lisp/magit-utils.el
kyleam: Hmm, I'll try to find time to look at this tonight, but when I was
stepping through `magit-completing-read` last night, I recall it being
bound to `magit-invoke-popup-action`.

```
(add-to-list 'magit-dwim-selection '(magit-branch-reset nil ask))
(edebug-instrument-function 'magit-completing-read)
```

lisp/magit-utils.el
kyleam: > I picked the current order so that all all of the "key" comes before the value - alist-style

I see.  Makes sense.

If we go with `(((COMMAND nil|PROMPT) DEFAULT)...)`, do you think it'd
be worth allowing `((COMMAND DEFAULT)...)` as short for `(((COMMAND
nil) DEFAULT)...)`?

lisp/magit-utils.el
kyleam: Testing this out again and found the same thing.

> No, magit-invoke-popup-action explicitly sets this-command to the invoked action.

It explicitly sets `this-command` to the command, but **after** calling
the command.

tarsius: Thanks for catching that. Simply setting `this-command` first seems to do the trick.

lisp/magit-utils.el
tarsius: That would be very confusing I think. I am sticking to what I started with.

tarsius:
I did spread the documentation around a bit. Wasn't too much work and I *think* I relocated all the misplaced bits.

lisp/magit-utils.el
kyleam: > That would be very confusing I think.

Hmm, I don't know.  I think "atom for simple behavior, list to specify
something more complex" is a common enough and useful pattern.  At any
rate, I don't have a strong preference, so ...

> I am sticking to what I started with.

... that's fine by me.

PR #3234, opened 2017-10-30
Support one-section selections

### LINKED ISSUE #2920: magit-branch-spinoff should support a single commit

### LINKED ISSUE #3026: Consider allowing one-section selections
Magit allows selecting two or more items using the region and then acting on that instead of only the current section. But #3013 got me thinking, maybe we should start supporting one-item selections.

Some thoughts on that:

* Already now it is undefined whether the next command will act on the selection or just ignore it and just use the current section. Some commands support it and some don't (or don't even act on any section at all).
* The behavior of some commands that do support the selection, differs in other ways than just acting on more things when it is active. They might for example require additional confirmation. And this, I believe, is why I originally decided that there cannot be any one-item selections.
* If we decide to visualize one-item selection, then that does not mean that each and every command that supports the selection has to honor it in that case. I.e. it can forgo asking for additional confirmation and behave the same as if there were no selection. Pseudo code: `(get-selection 'require-multiple-items)`.
* The only drawback as I can see right now is that the visualization indicates "special behavior" but depending on the invoked command there might not actually be any such behavior. But that isn't much different from having an active multi-section selection and then invoking a command that does not support the selection at all.

tarsius:
```
For a long time Magit supported selecting two or more sibling sections
using the region and then acting on that selection instead of only on
the current section.  Single-section selections were not supported and
a region that did not span multiple siblings was not visualized as a
selection, instead the underlying region was visualized as it is in
non-magit buffers.

The change users will notice first is that when they press C-SPC to
set the beginning of the region/magit-selection on a section heading,
then the magit-selection is visualized.  Previously not even the
region was shown after just pressing C-SPC, because the mark and the
point were identical, so there was no non-empty region to visualize.

For the time being most commands continue to behave exactly as before
when there is a one-section selection: the act on the current section
instead.  While a one-section selection is a selection that contains
nothing but the current section, this difference between the current
section and a set consisting of only the current section is still
relevant.  It affects whether and how commands ask for confirmation
and/or offer the user to act on something else instead.

The reason I decided against supporting one-section selections in the
past is that it is a bit unfortunate to visualize the selection and
then the invoked command does not actually use it.  But this is no
different from a multi-section selection being visualized and then
invoking a command that isn't magit-selection aware at all.  Or from
having the region visualized in any Emacs buffer and then invoking any
command that doesn't behave differently when the region is active.

Beginning with this commit only a handful of commands begin using a
one-section selection.  Most importantly, and this is what motivated
this change, it is now possible to use `magit-branch-spinoff' after
selecting just HEAD to create a new branch rewinding the previously
current branch by a single commit.  Previously it was only possible
to rewind it to its upstream or to rewind it by at least two commits.

The other commands that respect one-section selections are:
- magit-am-apply-patches
- magit-cherry-apply
- magit-cherry-pick
- magit-revert
- magit-revert-no-commit
- magit-stash-pop
- magit-tag-delete

The recently added commands `magit-previous-line' and
`magit-next-line' now forgo moving on the first shift-selection move
when point is on a section heading, not just when inside a hunk body.
```

Closes #2920.
Closes #3026.


Documentation/RelNotes/2.12.0.txt
```diff
@@ -7,11 +7,12 @@ Changes since v2.11.0
...
-  than the region.  This causes shift selection to select two lines on
-  the first invocation when using the vanilla commands.  When invoked
-  inside the body of a hunk the new Magit-specific variants don't move
-  point on the first invocation and thereby they only selects a single
-  line.  Which inconsistency you prefer is a matter of preference.  #2912
+  than the region.  This causes shift-selection to select two lines on
+  the first invocation when using the vanilla commands.  On section
+  headings and inside hunk bodies the new magit-specific variants
+  don't move point on the first invocation and thereby they only
+  selects a single section or line.  Which inconsistency you prefer
```
kyleam: s/selects/select/

Documentation/RelNotes/2.12.0.txt
```diff
@@ -110,6 +111,23 @@ Changes since v2.11.0
   `create-branch' variant and the branch exists, then it offers to
   simply checking it out instead of resetting it first.  #3009
 
+* For a long time Magit supported selecting two or more sibling
```
kyleam: s/supported/has supported/

Documentation/RelNotes/2.12.0.txt
```diff
@@ -110,6 +111,23 @@ Changes since v2.11.0
...
+  sections using the region and then acting on that selection instead
+  of only on the current section.  Single-section selections were not
+  supported and a region that did not span multiple siblings was not
+  visualized as a selection.
+
+  For the time being most magit-selection-aware commands are not
+  affected by this change; they continue to act on the current section
```
kyleam: I believe the above paragraph is just following the past tense "X
didn't Y" form, but it reads a bit weird here because it refers to
"this change" without explicitly saying single-section selections are
now supported.  Perhaps lead with "Single-section selections are now
supported, but for the ..."

Documentation/RelNotes/2.12.0.txt
```diff
@@ -110,6 +111,23 @@ Changes since v2.11.0
...
+  instead.  This relevant because many commands ask for confirmation
+  when there is a magit-selection, while they otherwise give the user
+  the opportunity to act on something else instead.  #3026
+
+* The command `magit-branch-spinoff' now spins off just HEAD, when
+  that constitutes the single-section selection.  Previously one could
+  only spin off all commits that aren't in the upstream yer or at least
```
kyleam: s/yer/yet

lisp/magit-section.el
```diff
@@ -1054,18 +1054,21 @@ or is not a valid section selection, then return nil.  If optional
...
 When the region is active and constitutes a valid section
 selection, then return a list of all selected sections.  This is
 the case when the region begins in the heading of a section and
-ends in the heading of a sibling of that first section.  When
-the selection is not valid then return nil.  Most commands that
-can act on the selected sections, then instead just act on the
-current section, the one point is in.
+ends in the heading of the same section or in that of a sibling
+section.  If optional MULTIPLE is non-nil, then the region has
+to begin and end in the headings of different sibling sections.
```
kyleam: Aren't all sibling sections different?  I think "... the region cannot
begin and end in the same section" would be clearer.

lisp/magit-section.el
```diff
@@ -1054,18 +1054,21 @@ or is not a valid section selection, then return nil.  If optional
...
+
+When the selection is not valid then return nil.  Most commands
```
kyleam: nit: s/valid/valid,/

lisp/magit-section.el
```diff
@@ -1054,18 +1054,21 @@ or is not a valid section selection, then return nil.  If optional
...
+that can act on the selected sections, then instead just act on
+the current section, the one point is in.
```
kyleam: I'm having trouble parsing this sentence.  Perhaps "In this case, most
commands that can act on the selected sections will instead act on the
section at point."

lisp/magit-stash.el
```diff
@@ -217,10 +217,11 @@ and forgo removing the stash."
 (defun magit-stash-drop (stash)
   "Remove a stash from the stash list.
 When the region is active offer to drop all contained stashes."
-  (interactive (list (--if-let (magit-region-values 'stash)
-                         (or (magit-confirm t nil "Drop %i stashes" it)
-                             (user-error "Abort"))
-                       (magit-read-stash "Drop stash"))))
+  (interactive
+   (list (--if-let (magit-region-values 'stash)
+             (or (magit-confirm "Drop %s" "Drop %i stashes" it)
```
kyleam: You've dropped the ACTION argument (or rather are now passing the
prompt string as ACTION).

tarsius:
>> magit-stash-pop

> should this be magit-stash-DROP?

Yes.

This keeps happening to me. I think that to prevent that I will have to post-process git output like this:

```
@@ -217,10 +217,11 @@ (defun magit-stash-pop (stash)
 (defun magit-stash-drop (stash)
```

> Is using "^" rather than "#" intentional?

Yes. I have gotten sick of Github counting how many times I have rebased commits that mention issues. So the intention is to `s/^/#` right before merge.

PR #3244, opened 2017-11-04
Always display a diff while committing and improve toggling

### LINKED ISSUE #3199: Open diff when editing a commit message during a rebase
It would be nice if the diff was shown when editing a commit message in a rebase (interactive, `r w`, etc.), just like it is when committing normally.

`Magit 20171007.431, Git 2.14.2, Emacs 26.0.60, gnu/linux`

tarsius:
Following the discussion in #3199 I have implemented this like so:

* Separate the task of initially displaying the most suitable diff and toggling.
* Displaying the initial diff is done by `magit-commit-diff` alone.
  * We considered to instead hand of that task to the black-box `magit-diff-while-committing` if we cannot decide based on `last-command`. Now that that function no longer is a black-box I was able to tell that that wouldn't have been a good idea :wink: 
* Toggling is done in `magit-diff-while-committing`.
  * Toggling is only done if the buffer previously displayed "the other" diff.
    * Previously we tried doing that too in most cases, but not in a reliable fashion and not in all cases.
  * Else the most appropriate diff is displayed. This can become necessary if the user previously killed the buffer.
* And, somewhat unrelated, `magit-diff-while-committing` now stays in the commit buffer.

Closes #3199.

### DIFF
--- a/lisp/magit-commit.el
+++ b/lisp/magit-commit.el
@@ -339,28 +339,31 @@ depending on the value of option `magit-commit-squash-confirm'."
 ;;; Pending Diff
 
 (defun magit-commit-diff ()
-  (-when-let (fn (and git-commit-mode
-                      magit-commit-show-diff
-                      (cl-case last-command
-                        (magit-commit
-                         (apply-partially 'magit-diff-staged nil))
-                        (magit-commit-all
-                         (apply-partially 'magit-diff-working-tree nil))
-                        ((magit-commit-amend
-                          magit-commit-reword
-                          magit-rebase-reword-commit)
-                         'magit-diff-while-amending))))
+  (when (and git-commit-mode magit-commit-show-diff)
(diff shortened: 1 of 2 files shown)


tarsius:
Apparently Github does not allow requesting a review from a non-member. @fice-t, please try this out too.

kyleam:
doh, I just started to look at this now

tarsius:
Sorry about that. Please keep looking. Bugs can still be fixed of course.

kyleam:
> Sorry about that. Please keep looking.

oh, no worries of course.  Didn't want you to think I was ignoring
this rather than just being slow to get to it.

And looking through the commits, I wouldn't have had any code comments
(just a few commit message typos and comments).

I'll let you know if I hit any edge cases while testing.

fice-t:
It works so far, thanks. I believe I uncovered another bug during testing, though (#3246).

PR #3245, opened 2017-11-05
Improve interface in case of 'git am' failure

npostavs:
Currently, when `git am` fails, e.g, [if the patch was corruped by line wrapping](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=28308#117), the sections magit presents are a bit unhelpful. It looks like this:

```
Applying patches
stop 0001
onto de8c782f34 Only match password prompts in part of line (Bug#28329)
```

where that funny "0001" represents the patch that failed. Git requires you to apply something to the index before `git am --continue` will succeed.

The first patch in this PR adds a title after the "0001", to make it more human friendly.
The next two allow applying it with `a`, although I'm not sure those represent the best way of doing that.

### DIFF
--- a/lisp/magit-sequence.el
+++ b/lisp/magit-sequence.el
@@ -573,9 +573,16 @@ If no such sequence is in progress, do nothing."
 
 (defun magit-sequence-insert-am-patch (type patch face)
   (magit-insert-section (file patch)
-    (insert (propertize type 'face face)
-            ?\s (propertize (file-name-nondirectory patch) 'face 'magit-hash)
-            ?\n)))
+    (let ((title
+           (with-temp-buffer
+             (insert-file-contents patch nil nil 4096)
+             (unless (re-search-forward "^Subject: " nil t)
+               (goto-char (point-min)))
+             (buffer-substring (point) (line-end-position)))))
+      (insert (propertize type 'face face)
+              ?\s (propertize (file-name-nondirectory patch) 'face 'magit-hash)
+              ?\s title
+              ?\n))))
 
 (defun magit-insert-rebase-sequence ()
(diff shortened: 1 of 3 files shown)


lisp/magit-sequence.el
```diff
@@ -573,9 +573,16 @@ If no such sequence is in progress, do nothing."
 
 (defun magit-sequence-insert-am-patch (type patch face)
   (magit-insert-section (file patch)
-    (insert (propertize type 'face face)
-            ?\s (propertize (file-name-nondirectory patch) 'face 'magit-hash)
-            ?\n)))
+    (let ((title
+           (with-temp-buffer
+             (insert-file-contents patch nil nil 4096)
+             (unless (re-search-forward "^Subject: ")
```
kyleam: shouldn't NOERROR be non-nil here?

npostavs: Oops, yes.

lisp/magit-sequence.el
```diff
@@ -573,9 +573,16 @@ If no such sequence is in progress, do nothing."
 
 (defun magit-sequence-insert-am-patch (type patch face)
   (magit-insert-section (file patch)
-    (insert (propertize type 'face face)
-            ?\s (propertize (file-name-nondirectory patch) 'face 'magit-hash)
-            ?\n)))
+    (let ((title
```
tarsius: I would not use a variable here and instead rearrange the existing code a bit. But it doesn't really matter and you can leave this as-is if you like it this way.

tarsius:
> The next two allow applying it with `a`, although I'm not sure those represent the best way of doing that.

It seems appropriate and I cannot think of anything better.

PR #3247, opened 2017-11-06
Add new command magit-branch-checkout

### LINKED ISSUE #2906: 'Checkout or create' branch
This is a feature request. Our team creates plenty of branches that are pushed to the remote. When starting to work on something I have to either:

1. create a local branch by checking out an existing remote branch (i.e., a colleague already worked on the task) (the name of the new branch is the same as in remote): <kbd>b c</kbd>
2. checkout a local branch that's already dedicated to the task (i.e., I already worked on the task earlier) (the branch already has a name): <kbd>b b</kbd>
3. create a local branch on top of master (i.e., nobody worked on the task yet) (in this case I have to give the branch a name): <kbd>b j</kbd> (personal shortcut that computes the new branch name by [looking at the Jira issue tracker](https://gist.github.com/DamienCassou/d5d05c2501271ac4b55281b4c7aef4fc))

The problem is that I have to know in which case I am (either 1, 2 or 3) before doing the right shortcut.

tarsius:
Closes #2906.

### DIFF
--- a/Documentation/RelNotes/2.12.0.txt
+++ b/Documentation/RelNotes/2.12.0.txt
@@ -129,6 +129,12 @@ Changes since v2.11.0
   itself was never visualized.  Now it is, but only if it stays within
   a single line.  #3249
 
+* Added new command `magit-branch-checkout' and bound "l" in the
+  branch popup to it.  This command always checks out a local branch.
+  Valid inputs are an existing local branch, a remote branch for which
+  no corresponding local branch exists, and a new branch name that
+  doesn't match the name of any existing local or remote branch.  #2906
+
 Fixes since v2.11.0
 -------------------
 
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -7,7 +7,7 @@
 #+TEXINFO_DIR_CATEGORY: Emacs
 #+TEXINFO_DIR_TITLE: Magit: (magit).
 #+TEXINFO_DIR_DESC: Using Git from Emacs with Magit.
-#+SUBTITLE: for version 2.11.0 (2.11.0-217-ge5f626de8+1)
+#+SUBTITLE: for version 2.11.0 (2.11.0-218-gb1db91fb1+1)
 #+BIND: ox-texinfo+-before-export-hook ox-texinfo+-update-version-strings
 
 #+TEXINFO_DEFFN: t
@@ -22,7 +22,7 @@ directly from within Emacs.  While many fine Git clients exist, only
 Magit and Git itself deserve to be called porcelains.
 
 #+BEGIN_QUOTE
-This manual is for Magit version 2.11.0 (2.11.0-217-ge5f626de8+1).
+This manual is for Magit version 2.11.0 (2.11.0-218-gb1db91fb1+1).
 
 Copyright (C) 2015-2017 Jonas Bernoulli <jonas@bernoul.li>
 
@@ -3670,6 +3670,28 @@ features are available from separate popups.
   checks it out.
 
   Also see option ~magit-branch-prefer-remote-upstream~.
+
+- Key: b l, magit-branch-checkout
+
+  This command checks out an existing or new local branch.  It reads a
+  branch name from the user offering all local branches and a subset
+  of remote branches as candidates.  Remote branches for which a local
+  branch by the same name exists are omitted from the list of
+  candidates.  The user can also enter a completely new branch name.
+
+  - If the user selects an existing local branch, then that is checked
+    out.
+
+  - If the user selects a remote branch, then it creates and checks
+    out a new local branch with the same name, and configures the
+    selected remote branch as the push target.
+
+  - If the user enters a new branch name, then it creates and checks
+    that out, after also reading the starting-point from the user.
+
+  In the latter two cases the upstream is also set.  Whether it is set
+  to the chosen starting point or something else depends on the value
+  of ~magit-branch-adjust-remote-upstream-alist~.
 
 - Key: b s, magit-branch-spinoff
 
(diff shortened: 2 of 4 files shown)


tarsius:
@DamienCassou please test this new command.

DamienCassou:
Thank you very much. Some feedback:

1. What about `--filter` instead of `--keep`?
2. The code crashes when `magit-branch-at-point` returns `nil`. This happens for example when point is on a blank line at the end of the status buffer.
3. When choosing remote branch, it would be nice to let the user choose the local branch name instead of imposing the same name as the remote branch; The same name is still a very good default (just like `magit-branch-and-checkout`).
4. Would it be possible to add a hook so that I could also use `b l` to decide branch name based on a Jira/github ticket number? I currently do with `b j` and this https://gist.github.com/DamienCassou/d5d05c2501271ac4b55281b4c7aef4fc#file-magit-jira-el-L71. I'm ok to advice a magit function to achieve that, but it would be nice if there was a function to advice :-).

tarsius:
Fixed (1) and (2).

tarsius:
- (3) Git's triangular workflow support insists on these branches having the same name. So I want to keep defaulting to silently using the same name as for the remote branch. Otherwise users might be tempted to needlessly(*) use another name and end up with some features missing.
- (*) It's not needless when some other software such as Jira impose *conflicting* restrictions.
- (4) I have pushed a draft with such a hook...
- (3b) ... which can also be used to be prompted for an alternative name.

-

DamienCassou:
Thank you for your work. I have the impression that we don't understand each other regarding Jira and what I am doing.

We use Jira as an issue tracker. Jira doesn't know anything about our git branches and doesn't care. We use github to version code. To track the relation between the branches in git and jira tickets, we don't use git commit comments (as you do in magit for example) but git branch names. Each branch name is prefixed with the jira ticket id and contains an underscore-separated version of the ticket description: for example: `FTGP-1281_Fix_the_bug`. Because it is annoying to copy/paste/underscore ticket descriptions, I wrote the gist referenced above.

My workflow is then:
1. create a ticket in jira
2. use `b j` in magit (see gist)
3. enter ticket id (e.g., "FTGP-1281")
4. enter starting point branch (usually `origin/master`)
5. gist asks Jira for description of corresponding ticket
5. magit creates a local branch named `FTGP-1281_Fix_the_bug`

I'm not sure your hook can help me as it seems it is based on the fact that the user chooses a remote branch. In my workflow, there is no remote branch yet.

> Git's triangular workflow

I don't know what you are talking about

> It's not needless when some other software such as Jira impose conflicting restrictions

I don't know what you are talking about

> it is your own preference (if I remember correctly) to put the issue number into the branch name

it is a convention of the team. Nothing imposed by Jira

> you might want to consider using another approach. (And yes, I plan some features in that direction.)

I can certainly wait. My `c j` works well for now.

> (3b) ... which can also be used to be prompted for an alternative name.

what do you mean?

tarsius:
> My workflow is then: [...]

I think I understand now.

The new command isn't a good place for that behavior.

IMO it would make more sense to teach `magit-branch` and `magit-branch-checkout` to do it by passing the provided input through a hook to, in your case, translate a number into a suitable branch name.

But I think that using the same prompt to read a ref *or* an issue identifier restricts the possibilities too much. For example it becomes harder to eventually offer completion candidates of the form `NUMBER DESCRIPTION` instead of just expecting the user to know the issue number ahead of time.

That's something I am currently working on for github issues, see #3134. In that case I implement the "given a number, create the appropriate branch" logic as a separate command, and I think that's what you should continue to do too.

But I realize that doesn't really address your original motivation:

> The problem is that I have to know in which case I am (either 1, 2 or 3) before doing the right shortcut.

It would be nice if that problem did not exist, and I will try to come up with something. While I will concentrate on Github I might be able to generalize that eventually. Meanwhile you might want to change your workflow and code to work like this:

- (1) create a ticket in Jira
- (2) invoke the dwim command. That makes an api request so that you can let you choose from a list of open issues. Choices look like `#123 Something is broken`. (That's the key difference to your approach - the api request is made earlier.)
- (3a) if a corresponding local branch already exists, check it out
- (3b) if a corresponding remote branch already exists, create and checkout a local branch
- (3c) otherwise create a new local branch

Such a command would be Jira-specific and couldn't be used to create or checkout branches that are not associated with a Jira issue. But it would always put you on the appropriate branch to work on a given issue without you having to worry about whether you or someone else has already worked on that.

The command that I have implemented for Github issues doesn't go that far, it basically only does (2)
 and (3b). But now that I have thought about this again, I might actually implement something like what I have outlined above for Jira but for Github.

That might eventually turn into something that is generic enough to be used with other issue trackers, but implementing a Jira-specific is beyond Magit's scope. Other people suffering from having to use Jira (I have been told that nobody actually likes it), would probably appreciate it you could make your command available in a `magit-jira` package.

tarsius:
>> Git's triangular workflow

> I don't know what you are talking about

There's pretty much nothing about it in Git's own documentation, but I have tried to provide an introduction at [The Two Remotes](https://magit.vc/manual/magit/The-Two-Remotes.html).

tarsius:
The command that I suggest you implement would be very similar to `magit-branch-checkout` in that it would have to handle the same three cases. But pretty much every detail would be different, since the starting point is an issue, not a branch. Which is why I think there should be two commands instead of trying to make one command do both.

In other words, I am going to merge this as-is and close your feature request.

DamienCassou:
Thank you very much for your work. I will try to follow your approach to reimplement my Jira code and we will see where it goes. I'm just a bit afraid that fetching all ticket numbers and descriptions might slow. I'm also not sure if we would gain anything compared to just entering the ticket number.

PR #3251, opened 2017-11-08
Linkify hashes in commit messages

### LINKED ISSUE #2949: Linkify issues and hashes in commit messages
Issue numbers and hashes should be linkified in commit messages. This applies both to `magit-revision-mode` buffers and to "COMMIT_EDITMSG` and friends.

For issue numbers we can use the built-in `bug-reference` package. For Magit for example one has to add the following to `./.dir-locals.el` (I'm gonna do that by the way):

```
((nil
  (bug-reference-bug-regexp . "#\\(?2:[0-9]+\\)")
  (bug-reference-url-format . "https://github.com/magit/magit/issues/%s")))
```

We might want to narrow this down instead of using the match-all `nil`. Additionally `bug-reference-mode` has to be added to `magit-revision-mode-hook` and `git-commit-setup-hook`. We might wanna do that by default or at least present it as an option.

This should also be documented in the Magit manual (and or the doc-strings of the mentioned hook variables) or maybe we could even provide tooling to set this up automatically.

---

For hashes I am not aware of an existing package. Also here we don't want to send the user to some web page, instead the revision should be shown using `magit-show-commit`.

Regarding finding hashes will avoiding to many false-positives and false-negatives: this should be good enough, at first at least:

```
(and (re-search-forward "\\<[0-9a-z]\\{7,40\\}\\>" nil t)
     (string-match-p "[0-9]" (match-string 0)))
```

---

We should also attempt to set this up automatically.

---

Some things to check out:

- http://oremacs.com/2016/02/15/counsel-ffap/
- https://melpa.org/#/bug-reference-github

tarsius:
Closes #2949.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -291,6 +291,30 @@ subject to option `magit-revision-insert-related-refs'."
   :group 'magit-revision
   :type 'boolean)
 
+(defcustom magit-revision-use-hash-sections 'quicker
+  "Whether to turn hashes inside the commit message into sections.
+
+If non-nil, then hashes inside the commit message are turned into
+`commit' sections.  There is a trade off to be made between
+performance and reliability:
+
+- `slow' calls git for every word to be absolutely sure.
+- `quick' skips words less than seven characters long.
+- `quicker' additionally skips words that don't contain a number.
+- `quickest' uses all words that are at least seven characters
+  long and which contain at least one number as well as at least
+  one letter.
+
+If nil, then no hashes are turned into sections, but you can
+still visit the commit at point using \"RET\"."
+  :package-version '(magit . "2.12.0")
+  :group 'magit-revision
+  :type '(choice (const :tag "Use sections, quickest" quickest)
+                 (const :tag "Use sections, quicker" quicker)
+                 (const :tag "Use sections, quick" quick)
+                 (const :tag "Use sections, slow" slow)
+                 (const :tag "Don't use sections" nil)))
+
 (defcustom magit-revision-show-gravatars nil
   "Whether to show gravatar images in revision buffers.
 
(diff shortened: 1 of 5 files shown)


lisp/magit-diff.el
```diff
@@ -291,6 +291,22 @@ subject to option `magit-revision-insert-related-refs'."
...
   :type 'boolean)
 
+(defcustom magit-revision-use-hash-sections 'quick
+  "Whether to turn hashes inside the commit message into sections.
+
+If non-nil, then hashes inside the commit message are turned
+into `commit' sections.  If `quick', which is the default, then
+only consider words that are at least seven characters long and
+contain at least one number.
+
```
kyleam: I wonder if it's worth adding another mode (`quicker`?) that makes any
instance of "\\_<[a-f0-9]\\{7,40\\}\\_>" a commit section without
verifying that a commit object exists, trading false positives
(including hashes from other repos) for eliminating a git call (maybe
a few, depending on the message).

lisp/magit-diff.el
```diff
@@ -1880,6 +1902,24 @@ or a ref which is not a branch, then it inserts nothing."
...
+            (while (not (eobp))
+              (re-search-forward "\\_<" nil 'move)
+              (let ((beg (point)))
+                (re-search-forward "\\_>" nil t)
+                (when (> (point) beg)
+                  (let ((text (buffer-substring-no-properties beg (point))))
+                    (when (and (or (not (eq magit-revision-use-hash-sections
+                                            'quick))
+                                   (and (>= (length text) 7)
+                                        (string-match-p "[0-9]" text)))
```
kyleam: For the non-`quick` case, would it be worth adding some additional
checks to reduce the git calls (length between 4 and 40, hex)?

PR #3252, opened 2017-11-09
Fix magit-submodule-deinit being copy-pasted

Fuco1:
The binding and command name were incorrectly copy-pasted from `magit-submodule-init`

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -7,7 +7,7 @@
 #+TEXINFO_DIR_CATEGORY: Emacs
 #+TEXINFO_DIR_TITLE: Magit: (magit).
 #+TEXINFO_DIR_DESC: Using Git from Emacs with Magit.
-#+SUBTITLE: for version 2.11.0 (2.11.0-227-gaeca03f04+1)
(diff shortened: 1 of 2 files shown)


tarsius:
Please edit `magit.org` instead. I will then regenerate `magit.texi` from that and amend.

Fuco1:
I've noticed way too late there's an org file.  It would be cool to have a note somewhere about how to contribute to documentation, I couldn't find it.

tarsius:
> It would be cool to have a note somewhere about how to contribute to documentation, I couldn't find it.

Yeah! I think that every time someone does it wrong :wink: 
I'll try to do it soon.

Fuco1:
Hmm, I'm quite curious how this org -> info export works, for example why it turns `- Key:` into the key markup.  Where are these things defined?

tarsius:
See https://github.com/tarsius/ox-texinfo-plus/blob/master/ox-texinfo%2B.el, but there's also some relevant stuff in the Makefile. I hope to get a rough overview on the wiki this weekend.

Fuco1:
Cool.  I've been tinkering with the idea of an info manual for Smartparens so this is definitely something that interests me.

PR #3253, opened 2017-11-09
On an uncommitted removed line visit HEAD

tarsius:
```
Since 050c8f070f this command visits the last commit that still had
that line when invoked with point on a removed line in a diff about a
committed change.

That commit also mentioned that we would do the same for unstaged and
staged changes and this commit finally implements that.

Setting `magit-diff-visit-previous-blob' to t reverts this change, but
I think the new behavior is very useful as it makes it much faster to
check why some line that is about to be removed was originally added:
"RET" to visit the blob from the commit that still had it (always
"HEAD" in this case) and then "b b" to find the commit that introduced
it.
```

050c8f070f1b3ec23f95d79dd2fe1a22423b2a73

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -214,17 +214,16 @@ many spaces.  Otherwise, highlight neither."
   :group 'magit-diff
   :type 'boolean)
 
-(defcustom magit-diff-visit-previous-blob t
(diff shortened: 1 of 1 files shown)


tarsius:
That needs more for reaching work.

PR #3256, opened 2017-11-10
Translate `pcase-lambda` to `lambda`/`pcase-let*`

### LINKED ISSUE #3255: pcase-lambda not available in emacs 24.5.1
The nice changes to completion and confirmation (#3232) check `magit-dwim-selection` with a `pcase-lambda` (https://github.com/magit/magit/commit/64059d311790b024dc839b393ab4a5fb989c107b#diff-bbef6f6dfe185467d9253695ced196e4R369), but that macro does not seem to be available in emacs 24.5.1.

On one machine (`M-x magit-version`: `Magit 20171109.1438, Git 2.7.4, Emacs 24.5.1, gnu/linux`) any operation that invokes `magit-completing-read` (e.g., `b b`) fails with
```
magit-completing-read: Invalid function: (\` ((\, cmd) (\, re) (\, _)))
```

Meanwhile, on another machine (`M-x magit-version`: `Magit 20171109.1438, Git 2.13.6, Emacs 25.3.1, gnu/linux`), operations succeed as expected.

MatthewFluet:
The `pcase-lambda` macro is not available in emacs 24.5.1 (e.g.,
provided by Ubuntu 16.04LTS), so we translate `pcase-lambda` to the
equivalent `lambda`/`pcase-let*`.

Closes #3255.

### DIFF
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -379,10 +379,11 @@ acts similarly to `completing-read', except for the following:
   `magit-builtin-completing-read'."
   (setq magit-completing-read--silent-default nil)
   (-if-let (dwim (and def
-                      (or (nth 2 (-first (pcase-lambda (`(,cmd ,re ,_))
-                                           (and (eq this-command cmd)
-                                                (or (not re)
-                                                    (string-match-p re prompt))))
+                      (or (nth 2 (-first (lambda (arg)
+                                           (pcase-let ((`(,cmd ,re ,_) arg))
+                                             (and (eq this-command cmd)
+                                                  (or (not re)
+                                                      (string-match-p re prompt)))))
                                          magit-dwim-selection))
                           (memq this-command
                                 (with-no-warnings magit-no-confirm-default)))))


lisp/magit-utils.el
```diff
@@ -379,10 +379,11 @@ acts similarly to `completing-read', except for the following:
   `magit-builtin-completing-read'."
   (setq magit-completing-read--silent-default nil)
   (-if-let (dwim (and def
-                      (or (nth 2 (-first (pcase-lambda (`(,cmd ,re ,_))
-                                           (and (eq this-command cmd)
-                                                (or (not re)
-                                                    (string-match-p re prompt))))
+                      (or (nth 2 (-first (lambda (arg)
+                                           (pcase-let* ((`(,cmd ,re ,_) arg))
```
tarsius: Since there is only one binding it would be better to use `pcase-let`.

tarsius:
Thanks. I have squashed the two commits.

PR #3269, opened 2017-11-16
Fix return value from magit-branch-read-args

### LINKED ISSUE #3268: magit-branch-and-checkout fails
when using `b` `c` in the status window to create and checkout a new branch, i get two prompts for the name (`foo`) and the upstream (`origin/master`) and then it fails:

```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  string-match("^stash@{[0-9]+}$" nil nil)
  magit-branch-and-checkout("foo" nil)
  funcall-interactively(magit-branch-and-checkout "foo" nil)
  call-interactively(magit-branch-and-checkout)
  magit-invoke-popup-action(99)
  funcall-interactively(magit-invoke-popup-action 99)
  call-interactively(magit-invoke-popup-action nil nil)
  command-execute(magit-invoke-popup-action)
```

i am using `magit-20171115.153` from melpa.

wbolster:
...when magit-branch-read-upstream-first is nil.

Fixes #3268

### DIFF
--- a/lisp/magit-branch.el
+++ b/lisp/magit-branch.el
@@ -352,7 +352,8 @@ when using `magit-branch-and-checkout'."
         (list branch
               (with-no-warnings
                 (let ((magit-no-confirm-default nil))
-                  (magit-read-starting-point prompt branch) args)))))))
+                  (magit-read-starting-point prompt branch)))
+              args)))))
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks!

wbolster:
yw, keep up the good work! :)

Issue #2958, opened 2017-01-12
Implement git-handler.el, a file-handler for blobs


tarsius:
*This description was taken from #2956. I intend to replace it with a more in-depth description at a later time.*

A number of Git related packages, including Magit, allow visiting blobs in a buffer.  These buffers are a lot like file-visiting buffers, except they are not.  The result is that the buffers created by one of these packages is only barely useful for that package that created them and not compatible with any of the other packages.

This can be fixed by turning blob-visiting buffers into actual file-visiting buffers.  That doesn't mean that a file has to be created, it only means that a *file-handler* has to be defined, a feature provided by Emacs which was originally implemented to allow accessing remote files *as if they were local files*.  In the case of blobs, a file-handler would allow accessing past versions of a file *as if it was the current version*.

This file-handler will likely be called just **`git-handler.el`**.

PR #3278, opened 2017-11-22
allow magit-file-mode to work in dired buffers

jpkotta:


tarsius:
At least the blame commands won't work in a dired buffer. The others commands in the popup have to be checked too.

jpkotta:
I mostly want to use `magit-status` and `magit-dispatch-popup`.  Diff and log commands in `magit-file-popup` applied to a directory would be handy.  The other commands in `magit-file-popup` might not be all that useful when applied to a directory.  I can try to generalize the diff and log commands in the file popup for directories, if you think that would be worthwhile.

jpkotta:
I mostly want to use `magit-status` and `magit-dispatch-popup`.  I think it would be handy to be able to diff and log a subdirectory from a dired buffer.  I can try to generalize the log and diff commands in the file menu if you think that would be worthwhile.

tarsius:
> I can try to generalize the diff and log commands in the file popup for directories, if you think that would be worthwhile.

Where it makes I intend to do that too, but only once I have implemented `git-handler.el` (#2958).

> I mostly want to use `magit-status` and `magit-dispatch-popup`.

Why not just bind those keys in `dired-mode-map` (or globally as recommended in [Getting Started](https://magit.vc/manual/magit/Getting-Started.html)?

PR #3296, opened 2017-12-08
Use a custom face for the process status indicator in the mode line

### LINKED ISSUE #3284: Some kind of status indicator for asynchronous activity?
Magit 2.11.0-4-g6b7dd53-dirty, Git 2.6.6, Emacs 25.2.1, gnu/linux

Sometimes I'll initiate an action (pushing or rebasing are both examples, I think) and Magit remains responsive, but it's waiting for the git command to finish before it can update the status buffer, and there's no visual feedback to tell me that this is happening unless I open the process log buffer.

If I'm expecting an action to happen faster than it does, it can be slightly confusing when the status buffer is not in the state I'm anticipating. Generally I'll open the process log to check what's happening, but I think it would be a useful improvement if we could tell at a glance that a process is running in the background.

Errors are displayed at the top of the status buffer, and I wonder whether other status notices should be displayed there too?  (There's the possibility that doing so would prove to be too distracting in general; or that it would make a change to an error message not distracting enough??)

phil-s:
New face `magit-mode-line-process' increases the visibility of the
running process when Git is run for side-effects.

In addition, `magit-process-sentinel' now forces the redisplay of all
mode lines, as otherwise the `mode-line-process' value set by Magit
could remain visible in other windows after it had been cleaned up in
the selected window.

Closes #3284.

### DIFF
--- a/Documentation/RelNotes/2.12.0.txt
+++ b/Documentation/RelNotes/2.12.0.txt
@@ -200,6 +200,15 @@ Changes since v2.11.0
   with `magit-branch-pull-request' or magit-checkout-pull-request',
   provided that remote has no other tracking branches.  #3134
 
+* Added new face `magit-mode-line-process' which is applied to the
+  mode line process status indicator shown while Git is running for
+  side-effects.  This improves the visibility of pending asynchronous
+  processes (in particular), as Magit remains responsive after
+  initiating such commands (for instance fetching or rebasing), but
+  will not update its buffers until the process has completed, which
+  might take longer than anticipated.  Customize this face if you wish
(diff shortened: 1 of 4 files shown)


tarsius:
Please split this into two separate commits.

phil-s:
Hmm. Is GitHub going to say `Merge branch 'phil-s/mode-line-process'` or `Merge branch 'phil-s/phil-s/mode-line-process'`? It's my habit to name-space my branch names, but the latter is going to look silly and inconsistent. Perhaps you can attend to that manually if need be?

tarsius:
Thanks!

PR #3299, opened 2017-12-09
manual: update pull-request key bindings

orivej:
Replace `b p` (the keys to configure push remote) with `b R` and `b r`.


tarsius:
Thanks!

PR #3302, opened 2017-12-10
magit-diff-tab-width fixups

basil-conto:
#3298 introduced a couple of issues:

1. Use of the `file-attribute-size` accessor, which was only added in Emacs 26.
2. Passing the new `defcustom` `magit-diff-adjust-tab-width`, whose value could be `t`, to an arithmetic operator.

The first commit addresses these two issues. The second fixes what I believe to be a docstring typo, and the last two commits are suggestions for trivial logic simplification and optimisation.

Let me know if you'd like anything written/split differently, and thanks for this cool feature. :)

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -2453,8 +2453,9 @@ are highlighted."
           (or (cdr it)
               tab-width)))
        ((or (eq magit-diff-adjust-tab-width 'always)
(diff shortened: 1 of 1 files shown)


basil-conto:
Ha, the commits appear in the order that I describe them on my branch, but GitHub is displaying them in an older order, with the docstring commit first.

tarsius:
Thanks!

PR #3306, opened 2017-12-14
magit-read-string: trim input string

### LINKED ISSUE #3305: Remove whitespace at the beginning and end in magit-read-string
I just tried to create a branch with a copied string that contained trailing whitespace. Instead of complaining about the whitespace, it would be nice if magit would trim the string.

I would use `s-trim` from s.el, but I'm not sure if you want it as a dependency.

brotzeit:
closes  #3305

There is a new function `string-trim`. I took the regular expressions from there.

### DIFF
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -517,7 +517,8 @@ back to built-in `completing-read' for now." :error)
 This is similar to `read-string', but
 * empty input is only allowed if DEFAULT-VALUE is non-nil in
   which case that is returned,
-* whitespace is not allowed if NO-WHITESPACE is non-nil,
+* whitespace is not allowed and leading and trailing whitespace is
+  removed automatically if NO-WHITESPACE is non-nil,
 * \": \" is appended to PROMPT, and
 * an invalid DEFAULT-VALUE is silently ignored."
   (when default-value
@@ -529,9 +530,17 @@ This is similar to `read-string', but
          (val (read-from-minibuffer
                (magit-prompt-with-default (concat prompt ": ") default-value)
                initial-input (and no-whitespace magit-minibuffer-local-ns-map)
-               nil history default-value inherit-input-method)))
+               nil history default-value inherit-input-method))
+         (trim (lambda (regexp string)
+                 (save-match-data
+                   (if (string-match regexp string)
+                       (replace-match "" t t string)
+                     string)))))
     (when (and (string= val "") default-value)
       (setq val default-value))
+    (when no-whitespace
+      (setq val (funcall trim "\\`\\(?:[ \t\n\r]+\\)"
+                         (funcall trim "\\(?:[ \t\n\r]+\\)\\'" val))))
     (cond ((string= val "")
            (user-error "Need non-empty input"))
           ((and no-whitespace (string-match-p "[\s\t\n]" val))


basil-conto:
> Unless `no-whitespace` is non-nil, you shouldn't do any of this.

I admit it took me a few tries to successfully parse this. :)

tarsius:
> closes #3305
> There is a new function string-trim. I took the regular expressions from there.

Please mention both of these things in the commit message also.

tarsius:
And then rebase onto `master`.

By the way, next time please use a dedicated feature branch. I just added a note about that to `CONTRIBUTING.md`.

lisp/magit-utils.el
```diff
@@ -529,9 +530,16 @@ This is similar to `read-string', but
          (val (read-from-minibuffer
                (magit-prompt-with-default (concat prompt ": ") default-value)
                initial-input (and no-whitespace magit-minibuffer-local-ns-map)
-               nil history default-value inherit-input-method)))
+               nil history default-value inherit-input-method))
+         (trim (lambda (regexp string)
+                 (if (string-match regexp string)
```
fbergroth: Wrap `save-match-data` around the body?

npostavs: @fbergroth Why?

fbergroth: Just because the code modifies the global match data, but it's a minor detail.

tarsius: That's always a good idea (except when you know it's not).

brotzeit:
Out of curiosity, are there downsides to using `push --force` when I'm the only one working on a branch ?

tarsius:
Sorry for the delay, I took a vacation.

Your commit message pretty weird, so I replaced it. Hope that is okay. The note about `string-trim` is a detail that does not belong in the summary and the issue number being in the middle of the line confused me quite a bit.

npostavs:
> Out of curiosity, are there downsides to using push --force when I'm the only one working on a branch ?

The downside is you lose the older version, and you might later actually want to see it (for trivial typo fixes it doesn't really matter).

tarsius:
But even then you should still be able to get the old version back from the reflog. But the maintainer might not be able to do so as that is only possible if that version ever existed locally. If the maintainer uses the new `magit-checkout-pull-request` command, then that does not only make the initial review more comfortable, it also means that s/he can easily compare the two versions after a forced push before resetting the local branch too.

brotzeit:
Thanks for the explanation. I wasn't sure which info you wanted in the commit message ;)

tarsius:
Speaking of information we would want in commit metadata:

Would you mind giving us your full name so that it can appear in https://github.com/magit/magit/blob/master/Documentation/AUTHORS.md?

brotzeit:
Fritz Stelzer brotzeitmacher@gmail.com

tarsius:
Thanks.

kyleam:
> it also means that the maintainer can easily compare the two versions after a forced push even before resetting the local branch

and, for comparing rebased branches, you may find {magit-,}[tbdiff] useful :)

[tbdiff]: https://github.com/trast/tbdiff

afrepues:
Gitlab offers the option to show the diff between "versions" of a merge request. Any push (forced or not) to the branch counts as a new "version".

So far I have used plain `git diff branch-v0..branch-v1` to see the difference between versions of topic branches, but I never overwrite my local branches, always create a new one, and force push to the same branch for merge requests.

PR #3309, opened 2017-12-15
Define magit-section as a class instead of as a struct

tarsius:
This is only a first step. The benefits will come once we start using generic functions.

### DIFF
--- a/lisp/magit-apply.el
+++ b/lisp/magit-apply.el
@@ -113,7 +113,7 @@ so causes the change to be applied to the index as well."
       (`(,_  files) (magit-apply-diffs  it args)))))
 
 (defun magit-apply--section-content (section)
-  (buffer-substring-no-properties (if (eq (oref section type) 'hunk)
+  (buffer-substring-no-properties (if (magit-hunk-section-p section)
                                       (oref section start)
                                     (oref section content))
                                   (oref section end)))
@@ -414,7 +414,7 @@ without requiring confirmation."
   (if (eq (magit-diff-type section) 'unstaged)
       (funcall apply section "--reverse")
     (if (magit-anything-unstaged-p
-         nil (if (eq (oref section type) 'file)
+         nil (if (magit-file-section-p section)
                  (oref section value)
                (magit-section-parent-value section)))
         (progn (let ((inhibit-magit-refresh t))
@@ -435,7 +435,7 @@ without requiring confirmation."
     (if (eq (magit-diff-type section) 'unstaged)
         (funcall apply sections "--reverse")
       (if (magit-anything-unstaged-p
-           nil (if (eq (oref section type) 'file)
+           nil (if (magit-file-section-p section)
                    (oref section value)
                  (magit-section-parent-value section)))
           (progn (let ((inhibit-magit-refresh t))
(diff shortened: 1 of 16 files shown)


vermiculus:
Will there be a need for `let-alist`-esque syntax? For example, I have a macro/function (somewhere) that works something like this:

```elisp
(oref* object prop subprop subsubprop)
```

tarsius:
> Will there be a need for `let-alist`-esque syntax?

Possibly. But it's to early to tell.

kyleam:
Huh, that Emacs 25 specific failure (25.1 and 25.3) on Travis is
interesting.  I can't trigger it locally with 25.3.

tarsius:
@kyleam @npostavs I am taking two weeks off.

tarsius:
Happy holidays!

npostavs:
> Huh, that Emacs 25 specific failure (25.1 and 25.3) on Travis is
> interesting. I can't trigger it locally with 25.3.

I can reproduce locally, but only if I `make lisp` first.

tarsius:
> I can reproduce locally

The issue was due to the use of `cl-typep`, which doesn't handle objects in Emacsen before 26.

tarsius:
Due to incompatibilities between Emacsen >= 25.1 and < 25.1 it is no longer possible to compile Magit using >= 25.1 and then run it using Emacs < 25.1. This is the error that you get when loading Magit if you try to do so:

```
Symbol's function definition is void: cl-struct-define
```

In other words, if you cross the 25.1 boarder, then you have to recompile Magit.

npostavs:
> The issue was due to the use of `cl-typep`, which doesn't handle objects in Emacsen before 26.

But the tests passed when running interpreted, so it's probably [Bug#27718 "(cl-typep instance-ab 'class-a) gives wrong answer when compiled"](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=27718).

tarsius:
Ah, then `cl-typep` is *supposed* to work, but doesn't do so reliably 😉 Thanks for the info.

DamienCassou:
> The benefits will come once we start using generic functions.

generic functions can be [refined on structs too](https://github.com/DamienCassou/mpdel/pull/1/files#diff-9c84db7631e9bd3bf209a9a33d068c9bR376).

DamienCassou:
> Use oref instead of dedicated slot accessors

I'm curious. Why do you do that?

tarsius:
> generic functions can be refined on structs too.

But a struct cannot inherit from another, can it?

> I'm curious. Why do you [prefer `oref`]?

It seems cleaner (though that might be a purely personal preference) and more informative. When I see an `oref` then I know that we just get some value, when I see a `magit-section-*` function, then I know stuff is being calculated or being done for side-effects.

DamienCassou:
> But a struct cannot inherit from another, can it?

I checked the documentation and it is indeed possible:

```emacs
(cl-defstruct person name)

(defgeneric my/entity-name (entity)
  "Return the name of ENTITY.")

(defmethod my/entity-name ((entity person))
  (person-name entity))

(cl-defstruct (astronaut (:include person))
  helmet-size)

(setq buzz (make-astronaut :name "Buzz"))

(my/entity-name buzz)
  => "Buzz"\
```

>> Why do you [prefer oref]?
> It seems cleaner [...]

thanks for your rationale. It makes sense. The downside is that it exposes the implementation:changing from a class back to a struct will be harder. I guess it's a trade-of. Thanks.

PR #3312, opened 2017-12-20
magit-submodule-add: check for Ido advice

johnmastro:
`magit-submodule-add` currently checks whether `ido-read-file-name` will be used by comparing with `eq` against `read-file-name-function`. However, as of Emacs 26, the `ido-everywhere` minor mode uses `add-function` rather than setting `read-file-name-function` directly. Thus, also check with `advice-function-member-p`.

On the other hand, I briefly tested `magit-submodule-add` with `ido-read-file-name` in Emacs 27.0.50 and it seemed to work as I expected, so maybe the kludge isn't necessary in newer Emacsen? I don't know what the original problem was though.

### DIFF
--- a/lisp/magit-submodule.el
+++ b/lisp/magit-submodule.el
@@ -132,7 +132,10 @@ PATH also becomes the name."
    (magit-with-toplevel
      (let* ((url (magit-read-string-ns "Add submodule (remote url)"))
             (path (let ((read-file-name-function
-                         (if (eq read-file-name-function 'ido-read-file-name)
+                         (if (or (eq read-file-name-function 'ido-read-file-name)
+                                 (advice-function-member-p
+                                  'ido-read-file-name
+                                  read-file-name-function))
                              ;; The Ido variant doesn't work properly here.
                              #'read-file-name-default
                            read-file-name-function)))


johnmastro:
Of course, right after submitting, I found the description of the original problem in 47eb9be, but I wasn't able to reproduce it in Emacs 27.0.50.

tarsius:
Would you mind checking if you can reproduce the issue that this kludge (any any of the incarnations) is supposed to fix with 25.3 and/or 26.0.90?

johnmastro:
I've now tried with 25.3 and 26.0.90 and was unable to reproduce the problem described.

However, I did find Ido's behavior in this scenario a bit annoying. I don't think it's the bug you described but rather a mismatch between my own expectations and how Ido works with initial input. When I tried to edit the initial input using standard keys like `M-f` and `M-d` it would instead run Ido commands (`ido-wide-find-file-or-pop-dir` and `ido-wide-find-dir-or-delete-dir` respectively) and land me in a recursive minibuffer.

So, though I didn't reproduce the original problem, I don't like how Ido works in this scenario and would personally prefer to use `read-file-name-default` here.

I'm not sure if my reaction is typical or not so it would be good to hear what others think as well.

tarsius:
Well it sounds like you have observed the same behavior as me but looked close enough to see that it is not a bug - just very annoying expected behavior. So I would say, lets merge this for a smoother experience on 26 (and 27, but I am not sure whether it is needed there based on what you have said). 

Overall it appears to be not very important. I was the one who noticed the weird behavior back when I was still using Ido and nobody has ever complained about the kludge or about the kludge having broken. (Except the issue that prompted this; the kludge also applying to other completion frameworks.)

johnmastro:
Sounds good to me. You're right, it probably is the same thing.  The main reason I didn't want to assume they're the same is that I didn't observe the initial input changing the second time I was prompted (upon landing in a recursive minibuffer), but that may depend on what you've done before triggering one of those commands. Also, I didn't want to discount the possibility that I was just confused about what Ido should be doing in that scenario :)

I re-tested in 27 now and the behavior is the same there as in 25 and 26, I just happened not to use any key bindings that landed me in a recursive minibuffer when I initially tried with 27.

tarsius:
Thanks and sorry for the delay. I went on vacation shortly after this arrived.

PR #3314, opened 2017-12-26
Use mode-line-process to indicate process errors

### LINKED ISSUE #3297: Use mode-line-process to indicate process errors
Following #3284 I'm considering enhancing this so that in the event of an error the `mode-line-process` value is *not* removed, and instead has an error face applied to it.  As before, this would increase the visibility of the process status.

In `magit-process-finish` I've changed:

       (when (buffer-live-p process-buf)
         (with-current-buffer process-buf
    -      (magit-process-unset-mode-line)
    +      (if (= arg 0)
    +          (magit-process-unset-mode-line)
    +        (magit-process-set-mode-line-error mode-line-process))

With the latter looking like:

    (defun magit-process-set-mode-line-error (str)
      "Apply an error face to the string set by `magit-process-set-mode-line'."
      (setq str (concat " " (propertize
                             (substring-no-properties str 1)
                             'face 'magit-mode-line-process-error)))
      (dolist (buf (magit-mode-get-buffers))
        (with-current-buffer buf
          (setq mode-line-process str)))
      (force-mode-line-update t))

Which seems to do the job, but does leaves the question of how and when this `mode-line-process` would be *removed*?

I could put it on a timer, which would at least ensure that it *does* get removed; but I don't really like that idea, because the user might not see it at all.

I thought about using `magit-insert-error-header` to remove it if there was *no* error, and therefore keep the mode line status in sync with the error notice in the Magit status buffer. I liked that idea, but realised there's no guarantee that a status buffer will exist, so this function isn't necessarily called at all.

I thought about removing it as soon as *any* magit buffer was interactively refreshed, and I think this might be the way to go, but I wasn't sure of the correct way to detect that.

What are your thoughts?

(n.b. as an aside, I've also realised my `(force-mode-line-update t)` call belongs in `magit-process-unset-mode-line` rather than `magit-process-sentinel`, and I'm guessing that this should happen when setting the value as well as unsetting it.)

phil-s:
Closes #3297

This provides an error state for `mode-line-process`, when the Git process exits with an error.

This includes an error face, a mouse tooltip with the error details, and a keymap so users can click to open the process log.

A by-product of this was a new general-purpose "repository-local" value cache which provides a more persistent alternative to buffer-local variables for storing data which is relevant to the repository in general.  I added the code for this in `magit-mode.el` primarily because that is where the `magit--default-directory` variable is defined, as this value is used to identify the repository.

Questions:

* I've made `magit-repository-local-repository` call `magit--not-inside-repository-error` if it can't establish a repository key. Is that sensible?

* I've limited the number of *lines* displayed in the tooltip for an error, but not the number of columns.  Is that likely to be a problem?

* In `magit-process-finish` it used to be the case that calling `magit-process-unset-mode-line` was one of the very first things to happen, which meant that even if an elisp error occurred in subsequent code, the mode line would still have been cleaned up.  Now that I'm leaving this until later, there is more potential for a stale `mode-line-process` to be left behind (but presumably only if there are bugs).  Is that ok, or should I split the code in order to take care of the no-error case back where it used to happen?

### DIFF
--- a/Documentation/RelNotes/2.12.0.txt
+++ b/Documentation/RelNotes/2.12.0.txt
@@ -200,14 +200,26 @@ Changes since v2.11.0
   with `magit-branch-pull-request' or `magit-checkout-pull-request',
   provided that remote has no other tracking branches.  #3134
 
-* Added new face `magit-mode-line-process' which is applied to the
-  mode line process status indicator shown while Git is running for
-  side-effects.  This improves the visibility of pending asynchronous
+* The mode line process indicator, displayed in the mode line when Git
+  is run for side-effects, is now more visible than before, and is
+  additionally used to highlight process errors.
+
+  New face `magit-mode-line-process' is applied to the mode line
+  process indicator to improve the visibility of pending asynchronous
   processes (in particular), as Magit remains responsive after
   initiating such commands (for instance fetching or rebasing), but
   will not update its buffers until the process has completed, which
   might take longer than anticipated.  Customize this face if you wish
   to make this indicator more (or less) visible.  #3284
+
+  If the Git process returns an error, the mode line process indicator
+  is no longer removed, and new face `magit-mode-line-process-error`
+  is applied to highlight the error status.  Details of the error from
+  the process buffer are also provided as a tooltip.  The error
+  indicator will remain visible in the mode line until a magit buffer
+  is refreshed.  If you do not want errors to be indicated in the mode
+  line, customize the `magit-process-display-mode-line-error' user
+  option.  #3297
 
 * When staging and unstaging at the file level and there are arguments
   that cause whitespace differences to be hidden, then apply the
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -1551,12 +1551,14 @@ will also have to install the ~ido-completing-read+~ package and use
 *** Viewing Git Output
 
 Magit runs Git either for side-effects (e.g. when pushing) or to get
-some value (e.g. the name of the current branch).  When Git is run for
-side-effects, the mode line displays a process status indicator (using
-the ~magit-mode-line-process~ face), and the process output goes into
-a per-repository log buffer, which can be consulted when things don't
-go as expected.  Process errors are indicated at the top of the status
-buffer.
+some value (e.g. the name of the current branch).
+
+When Git is run for side-effects, the process output is logged in a
+per-repository log buffer, which can be consulted using the
+~magit-process~ command when things don't go as expected.
+
+The output/errors for up to `magit-process-log-max' Git commands are
+retained.
 
 - Key: $, magit-process
 
@@ -1575,6 +1577,26 @@ sections are available.  There is one additional command.
   When this is non-nil then the output of all calls to git are logged
   in the process buffer.  This is useful when debugging, otherwise it
   just negatively affects performance.
+
+*** Git Process Status
+
+When a Git process is running for side-effects, Magit displays an
+indicator in the mode line, using the ~magit-mode-line-process~ face.
+
+If the Git process exits successfully, the process indicator is
+removed from the mode line immediately.
+
+In the case of a Git error, the process indicator is not removed, but
+is instead highlighted with the ~magit-mode-line-process-error~ face,
+and the error details from the process buffer are provided as a
+tooltip for mouse users.  This error indicator persists in the mode
+line until the next magit buffer refresh.
+
+If you do not wish process errors to be indicated in the mode line,
+customize the ~magit-process-display-mode-line-error~ user option.
+
+Process errors are additionally indicated at the top of the status
+buffer.
 
 *** Running Git Manually
 
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -1011,6 +1011,7 @@ Run hooks `magit-pre-refresh-hook' and `magit-post-refresh-hook'."
         (setq magit-section-highlighted-section nil)
         (setq magit-section-highlighted-sections nil)
         (setq magit-section-unhighlight-sections nil)
+        (magit-process-unset-mode-line-error-status)
         (let ((inhibit-read-only t))
           (erase-buffer)
           (save-excursion
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -1230,5 +1230,62 @@ Currently `magit-log-mode', `magit-reflog-mode',
                    (float-time (time-subtract (current-time) start))))
       (run-hooks hook))))
 
+(defvar magit-repository-local-cache nil
+  "Alist mapping `magit-toplevel' paths to alists of key/value pairs.")
+
+(defun magit-repository-local-repository ()
+  "Return the key for the current repository."
+  (or (bound-and-true-p magit--default-directory)
+      (magit-toplevel)
+      (magit--not-inside-repository-error)))
+
+(defun magit-repository-local-set (key value &optional repository)
+  "Set the repository-local VALUE for KEY.
+
+KEY should be a symbol, or otherwise comparable by `eq'.
+
+Unless specified, REPOSITORY is the current buffer's repository."
+  (let* ((repokey (or repository (magit-repository-local-repository)))
+         (cache (assoc repokey magit-repository-local-cache)))
+    (if cache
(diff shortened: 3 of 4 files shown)


tarsius:
Sorry for the delay, I took a vacation.

This is great. Thanks a lot!

tarsius:
I use the [`smart-mode-line`](https://github.com/Malabarba/smart-mode-line) package, which appears to strip the face and additional properties. Could you please look into that and submit a pull-request for that package if necessary?

/cc @Malabarba

phil-s:
Interesting. With `smart-line-mode` installed and enabled I can see the text properties *are* still present. e.g.:

    (with-current-buffer "*magit: .emacs.d"
      mode-line-process)

> #(" git rebase" 1 11 (mouse-face highlight keymap (keymap (mode-line keymap (mouse-1 . magit-process-buffer))) help-echo "Cannot rebase: You have unstaged changes.
Please commit or stash them.
> 
> mouse-1: Show process buffer" face magit-mode-line-process-error))

But I can confirm that none of the properties have any obvious effect when this library is used.

Malabarba:
It's probably because sml wraps its own properties around mode-line-process. Back when I wrote it (years ago) I wasn't aware of any package that added the properties to this string. I wouldn't mind a PR to fix that issue, but it would have to be a solution that doesn't sacrifice current functionality.

tarsius:
@phil-s I am currently generalizing the section visibility cache. It uses a buffer-local variable but I would like to, uhm, cache the cache when killing the buffer using repository-local values. Because the value is specific to a buffer/mode ^1 the key has the form `(MODE . magit-section-visibility)` so I have to change your accessor functions to compare using `equal`. I don't think that is a problem - do you see any?

I am changing the visibility cache to use `alist-get` and `(setf (alist-get ...) ...)`. Could you please look into doing the same for the repository-local cache?

_1 Only modes that cannot be used more than once per repository are supported at the moment.

phil-s:
I'll take a look, @tarsius.

The notable reason for using `eq` for equality in the `magit-repository-local-*` cache was because I was using an alist for the cache (which I had concluded was fine on account of expecting it to remain small), and for some reason `assq-delete-all` is a function which exists while `assoc-delete-all` does not exist.

It should be trivial enough to write equivalent code, though. 

I'm not getting a hit for `assoc-delete-all` searching the bug-gnu-emacs archives, so I guess I'll propose that as well.

basil-conto:
@tarsius 

> I am changing the visibility cache to use `alist-get`

This accessor was only added in Emacs 25, sadly.

npostavs:
> I'm not getting a hit for `assoc-delete-all` searching the bug-gnu-emacs archives, so I guess I'll propose that as well.

`(cl-delete KEY ALIST :key #'car :test #'equal)` should already cover it, I think.

tarsius:
> I've made `magit-repository-local-repository` call `magit--not-inside-repository-error` if it can't establish a repository key. Is that sensible?

Turns out it is not (and that we both didn't run the tests). It is allowed to run git outside any repository (e.g. to create a repository) and a macro used by many tests does exactly that. I have just changed the function to return `nil` instead of raising an error. This obviously causes a lookup with the invalid key `nil`, which should result in `nil` being returned, right? That's not a problem is it?

phil-s:
In that case we'll possibly want to inhibit the setting of repository-local values when the repokey is `nil`, as that's a valid key, so values could be inadvertently stored under that and subsequently retrieved in some entirely different context?

phil-s:
In the case of `magit-init` we could potentially set (or let-bind?) `magit--default-directory` for the git call, as we know what we expect the value to end up being if the `git init` succeeds.

As far as my own repo-local usage goes, if we do inhibit setting repo-local values for a nil repokey, then the `inhibit-magit-process-unset-mode-line` value never gets set if a Git error occurs for a Git call outside of a repo, which means the mode line wouldn't retain the error state long enough for anyone to see it.

(Which of course is no different to how things used to be.)

Other than the git init case (which can possibly be worked around as per the initial paragraph), I'm not sure whether this is an actual problem?

tarsius:
I am fine with keeping things as is but making this edge-case consistent with how things behave normally would certainly be nice. Please do so if you have the time.

> Depending on the situation it might also be useful to permit the `nil` key, but my gut says that the potential problems outweigh the potential benefits.

Maybe. But lets try to find out. If it doesn't work or things get to complicated then we can still give up on this.

> In the case of `magit-init` we could potentially set (or let-bind?) `magit--default-directory` for the git call, as we know what we expect the value to end up being if the `git init` succeeds.

Let-bind probably. Sounds a bit risky but worth a shot.

> Other than the git init case (which can possibly be worked around as per the initial paragraph), I'm not sure whether this is an actual problem?

We won't succeed anticipating all potential scenarios where that could be a problem. So this had to work outside a repository even if we explicitly handle some special cases that we know and care about.

> Are there other outside-of-a-repository scenarios in the tests where the new `mode-line-process` error state behaviour would be desirable?

For the tests all that matters is that it does not fail due to an unrelated bug. So as far as those go we are done after my fix.

Other non-test outside-of-a-repository scenarios include cloning, `git version`, and `git` commands that query a remote repository directly.

phil-s:
I reached the conclusion that, at least for the purposes of the `mode-line-process` work, `magit-init` doesn't actually need a workaround, as the current behaviour seems best (and indeed I created bugs when I tried let-binding `magit--default-directory`).

Currently, if you are not in a magit-mode buffer when issuing `magit-init`, then `magit-process-set-mode-line` does nothing (because it only acts on `magit-mode-get-buffers`), so there's nothing in the mode line for us to convert to the error state, should one occur.  So there's nothing to be gained when starting in this state.

If you *are* in a magit-mode buffer when issuing `magit-init` then you will be obtaining the repokey for the repository that you are in at that time (rather than the repository you are trying to create), but so far as `mode-line-process` is concerned I think that's actually desirable, because it means that `magit-process-set-mode-line` will do something (indicating the "git init" call), and therefore if there is an error, we can set the `mode-line-process` error state (albeit in the buffers for the original repository -- but as the git init failed, that seems reasonable).

I played with let-binding `magit--default-directory` in `magit-init` like so:

    (let ((magit--default-directory (expand-file-name directory)))
      (magit-call-git "init" (magit-convert-filename-for-git
                              (expand-file-name directory)))
      (magit-status-internal directory)))

but as well as being uncertain that this value was always guaranteed to match what Emacs would ultimately set as `default-directory`, doing this caused bugs.  When the new repo was created, it was associated with directory of the original repository.  Adding a binding for `default-directory` as well caused other problems.

Your original feeling was that this was a risky approach, which was clearly correct, so I'm not pursuing it any further.

phil-s:
> > Depending on the situation it might also be useful to permit the nil key, but my gut says that the potential problems outweigh the potential benefits.

> Maybe. But lets try to find out. If it doesn't work or things get to complicated then we can still give up on this.

I'm fairly keen to give up on allowing a `nil` repokey.

I don't have a specific example problem in mind, but it seems generally dangerous to me.  I can simply imagine that a value gets cached against the `nil` repokey, sits in the cache for ages, and at some point in the future the user triggers some other not-in-a-repository action and Magit unwittingly retrieves and uses what it thinks is an appropriate cached value, causing some bug which would probably be really difficult to track down.

Conversely, if values are *not* cached against the `nil` repokey and that resulted in a bug, I think that bug would be dramatically easier to isolate.

If there's a genuine need to put something in the repository-local cache when we're not currently in a repository, then an explicit REPOSITORY argument can be passed.  It's not clear to me that such a need exists or would arise, though.

Also, by ignoring the `nil` repokey, the 'set' function will itself return `nil` in that scenario which means we can *detect* if we didn't cache a value and react accordingly, if necessary.

phil-s:
> > I am changing the visibility cache to use alist-get

> This accessor was only added in Emacs 25, sadly.

That aside, I note that `alist-get` says it uses `assq`, so I'm not sure that's what you wanted if we need to be comparing with `equal`?

basil-conto:
>>> I am changing the visibility cache to use `alist-get`

>> This accessor was only added in Emacs 25, sadly.

> That aside, I note that `alist-get` says it uses `assq`, so I'm not sure that's what you wanted if we need to be comparing with `equal`?

Emacs 26 adds an optional `TESTFN` argument to `alist-get`. :)

tarsius:
@phil-s I agree. Thanks for the detailed analysis.

PR #3315, opened 2017-12-28
Add trailing space to yes-or-no-p prompts

phil-s:
A couple of yes-or-no-p prompts were missing a space after the prompt.


kyleam:
thanks

PR #3316, opened 2017-12-31
Update manual installation to include ghub dependency

cjlarose:
`make` does not succeed without `ghub` being available on the load path.


cjlarose:
Sorry about that! Updated `magit.org` instead.

kyleam:
thanks

PR #3317, opened 2018-01-01
git-commit.el: declare diff-default-read-only

npostavs:
When doing `make lisp` I get:
```
In toplevel form:
git-commit.el:802:1:Warning: Unused lexical variable ‘diff-default-read-only’
```

Normally, I would just go ahead and push such a trivial fix directly, but I don't see this warning in the travis logs, so I'm wondering if there is something wrong on my end.

### DIFF
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -127,6 +127,7 @@
 (defvar flyspell-generic-check-word-predicate)
 (defvar font-lock-beg)
 (defvar font-lock-end)
+(defvar diff-default-read-only)
 
 (declare-function magit-expand-git-file-name 'magit-git)
 (declare-function magit-list-local-branch-names 'magit-git)


tarsius:
I wasn't able to reproduce this issue, but think we should make the dependency explicit regardless of this issue.

npostavs:
I checked this more closely; without your update, the warning goes away if I do `make clean` before `make lisp`. The dependency chain was: 
* git-commit.el `(require 'magit-git nil t)`
* magit-git.el `(require 'magit-utils)`
* magit-utils.el `(eval-when-compile (require 'vc-git))`
* vc-git.el `(require 'diff-mode)`

So if `magit-utils.elc` was lying around then `diff-mode` wouldn't be loaded.

tarsius:
Thanks! I followed the chain too, but failed to notice that it mattered whether we are compiling or not. That of course means that my previous change results in `diff-mode` being needlessly loaded in the normal case.

PR #3322, opened 2018-01-09
magit-emacs-Q-command: Add ghub

xuchunyang:
I run command provided by `magit-emacs-Q-command` then Emacs enters the debugger and complains ghub is missing.


tarsius:
Thanks!

PR #3326, opened 2018-01-11
Bug fix for Git error summary messages

phil-s:
It was possible for Magit to report an earlier error instead of the current error.

I noticed this on account of "Cannot rebase" not being recognised as an error message, so I've added that to the list.


tarsius:
Thanks!

Issue #2941, opened 2017-01-09
Allow users to select arguments to be shown in a popup


tarsius:
This depends on the complete rewrite of the code that implements the popups, but is worth mentioning separately because "please add argument `--foo` to the `bar` popup" is something that comes up a lot.

Usually I respond by saying that the we intentionally only provide a subset of the available arguments to avoid overwhelming users for whom most of the other arguments would be overkill, and by pointing the user to the [documentation](https://magit.vc/manual/magit-popup/Customizing-Existing-Popups.html) on how s/he can add arguments to existing popups. I also encourage user to add an entry to [this wiki page](https://github.com/magit/magit/wiki/Additional-proposed-infix-arguments-and-suffix-commands).

However we could do better than that. Popup definitions should define all the available arguments (with some exceptions, and while adding specialized completion functions where necessary), but mark most of those arguments to not be shown by default.

Users should be able to press some keys in the popup to show all arguments in order to use a non-standard one this one time. At the same time it should be possible to select the arguments to be shown by default (directly from the popup but also through the Custom interface).

---

This description left out a lot of details, and I might fill them in at a later time. For now this should serve to show those users who had their proposed arguments rejected that I am taking their needs seriously and am thinking about ways to satisfy them (without overwhelming other users).

tarsius:
Transient adds support for this, see https://magit.vc/manual/transient/Enabling-and-Disabling-Suffixes.html.

I haven't added many new infixes and suffixes on higher levels yet. I plan to add more gradually.

PR #3329, opened 2018-01-12
Add subject-prefix to magit-format-patch

kennyballou:
This adds the ability to specify a subject prefix for patches created by
the `git-format-patch` command.  Adding this ability is especially
useful when creating versioned patch sets to be mailed.

Regarding the shortcut key, I've debated either `P` or `S`.  I'm not
aware of any other good candidates.

Signed-off-by: Kenny Ballou <kballou@devnulllabs.io>

### DIFF
--- a/lisp/magit-remote.el
+++ b/lisp/magit-remote.el
@@ -869,6 +869,7 @@ To add this command to the push popup add this to your init file:
               (?t "To"               "--to=")
               (?c "CC"               "--cc=")
               (?r "In reply to"      "--in-reply-to=")
+              (?P "Subject Prefix"   "--subject-prefix=")
               (?v "Reroll count"     "--reroll-count=")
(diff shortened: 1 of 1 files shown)


kennyballou:
Another comment, I noticed another [pull request](https://github.com/magit/magit/pull/2512) that implements the same change, but was closed.  Is there an updated reason for not adding subject prefix?

kyleam:
> Is there an updated reason for not adding subject prefix?

I'm not aware of any updated reason.  In general, when a popup
argument gets requested, the response is that the intention is not to
include every argument (but see #2941) and that it is easy for users
to [add] their own.

[add]: https://github.com/magit/magit/wiki/Additional-proposed-infix-arguments-and-suffix-commands

tarsius:
Beginners won't use the popup, so we don't have to avoid confusing them with too many arguments.

PR #3332, opened 2018-01-16
Use thing-at-point to guess which GitHub pull request is desired

vermiculus:
There's a good bit of functionality in Magithub that could be taken advantage of in Magit to be more DWIM. In vermiculus/magithub#201, I introduced thingatpt support for a few symbols like github-pull-request. I'm not totally sold on using the 'github' prefix here -- perhaps something more generic like 'gitforge' could be used instead.

This is a very simple patch right now and I expect it will change before it's merged. Specifically, I'm not sure how to handle maintaining the ability to choose a PR that is *not* the one at point. As it's set up right now, if point is on something that defines github-pull-request, the user won't have the opportunity to override that value.

### DIFF
--- a/lisp/magit-collab.el
+++ b/lisp/magit-collab.el
@@ -91,7 +91,10 @@ only supports Github, but that will change eventually."
                            (cdr (assq 'number pull-request))
                            (cdr (assq 'title  pull-request)))))
          (prs    (ghub-get (format "/repos/%s/pulls" id) nil :auth 'magit))
-         (choice (magit-completing-read prompt (mapcar fmtfun prs)))
+         (choice (magit-completing-read
+                  prompt (mapcar fmtfun prs) nil nil nil nil
+                  (let ((default (thing-at-point 'github-pull-request)))
+                    (and default (funcall fmtfun default)))))
          (number (and (string-match "\\([0-9]+\\)" choice)
                       (string-to-number (match-string 1 choice)))))
     (and number
--- a/lisp/magit-collab.el
+++ b/lisp/magit-collab.el
@@ -86,13 +86,12 @@ only supports Github, but that will change eventually."
          (id     (and (string-match "github.com[:/]\\(.+?\\)\\(?:\\.git\\)?\\'"
                                     url)
                       (match-string 1 url)))
+         (fmtfun (lambda (pull-request)
+                   (format "%s  %s"
+                           (cdr (assq 'number pull-request))
+                           (cdr (assq 'title  pull-request)))))
          (prs    (ghub-get (format "/repos/%s/pulls" id) nil :auth 'magit))
-         (choice (magit-completing-read
-                  prompt
-                  (--map (format "%s  %s"
-                                 (cdr (assq 'number it))
-                                 (cdr (assq 'title  it)))
-                         prs)))
+         (choice (magit-completing-read prompt (mapcar fmtfun prs)))
          (number (and (string-match "\\([0-9]+\\)" choice)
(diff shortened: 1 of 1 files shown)


vermiculus:
After a little more thought, simply changing the name won't be appropriate; this would preclude packages (say, a gitlab integration package) from providing its own definition. Ideally, there should be some way to get `(thing-at-point 'gitforge-pull-request)` to detect the forge in use from context and defer to the appropriate symbol (`github-pull-request`, `gitlab-pull-request`, …).

tarsius:
> I'm not totally sold on using the 'github' prefix here -- perhaps something more generic like 'gitforge' could be used instead.

Since neither `magithub` nor `magit` support other git forges currently I would stick to `github`.

> Ideally, there should be some way to get `(thing-at-point 'gitforge-pull-request)` to detect the forge in use from context and dispatch to the appropriate symbol (`github-pull-request`, `gitlab-pull-request`, …).

Probably 😉 but I guess we can worry about that later.

> As it's set up right now, if point is on something that defines github-pull-request, the user won't have the opportunity to override that value.

See `magit-dwim-selection` and `magit-completing-read`. The default should probably be to prompt and offer the thing at point as the default.

vermiculus:
As I understand it, `magit-completing-read` can do that for us if we provide the appropriate value for `DEF`. I've taken that approach above – what do you think?

lisp/magit-collab.el
```diff
@@ -86,13 +86,11 @@ only supports Github, but that will change eventually."
...
+                   (let-alist pull-request
+                     (format "%s  %s" .number .title))))
          (prs    (ghub-get (format "/repos/%s/pulls" id) nil :auth 'magit))
-         (choice (magit-completing-read
-                  prompt
-                  (--map (format "%s  %s"
-                                 (cdr (assq 'number it))
-                                 (cdr (assq 'title  it)))
-                         prs)))
+         (choice (magit-completing-read prompt (mapcar fmtfun prs)))
```
tarsius: Lets just leave this as-is, shall we.

* I don't like adding a variable. That's my main dislike for this change.
* I kinda prefer `--map`.
* These two changes are not really the same. We could use `mapcar` and an inline `lambda`. But I don't quite see the point. If you contributed new code like that I would not have said anything, but I don't see a need for a change.
* I kinda like using `let-alist`, but since elsewhere in `magit-read-pull-request` we use `(cdr (assq ...))` just *once*, using `let-alist` there would be kinda silly. And if we only do it here but not there, then it would be inconsistent.

tarsius:
I think this should be done *inside* `magit-read-pull-request` instead of expecting every caller to provide the default.

lisp/magit-collab.el
tarsius: In what way is that not ensured now? Isn't this a purely cosmetic change?

vermiculus: No, unfortunately. In the commit we're commenting on right now -- yes it's a cosmetic change. I split it out though to make the change more straightforward and digestible, but it's necessary to make sure DEFAULT (an alist) is correctly formatted into a string for magit-completing-read.

vermiculus:
I've addressed some of your comments above. I'm not sure where we left the `--map`/`lambda` discussion, though.

lisp/magit-collab.el
tarsius: Ah! Then it makes perfect sense.

The problem I have with let-binding something and then only using it once is that I then end up looking for the second use. If there is a second use, then that obviously isn't a problem. 😄

tarsius:
Thanks!

PR #3335, opened 2018-01-18
git-commit-turn-on-flyspell: respect customized comment character

dakra:
I'm not sure about my commit message if it should be more elaborate.
Guess that one line is enough?!

I customized `core.commentchar` as sometimes I just put
`#123` as a reference to a issue in it's own line so I
set comment char to `;`.

### DIFF
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -498,12 +498,13 @@ finally check current non-comment text."
   (turn-on-flyspell)
   (setq flyspell-generic-check-word-predicate
         'git-commit-flyspell-verify)
-  (let (end)
+  (let ((end)
+        (comment-start-regex (format "^\\(%s\\|$\\)" comment-start)))
     (save-excursion
       (goto-char (point-max))
-      (while (and (not (bobp)) (looking-at "^\\(#\\|$\\)"))
+      (while (and (not (bobp)) (looking-at comment-start-regex))
         (forward-line -1))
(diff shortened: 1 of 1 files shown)


tarsius:
By the time this function is run as a member of `git-commit-setup-hook`, `git-commit-setup-font-lock` has already been run, which also consults the value of `core.commentchar` but then also stores it in `comment-start`. So should be able to just use the value of `comment-start` here.

> I customized `core.commentchar` as sometimes I just put `#123` as a reference to a issue in it's own line so I set comment char to `;`.

Not sure that's a good idea. It will make it harder for others to fix a typo in your commit messages before merging.

dakra:
> So should be able to just use the value of comment-start here

Great. I updated the commit.

> Not sure that's a good idea. It will make it harder for others to fix a typo
> in your commit messages before merging

Good point, didn't think about others editing my message and will
avoid it in the future.
I think changing the comment char is still useful as maybe others start
a line with an issue number and that makes it easier for me to edit their
message. And I can't think of a case where someone wants to start
a line with `;`.

tarsius:
Thanks!

> I think changing the comment char is still useful as maybe others start a line with an issue number and that makes it easier for me to edit their message.

As a default Magit will continue to stick to ... the default (Git's) 😉 

> And I can't think of a case where someone wants to start a line with `;`.

I think the Emacs developers use it to mark "extremely boring" commits.

PR #3338, opened 2018-01-19
magit-submodule-update: Make operation resursive

nikital:
`git submodule update` is invoked in order to match the submodules with the
currently checked-out superproject commit. It makes sense to perform the update
operation recursively. Otherwise, the working tree won't fully match the
superproject commit because nested submodules won't be updated.

Magit is great software, thanks! :heart:


tarsius:
I don't think that is universally true, so this argument should be added to the popup instead of being hard-coded. The only other open pull-request is about submodules too and while it is pretty old by now, it just so happens that I plan to finish that next week. (As you can see I have already added  `--recursive` to the popup, but other than that some big changes may be in order before I can merge that pr.)

Issue #3300, opened 2017-12-09
Magit pull-request can only check out 30 most recent PRs


orivej:
When the repository contains many outstanding PRs, all but the 30 most recent ones can not be selected at the `b r` prompt.

(I am working with https://github.com/NixOS/nixpkgs/ that have 530 open PRs at the moment. I don't need the completion functionality in the PR selection prompt, and would rather just paste the number without waiting for GitHub to respond to the `/pulls` API call.)

vermiculus:
Needs to be unpaginated at `magit-read-pull-request`.

orivej:
It takes GitHub 4 to 5 seconds to return the maximum of 100 PRs per page ([test](https://api.github.com/repos/NixOS/nixpkgs/pulls?per_page=100)). I'd rather not wait 20–30 seconds to select the PR.

vermiculus:
Perhaps [a search](https://developer.github.com/v3/search/#search-issues) could be used here.

I'm just not sure pulling by number only is the best option for most people. We could try to do some heuristics with number of responses that will be returned (looking at last page per `rel="last"` and the number of responses per page) – a method could be selected like that.

vermiculus:
Actually, it wouldn't be crazy for `b C-u r` to open a popup of other pull request selection methods and maybe have a customizable option (or probably rather a git variable) for the default method.

orivej:
By default Magit could keep suggesting 30 PRs, but also accept any numeric input and query an unknown number at `/pulls/%d`. (This opens the path for an option to disable PR suggestions, to speed up PR selection even further.)

orivej:
> This opens the path for an option to disable PR suggestions, to speed up PR selection even further.

That option would not be needed if Magit queried `/pulls` after displaying `Checkout pull request:` prompt: either when we user presses `TAB`, or automatically but asynchronously.

tarsius:
> It takes GitHub 4 to 5 seconds to return the maximum of 100 PRs per page (test). I'd rather not wait 20–30 seconds to select the PR.

That's pretty shocking. But appears to be true.

> By default Magit could keep suggesting 30 PRs, but also accept any numeric input and query an unknown number at `/pulls/%d`.

Could you implement that please?

> (This opens the path for an option to disable PR suggestions, to speed up PR selection even further.)

Yes, but let's not do that yet.

There are at least three ways we could get the data:

1. Get it when needed. That's what Magit currently does, but that's just a first iteration.
[...]

PR #3346, opened 2018-01-26
magit-read-pull-request: Fetch all pull requests from GitHub

CyberShadow:
Use :unpaginate to fetch all pages, not just the first one.

And thanks for this feature (`magit-checkout-pull-request`), it's awesome!

### DIFF
--- a/lisp/magit-collab.el
+++ b/lisp/magit-collab.el
@@ -90,7 +90,8 @@ only supports Github, but that will change eventually."
                    (format "%s  %s"
                            (cdr (assq 'number pull-request))
                            (cdr (assq 'title  pull-request)))))
-         (prs    (ghub-get (format "/repos/%s/pulls" id) nil :auth 'magit))
+         (prs    (ghub-get (format "/repos/%s/pulls" id) nil
+                           :auth 'magit :unpaginate t))
          (choice (magit-completing-read
                   prompt (mapcar fmtfun prs) nil nil nil nil
                   (let ((default (thing-at-point 'github-pull-request)))


npostavs:
> Use :unpaginate to fetch all pages, not just the first one.

If there are enough open PRs that you need to fetch multiple pages, wouldn't fetching them all be painfully slow?

CyberShadow:
I tested with https://github.com/dlang/phobos, which has 102 open pull requests currently. It's slower, but not unbearably slow. Perhaps `ghub` could show the unpagination progress in the echo area?

vermiculus:
In magit/ghub#43, I mentioned ghub needs some means of controlled pagination. I know some magithub users are often on what I'd consider slow internet, which means they're impacted by decisions such as these  more than I'd realize (with my relatively fast 20-30 mbps).

The best approach for users would probably be to grab the 30 most recent PRs and allow for pulling down the next page of results interactively during read.

tarsius:
Also see #3300 and bd5da98738feb202d29f439eb2a167c68627f9e3.

CyberShadow:
Thanks. Maybe this can be exposed as a preference? Not sure if that would conflict with your plan to allow explicitly fetching all pull requests.

Typing in the PR number is a sensible workaround, though ideally the user ought to be somehow informed that the list is incomplete (due to a mostly-arbitrary limitation in place for performance reasons).

CyberShadow:
> ideally the user ought to be somehow informed that the list is incomplete

For instance:
- `ghub-get` and friends somehow communicate to the caller (e.g. by setting a global) whether the last invocation returned all results, or just the first page
- `magit-read-pull-request` detects this situation, and modifies the `prompt` by appending " (showing first %d out of about %d)" to it

vermiculus:
Some other options are available on magit/ghub#43.

tarsius:
> Thanks. Maybe this can be exposed as a preference? Not sure if that would conflict with your plan to allow explicitly fetching all pull requests.

It conflicts in so far as that option would be fairly short-lived. I would rather not add it.

> Typing in the PR number is a sensible workaround, though ideally the user ought to be somehow informed that the list is incomplete (due to a mostly-arbitrary limitation in place for performance reasons).

Likewise I feel any effort adding a kludge now would be fasted once Magit switches to an explicit "fetch all github metadata now" approach. But...

> `magit-read-pull-request` detects this situation, and modifies the `prompt` by appending " (showing first %d out of about %d)" to it

... that sounds reasonable. Feel free to implement and submit it once `ghub` supports it.

vermiculus:
One alternative I just thought of is to use magithub in the meantime; pull requests are unpaginated before they're inserted into the buffer and it teaches `thing-at-point` to recognize that it's a `github-pull-request`. As of #3332, `magit-read-pull-request` knows to use the PR at point as the default.

PR #3347, opened 2018-01-26
git-commit-prev-message: Fix error in buffers without comments

CyberShadow:
When attempting to invoke `git-commit-prev-message` or `git-commit-next-message` in a buffer without comments, the functions would fail with the message:

    git-commit-prev-message: Search failed: "^#"

Such a buffer could be the result of a `prepare-commit-msg` git hook which does not emit any comment lines.

The rest of the code was already prepared to handle this scenario, so just fix the `re-search-forward` invocation to not error when the regexp was not found, and allow the caller to handle the failure.


tarsius:
Thanks!

PR #3355, opened 2018-01-31
magit-worktree-delete: check for worktree directory before deleting

kyleam:
```
'git worktree list' (and thus magit-list-worktrees) may return
worktrees for which a directory no longer exists (e.g., if the
directory was deleted outside of magit).  Avoid the delete-directory
call in this case because it leads to an error and because the "git
worktree prune" call that follows is sufficient cleanup.
```


kyleam:
Right after I submitted that I thought "my 'this is trivial/should
just commit directly gauge' is off".  Sorry, noise made and damage
done.

PR #3358, opened 2018-02-02
Add commands to absorb (merge and remove) a branch

### LINKED ISSUE #2922: Add a command to merge the current branch into another and then remove it
When I am done with a feature branch, then the sequence of steps are often the same:

1. Checkout the upstream branch (usually `master`).
2. Merge the feature branch.
3. Force push the feature branch.
4. Delete the feature branch.
5. Delete the remote feature branch.

A command `magit-merge-into` (defaulting to the upstream) could take care of all this.

There are complications:

* (5) could cause issues for CI. But the general idea seams useful and initially just doing (1-4) would already be a big step forward.
* If the upstream is e.g. `origin/master`, then merge into the local branch that has the same name (provided its upstream is the former?).

tarsius:
Closes #2922.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -4589,6 +4589,30 @@ following commands.
   user can then further adjust the merge, even when automatic conflict
   resolution succeeded and/or adjust the commit message.
 
+- Key: m a, magit-merge-absorb
+
+  This command merges another local branch into the current branch and
+  then removes the former.
+
+  Before the source branch is merged, it is first force pushed to its
+  push-remote, provided the respective remote branch already exists.
+  This ensures that the respective pull-request (if any) won't get
+  stuck on some obsolete version of the commits that are being merged.
+  Finally, if ~magit-branch-pull-request~ was used to create the merged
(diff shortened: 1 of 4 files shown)


Documentation/magit.org
```diff
@@ -4589,6 +4589,30 @@ following commands.
...
+  then removes the former.
+
+  Before the source branch is merged, it is first force pushed to its
+  push-remote, provided the respective remote branch already exists.
+  This ensures that the respective pull-request (if any) won't get
+  stuck on some obsolete version of the commits that are being merged.
+  Finally, if ~magit-branch-pull-request~ was used to create the merged
+  branch, then the respective remote branch is also removed.
+
+- Key: m a, magit-merge-into
```
vermiculus: is this the right keybinding?

tarsius: Nope...

tarsius:
In the future I might add variants that allow editing the message and the changes before making the merge commit. And also a variant that doesn't require the use of `--no-ff` to make sure that no fast-forward merge is performed.

PR #3359, opened 2018-02-02
Add commands to move cherries between branches

### LINKED ISSUE #2921: Add a command to extract some commits from a branch
Similar to `magit-branch-spinoff` a new command (`magit-cherry-spinoff`?) could extract commits from the current branch and put them on another, possibly new, branch. And then obviously remove them from the current branch. Unlike for `magit-branch-spinoff` these commits would not have to include `HEAD`.

1. If the current branch is `master` or another long-lived branch, then the commits would be extracted to a new feature branch (at least by default).

2. If the current branch is a feature branch, then they would be extracted to `master` (at least by default). This is the primary use-case. You are working an a new feature and run into typos and bugs and fix them as you go using separate commits, without having to break the flow of things. Later you can extract the fixes onto master, without having to do an explicit rebase.

A rebase is still necessary though.

This command would work like this:
* Checkout the "destination branch".
* Cherry-pick the commits.
* Checkout the "source branch" again.
* Rebase to actually remove the commits.

Obviously both the cherry-picking and rebasing can fail and the challenge is to determine how to continue after the user resolved those conflicts. Maybe simply leaving it to the user is the best we can do, after all that's what s/he would have had to do anyway (in all cases) did this command not exist.

tarsius:
Closes #2921.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -5159,6 +5159,59 @@ features the following commands.
   This command also has a top-level binding, which can be invoked without
   using the popup by typing ~a~ at the top-level.
 
+The following commands not apply some commits to some branch, but also
+remove them from some other branch.  The removal is performed using
+either ~git-update-ref~ or if necessary ~git-rebase~.  Both applying
+commits as well as removing them using ~git-rebase~ can lead to
+conflicts.  If that happens, then these commands abort and you not
+only have to resolve the conflicts but also finish the process the
+same you would have to if these commands didn't exist at all.
+
+- Key: A h, magit-cherry-harvest
+
+  This command moves the selected COMMITS that must be located on
+  another BRANCH onto the current branch instead, removing them the
+  former.  When this command succeeds, then the same branch is current
+  as before.
+
+  Applying the commits on the current branch or removing them from the
+  other branch can lead to conflicts.  When that happens, then this
+  command stops and you have to resolve the conflicts and then finish
+  the process manually.
+
+- Key: A d, magit-cherry-donate
+
+  This command moves the selected COMMITS from the current branch onto
+  another existing BRANCH, removing them from the former.  When this
+  command succeeds, then the same branch is current as before.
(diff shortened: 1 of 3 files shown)


Documentation/magit.org
```diff
@@ -5159,6 +5159,59 @@ features the following commands.
   This command also has a top-level binding, which can be invoked without
   using the popup by typing ~a~ at the top-level.
 
+The following commands not apply some commits to some branch, but also
```
basil-conto: "The following commands not **only** apply..."?

Documentation/magit.org
```diff
@@ -5159,6 +5159,59 @@ features the following commands.
...
+remove them from some other branch.  The removal is performed using
+either ~git-update-ref~ or if necessary ~git-rebase~.  Both applying
+commits as well as removing them using ~git-rebase~ can lead to
+conflicts.  If that happens, then these commands abort and you not
+only have to resolve the conflicts but also finish the process the
+same you would have to if these commands didn't exist at all.
```
basil-conto: "...the same **way** you would have to..."?

Documentation/magit.org
```diff
@@ -5159,6 +5159,59 @@ features the following commands.
...
+
+- Key: A h, magit-cherry-harvest
+
+  This command moves the selected COMMITS that must be located on
+  another BRANCH onto the current branch instead, removing them the
```
npostavs: "...removing them **from** the former."

tarsius: Any more? 🙄

npostavs: Nope, I think that's the last one :)

tarsius: Thanks

basil-conto: > Nope, I think that's the last one :)

To borrow a phrase from Homer (not that one, the other one): "The last one *so far*." :)

PR #3365, opened 2018-02-05
Don't apply w32 env hacks for remote repositories

### LINKED ISSUE #3345: magit can't find "git" in Tramp from local PC to remote nix-like
(1) what behavior you expected
In Tramp mode (plink or ssh, local Windows 7 PC, remote linux/unix), I expect magit can find "git" remotely and works as normal after setting
(add-to-list 'tramp-remote-path 'tramp-own-remote-path)

  (2) what behavior you observed
The magit can't find "git".

  (3) and how we can reproduce the issue.
Step 1. (setq magit-git-executable  "git") and (add-to-list 'tramp-remote-path 'tramp-own-remote-path)
Step 2. Goto git repo in the remote via tramp, and magit-status

I have looked in the code, and it seems due to my-magit-process-file where process-environment is let-assigned. In windows, there is a PATH env variables in process-environment, and if it is let-assigned here, it will bring the local window PC PATH to the remote linux/unix environment. Therefore, magit can't find "git" in the remote server.
I don't think it is proper to take local process-environment to the remote.
Maybe it is just benefit Cygwin's noglob option and during the call and ensure unix eol conversion as it is documented?

Currently I use an advice as a temporary solution as reference for those meet same issue:
````emacs
    (defun my-magit-process-file (&rest args)
      "Don't set process-environment in tramp-like mode"
      (let ((default-process-coding-system (magit--process-coding-system)))
        (apply #'process-file args)))
    (advice-add 'magit-process-file :override #'my-magit-process-file)
````

npostavs:
This should fix #3345. The first two commits fix some other issues that I noticed while looking at that bug (https://github.com/magit/magit/issues/3345#issuecomment-360813859 and https://github.com/magit/magit/issues/3345#issuecomment-361331270 (bottom half)).

I want to double check the thing about tramp using `process-environment` though, as I would have assumed that `tramp-remote-process-environment` would be used instead...

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -70,6 +70,12 @@ this."
   :type '(choice (coding-system :tag "Coding system to decode Git output")
                  (const :tag "Use system default" nil)))
 
+(defvar magit-git-w32-path-hack nil
+  "Alist of (EXE . (PATHENTRY)).
+This specifies what additional PATH setting needs to be added to
+the environment in order to run the non-wrapper git executables
+successfully.")
+
 (defcustom magit-git-executable
   ;; Git might be installed in a different location on a remote, so
   ;; it is better not to use the full path to the executable, except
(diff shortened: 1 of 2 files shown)


tarsius:
Sorry for not replying earlier. Looks good to me. Please merge once you have checked the tramp thing.

npostavs:
I dusted off my old netbook and started up sshd on it. I could confirm tramp's behaviour as I described in https://github.com/magit/magit/issues/3345#issuecomment-362640311. The TRAMP manual also talks about the non-default value of `process-environment` right after it explains `tramp-remote-process-environment`, but I'm a bit confused as to how it fits together so I opened [Emacs Bug#30419](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=30419).

PR #3377, opened 2018-02-14
Allow 'submodule update' to use --recursive switch

DanielG:
`git submodule update` allows the --recursive flag but when set in the popup it does not get passed down to it since it is filtered out. This PR simply adds `--recursive` to `update`'s filter list.


tarsius:
Thanks!

PR #3379, opened 2018-02-16
Translate `pcase-lambda` to `lambda`/`pcase-let`

MatthewFluet:
The pcase-lambda macro is not available in emacs 24.5.1 (e.g.,
provided by Ubuntu 16.04LTS), so translate pcase-lambda to the
equivalent lambda/pcase-let.


tarsius:
Thanks!

PR #3407, opened 2018-03-31
Create new worktree for pull-request branches

tarsius:
In response to https://www.reddit.com/r/emacs/comments/88b51s/putting_the_i_back_in_ide_jane_sts_emacsbased/.

This is only a quick draft. However limited testing showed that finishing this up with <kbd>m i</kbd> should work. You will still have to delete the worktree itself in a second step.


tarsius:
I have pushed this a little beyond a minimal viable implementation, but there is still room for improvement.

* You still have to delete the worktree manually before you can delete the branch.
* I have experimented with how to read the location of the new working tree quite a bit, but all the methods that I could come up with had downsides, some of which are dependent on the used completion framework. I have gone with a rather simple approach for now.

PR #3408, opened 2018-03-31
Add browse-url support

tarsius:
This is a *very* early draft.

### DIFF
--- a/default.mk
+++ b/default.mk
@@ -63,6 +63,7 @@ ELS += magit-status.el
 ELS += magit-refs.el
 ELS += magit-files.el
 ELS += magit-collab.el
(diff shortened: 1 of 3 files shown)


eoranged:
Maybe better just integrate with [browse-at-remote](https://github.com/rmuslimov/browse-at-remote)?

stardiviner:
This function is great, I like it pretty much!

tarsius:
Instead of doing the dispatching inside a single command as done here, I will define a dummy command `magit-browse-thing`, similar to other `magit-*-thing` commands, and then remap its keybinding `C-c C-b` to specialized commands in section-specific keymaps. I'll do that as part of #3490.

PR #3412, opened 2018-04-03
Use relative paths for dired buffers

marsam:
aac74cc466c40adbbeba373d85e13f72b175bc5e added support for `magit-log-buffer-file` in dired buffers, but `default-directory` returns the full directory path instead of relative path. 

This change adds dired support for `magit-file-relative-name`, tell me if this is not desirable.

### Scenario
Calling `magit-log-buffer-file` in a dired buffer sets:
```elisp
(magit-log-arguments)    ;; => (("-n256" "--graph" "--decorate") ("~/code/magit/lisp/"))
```
expected:
```elisp
(magit-log-arguments)    ;; => (("-n256" "--graph" "--decorate") ("lisp/"))
```

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -644,7 +644,8 @@ tracked file."
   (unless file
     (with-current-buffer (or (buffer-base-buffer)
                              (current-buffer))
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks!

PR #3413, opened 2018-04-03
manual: update pull-request key bindings

basil-conto:
This is a repeat of #3299, but following c741dc3d035aeac060fa3992e0016e31c606b813.


tarsius:
Thanks!

PR #3418, opened 2018-04-09
magit-log-buffer-file: Use absolute filename in call to file-regular-p

jguenther:
Calling `magit-log-buffer-file` interactively with an active region would previously call `magit-log` with a corresponding `-L<beg>,<end>:<file>` argument, to display the log of the selected region. This currently does not work for any files that aren't in the base dir of the repo.

It looks like this function was changed in 293bf0d6c to only enable `-L` if the file is a regular file, using `file-regular-p`. However, it calls `file-regular-p` with the filename *relative to the repo's root directory*, but the `default-directory` of this command is the dir that the file is located in.

e.g. in the repo `/home/user/repo`, if editing the file `/home/user/repo/some/example.txt` with an active region, calling `magit-log-buffer-file` would end up checking `(file-regular-p "some/example.txt")`, which refers to the absolute filename `"/home/user/repo/some/some/example.txt"` which doesn't exist, and therefore `-L` arg isn't enabled.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -3493,8 +3493,13 @@ buffers that are being blamed.
 - User Option: magit-blame-show-headings
 
   This option controls whether blame block headings are initially
-  shown.  The headings can also be toggled locally using the command
-  ~magit-blame-toggle-headings~.
+  shown.  Whether blame headings are shown in the current buffer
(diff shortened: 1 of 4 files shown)


tarsius:
Thanks!

(The unreleated commits appear here because I rebased this onto `master` etc. without realizing that I had not yet pushed that to `origin/master`.

Issue #3263, opened 2017-11-15
Forward blame shows unexpected commit details


kvaneesh:
Using forward blame I am finding wrong commit details mapped to lines.

-aneesh

tarsius:
Please elaborate.

kvaneesh:
with latest magit from melpa, i do a git blame and after some time want to froward blame 'key binding f'. I see commit details show which doesn't touch the line at all. For example see the attached file. The commits shown around do_fork are not related to that file at all
![screenshot from 2017-11-15 21-47-38](https://user-images.githubusercontent.com/19270/32846971-b05ca244-ca4e-11e7-9f03-c10d1cd2bd2f.png)

tarsius:
Please run the equivalent git command in a shell and verify that it doesn't display the same information. Have you been using reverse blaming before support was added to Magit? Are you sure you understand what information it is supposed to display and that what is being displayed is not exactly that?

```
       --reverse <rev>..<rev>
           Walk history forward instead of backward. Instead of showing the
           revision in which a line appeared, this shows the last revision in
           which a line has existed. This requires a range of revision like
           START..END where the path to blame exists in START.  git blame
           --reverse START is taken as git blame --reverse START..HEAD for
           convenience.
```

tarsius:
If Magit really displays something that is different from what Git displays, then find a public replsitory in which you can reproduce the issue and the provided detailed steps on how to produce it. Also include the steps you performed on the command line to see whether Git shows the same information.

fice-t:
Could Magit provide a distinction between "forward" blaming and "reverse" blaming?

By "forward blaming" I am referring to the description used in `magit-blame-reverse`:

> For each line show the subsequent revision that removes it.

"reverse" blaming instead shows the revision that last contains the lines (a parent of the above).

I think it would be useful to find the commit that actually removes the given lines.

kyleam:
@fice-t I've read your comment a couple times and don't understand
what distinction you're asking for, but FWIW I think the description
in `magit-blame-reverse`'s doc-string is inaccurate (and maybe that's
what is leading to @kvaneesh's confusion).  `--reverse` doesn't show
"the subsequent revision that removes it" but instead "the last
revision in which a line has existed".

kyleam:
@kvaneesh try running `git blame --reverse HEAD^.. -- <file` on a file
that you know wasn't touched in the last commit.  You'll see that
every line is assigned HEAD's commit.

fice-t:
[...]

PR #3420, opened 2018-04-10
Support blaming the removal of lines

tarsius:
See #3263.


lisp/magit-blame.el
```diff
@@ -360,19 +369,37 @@ modes is toggled, then this mode also gets toggled automatically.
             (cond ((looking-at "^filename \\(.+\\)")
                    (oset chunk orig-file (match-string 1)))
                   ((looking-at "^previous \\(.\\{40\\}\\) \\(.+\\)")
+                   (setq       prev-rev  (match-string 1))
```
kyleam: You're binding prev-rev  globally here.

kyleam:
I would've liked to see some analysis/discussion about the choice to assign blame's "previous" field as the "deletion" revision in the case where the "last containing" revision has multiple children.  Based on the Git commit that added `blame --reverse` and on the ML post that I [mentioned before], Junio viewed the "deletion" revision as the more useful thing to report, but it was left as a todo.  So, it seems fair to say that blame's "previous" field wasn't/isn't viewed as an appropriate value for the "deletion" revision.

[mentioned before]: https://github.com/magit/magit/issues/3263#issuecomment-345154599

tarsius:
Two reasons:

1. It's what git currently provides. Finding another commit would mean having to implement another algorithm for finding another commit.
2. `git blame --reverse` finds a *pair* of commits related to a chunk. If the "last commit that still has this line" commit of this pair is good enough (which it appears to be, after all the end-user variant of `git blame --reverse` does display it), then why should that not be the case for the "first commit that doesn't have that commit anymore"?

PR #3421, opened 2018-04-11
magit-gitk-executable: fix default gitk filename, on Windows.

bustercopley:
Avoid an error where we try to open "git.exe/gitk", which triggers Tramp to try to extract a file "gitk" from an archive "git.exe", which causes an error, "package 'tramp-archive' not supported".

To reproduce the error (on Windows), have magit installed, then run "emacs -Q", then:

    M-x package-initialize RET
    M-x load-library RET magit-extras RET

### DIFF
--- a/lisp/magit-extras.el
+++ b/lisp/magit-extras.el
@@ -36,8 +36,9 @@
 
 (defcustom magit-gitk-executable
   (or (and (eq system-type 'windows-nt)
-           (let ((exe (expand-file-name
-                       "gitk" (file-name-directory magit-git-executable))))
+           (let ((exe (magit-git-string
+                       "-c" "alias.X=!x() { which \"$1\" | cygpath -mf -; }; x"
+                       "X" "gitk.exe")))
              (and (file-executable-p exe) exe)))
       (executable-find "gitk") "gitk")
   "The Gitk executable."
--- a/lisp/magit-extras.el
+++ b/lisp/magit-extras.el
@@ -37,7 +37,7 @@
 (defcustom magit-gitk-executable
   (or (and (eq system-type 'windows-nt)
            (let ((exe (expand-file-name
-                       "gitk" (file-name-nondirectory magit-git-executable))))
+                       "gitk" (file-name-directory magit-git-executable))))
              (and (file-executable-p exe) exe)))
       (executable-find "gitk") "gitk")
   "The Gitk executable."


npostavs:
Thanks, `file-name-nondirectory` is definitely the wrong thing there. You're running Emacs 27 right? I think before the new Tramp feature, this was just silently failing all along.

We should also be checking for `gitk.exe`, and not just plain `gitk`, I think.

bustercopley:
>You're running Emacs 27 right?

Right!
>I think before the new Tramp feature, this was just silently failing all along.

I think so.
>We should also be checking for gitk.exe, and not just plain gitk, I think.

Actually there's another issue with the search: magit-git-executable ends up pointing at git.exe in the git-core directory, and there is no gitk alongside it (at least, on Git for Windows). Perhaps we should be re-using the logic from magit-git-executable instead?
For info, this returns a usable filename for "gitk.exe" on Git for Windows:

    (process-lines "git" "-c" "alias.X=!x() { which \"$1\" | cygpath -mf -; }; x" "X" "gitk.exe")

This returns the filename for the "gitk" script, on Git for Windows and on MSYS2 Git:

    (process-lines "git" "-c" "alias.X=!x() { which \"$1\" | cygpath -mf -; }; x" "X" "gitk")

npostavs:
> Actually there's another issue with the search: magit-git-executable ends up pointing at git.exe in the git-core directory, and there is no gitk alongside it (at least, on Git for Windows).

Ah, with Git for Windows 1.9.5 `magit-git-executable` points to the `bin/` directory, and there is a `gitk` (not `gitk.exe`!) alongside it. There's also a `gitk.cmd` under `cmd/`. But we probably shouldn't worry too much about such an old version, since nobody even noticed this bug up until now anyway.

tarsius:
And I will drop support for versions before `v2.4.*` soon anyway. (That's the oldest version that still receives security patches.)

bustercopley:
Is there anything you want me to do? Should I close this PR (and maybe open an issue instead), or add another commit to this PR to fix the issue the way I suggested (re-using the logic from magit-git-executable)? Or something else?

npostavs:
> add another commit to this PR to fix the issue the way I suggested (re-using the logic from magit-git-executable)?

Yes, I think that's the right thing. But instead of `(car (process-lines "git" ...)` we should be able to use `(magit-git-string ...)` since this stuff gets loaded later.

tarsius:
> add another commit to this PR to fix the issue the way I suggested

Please do that.

bustercopley:
Thanks, done. I kept the file-executable-p check and fallback to (executable-find ...) to cater for Cygwin and MSYS2. I'm not sure that's necessary but it shouldn't hurt. I can't test it there. With native Emacs and Git for Windows, this gives cmd/gitk.exe as expected.

npostavs:
Thanks, I edited the commit messages a bit and pushed to master.

PR #3427, opened 2018-04-20
Fix typo: reselve -> reshelve

jonathn:
Fixing a typo. A cosmetic-only change for consistency.


tarsius:
Would you mind giving us your full name so that it can be added to `AUTHORS.md`?

jonathn:
vanderwijk

PR #3428, opened 2018-04-23
magit-read-starting-point: allow completion for FETCH_HEAD, ...

christophejunke:
This commit adds a list of predefined Git names for temporary branches
in addition to the already existing HEAD: FETCH_HEAD, ORIG_HEAD,
MERGE_HEAD and CHERRY_PICK_HEAD.

I mostly use FETCH_HEAD as a starting point for branch creation, when
fetching commits from Gerrit (I can write the name and create a
branch, but completion does not recognize it). The other names are
added for completeness.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1864,7 +1864,9 @@ the reference is used.  The first regexp submatch becomes the
                       (concat " "
                               (propertize branch 'face 'magit-branch-local))))
                " starting at")
-       (cons "HEAD" (magit-list-refnames))
+       (nconc (list "HEAD")
+              (magit-list-refnames)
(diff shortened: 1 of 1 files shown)


tarsius:
An important difference between `HEAD` and the other symbolic refs is that the former always exists while the latter do not. We shouldn't offer invalid completion candidates.

christophejunke:
That's a good point. I updated the patch, but I am not sure how much it covers all corner cases.

tarsius:
I've pushed a third commit, which places most of the symrefs after the refs. Could you please squash all these three commits into a single commit and then force push?

christophejunke:
Looks good to me.

Thank you.

PR #3442, opened 2018-05-09
Make the reset command prompts more explicit

### LINKED ISSUE #3438: Improved reset command confirmation message?
If I'm resetting the current branch to a commit from the log of some other branch, it would be reassuring if the confirmation message was explicit about the branch that I'm about to modify.

Would this change ever be incorrect?

    diff --git a/lisp/magit-reset.el b/lisp/magit-reset.el
    index 3dba6219..4f19e8d8 100644
    --- a/lisp/magit-reset.el
    +++ b/lisp/magit-reset.el
    @@ -55,9 +55,14 @@ head this effectively unstages all changes.
     With a prefix argument also reset the working tree.
     \n(git reset --mixed|--hard COMMIT)"
       (interactive (list (magit-read-branch-or-commit
    -                      (if current-prefix-arg
    -                          "Hard reset to"
    -                        "Reset head to"))
    +                      (let ((currentbranch (magit-get-current-branch)))
    +                        (if current-prefix-arg
    +                            (format "Hard reset %s to"
    +                                    (or currentbranch "detached head"))
    +                          (format "Reset %s to"
    +                                  (if currentbranch
    +                                      (format "head of %s" currentbranch)
    +                                    "detached head")))))
                          current-prefix-arg))
       (magit-reset-internal (if hard "--hard" "--mixed") commit))

phil-s:
Closes #3438 

* Makes the prompts for all magit-reset-* commands more explicit about what will be done.
* Hard resets are further highlighted by making the word "Hard" bold.

The new prompts are more verbose than I originally suggested in #3438, but I think it's worthwhile verbosity.

### DIFF
--- a/lisp/magit-reset.el
+++ b/lisp/magit-reset.el
@@ -54,10 +54,11 @@ head this effectively unstages all changes.
   "Reset the head and index to COMMIT, but not the working tree.
 With a prefix argument also reset the working tree.
 \n(git reset --mixed|--hard COMMIT)"
-  (interactive (list (magit-read-branch-or-commit
+  (interactive (list (magit-reset-read-branch-or-commit
                       (if current-prefix-arg
-                          "Hard reset to"
-                        "Reset head to"))
+                          (concat (propertize "Hard" 'face 'bold)
+                                  " reset %s to")
+                        "Reset %s to"))
                      current-prefix-arg))
   (magit-reset-internal (if hard "--hard" "--mixed") commit))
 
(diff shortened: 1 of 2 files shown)


tarsius:
In this prompt:

```
Hard reset phil-s/reset-confirmation, index & files to: 
```

`confirm` has the color used for local branches, and `Hard` is not bold.

tarsius:
A branch doesn't have a `head` - there is only one `HEAD`. A branch has a `tip`, but since a branch is nothing but a pointer `tip of branch` is actually equivalent of just `branch`. I don't think the extra verbosity actually helps those less familiar with the internals. It's just more to read and when you don't know what that extra information actually means that *increases* the confusion. And I don't want to use the prompt to try to teach users how resetting works, I just want to clarify what is going to happen to prevent user mistakes.

tarsius:
I am undecided on the "branch, index & files" part instead of just using the words hard and soft. That falls into "teaching about what rebase does" category. I think the extra verbosity would make it less that a user would note that s/he accidentally used the wrong reset variant or is on a different branch than expected, at least I would have that problem.

kyleam:
> I am undecided on the "branch, index & files" part instead of just using the words hard and soft

I vote for just saying "hard" and "soft".

phil-s:
No problem, will revert to the simpler wording.

> `Hard reset phil-s/reset-confirmation, index & files to:`
> confirm has the color used for local branches, and Hard is not bold.

I can't replicate that here, either with my normal confirm, or with a fresh config with only Magit installed (with latest Melpa + this change).  Something is post-processing the text and treating "confirm" as a keyword?  My code is doing this:

    (format (concat (propertize "Hard" 'face 'bold)
                    " reset %s, index & files to")
            (format "%s" (magit-get-current-branch)))

phil-s:
Prompts are now just:

* Reset BRANCH to
* Soft reset BRANCH to
* **Hard** reset BRANCH to

tarsius:
> If you add text including "confirm" into the prompt without using this patch, does the same thing still happen?

Yep. I guess `ivy` is doing that.

lisp/magit-reset.el
```diff
@@ -49,37 +49,48 @@ head this effectively unstages all changes.
...
 \n(git reset --mixed|--hard COMMIT)"
-  (interactive (list (magit-read-branch-or-commit
-                      (if current-prefix-arg
-                          "Hard reset to"
-                        "Reset head to"))
+  (interactive (list (if current-prefix-arg
+                         (magit-reset-read-commit
+                          (concat (propertize "Hard" 'face 'bold)
+                                  " reset %s to"))
+                       (magit-reset-read-commit "Reset %s to"))
```
tarsius: Since there's exactly one argument in both cases I would keep the reader function "on the outside".

phil-s: Agreed.  That had made more sense when the more elaborate version of the prompt was being used, but not now.  Pushed.

tarsius: That was my guess :wink:

tarsius:
Oh and I usually place "implementation details" later in the file, so this should go after `magit-reset-hard`, next to `magit-reset-internal`.

phil-s:
I'll try to remember that for future.

Shall I change the name `magit-reset-read-commit` to `magit-reset-read-branch-or-commit`?  (Or something different?)  I can't remember having a particular reason for the name I used, but it's striking me as inconsistent now.  I can see that I won't push any lines over 80 columns if I switch to the longer name, so it wouldn't mess with any formatting.

tarsius:
Either is find, but sure.

PR #3443, opened 2018-05-09
magit-blame--format-time-string: Correct timezone handling

ynilu:
Because the time variable is an Unix time which is not related to time zone,
we don't have to update time according to time zone.

If time-format contains the time zone symbol ( %z or %Z ),
use tz-in-second to format the time in committer time zone or author time zone.

If time-format doesn't contain the time zone symbol ( %z or %Z ),
assign tz-in-second to nil to format the time in local time zone.

### DIFF
--- a/lisp/magit-blame.el
+++ b/lisp/magit-blame.el
@@ -673,12 +673,14 @@ modes is toggled, then this mode also gets toggled automatically.
                (face-attribute 'magit-blame-heading :background nil t))))
 
 (defun magit-blame--format-time-string (time tz)
-  (setq time (string-to-number time))
-  (setq tz   (string-to-number tz))
-  (format-time-string
-   (or (magit-blame--style-get 'time-format)
-       magit-blame-time-format)
-   (seconds-to-time (+ time (* (/ tz 100) 60 60) (* (% tz 100) 60)))))
+  (let* ((time-format (or (magit-blame--style-get 'time-format)
+                          magit-blame-time-format))
+         (tz-in-second (and (not (version< emacs-version "25"))
+                            (string-match "%z" time-format)
+                            (car (last (parse-time-string tz))))))
+    (format-time-string time-format
+                        (seconds-to-time (string-to-number time))
+                        tz-in-second)))
 
 (defun magit-blame--remove-overlays (&optional beg end)
   (save-restriction
--- a/Documentation/RelNotes/2.13.0.txt
+++ b/Documentation/RelNotes/2.13.0.txt
@@ -9,3 +9,6 @@ Changes since v2.12.0
 
 Fixes since v2.12.0
 -------------------
+
+* Time zones were not handled correctly when calculating times for
+  blame headings.  #3443


tarsius:
> a major issue with your change is that `format-time-string` didn't get its time zone argument until Emacs 25, and Magit has Emacs 24.4 as its minimum version.

Since I am going to drop support for the latter soon anyway, it would be okay to just catch the `wrong-number-of-arguments` error and then try again without that argument.

kyleam:
> it would be okay to just catch the wrong-number-of-arguments error and then try again without that argument.

Both versions of `format-time-string` take 3 arguments.  (The third argument in the previous version is UNIVERSAL, which can be nil or non-nil.)  An `emacs-version` check might be the best we can do here.

tarsius:
> An `emacs-version` check might be the best we can do here.

@ynilu Could you please do that?

tarsius:
`Correct time & time zone in magit-blame--format-time-string`

We would write this as somethink like:

`magit-blame--format-time-string: Correct timezone handling`

ynilu:
> @ynilu Could you please do that?

Sure, I was trying to handle the old version `format-time-string` issue

But it's not easy to achieve the same behavior in both version

I had done a patch which has different behavior in these two version

tarsius:
Its not necessary that the two version have the same behavior, they just both have to do something reasonable and the differences should be documented. Support for Emacs 24 will be dropped very soon.

If you can get the same behavior and a old and new Emacsen, then that would still be nice of course, just not absolutely necessary.

ynilu:
Since Emacs 24 will be dropped very soon, I just format the time string in local time zone in Emacs version < 25 to deal with the different behavior of `format-time-string`

lisp/magit-blame.el
```diff
@@ -653,12 +653,17 @@ modes is toggled, then this mode also gets toggled automatically.
...
-  (setq time (string-to-number time))
-  (setq tz   (string-to-number tz))
-  (format-time-string
-   (or (magit-blame--style-get 'time-format)
-       magit-blame-time-format)
-   (seconds-to-time (+ time (* (/ tz 100) 60 60) (* (% tz 100) 60)))))
+  (if (version< emacs-version "25")
+      (let ((time-format (or (magit-blame--style-get 'time-format)
+                             magit-blame-time-format))
+            (time-in-second (string-to-number time)))
```
kyleam: I've tested this on Emacs 24.5.1.  At first it was failing with the error fixed by adbf6962 (magit-blame-process-filter: Uglify for the benefit of Emacs 24, 2018-05-15).  After rebasing, it failed with "error in process filter: Invalid time specification".

The issue is that `format-time-string` didn't accept plain seconds back then.

```elisp
(format-time-string "%F %H:%M %z" 1526851297)  => (error "Invalid time specification")
(format-time-string "%F %H:%M %z" (seconds-to-time 1526851297)) => "2018-05-20 17:21 -0400"
```

So the `string-to-number` call should be followed by `seconds-to-time` for compatibility.

Just as a minor comment: I'd prefer the scope of your version check to be more restricted to avoid duplication of the let bindings and `format-time-string` call.  Something like (untested)

```elisp
(let* ((time-format (or (magit-blame--style-get 'time-format)
                        magit-blame-time-format))
       (tz-in-second (and (not (version< emacs-version "25"))
                          (string-match "%z" time-format)
                          (car (last (parse-time-string tz))))))
  (format-time-string time-format
                      (seconds-to-time (string-to-number time))
                      tz-in-second))
```

ynilu:
@kyleam, thank you for the suggestions
The code is now much cleaner.
I have tested the change on Emacs 24.5.1 and 27.0.50

kyleam:
> I have tested the change on Emacs 24.5.1 and 27.0.50

Thanks, LGTM.

Would you mind giving us your full name so that it can be added to AUTHORS.md?

ynilu:
Sure, my name is `Li-Yun Chang`

PR #3448, opened 2018-05-14
Enable magit-log-buffer-file to show log of deleted files

ivarru:
When looking at a commit, selecting a deleted file opens an empty buffer
which does not visit any file.


tarsius:
Thanks! I pushed another fix.

Issue #324, opened 2012-01-08
*magit-key* buffer


dudebout:
The _magit-key_ buffer does not indicate either the key that was pressed to get it, the name of the category or the function that is going to be called. This can be confusing.

Should we:
- rename the buffer _magit-key: branching_ _magit-key: merging_, etc.?
- put the name of the category somewhere in the buffer?

PR #3450, opened 2018-05-15
Use `git remote get-url` to find actual remote url.

jleechpe:
Using the value from the repo's `.git/config` does not expand in cases
where an alias is being used as the remote name.  Using `git remote
get-url` instead expands the url into what is actually being used (and
then supports both the alias and a more complete github url to match
against).

---

In my git config I have various alternate urls for `insteadOf` and `pushInsteadOf` purposes to differentiate between various github account requirements (personal, work, ...).

Since I'm using these features I take advantage and use abbreviations in the remote names and leverage `.gitconfig` to resolve them into the actual urls used in my `.ssh/config`.

```ini
[url "git@github.com:"]
     insteadOf = "gh:"
     PushInsteadOf = "https://github.com/"
     PushInsteadOf = "http://github.com/"
     PushInsteadOf = "git:github.com/"
     PushInsteadOf = "gh:"

[url "git@github.com:jleechpe"]
     insteadOf = "ghj:"
     PushInsteadOf = "https://github.com/jleechpe/"
     PushInsteadOf = "http://github.com/jleechpe/"
     PushInsteadOf = "git:github.com/jleechpe/"
     PushInsteadOf = "ghj:"
```

Because of this magit is unable to identify any of my `gh:` or `ghj:` remotes as being valid github repositories.

Changing `magit--github-remote-p` to use `git remote get-url --push <remote>` allows it to get the actual url being used and not just the key in the repo config.

### DIFF
--- a/lisp/magit-collab.el
+++ b/lisp/magit-collab.el
@@ -139,9 +139,9 @@ exist, then raise an error."
                 host)))))
 
 (defun magit--github-remote-p (remote)
-  (or (--when-let (magit-git-str "remote" "get-url" "--push" remote)
+  (or (--when-let (magit-git-string "remote" "get-url" "--push" remote)
         (magit--github-url-p it))
-      (--when-let (magit-git-str "remote" "get-url" "--all" remote)
+      (--when-let (magit-git-string "remote" "get-url" "--all" remote)
         (magit--github-url-p it))))
 
 (defun magit--github-url-equal (r1 r2)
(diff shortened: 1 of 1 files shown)


vermiculus:
What a coincidence -- related to #324.

tarsius:
Any reason why you are using `magit-git-str` instead of `magit-git-string`. The former is only intended for cases where debugging might get us into an infinite loop.

jleechpe:
Because I hadn't been aware of the difference.  I'll update to use `magit-git-string`

tarsius:
That's what I expected, just wanted to make sure.

I have squashed and rebased the commits, and changed the commit summary a bit.

Thanks!

jleechpe:
Thanks

Issue #3015, opened 2017-03-02
Verifying tag signatures


nidrissi:
Magit provides some support for verifying commit signatures (they are shown either in green or red in the log if the option `-S` is provided). However no such feature exists for tags. There is a tool [`git-verify-tag`](https://git-scm.com/docs/git-verify-tag) whose output looks like this:

```
$ git verify-tag v0
gpg: Signature made 03/02/17 16:38:10 Paris, Madrid
gpg:                using RSA key 8DD4D303141F56B7
gpg: Can't check signature: No public key
```

Alternatively, the `-v` option of `git-tag` provides an output similar to this one. Could it be possible to integrate that into Magit?

tarsius:
Yes, that would be possible ;-)

In what places would you expect the tags to be verified?

nidrissi:
I think the logs (show the tag name in red/green like the commits perhaps) and the `show-refs` screen (the `y` key) would be the best places.

fice-t:
There's a couple issues I see:

- There's already a face for tags to use, so how should they be coloured differently? Should there be new `magit-tag-signature-*` faces to match all of the `magit-signature-*` faces? If so, then should the log buffer disambiguate tags from commit messages? e.g., `git log` puts all tags inside `(tag: ...)`.

- There doesn't appear to be a tag analogue for the `%G?` format, which helps with determining what face to use. So between `git tag` and `git verify-tag`:

  - I don't think `git-tag` will allow for an easy equivalent of "%G?". Also, maybe it's just me, but I can't seem to get the `--format` argument of `git tag -v` to work. Can anyone else get it to work?
  - `git-verify-commit` doesn't have a format argument, so if that's to be used, then we'd have to parse its `--raw` output ourselves. I guess this is what Git does itself, so while a bit annoying it's probably the best approach.

tarsius:
> Also, maybe it's just me, but I can't seem to get the `--format` argument of `git tag -v` to work.

It expects a format containing "field names" like `%(refname)` as described in `git-for-each-ref(1)`, not "placeholders" like `%H` as described in `git-log(1)`. There does not appear to be a "field name" equivalent to the `%G?` "placeholder".

We should parse the output of `git verify-tag`. Unfortunately I was unable to find a reliable list of all the things this command might tell us. Any help locating or creating such a list would be appreciated. This is what I have got so far.

```
|   | git-log(1) description                | found in gnupg's po/en@quod.po |
|---+---------------------------------------+--------------------------------|
[...]

PR #3451, opened 2018-05-16
Add PGP verification support

thblt:
This PR adds two functions, `magit-verify-commit` and `magit-verify-tag`, for verifying the PGP signature of commits and tags.  Both take an object ID (a commit ID or a tag name, respectively) and return either nil (if object is unsigned, or signature is invalid) or a list with the following elements:

 1. The full key fingerprint; (string)
 2. The key's primary UID; (string)
 3. The signature validity; (bool, non-nil means valid) [see note below]
 4. The key's owner trust; (a symbol, or nil if *never*) [see note below]
 5. The key's expiration status (non nil means expired)
 5. The signature's expiration status (non nil means expired)
  
Note: fields 3 and 4 are never nil by default, because invalid signatures don't return a list, but nil --- this allows to use the return value of these functions as a boolean.  Yet the functions take an optional `include-invalid` argument which overrides this behavior.

The implementation tries to be reasonably paranoid, and should fail rather than returning false positives.   

This PR still lacks:
 
 - [ ] An UI exposing the new functionality 
 - [ ] Tests
 - [ ] Documentation
 
I'm willing to add all that, but I wanted to get an early review of
the base implementation.  

Thanks!

### DIFF
--- a/lisp/magit-verify.el
+++ b/lisp/magit-verify.el
@@ -1,0 +1,172 @@
+;;; magit-verify.el --- cryptographic signature support for Magit  -*- lexical-binding: t -*-
+
+;; Copyright (C) 2012-2018  The Magit Project Contributors
+;;
+;; You should have received a copy of the AUTHORS.md file which
+;; lists all contributors.  If not, see http://magit.vc/authors.
+
+;; Author: Jonas Bernoulli <jonas@bernoul.li>
+;; Maintainer: Jonas Bernoulli <jonas@bernoul.li>
+
+;; Magit is free software; you can redistribute it and/or modify it
+;; under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 3, or (at your option)
+;; any later version.
+;;
+;; Magit is distributed in the hope that it will be useful, but WITHOUT
+;; ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+;; or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+;; License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with Magit.  If not, see http://www.gnu.org/licenses.
+
+;;; Commentary:
+
+;; Verifies PGP signatures of commit and tag objects.
+
+;;; Code:
+
+(require 'magit)
+
+(defun magit-verify-commit (id &optional include-invalid)
+  "Verify cryptographic signature of commit ID.
+
+Return value is nil if ID is not signed, or a list of the form
+\(KEY-FP KEY-UID VALID OWNERTRUST KEY-EXPIRED SIG-EXPIRED), where:
+
+ - KEY-FP is a PGP key fingerprint.
+ - KEY-UID is the PGP key's primary UID.
+ - VALID is t if the signature is valid.  Unless INCLUDE-INVALID
+   is non-nil, this is always the case.
+ - OWNERTRUST is either a symbol between 'ultimate, 'full,
+   'unknown, 'undefined, 'marginal or nil (never)
+ - KEY-EXPIRED is non-nil if the key or a subkey has expired.
+ - SIG-EXPIRED is non-nil if the signature has expired.
+
+By default, this function returns nil for missing or invalid
+signature.  If INCLUDE-INVALID is non-nil, however, a list is
+always returned even if the signature is invalid.
+
+This function is implemented following GnuPG documentation at:
+https://www.gnupg.org/documentation/manuals/gnupg/Automated-signature-checking.html"
+  (magit-verify--parse-output (magit-git-lines "verify-commit" "--raw" id) include-invalid))
+
+(defun magit-verify-tag (name &optional include-invalid)
+  "Verify cryptographic signature of tag NAME.
+
+The return value has the same format as `magit-verify-tag', which
+see."
+  (magit-verify--parse-output (magit-git-lines "verify-tag" "--raw" name) include-invalid))
+
+(defun magit-verify--ownertrust-symbol (level)
+    "Internal, DO NOT USE.
+
+Convert owner trust LEVEL, a string, to a symbol.
+
+Accepted inputs are ULTIMATE ('ultimate), FULLY ('full),
+UNKNOWN ('unknown), UNDEFINED ('undefined), MARGINAL ('marginal)
+and NEVER (nil).  Any other value will raise an error.
+
+Input is case-insensitive."
+    (setq dlevel (downcase level))
+
+    (cond
+     ((equal dlevel "ultimate") 'ultimate)
+     ((equal dlevel "full") 'full)
+     ((equal dlevel "fully") 'full)
+     ((equal dlevel "undefined") 'undefined)
+     ((equal dlevel "marginal") 'marginal)
+     ((equal dlevel "never") nil)
+     (t (error "Unknown owner trust %s" level))))
+
+(defun magit-verify--parse-output (lines &optional include-invalid)
+  "Internal, DO NOT USE.
+
+The common part of `magit-verify-commit' and
+`magit-verify-tag'."
+
+  "Parse LINES as output of git verify-[tag,commit] --raw ...
+Returns a possibly empty list of (KEYID OWNERID)."
+  (when lines
+    (let ((keydata (-non-nil
+                    (mapcar
+                     (lambda (str)
+                       (when (string-match
+                              (rx
+                               line-start
+                               "[GNUPG:] "
+                               (group (or "GOODSIG" "BADSIG" "EXPKEYSIG" "EXPSIG"))
+                               " "
+                               (one-or-more hex-digit) ;; This a short ID, we'd rather not use it.
+                               " "
+                               (group (one-or-more any))
+                               line-end)
+                              str)
+                         (list (match-string 1 str) ;; GOOD/BAD/EXP/EXPSIG
+                               (match-string 2 str) ;; UID
+                               )))
+                     lines)))
+          (fingerprint (-non-nil
+                       (mapcar
+                        (lambda (str)
+                          (when (string-match
+                                 (rx
+                                  line-start
+                                  "[GNUPG:] VALIDSIG "
+                                  (group (one-or-more hex-digit))
+                                  " "
+                                  (one-or-more any)
+                                  line-end)
+                                 str)
+                            (match-string 1 str)))
+                        lines)))
+          (ownertrust (-non-nil
+                       (mapcar
+                        (lambda (str)
+                          (when (string-match
+                                 (rx
+                                  line-start
+                                  "[GNUPG:] TRUST_"
+                                  (group (one-or-more alpha))
+                                  " "
+                                  (one-or-more any)
+                                  line-end)
+                                 str)
+                            (match-string 1 str)))
+                        lines))))
+
+          ;; We should have processed exactly one
+          ;; GOODSIG/BADSIG/EXPSIG/EXPKEYSIG line and exactly one TRUST_
+          ;; line.  If this is not the case, panic.
+
+      (unless (= 1 (length keydata))
+        (error "Abnormal state 1 in magit-verify--parse-output"))
+
+      (unless (= 1 (length fingerprint))
+        (error "Abnormal state 2 in magit-verify--parse-output"))
+
+      (unless (= 1 (length ownertrust))
+        (error "Abnormal state 3 in magit-verify--parse-output"))
+
+      (let* ((keydata (car keydata))
+             (fingerprint (car fingerprint))
+             (ownertrust (car ownertrust))
+
+             (sig-validity (nth 0 keydata))
+             (key-uid (nth 1 keydata))
+             (valid (and
+                     (not (equal "BADSIG" sig-validity))
+                     (not (equal "NEVER" ownertrust)))))
+
+        (when (or valid include-invalid)
+          (list
+           fingerprint
+           key-uid
+           valid
+           (magit-verify--ownertrust-symbol ownertrust)
+           (equal "EXPKEYSIG" sig-validity)
+           (equal "EXPSIG" sig-validity)))))))
+
+(provide 'magit-verify)
+;;; magit-verify.el ends here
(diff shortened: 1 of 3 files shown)


thblt:
(Also I'm this probably doesn't deserve a separate elisp file.  `magit-log.el` maybe?)

thblt:
> Can `(unless (= 1 (length keydata))` et al. actually happen?

 - **(null keydata)**: from my tests, `(and (null keydata) (null fingerprint) (null ownertrust))` *may* happen if the signature data is corrupted (not just invalid).  This should be excessively rare, but it still may happen, so I'd rather abort in this case.

 - **(< 1 (length keydata))**: from my understanding of the semantics of GnuPG's `--status-fd` output (git's `--raw`), the case where these lengths is > 1 is impossible, since git doesn't allow to sign the same object with multiple keys (as of git 2.16.2).  But the documentation on this format is really too scarce to be absolutely certain it will never happen, so I'd rather keep some sanity check at this point.  
 
The idea behind all these tests is to make as few unchecked assumptions as possible about the output.  The stupid `magit-verify--ownertrust-symbol` function is there for the same reason: it could theoretically be replaced with `(if (string= NEVER "ownertrust") nil (intern (downcase ownerlevel)))`, but the current implementation makes sure it only works with values it actually expects.

tarsius:
This seems way to verbose. I have started making it shorter using two fixup commits.

Please stop using `rx` and just use a regexp, that should make it shorter still. (Also one does not have to consult the `rx` documentation to be sure one knows what the equivalent regexp would be.) In the same step you might also want to replace `-not-nil` and `mapcar` with `cl-mapcan` or possibly `-mapcat`.

tarsius:
Related:  #3015.

thblt:
> What is the use-case for claiming that the signature is valid when it is not? 

This is a case of very poor wording.  Is this more clear?:

```
VALID is t if and only if the signature is valid.  
Notice that unless INCLUDE-INVALID is non-nil, 
you don't need to verify this value: the function 
would have returned nil if the signature was invalid. 
```

thblt:
> one does not have to consult the rx documentation to be sure one knows what the equivalent regexp would be.

FWIW `rx` is just a macro, so C-x C-e is enough to see the generated regexp.  

> Please stop using rx and just use a regexp, ... In the same step you might also want to replace -not-nil and mapcar with cl-mapcan or possibly -mapcat.

Doing this, I'll push in a few minutes

tarsius:
> FWIW `rx` is just a macro, so C-x C-e is enough to see the generated regexp.

I considered editing my message to make it clear that I know that. The point is, I find simple regexps easy to read, while `rx` is very verbose and difficult to read. `rx` may make sense for complex stuff though.

thblt:
I haven't found a function equivalent to `(-non-nil (mapcar FN SEQ))` that would work with function returning strings.  Both cl-mapcan and -mapcat require than FN returns a list.

lisp/magit-verify.el
```diff
@@ -133,17 +116,17 @@ Returns a possibly empty list of (KEYID OWNERID)."
...
-                      (pcase (downcase level)
-                        ("ultimate"  'ultimate)
-                        ("fully"     'full)
-                        ("undefined" 'undefined)
-                        ("marginal"  'marginal)
-                        ("never"     nil)
-                        (_ (error "Unknown owner trust %s" level)))
-                      (equal "EXPKEYSIG" sig-validity)
-                      (equal "EXPSIG" sig-validity)))))))
+           (when (or valid include-invalid)
```
tarsius: This is about returning a value, so we should not use `when` here. That should only be used when the body is about side-effects.

thblt: Reverting

thblt:
Unrelated: I'm wondering if we should return the key's UID and the owner trust.  Technically, these are properties of the key, not of the signature, and we're checking a signature.  On the other hand though, they'll probably be needed when displaying information about the signature, and grabbing them here would prevent having to run a second subprocess just to grab key information.  Do you have an opinion on this?

tarsius:
I have pushed an alternative, shorter, but completely untested version to https://github.com/magit/magit/compare/pgp-verification-alt?expand=1.

tarsius:
> Do you have an opinion on this?

We probably should include all the relevant data.

tarsius:
I think we should always return the object. A caller who only wants the big picture can then use `(oref (magit-verify-commit rev) valid)`.

thblt:
> I think we should always return the object.

Why not, but only if we return an object even for unsigned commits.  The reason I wanted to return nil for everything but a valid signature is that invalid signatures are extremely rare.  It means that people will automatically write things like below,  test their code on unsigned commits and see that it works:

```elisp
(if (magit-verify-tag ID) 
 (do-this-if-tag-is-signed)
 (do-this-if-not))
```

This will accept invalid signatures silently, which is *not* the intended result.  I really prefer your idea of using  a class over a list, but this class would need a first field `is-signed` (or something similar) and we could add a convenience method like `magit-get-valid-signer` which would return the key fingerprint of a valid, trusted signature object or nil.

> but completely untested 

I'll start working on a test repository (with bad signatures and everything) and a test suite tomorrow or in the week-end.

tarsius:
> Why not, but only if we return an object even for unsigned commits. [...]

Lets go with that approach at least for the time being.

> It means that people will automatically write things like below,

Well, I doubt that too many people are going to use it at all, let alone people who didn't bother to read the docs, but better safe than sorry I guess.

> this class would need a first field `is-signed`

Yeah, I was just modifying your code hoping I was not making too many translation mistakes. Since it sounds like you have already embraced these changes (in general), just go ahead and add whatever slot is necessary and remove/rearrange... as you see fit.

thblt:
I promised something for the weekend, it isn't actually complete yet, but here's what I have now.  It seems to work reasonably well, the test suite is not yet complete but will be soon and it still lacks a few convenience functions so that people *don't* play with the lower-level features.

I've built upon the EIEIO class you drafted, but completely rewrote the parsing logic, converting it back from the buffer-based approach to a string-list-based parser.   This is because I've found (at last, and by sheer luck) the (well-hidden) [specification for the output format](http://git.gnupg.org/cgi-bin/gitweb.cgi?p=gnupg.git;a=blob;f=doc/DETAILS;h=52051ed2c4eea5547b51b211ef5fb9d88c35bea9;hb=refs/heads/master), so I've been able to implement a (reasonably) complete parser.  (With this limitation that no parser can really be complete, since the documentation states that the format is indefinitely extensible, and unknown keywords should be dropped).

Another important change is that the new parser supports signatures with multiple signers.  I've pondered this back and forth, since the "official" git implementation doesn't support more than one signer per object for signing (the `-S,--gpg-sign` field can be repeated, but only the last value is used), although it has no such limitations when verifying.  But other implementations do not have such a limitation:  libgit2 lets the user create signature directly from GnuPG and [only adds them as a field in existing buffers](https://libgit2.github.com/libgit2/#HEAD/group/commit/git_commit_create_with_signature).   Libgit2 users may then very well create signatures with multiple signers. It should be rare, but it follows that git objects with multiple signers are not impossible, and should be supported. 

This is not yet mergeable, but should be readable, less gratuitously verbose, reasonably well documented and safe.  

I'll try to push something more complete and polished in the next couple days.

thblt:
I just noticed I've changed indentation in existing magit-tests.el code, I'll revert this of course.

tarsius:
It might be worth looking at related code in gnus and mu4e.

thblt:
> It might be worth looking at related code in gnus and mu4e.

mu4e performs the verification from C code, the elisp client just calls `mu verify`.  Gnus uses Emacs' EPG, which would be the obvious choice for Magit too, but is harder to use than it should:

 - `epg--process-filter` is internal and isn't meant to be used from outside epg.
 - the rest of EPG starts GnuPG processes in a way that doesn't permit to use git instead of the "real" gnupg, because it passes a number of arguments that git rejects (`--status-fd 1` and the like).  These arguments are hardcoded in EPG source code.

A solution that would allow to use EPG would be to manually extract the signature and the signed data from the commit/tag object and call epg for the verification.  It would make code shorter, but will probably [break when `newhash` will be introduced](https://github.com/git/git/blob/26e47e261e969491ad4e3b6c298450c061749c9e/Documentation/technical/hash-function-transition.txt#L389-L404) (although the fix should then be easy).  Would you prefer that? It feels a bit hacky, but could be considered cleaner than reimplementing a status-fd parser that already exists in Emacs.

tarsius:
> Would you prefer that?

Doesn't sound appealing. I just wanted to make sure you had looked at some other packages :wink: 

I am rather busy right now and it will likely take a week until I get back to you again.

thblt:
This should be an almost final version of the elisp module.  If it looks good to you, I'll finalize the test suite and update the manual.  The mot important change is in the tag verification logic.  The new version works in two steps checking that the NAME argument it receives matches the value of the `tag` field in the object *before* processing the signature.

tarsius:
The symbols and file name should contain "pgp" because completely unrelated things can also be "verified" using git/magit.

thblt:
First, apologies for the two stupid mistakes you had to fix.  I  *really*  took the time to re-read my changes this time, and to run the code and tests from a clean Emacs instance, so I hope there won't be missing/misplaced parentheses anymore...

"A few" more questions :)

 - The two last test cases assume the tests are run from inside the Magit repository.  Is this a reasonable assertion?  
 - They also assume both our keys are in the keyring.  Is this OK?  If not, could you sign a commit so only your key is required?
 - Should I update the manual, and if yes, how?  I drafted a "PGP signature verification" section in "Miscellaneous" presenting the public functions of `magit-pgp.el`, would that do?
 - Should I leave your name as the author and maintainer in the header of the `magit-pgp` module and/or should I add my name to AUTHORS.md?

I've also updated the TODO list in the original message.

Thanks a lot for your patience with this PR...

-- 
Thibault

tarsius:
1. Seems okay but...
2. It would be best to create a new keyring during testing and populate that with ad hoc keys.
3. That should do.
4. You are the author and the library should say so. You will get added to AUTHORS.md eventually when I run `make authors`.

> An UI exposing the new functionality

Looking forward to that.

tarsius:
Yes and yes.

See https://github.com/magit/magit/pull/3451/commits/717432c3dba0bd0221a227573962ac2b55ebabb5#diff-dc8615422025d4d47c20f4637d1571b7R80 for inspiration.

thblt:
Working on the UI, and I have a small issue.  I added `(?H "PGP verification" magit-pgp-popup)` to `(magit-define-popup magit-dispatch-popup ... ` in magit.el, and the binding works only when the help screen is displayed.  Ie, from `M-x magit-status RET`, `H` displays "H is undefined", but `? H` shows the verification popup.  Am I missing something stupid?  

Complete modified code is

``` elisp
(magit-define-popup magit-dispatch-popup
  "Popup console for dispatching other popups."
  :actions '("Popup and dwim commands"
             (?A "Cherry-picking"   magit-cherry-pick-popup)
             (?b "Branching"        magit-branch-popup)
             (?B "Bisecting"        magit-bisect-popup)
             (?c "Committing"       magit-commit-popup)
             (?d "Diffing"          magit-diff-popup)
             (?D "Change diffs"     magit-diff-refresh-popup)
             (?e "Ediff dwimming"   magit-ediff-dwim)
             (?E "Ediffing"         magit-ediff-popup)
             (?f "Fetching"         magit-fetch-popup)
             (?F "Pulling"          magit-pull-popup)
             (?l "Logging"          magit-log-popup)
             (?L "Change logs"      magit-log-refresh-popup)
             (?m "Merging"          magit-merge-popup)
             (?M "Remoting"         magit-remote-popup)
             (?o "Submodules"       magit-submodule-popup)
             (?O "Subtrees"         magit-subtree-popup)
             (?P "Pushing"          magit-push-popup)
             (?r "Rebasing"         magit-rebase-popup)
             (?t "Tagging"          magit-tag-popup)
             (?T "Notes"            magit-notes-popup)
             (?V "Reverting"        magit-revert-popup)
             (?w "Apply patches"    magit-am-popup)
             (?W "Format patches"   magit-patch-popup)
             (?X "Resetting"        magit-reset-popup)
             (?y "Show Refs"        magit-show-refs-popup)
             (?z "Stashing"         magit-stash-popup)
             (?! "Running"          magit-run-popup)
             (?% "Worktree"         magit-worktree-popup)
             (?H "PGP verification" magit-pgp-popup)
             (lambda ()
               (and (with-current-buffer magit-pre-popup-buffer
                      (derived-mode-p 'magit-mode))
                    (propertize "Applying changes" 'face 'magit-popup-heading)))
             (?a "Apply"           magit-apply)
             (?s "Stage"           magit-stage)
             (?u "Unstage"         magit-unstage)
             (?v "Reverse"         magit-reverse)
             (?S "Stage all"       magit-stage-modified)
             (?U "Unstage all"     magit-unstage-all)
             (?k "Discard"         magit-discard)
             (lambda ()
               (and (with-current-buffer magit-pre-popup-buffer
                      (derived-mode-p 'magit-mode))
                    (propertize "Essential commands" 'face 'magit-popup-heading)))
             (?g  "    refresh current buffer"   magit-refresh)
             ;; These bindings only work because of :setup-function.
             (?\t   "  toggle section at point"  magit-section-toggle)
             (?\r   "  visit thing at point"     magit-visit-thing)
             ;; This binding has no effect and only appears to do
             ;; so because it is identical to the global binding.
             ("C-h m" "show all key bindings"    describe-mode))
  :setup-function 'magit-dispatch-popup-setup
  :max-action-columns (lambda (heading)
                        (pcase heading
                          ("Popup and dwim commands" 4)
                          ("Applying changes" 3)
                          ("Essential commands" 1))))
```

tarsius:
You have to add a binding to 'magit-mode-map` too.

alphapapa:
> You have to add a binding to 'magit-mode-map` too.

Could that be done automatically by `magit-define-popup`?

tarsius:
We could but I don't think we should. Lets not complicate things for the sake of this special case. If we did, then looking at the definition of `magit-mode-map` would no longer be enough to learn about all its bindings.

vermiculus:
Please also check the dispatch key registry; `H` is already in use.

https://github.com/magit/magit/wiki/Plugin-Dispatch-Key-Registry

tarsius:
@thblt This has been open for too long for my taste, so I am closing it. I would still be interested in merging this once there is an UI and would be happy to open this again once you start working on it again.

PR #3455, opened 2018-05-23
Avoid error if gitk not found on windows-nt

bennorth:
Under windows-nt, I was getting error messages on trying to use commands defined in `magit-extras.el`.  Digging in, it was the `defcustom` of `magit-gitk-executable` causing the problem.  It might be something unusual in my setup (which I'm not in front of right now), but the `which` was not finding `gitk.exe`, and a `nil` was getting through to `file-executable-p`, about which it complained.

This PR puts in a check for `nil`, which avoids the error and so fixes things for me.

### DIFF
--- a/lisp/magit-extras.el
+++ b/lisp/magit-extras.el
@@ -39,7 +39,7 @@
            (let ((exe (magit-git-string
                        "-c" "alias.X=!x() { which \"$1\" | cygpath -mf -; }; x"
                        "X" "gitk.exe")))
-             (and (file-executable-p exe) exe)))
+             (and exe (file-executable-p exe) exe)))
       (executable-find "gitk") "gitk")
   "The Gitk executable."
   :group 'magit-extras


lisp/magit-extras.el
```diff
@@ -39,7 +39,7 @@
            (let ((exe (magit-git-string
                        "-c" "alias.X=!x() { which \"$1\" | cygpath -mf -; }; x"
                        "X" "gitk.exe")))
-             (and (file-executable-p exe) exe)))
+             (and exe (file-executable-p exe) exe)))
```
npostavs: Thanks, I think this is the right fix, but having `exe` twice in the same `and` looks weird.
So maybe use an `if` and push the second `exe` into the `THEN` clause?

tarsius:
>> What @npostavs suggests there is that the outer or be replaced with an if.

> Actually, I didn't mean for the outer `or` to change, just to add an `if` around the `and`.

Ah well, that might be even weirder, I think. `and-let` might be nice, but since that isn't available in all the Emacsen we still support we would have to use `when-let`, which is also weird, because this isn't about side-effects.

I have just merged this as is.

PR #3456, opened 2018-05-23
magit-gitk-executable: fix runtime error when gitk is not installed

kostya-sh:


tarsius:
Hehe. Either a big coincidence or something changed in a new Git release: #3455.

tarsius:
Actually, you might beat @bennorth to it.

What @npostavs suggests there is that the outer `or` be replaced with an `if`.

tarsius:
Hm, @bennorth looked at the issue in more detail, so I prefer their commit message.

npostavs:
> What @npostavs suggests there is that the outer or be replaced with an if.

Actually, I didn't mean for the outer `or` to change, just to add an `if` around the `and`.

kostya-sh:
Glad that this is now fixed! https://github.com/magit/magit/pull/3455 is a great coincedence indeed ;)

PR #3466, opened 2018-06-01
magit-get-previous-branch: Add timeout

### LINKED ISSUE #3457: magit-get-previous-branch can take a long time, depending on repo state
I was recently doing a long (i)merge followed by some rebasing, all of which means that finding the previous branch visited by `HEAD` takes a long time to find when going back one ref at a time. So when I did <kbd>l o</kbd> ("Log other"), which calls `magit-get-previous-branch` via `magit-log-read-revs`, my CPU started spinning, seemingly indefinitely (it was a bit confusing at first, because neither Emacs nor git would show up in `top` as using much CPU, each `git` invocation was over too quickly I guess).

So either `magit-get-previous-branch` needs to have a (configurable?) limit to how far back it searches, or it should be implemented in a way which takes a constant amount of time (if that's possible?).

npostavs:
Fixes #3457.

I somewhat arbitrarily chose 0.5 seconds as the default timeout.

By the way, I'm not sure exactly how I got my repo into a state where `magit-get-previous-branch` can take a long time (I still have it in that state currently though). I tried a test script which creates and deletes many branches, but that doesn't seem to do the trick.

### DIFF
--- a/Documentation/RelNotes/2.13.0.txt
+++ b/Documentation/RelNotes/2.13.0.txt
@@ -44,6 +44,9 @@ Changes since v2.12.0
 
 * Added new command `magit-worktree-checkout-pull-request'.  a3d788ec
 
+* New variable `magit-get-previous-branch-timeout' limits the amount
+  of time spent in `magit-get-previous-branch'.  #3457
+
 Fixes since v2.12.0
 -------------------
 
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks!

PR #3467, opened 2018-06-01
Fix unquoted apostrophe in code sample

christophejunke:
The change fixes the documentation string of ido-enter-magit-status, where 
an apostrophe in a code sample is written as-is, which renders it as 
RIGHT SINGLE QUOTATION MARK in the Help buffer (making the code
snippet invalid).

No need to update the manual.


tarsius:
Thanks!

Cherry-picked.

Issue #3337, opened 2018-01-19
"pcase-memoize: equal first branch, yet different" for custom sections


vermiculus:
This only started happening after #3309, I believe.  I can't be completely certain, though.

https://github.com/magit/magit/blob/a25d36b794499f05a6c7c50b48e6e13f2a2de7d7/lisp/magit-section.el#L718-L726

This line is causing runtime and compilation messages for pull-request- and issue-list sections. I edebug'd pcase and came up with this both times the message is shown:
```
value of `data':
(type2858 (((\` file) . #1=((quote magit-file-section)))
           ((\` hunk) . #2=((quote magit-hunk-section)))
           (_ . #3=((quote magit-section))))
          cond
          ((eq type2858 (quote file)) (let nil . #1#))
          ((eq type2858 (quote hunk)) (let nil . #2#))
          (t (let nil . #3#)))
value of `cases':
(((\` file) (quote magit-file-section))
 ((\` hunk) (quote magit-hunk-section))
 (_ (quote magit-section)))
```

Backtrace:
```
  (closure (pcase--memoize-1 pcase--memoize-2 t) (exp &rest cases) "[docstring removed]" (edebug-enter (quote pcase) (list exp cases) (function (lambda nil (edebug-after (edebug-before 0) 49 (let ((data ...)) (edebug-after (edebug-before 7) 48 (if ... ... ... ...))))))))(type3934 ((\` file) (quote magit-file-section)) ((\` hunk) (quote magit-hunk-section)) (_ (quote magit-section)))
  (pcase type3934 ((\` file) (quote magit-file-section)) ((\` hunk) (quote magit-hunk-section)) (_ (quote magit-section)))
  (funcall (pcase type3934 ((\` file) (quote magit-file-section)) ((\` hunk) (quote magit-hunk-section)) (_ (quote magit-section))) "" :type type3934 :value nil :start (point-marker) :parent magit-insert-section--parent)
[...]

PR #3469, opened 2018-06-04
magit-dired-jump: Don't append slash to directory

basil-conto:
1. `make emacs-Q`
2. <kbd>C-x</kbd><kbd>g</kbd>
3. <kbd>C-M-i</kbd>
   * Expected result: Pop to Dired buffer visiting Magit's directory.
   * Actual result: Pop to Dired buffer visiting parent of Magit's directory.

This is caused by `magit-dired-jump` appending `"/."` to directory names, which `dired-jump` expands to a file, rather than directory, name. A subsequent `file-name-directory` yields the parent, rather than current, directory.

Given that `dired-jump` defaults to `default-directory`, I think directory name manipulation can be removed from `magit-dired-jump` altogether, as in the proposed patch.

### DIFF
--- a/lisp/magit-extras.el
+++ b/lisp/magit-extras.el
@@ -126,12 +126,11 @@ like pretty much every other keymap:
 With a prefix argument, visit in another window.  If there
 is no file at point, then instead visit `default-directory'."
   (interactive "P")
-  (dired-jump other-window (if-let (file (magit-file-at-point))
-                               (progn (setq file (expand-file-name file))
-                                      (if (file-directory-p file)
-                                          (concat file "/.")
-                                        file))
-                             (concat default-directory "/."))))
+  (dired-jump other-window
+              (when-let (file (magit-file-at-point))
+                (expand-file-name (if (file-directory-p file)
+                                      (file-name-as-directory file)
+                                    file)))))
 
 ;;;###autoload
 (defun magit-dired-log (&optional follow)


basil-conto:
> Maybe things have changed between Emacs versions?

Looks like that's the case, because with

    Magit 2.13.0-28-gf67c8383-dirty, Git 2.17.1, Emacs 27.0.50, gnu/linux

I can reproduce on the following two versions:

    In GNU Emacs 27.0.50 (build 10, x86_64-pc-linux-gnu, X toolkit, Xaw3d scroll bars)
     of 2018-06-04 built on thunk
    Repository revision: 1dafa4a02ed45bb4d02c6dc34c55518858422088
    Windowing system distributor 'The X.Org Foundation', version 11.0.11906000
    System Description: Debian GNU/Linux buster/sid


    In GNU Emacs 26.1.50 (build 1, x86_64-pc-linux-gnu, X toolkit, Xaw3d scroll bars)
     of 2018-06-03 built on thunk
    Repository revision: 16e8541b743a6beb1926fef53cf1bfaed9c346db
    Windowing system distributor 'The X.Org Foundation', version 11.0.11906000
    System Description:	Debian GNU/Linux testing (buster)

but not on

    In GNU Emacs 25.2.1 (x86_64-pc-linux-gnu, X toolkit, Xaw3d scroll bars)
     of 2018-05-07, modified by Debian built on binet
    Windowing system distributor 'The X.Org Foundation', version 11.0.11906000
    System Description:	Debian GNU/Linux testing (buster)

Sorry, I thought I had checked all available versions, but must have gotten confused in what I was looking for. In any case, the proposed patch (which I've changed to use `--when-let` as per the older version you mention) works for all aforementioned versions, and simplifies the code.

---

By the way, `make EMACSBIN=emacs25 emacs-Q` (with the aforementioned version of Emacs 25) gives the following in `*Messages*`:

    pcase-memoize: equal first branch, yet different [3 times]
    Turning on magit-auto-revert-mode...done
    pcase-memoize: equal first branch, yet different [90 times]

basil-conto:
> re: the pcase-memoize messages, see #3337

Thanks @kyleam. I had seen that at the time, but managed to completely forget about it.

basil-conto:
> I've changed to use `--when-let` as per the older version you mention

And now I've changed it to use `when-let`, in line with e6a03524e2ef523bbc400e9c39fbc34dfb62be27.

tarsius:
I am not sure, but the thing you are fixing might actually be the intended behavior. I will try to look into this in more depth soon, but not immediately.

basil-conto:
> I am not sure, but the thing you are fixing might actually be the intended behavior.

If it is, then I understand neither its rationale nor its documentation.

> I will try to look into this in more depth soon, but not immediately.

Thanks, let me know if there is anything I can help with. In the meantime, I can live with advice (wouldn't be the first time ;).

kyleam:
>> I am not sure, but the thing you are fixing might actually be the intended behavior.
>
> If it is, then I understand neither its rationale nor its documentation.

I don't believe it's intentional.  It seems to be the result of 3988ceaa1c (dired-jump: Expand file-name before dired-goto-file call, 2016-09-14) ([bug24409]).  The issue there is that dired-goto-file expects an absolute file name, but read-file-name may not return one, so a call to expand-file-name was added.

So here's in the issue: In magit-dired-jump, we convert a default directory of "/path/to/src/magit/" to "/path/to/src/magit/.", but the expand-file-name call added in 3988ceaa1c just converts that back to "/path/to/src/magit" (i.e. no trailing "/.").

[bug24409]: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=24409

tarsius:
@kyleam I'll trust your judgement on this. Please merge if/when you think this is ready.

basil-conto:
@kyleam Thank you very much for looking into this. Is the latest wording of the commit message up to scratch? (Feel free to force-push to my branch as you see fit.)

kyleam:
> Is the latest wording of the commit message up to scratch?

Thanks.  Looks good.  But going back to my previous comment...

> The appended "/." guarded against the file-name-as-directory call chopping off the trailing name, so we must have been passing a directory without a trailing slash at some point. Anyway, at this point, I think we're fine because magit-file-at-point appends a slash for directories.

I later realized the case where magit-file-at-point doesn't return directories with a slash: submodules.  I modified your commit to address that case and adjusted the commit message accordingly.

```diff
1: 27cdfc50 ! 1: f5a11574 magit-dired-jump: Don't append slash to directory
@@ -8,10 +8,15 @@
     magit-dired-jump previously appended to directory name arguments,
     resulting in a jump to the parent of the desired directory.

-    In addition to mostly reverting Magit commit [2: 5665ba94],
-    magit-dired-jump can now be further simplified by passing a
-    FILE-NAME argument of nil to dired-jump when no file is detected at
-    point, as dired-jump will default to default-directory in this case.
+    In place of the "/." approach of Magit commit [2: 5665ba94], call
+    file-name-as-directory when magit-file-at-point returns a directory.
+    This is necessary because magit-file-at-point returns a submodule path
+    without a trailing slash, and we want to jump to the submodule
+    directory not the parent directory.
+
+    In addition, simplify magit-dired-jump by passing a FILE-NAME argument
+    of nil to dired-jump when no file is detected at point, as dired-jump
+    will default to default-directory in this case.

     [1: 3988ceaa1c]: 2016-09-14 00:04:32 +0900
       dired-jump: Expand file-name before dired-goto-file call
@@ -36,7 +41,9 @@
 -                             (concat default-directory "/."))))
 +  (dired-jump other-window
 +              (when-let (file (magit-file-at-point))
-+                (expand-file-name file))))
++                (expand-file-name (if (file-directory-p file)
++                                      (file-name-as-directory file)
++                                    file)))))

  ;;;###autoload
  (defun magit-dired-log (&optional follow)
```

basil-conto:
Thanks!

PR #3471, opened 2018-06-05
Change magit-file-rename to open in the current directory

ar1a:
Normally it would open in the base directory, which for projects like java projects, would require you to type out the full path which is a bitch.


tarsius:
Thanks! That makes sense.

You should use `expand-file-name` instead of `concat` and make sure lines don't get longer than 80 characters.

You might also want to create the commit from scratch using Emacs and Magit instead of Github. The third line of the commit message is also too long, which is something that Github encourages. The author email is some weird fake address, it would be better to use your real email, unless you don't want that. The committer is even weirder - it's basically a Github advertisement. And the commit message doesn't end with a newline character, which is something that cannot happen when using the Git porcelain but almost always when using Github. Github isn't a suitable interface for creating commits.

ar1a:
All done! Made it from magit this time too, so hopefully it won't be all fucked.

Issue #3476, opened 2018-06-07
Fuzzy selection of branches with Helm


m-cat:
(1) what behavior you expected
  (2) what behavior you observed
  (3) and how we can reproduce the issue.

1. When selecting a branch (`b b`) with the Helm interface, I would like to be able to type part of the branch name and have it available as a selection. For example, if I have a branch named `2650`, I would like to be able to type `50` and have the branch appear. **This used to work, but it seems the behavior has changed in a regression.**

2. Typing `50` in the example above results in no Helm candidates.

3. Try to fuzzy-select a branch with Helm installed and enabled.

tarsius:
I don't think anything changed on Magit's side. I don't use Helm and don't know whether it does fuzzy matching out-of-the-box or if it needs another package, which you might no longer have installed.

kyleam:
I use helm, but I hadn't yet upgraded to a version that has this issue.

I've bisected it to helm's

```
commit 1aca3e84501955b98b824c7ddd221e88f9f2563d
Date:   Thu May 24 12:25:49 2018 +0200
    Always use a sync handler when COLLECTION is a function

    in completing-read.

    * helm-mode.el (helm--completing-read-default): Do it.
```

I haven't looked in detail at that change yet.

m-cat:
> I don't think anything changed on Magit's side. I don't use Helm and don't know whether it does fuzzy matching out-of-the-box or if it needs another package, which you might no longer have installed.

That's odd, Magit is the only package where I've encountered this behavior so far. I use Helm in many places and fuzzy finding has always worked.

If it helps, the revert dialogue (`V V`) is able to fuzzy-select the *current* branch, but no other branches. Weird.

kyleam:
> That's odd, Magit is the only package where I've encountered this behavior so far.

The helm commit that I pointed to changes the behavior based on whether the collection is a function.  `magit-builtin-completing-read` wraps the collection in a function to prevent the collection from being sorted alphabetically by Emacs's built-in completion.  See aa5f098a (completion: override default alphabetical sorting, 2017-03-30).

So if you did

```diff
diff --git a/lisp/magit-utils.el b/lisp/magit-utils.el
index dd1869a4..fcd92c95 100644
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -439,7 +439,7 @@ (defun magit-builtin-completing-read
                              (bound-and-true-p ivy-mode))
                          prompt
[...]

PR #3477, opened 2018-06-09
completion: Update for latest helm/ivy changes

kyleam:
* Helm now handlers completion functions differently.  We need to account for that in order to get fuzzy matching.  #3476 

  * Ivy now sorts alphabetically by default.  Whether this commit (or some other fix) needs to be included depends on the outcome of abo-abo/swiper#1611


tarsius:
We should probably just merge this now (into `maint` and that into `master`) and then revert the second commit if an eventual upstream change makes that possible.

PR #3478, opened 2018-06-09
Stage and commit **all unstaged files**

ericprud:
For the last couple years, I assumed that "stage everything" included untracked files. I now assume it's more like `commit --all`.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -3858,7 +3858,7 @@ Also see [[man:git-commit]]
 
 - User Option: magit-commit-ask-to-stage
 
-  Whether to ask to stage everything when committing and nothing is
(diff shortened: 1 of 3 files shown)


tarsius:
I think "all unstaged **files**" is still ambigious - untracked files are also "not staged". "All unstaged **changes**" might be better. What do others think?

ericprud:
That makes sense to me. I've optimistically updated the PR to be "all unstaged changes".

kyleam:
s/everything/all staged changes/ sounds fine to me

kyleam:
> @kyleam, I think you meant "all unstaged changes".

Indeed.  Thanks.

tarsius:
Thanks!

I rebased, squashed, bumped version strings, and improved commit message.

PR #3480, opened 2018-06-12
Fix magit-commit-add-log with commit.verbose=true

Miciah:
When Git was configured with `commit.verbose=true`, `magit-commit-add-log` would append lines after the diff, past the cut line.  It should avoid appending lines after the first comment.

* `lisp/magit-commit.el` (`magit-commit-add-log-insert`): Narrow the region to include everything up to the first comment.


tarsius:
Thanks!

PR #3483, opened 2018-06-14
Use which-function instead of add-log-current-defun

kidd:
`which-function` uses `add-log-current-defun` as one of many helpers so it
should only improve the experience.

I've tried it with lua files and it works good enough. elisp work equally well as before so far.

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -50,6 +50,8 @@
 
 (require 'ansi-color)
 (require 'crm)
+(require 'which-func)
+
 (defvar bookmark-make-record-function)
 
 ;;; Options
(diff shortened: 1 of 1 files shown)


tarsius:
I think `which-func` has to be required explicitly.

kidd:
oh, you're right... Do you think it makes sense to:
1) test for (fboundp which-function) and do one thing or the other?
2)  require which-func first
3)  not even worth it

tarsius:
Nah, just require it.

kidd:
Done. Thanks!

Btw, how do you usually check if a function has to be required manually or it's already required by default? is `emacs -Q` the usual way?

PR #3490, opened 2018-06-18
Add a basic database to store forge (Github/Gitlab...) topics

tarsius:
This is the second step toward implementing #2972. It won't be the last. The first was the addition of `magit-branch-pull-request` and related commands, and teaching `magit-branch-delete` to cleanup pull-requests.

This pull-request brings `magit` to about the level of `magit-gh-pulls` feature-wise. Unlike that it also support Gitlab and there is a lot more going on in the background.

The following features and limitations are currently present.

* A database for storing issue and pull-requests.
  * The table schemata are still very much subject to change. A lot of information that should be stored in it isn't currently being stored yet. We will likely end up making breaking changes after the initial release, but I would still like to limit the number of such changes.
  * Currently only open topics are stored in the database. Updating a project works by deleting all existing information and fetching and storing up-to-date information again. That will change, I want the database to contain all topics, but that is not necessary at this point.
  * The database can be queried using SQL through the dsl provided by `emacsql` or by providing SQL statements as strings. Note however that `emacsql` stores everything as a string, which is then `read`, so a string is stored as `"\"string\""` for example.
* Two interfaces to access the data.
  * Access is also possible through the `closql` ORM that I originally wrote for `epkg`. It's a very limited ORM but it worked well for `epkg`, so I am using it here too. Basically you have to get the current project using `(magit-forget-get-project nil)` or `... t)` if you want to create it if it is missing, and then you can query that using `oref`, e.g. `(oref project pullreqs)`.
  * Closql was not really designed as an ORM it was creates as a replacement for the previous serializing approach used by `epkg`, which stored every object in a separate file. Additionally it gives us some SQL query capabilities for free, but it was designed as "a better method to store objects persistently" and that shows.
* Fetching as Git intended.
  * Like Git has `git fetch`, `magit-forge` also has an explicit fetch step, currently available at `f t`.
  * This feature, which some might see as a limitation, is very important to me. Having an explicit fetch step prevents both delays when updating data at inopportune moments as well as outdated caches.
* `magit-branch-pull-request` and related commands have been updated to use the db.
* The status buffer now lists issues and pull-requests, provided an object for the current project/repository exists in the database.
* To create the project object and fetch related data just use `f t`.
* All the reading of configuration happens in `magit-forge-get-project`. It's a long and yucky function, but keeping all that logic in one place seems like a good idea to me.
* This adds a few new dependencies `emacsql`, `emacsql-sqlite`, `glab` and `closql`.

I haven't implemented much new end-user features for now - basically the simple issue and pull-request lists in the status buffer. The reason for that is that `magithub` already implements a lot of that stuff and I would rather avoid too much duplication.

I am very interested in feedback from @vermiculus, the author of `magithub`. In particular I would like to know whether you are going "rebase" `magithub` on top of the functionality implemented here, at least for the read-only parts. Edit: First step would be to make sure that all the information that `magithub` needs is available in the database.


tarsius:
For now you'll need the `forge` branch of `closql`.

cmccloud:
Am I correct in thinking that only the first 100 open issues and pull requests are pulled down at the moment?

tarsius:
Yes, except I don't know if that's the limit for the Github *Graphql* API and the *Gitlab* API. But there is *some* limit.

cmccloud:
Great. Magithub currently grabs all the issues/prs, and the lack of pagination makes interacting with repos with lots of issues or pull requests very difficult.

vermiculus:
> Currently only open topics are stored in the database. Updating a project works by deleting all existing information and fetching and storing up-to-date information again. That will change, I want the database to contain all topics, but that is not necessary at this point.

Thinking forward, it would be nifty if we could teach each table column how to retrieve that piece of data.  When a query is made against a table with missing information, the database will determine the columns that need to be re-retrieved (either missing or too old) and execute the appropriate queries for that information.  For GitHub, it's not hard to imagine using GraphQL to compress those queries (assuming I make the time to finish `graphql.el`).  I'm not sure what this would look like for other REST methods to avvount for performance.  (Non-API forges, like a mailing list, might not need to be so mindful of performance if the entire thread is already local.)

This may be in conflict with my 'fetch' notes below.

> Note however that `emacsql` stores everything as a string, which is then `read`, so a string is stored as `"\"string\""` for example.

Are we worried about performance here?  You may be familiar that magithub uses `prin1-to-string` right before it writes its cache.  Otherwise, it keeps all information in memory as real, usable data.  I worry that `read`ing all the time will be an unnecessary slowdown, but this may be unwarranted.

> Fetching as Git intended. [...] `f t`

I like this.  Really, I do.  My experience with magithub with trying to make folks happy with how the cache behaves (and how it handles missing information) is something of a nightmare in wrangling all the technical restrictions.

There is one piece of this that's important to consider, though: there must be a way to notify the user when there is (or may be) updated information on the remote.  When I took the same `f t` route -- making all API requests only *semi*-automatic instead of *fully* automatic -- this was by far the strongest reason for reconsideration of that approach.  I still don't think I've got it 'right', but people are using magithub, so I guess that counts for something.

> The status buffer now lists issues and pull-requests, provided an object for the current project/repository exists in the database.

How should magithub react to this change?  I guess in general, if a third-party application wants to 'replace' a hook function in some scenarios, should that be done using the advice system?

> Edit: First step would be to make sure that all the information that `magithub` needs is available in the database.

Realistically, I don't think we'll be able to distill all forges into the same set of features.  It's important that the database be somewhat extensible by third-party applications for forge-specific features.  For example, GitHub has a standard way to approve pull requests, but a mailing list may have several different ways to go about it (hopefully all standardized on a project-by-project basis) that may not be amenable to the same representation.

> In particular I would like to know whether you are going "rebase" `magithub` on top of the functionality implemented here, at least for the read-only parts.

I would plan on doing such a 'rebase' :-) I think it's important that magithub not try to re-invent the wheel -- especially when the user-base of magit so fantastically exceeds that of magithub (i.e., adoption of magithub depends on its integration with magit).

I've thought for some time about creating such a database on the magithub layer, but I've not had the time with recent life events :-) For me, it solves one big problem: the resources retrieved from any forge do not exist independently of other resources.  They are interconnected.  Retrieving a list of issues from GitHub, for example, does some of the same work as retrieving one issue (except that the latter usually has more necessary information, like 'body' and 'comments').  This results in many mostly-duplicative API hits (and network round-trips) resulting in poorer performance.

This does only impact reading (not writing), so I think the database as a concept covers everything it needs to cover.

vermiculus:
@cmccloud

> Great. Magithub currently grabs all the issues/prs, and the lack of pagination makes interacting with repos with lots of issues or pull requests very difficult.

I can't put my finger on the GH issue right now, but this is on my to-do list :-)

vermiculus:
> Yes, except I don't know if that's the limit for the Github *Graphql* API and the *Gitlab* API. But there is *some* limit.

For GitHub, it's the limit for one request, but you can make successive requests.  IIRC, there's some state on the GraphQL response that can tell us if there's more to grab.

tarsius:
I have decided to implement this as a separate though tightly integrated package called [`forge`](https://github.com/magit/forge).

Currently I am working on fetching asynchronously and abstracting asynchronous unpagination enough so that most of that can be put into `ghub.el` and `forge` only has to deal with converting the retrieved data into tables.

Issue #3496, opened 2018-06-25
Consider providing hooks for commiting & staging


Alexander-Miller:
Hello there,

I maintain [treemacs](https://github.com/Alexander-Miller/treemacs), and I could use some help in making it and magit (pardon the buzzword) synergize together.

One of the features of treemacs is its `filewatch-mode`, a minor mode that allows treemacs to automatically refresh its view when it detects file system changes in the projects it is displaying. Another is `git-mode` which will fontify files and directories based on their git status. It's obvious how the two work together: you edit or create an untracked file and treemacs displays it in the appropriate color all on its own.

There's just one gap - git commands that don't create, delete or change files in the underlying file system. Making commits and staging untracked files are two actions that I can think of that treemacs cannot detect on its own and it is always slightly inconvenient to have to make that refresh yourself.

I've no solution on how to handle this in general, including the bare command line (I assume you know a bit more about these things) so hooking into magit is the obvious choice. I've also tried watching the .git directory itself for changes, but all that seems to get me is dozens of create/delete events for index.lock for *every* git command, including those I start myself, without really knowing what's actually happening.

So here's the details of I would need: Hooks for staging & committing and other git commands that would fit into the change git state, but not the file system category. They should run after the git command is finished so treemacs can successfully pick up the changes. Ideally the hooks would be called with an exact list of files that were staged or committed, but if that kind of information is not readily available just the git repo root is fine, too.

If you're okay with the proposal I can take a stab at making the changes myself, but I'll probably need some help with the details.

vermiculus:
One idea: you could advise the core magit process communication commands (`magit-git.el`) to give you a notification. `magit-top-dir` (IIRC) is your repository root.

Something I've seen people do in the past is to just open a PR with some proposed changes and use that for discussion/iteration.

Alexander-Miller:
Maybe it's just me, but I think advising is a workaround when the "proper" interface is not available, and I'd rather have the proper interface.

I am also not sure if my ideas are the right way to go about this - there's still the gap of directly using the command line, and the people maintaining magit certainly know more about git than I do - so I'd like to do some talking before I start implementing.

vermiculus:
If you want a proper interface that'll work for all users of git, your best bet is a git-hook. `pre-commit-hook` exists, but I can't find anything for staging. You could feasibly populate a .git/ file with a list of affected project files.

Creating something for staging would involve git development.

tarsius:
[...]

PR #3501, opened 2018-06-29
magit-apply-el: Add hooks for staging and unstaging.

Alexander-Miller:
As discussed in https://github.com/magit/magit/issues/3496#issuecomment-401170175 I have implemented a hook for staging commands.

I went with the option of passing the list of (un)staged files to the hooks because that is what I think such a hook will most likely need anyway. If you don't like the performance implications of that choice I am fine with running the hooks without that information.

### DIFF
--- a/lisp/magit-apply.el
+++ b/lisp/magit-apply.el
@@ -271,7 +271,10 @@ requiring confirmation."
                                       nil t nil nil default)
              default))))
   (magit-with-toplevel
-    (magit-stage-1 nil (list file))))
+    (magit-stage-1 nil (list file))
+    (when magit-post-stage-functions
+      (let ((staged-files (magit-staged-files)))
+        (run-hook-with-args 'magit-post-stage-functions staged-files)))))
 
 ;;;###autoload
 (defun magit-stage-modified (&optional all)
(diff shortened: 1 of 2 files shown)


lisp/magit-apply.el
```diff
@@ -234,7 +248,10 @@ at point, stage the file but not its content."
         (`(staged        ,_  ,_) (user-error "Already staged"))
         (`(committed     ,_  ,_) (user-error "Cannot stage committed changes"))
         (`(undefined     ,_  ,_) (user-error "Cannot stage this change")))
-    (call-interactively 'magit-stage-file)))
+    (call-interactively 'magit-stage-file))
+  (magit-with-toplevel
+    (let ((staged-files (magit-staged-files)))
```
vermiculus: Can we get the list of staged files only when there's something in the hook? That will mitigate any unnecessary performance concerns.

Alexander-Miller: Done and done.

Alexander-Miller:
Pushed all the requested changes.
The conditional hook call with input gathering is becoming a bit of a handful. If you think this pattern may be useful elsewhere I can extract this into a macro.

tarsius:
> I went with the option of passing the list of (un)staged files to the hooks because that is what I think such a hook will most likely need anyway.

I don't think we can jump to that conclusion, so I think it would be better to do that work inside your hook function. Once other hook functions from other packages also need to call `magit-{,un}tracked-files` we can consider passing this information as an argument.

But I would like to delay adding these hooks for a while anyway. Please use advice for the time being.

Like I said:

> I plan to improve Magit's own refreshing mechanism in a way that would allow other packages to query the current state, but it will be at least another month until I get to that.

And that would likely make any such hook redundant. No, I don't have more information about my plans at this point.

Alexander-Miller:
Let's close this then.

PR #3505, opened 2018-07-03
git-commit-setup-font-lock: Don't fail in non-existent directory

### LINKED ISSUE #3491: magit-commit not working (again) under cygwin
In a git repository I try to commit like this:

    M-x magit-status
    c c

and the response in the magit-process buffer is:

```
  1 C:/cygwin64/usr/libexec/git-core/git.exe … commit --

*ERROR*: Setting current directory: Permission denied, c:/cygdrive/c/my-repo/.git/error: There was a problem with the editor 'c\:/emacs/bin/emacsclient.exe'.
Please supply the message using either -m or -F option.
```
Note the extra `cygdrive/c/` part in the path. The seems to be the same issue as in #1318 but none of the workarounds there seem to work for me, hence the new issue. 

The same happens when trying to interactively rebase as well.

I'm using:

    M-x magit-version
    Magit 20180618.1934, Git 2.17.0, Emacs 25.2.1, windows-nt
    M-x emacs-version
    GNU Emacs 25.2.1 (x86_64-w64-mingw32) of 2017-04-24

npostavs:
Should fix #3491. I think we should additionally move the cygwin filename hack from `git-commit-setup` to `find-file-not-found-functions`, but I want to make sure just the `git-commit-setup-font-lock` change on its own will fix the original problem.

### DIFF
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -421,7 +421,9 @@ This is only used if Magit is available."
   ;; cygwin git will pass a cygwin path (/cygdrive/c/foo/.git/...),
   ;; try to handle this in window-nt Emacs.
   (--when-let
-      (and (string-match-p git-commit-filename-regexp buffer-file-name)
+      (and (or (string-match-p git-commit-filename-regexp buffer-file-name)
+               (if (boundp 'git-rebase-filename-regexp)
+                   (string-match-p git-rebase-filename-regexp buffer-file-name)))
            (not (file-accessible-directory-p
                  (file-name-directory buffer-file-name)))
            (if (require 'magit-git nil t)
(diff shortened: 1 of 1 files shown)


tarsius:
You might want to just merge this for now since this worked for @mgrunwald, instead of waiting for a reply from @philippmoeller.

npostavs:
Okay, I've installed cygwin git to try this out, and I found interactive rebase wasn't working either.

I'll wait a couple more days and push this weekend; @mgrunwald and @philippmoeller if you can test before then it would be useful to have a double check.

mgrunwald:
Thanks for that work! To test it, I'd have to clone the magit repo and convince Emacs to use that instead of elpa, right? I'm not so used in GitHub workflows...
-- 
Markus Grunwald

tarsius:
You can follow these instructions: https://magit.vc/manual/magit/Debugging-Tools.html.

tarsius:
Clone the repository and then do this:

> If you run Magit from its Git repository, then you should be able to use make emacs-Q instead of the output of this command. 

However I just realized that you might not have `make` installed, in which case that obviously won't work.

PR #3509, opened 2018-07-07
add --gpg-sign to all popups where it can be correctly used

cg505:
Currently the only popup where --gpg-sign is present is the commit popup. This adds --gpg-sign to all relevant popups, which are merge, pull, cherry-pick, am, and rebase.

### DIFF
--- a/lisp/magit-merge.el
+++ b/lisp/magit-merge.el
@@ -37,7 +37,8 @@
   :man-page "git-merge"
   :switches '((?f "Fast-forward only" "--ff-only")
               (?n "No fast-forward"   "--no-ff"))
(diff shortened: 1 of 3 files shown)


tarsius:
Please see https://github.com/magit/magit/wiki/Additional-proposed-infix-arguments-and-suffix-commands.

cg505:
@tarsius My bad. I take it these changes are not wanted, then. Should I generalize entry in the wiki page to include the popups other than `magit-rebase-popup`?

tarsius:
That would be useful, please do.

tarsius:
Thanks!

PR #3518, opened 2018-07-17
 Change order and initial visibility of logs in status buffer

tarsius:
Please see the commit messages, give this a try, and then report on whether you like these changes or not.


kyleam:
The rationale given for both commits makes sense to me.  Testing it out, I can't really say whether I like the order better because I don't have a baseline; for a while now I've removed these hooks in my personal config :/ (but the default hiding may make me reconsider that).

Issue #3523, opened 2018-07-25
magit-log-trace-definition doesn't work with Go


zvbuhl:
When run with Go code, magit-log-trace-definition can't find methods; it does this because `which-function` returns  a string of the form `(r *Receiver) Method(args …)` which contains characters with special meaning to POSIX-regexps. Git's unable to find anything which matches that regular expression, so the output is empty.

# Steps to reproduce

  1.  `git clone https://github.com/golang/go.git`
  2.  `emacs go/src/net/http/client.go`
  3. `M-x magit-log-trace-definition`

You'll see an empty log.

This is with Magit 20180723.1158, Git 2.11.0, Emacs 26.1.50, gnu/linux.

tarsius:
> `which-function` returns  a string of the form `(r *Receiver) Method(args …)`

Is this a reasonable thing for `which-function` to return? Will other callers be able to make sense of it? I suspect `go-mode` doesn't provide a suitable `which-func-functions`.

Please provide an example that works on the command-line, by modifying `git log -L:FUNCTION:FILE`.

/cc @dominikh `go-mode`'s maintainer.

tarsius:
Ah, #3524 mostly answers these questions I guess.

dominikh:
https://github.com/dominikh/go-mode.el/issues/253 may be relevant, too? I haven't yet had the time to look into these features, what they do or what magit is expecting from us.

zvbuhl:
Here's a failing example, again with the Go source:

```
$ cd go
$ git log '-L:(c *Client):src/net/http/client.go'
fatal: -L parameter '(c *Client)' starting at line 1: no match
```

And here's one which works (when run from the Go source directory):

```
$ git log '-L:(c \*Client):go/src/net/http/client.go'
# lots of output elided
```

PR #3524, opened 2018-07-25
Quote regexp-meaningful characters in function name

zvbuhl:
This addresses #3523; it resolves a bug in which `magit-log-trace-definition` doesn't work with Go methods on a pointer receiver (and in general, anything for which `which-function` returns a string containing a regular expression metacharacter: the issue is that git expects a regular expression).

I was a bit concerned that `regexp-quote` might not be suited to
escaping the regular expressions git expects, but it appears to do the
right thing: escaping `[*.\?+^$` but not `(`.


tarsius:
Looks safe enough. Thanks!

PR #3532, opened 2018-08-01
Autodisplay process buffer, whenever output is added to it

dpapavas:
This pull request introduces an option, named `magit-process-display-automatically`, which can be used to enable automatic display of the process buffer, whenever there's some Git output appended to it.  Perhaps my desire for such a feature, is due to the fact that I've only just switched to Magit, after using Git through the command line and am neither proficient enough with Magit to be able to tell that it's done what I wanted it to do, nor acquainted enough with it to blindly trust it and the need for it will soon go away.

Still, Git-related errors have the potential to inflict pain, so closely monitoring operations isn't necessarily bad and having this option isn't necessarily useless, hence this PR.

This feature also made the error header (alerting the user on a Git error and advising to type `$` to display the process buffer) redundant, so I disabled its insertion when the process buffer is already visible.  This seems to make sense regardless of process buffer autodisplay, so I've left it always on, without a configuration option.

Please feel free to merge any part of this PR you see fit (or none of it, of course) and to rewrite/reimplement parts of it as necessary (or request me to do so).  My Emacs lisp is of the pick-it-up-as-you-go variety and hence, in all probability, horrible.  Likewise, please excuse any glaring problems or deficiencies of the implementation, that I might have missed, due to my limited experience with both Magit and Emacs lisp.  All I can say is, that I've been using the functionality introduced in this PR for about a week, without discovering any issues.

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -243,12 +243,13 @@ Used when `magit-process-display-mode-line-error' is non-nil."
   (setq imenu-extract-index-name-function
         'magit-imenu--process-extract-index-name-function))
 
-(defun magit-process-buffer (&optional nodisplay)
+(defun magit-process-buffer (&optional nodisplay nocreate)
   "Display the current repository's process buffer.
 
-If that buffer doesn't exist yet, then create it.
-Non-interactively return the buffer and unless
-optional NODISPLAY is non-nil also display it."
+If that buffer doesn't exist yet, then create it, unless optional
+NOCREATE is non-nil.  Non-interactively return the buffer, or nil
+if it neither existed, nor was created and unless optional
+NODISPLAY is non-nil also display it."
   (interactive)
   (let ((topdir (magit-toplevel)))
     (unless topdir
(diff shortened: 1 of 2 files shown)


tarsius:
`(setq magit-process-popup-time 0)` only causes the buffer to be shown when a command actually added something to it. Commands that run `git` and then consume its output do not write anything to the buffer, this includes e.g. `magit-stage`. If `git` is run for side-effects only, as when e.g. pushing, then something is inserted into the buffer.

If you have set `magit-git-debug` to `t`, then some commands write to the buffer but without ever showing. The various wip modes also write to the buffer without showing it.

It seems you want to buffer to be shown whenever you do anything. Why don't you just show it right away by pressing `$` or using `magit-status-mode-hook`?

vermiculus:
> `(setq magit-process-popup-time 0)` only causes the buffer to be shown when a command actually added something to it.

> Why don't you just show it right away by pressing `$`

Wouldn't bringing up the popup buffer manually (by definition) still only show only what was added to the buffer?

Still, if this option is kept, it may make more sense to use the existing variable rather than introducing a new one with the potential for weird interactions.

tarsius:
Output that usually isn't inserted into the buffer is inserted when `magit-git-debug` is `t` **by `magit-git-string`**. Unlike the function that always inserts into the buffer (`magit-start-process`) does not call the code that may cause the buffer to be displayed depending on the value of `magit-process-popup-time`.

So when `magit-git-debug` is `t` and `magit-process-popup-time` is `0`, then it is possible for new output to arrive in the buffer without it being displayed.

But `magit-git-debug` is only intended for debugging purposes. It may also be useful for learning, but it comes with a heavy performance penalty. If a user decides to pay that price in order to learn what `git` commands Magit runs when doing certain things, then I think it is perfectly okay if s/he first has to press `$`.

In other words I don't plan to merge this or some variation.

dpapavas:
Perhaps I'm missing something, but both with the version of Magit installed via elpa, as well as with the current master branch checkout and `make emacs-Q`, all (as far as I've seen) Git commands that were carried out at my request, including staging, branching operations, etc. produce output on the process buffer, even though `magit-git-debug` has its default value, i.e. `nil`.

>It seems you want to buffer to be shown whenever you do anything. Why don't you just show it right away by pressing `$` or using `magit-status-mode-hook`?

Almost. I want the buffer to be shown whenever _Git is asked to do anything_.  For instance, I often visit the status buffer just to see which files are modified, in order to visit them, or find a specific change.  In that case, having the status buffer take up a window could be bothersome.  Similarly, pressing `$`, needs to be done manually and steals the focus from the status window, which you don't necessarily want.

In any case, I don't want to press the issue.  If you consider this PR undesirable in its entirety (i.e. including the somewhat independent change to avoid displaying redundant error headers), feel free to close it.

tarsius:
> Perhaps I'm missing something, [...]

I wasn't precise. Additionally to what I already said the automatic appearance of the buffer only happens when `git` is run *asynchronously* for side-effects.

> Almost. I want the buffer to be shown whenever Git is asked to do anything.

This buffer wasn't really designed for that purpose. Nothing wrong with trying to use it like that but I am weary of adding yet another option. Looking at the relevant commit, it can easily be seen that this can easily be implemented using advice. Please do that.

Please also add that advice to https://github.com/magit/magit/wiki/Tips-and-Tricks#useful-code-snippets.

> (i.e. including the somewhat independent change to avoid displaying redundant error headers)

I don't really want that either. Sure this may be redundant, but it is also just one line and many users will have grown accustomed to it. If the process buffer poped up a user because e.g. a pull took a long time, then s/he might not expect a subsequent error *not* to be displayed in the status buffer just because s/he did not get around to hiding the process buffer again yet.

PR #3538, opened 2018-08-05
magit-perl-executable: New variable

### LINKED ISSUE #3537: Add defvar for perl command
Magit 20180731.2119, Git 2.18.0, Emacs 26.1, gnu/linux

Some parts of Magit assume that `perl` is in the user's PATH.
For instance, the `magit--cherry-move' function:

```elisp
                   (push (format "%s=perl -i -ne '/^pick (%s)/ or print'"
                                 "GIT_SEQUENCE_EDITOR"
                                 (mapconcat #'magit-rev-abbrev commits "|"))
                         process-environment)
```
will fail if `perl` is not in the `PATH`.

On systems running [Nix](https://nixos.org) or [Guix](https://www.gnu.org/software/guix), Perl can be installed in the store but not necessarily in the user profile (i.e. in the PATH).  This is why it's good practice to make every system command customizable so that all users and packagers can easily adapt those dependencies.

My suggestion:
```elisp
(defvar magit-perl-command "perl"
  "The Perl interpreter.")

...

                   (push (format "%s=%s -i -ne '/^pick (%s)/ or print'"
                                 magit-perl-command
                                 "GIT_SEQUENCE_EDITOR"
                                 (mapconcat #'magit-rev-abbrev commits "|"))
                         process-environment)
```

N.B.: I'm not sure why "GIT_SEQUENCE_EDITOR" is not directly written into the format string.  Mistake?

Ambrevar:
Fix #3537.

### DIFF
--- a/lisp/magit-sequence.el
+++ b/lisp/magit-sequence.el
@@ -116,6 +116,9 @@ This discards all changes made since the sequence started."
 
 ;;; Cherry-Pick
 
+(defvar magit-perl-executable "perl"
+  "The Perl executable.")
+
 ;;;###autoload (autoload 'magit-cherry-pick-popup "magit-sequence" nil t)
 (magit-define-popup magit-cherry-pick-popup
   "Popup console for cherry-pick commands."
@@ -275,8 +278,9 @@ the process manually."
                 (t
                  (magit-call-git "checkout" src)
                  (let ((process-environment process-environment))
-                   (push (format "%s=perl -i -ne '/^pick (%s)/ or print'"
+                   (push (format "%s=%s -i -ne '/^pick (%s)/ or print'"
                                  "GIT_SEQUENCE_EDITOR"
+                                 magit-perl-executable
                                  (mapconcat #'magit-rev-abbrev commits "|"))
                          process-environment)
                    (magit-run-git-sequencer "rebase" "-i" keep))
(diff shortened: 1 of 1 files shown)


lisp/magit-sequence.el
```diff
@@ -648,7 +656,8 @@ START has to be selected from a list of recent commits."
                      (magit-rebase-arguments)))
   (magit-rebase-interactive-1 commit args
     "Type %p on a commit to remove it,"
-    "perl -i -p -e '++$x if not $x and s/^pick/# pick/'"
+    (concat magit-perl-executable
+            "%s -i -p -e '++$x if not $x and s/^pick/# pick/'")
```
fbergroth: The `%s` seems to be an accident here

Ambrevar: Duh!  Thanks for catching this!

lisp/magit-sequence.el
```diff
@@ -116,6 +116,11 @@ This discards all changes made since the sequence started."
 
 ;;; Cherry-Pick
 
+(defcustom magit-perl-executable "perl"
```
kyleam: Based on #3537, I expected a `defvar`.  I don't have a strong preference, but considering the "perl" calls in `magit-rebase-{edit,reword}-commit` have been around since 2014 and users haven't asked for an option, it doesn't seem worth exposing as a defcustom.

kyleam:
> I also thought of a defvar initially but then looking at the other `-executable` variables I noticed they all were defcustoms.

Yep,  `magit-git-executable` and `magit-gitk-executable`.  They both have some windows-specific logic in there too.  I think Magit certainly needs to expose the git executable as an option to the user.

> I've decided to use that for consistency.

Fair enough.  I still lean towards a defvar because a plain "perl" has been there since 2014 and nobody has asked for an option.  And it's easy to later go from defvar to defcustom if needed.  But again I don't have a strong preference.  (If you do stick with a defcustom, it should have a :package-version.)

Ambrevar:
Alright, I've changed it to a `defvar` and I've added a few lines to the commit log.

tarsius:
Thanks!

Issue #3308, opened 2017-12-15
magit-describe-section (or similar) for users


npostavs:
The existing `magit-describe-section` gives some info useful for developers, but I think users could benefit from easy access to some information about a given section too. Namely:

- Which hook produced it.
- Which function in the hook produced it.
- What git command produced the output (possibly just displaying the docstring of the function would be enough).

tarsius:
I don't plan to implement. Users aren't really asking for it.

npostavs:
I will probably try to implement this at some point, but I'm too busy right now; I might get to this in a month or two.

> Users aren't really asking for it.

Not in so many words, but I perceive that they are "asking for it" when I see questions like
[Meaning of magit status buffer: Head, Merge, Push](https://emacs.stackexchange.com/questions/26247/meaning-of-magit-status-buffer-head-merge-push) and 
[What's the meaning of rebase in magit status buffer?](https://emacs.stackexchange.com/questions/19162/whats-the-meaning-of-rebase-in-magit-status-buffer). Maybe even [why does magit say I haven't pushed the commit?](https://emacs.stackexchange.com/questions/14543/why-does-magit-say-i-havent-pushed-the-commit).

PR #3539, opened 2018-08-05
magit-describe-section for users

npostavs:
v2 of #3505 (see also #3308 for initial suggestion).

### DIFF
--- a/Documentation/RelNotes/2.90.0.org
+++ b/Documentation/RelNotes/2.90.0.org
@@ -13,4 +13,10 @@
   Use ~oref~ and ~oset~ instead.
 
 ** Changes since v2.13.0
+
+- ~magit-describe-section~ now pops up a ~*Help*~ buffer showing which
+  hook and function produced the section.  The old
+  ~magit-describe-section~ is renamed to
+  ~magit-describe-section-briefly~.
+
 ** Fixes since v2.13.0
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -8,7 +8,7 @@
 #+TEXINFO_DIR_CATEGORY: Emacs
 #+TEXINFO_DIR_TITLE: Magit: (magit).
 #+TEXINFO_DIR_DESC: Using Git from Emacs with Magit.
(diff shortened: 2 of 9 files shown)


npostavs:
I think this should be ready now, modulo regenerating the manual.

While testing this out, I did notice some slightly odd section navigation in the revision buffer:

> `Parent:     6d72f0ee magit-no-confirm: Fix description of Custom choice item`
> Merged:     `maint` `master`
> Containing: `master` `np/section-desc-v2`
> `Follows:    2.13.0 (107)`

The entries after "Merged" and "Containing" each act as their individual sections, whereas the "Parent" and "Follows" are just single sections over the whole line. Is that on purpose? It seems like it would be more consistent to have the word before the colon not be part of the section every time, or else it should always be part of the section.

tarsius:
> While testing this out, I did notice some slightly odd section navigation in the revision buffer:

I think that's okay. There's only one parent and git only spits out one tag that follows. It's the other lines that are unusual, so I don't think we should make the regular lines behave in an unusual way just because there are unusual lines close by. Of course doing that would also make sense, i.e. both approaches make sense and I favor sticking with what we have.

tarsius:
I added a fixup commit that wraps lines a bit differently just in case the name of the section inserted is long (I just tried it with `magit-insert-unpushed-to-upstream-or-recent`).

(I already rebased and updated the manual before I noticed that I wanted to make that change.)

npostavs:
> both approaches make sense and I favor sticking with what we have.

Fine with me.

> I added a fixup commit that wraps lines a bit differently just in case the name of the section inserted is long (I just tried it with magit-insert-unpushed-to-upstream-or-recent).

Hmm, should the `called by ...` get the same treatment to stay consistent? (this shows up when describing any of the header sections)

tarsius:
> Hmm, should the called by ... from ... get the same treatment to stay consistent?

I think so.

> `#<magit-section #("master" 0 6 (face magit-branch-local)) [branch branch status] 73-148>`

If the section value is a string, then we should remove all properties.

npostavs:
Okay, done those. I will merge to master tomorrow, since it's getting a bit late here, and I've found it's better to do that only when I'm fully awake.

PR #3540, opened 2018-08-05
Show executable-name without directory.

qwerbzuio:
Make this behaviour configurable via magit-process-display-executable-basename.

The path to the git-executable might be quite long on some systems. It then eats
up a large part of the modeline, hiding more interesting information like the
git-subcommand that has been run. With this change, an option is introduced to
show only the executable basename. By default, the old behaviour is preserved.


tarsius:
I don't think we even need an option for this. Just always drop the directory part.

PR #3541, opened 2018-08-05
Fix revision message placement in case of risky var query

npostavs:
It seems a side effect of calling `hack-dir-locals-variables` can mess up the point while inserting, here's a before screenshot:

bad-rev-highlight


tarsius:
Thanks. I saw that only once and figured it had something to do with with some incomplete changes I had loaded at the time.

phil-s:
Should that be reported as an Emacs bug?

npostavs:
> Should that be reported as an Emacs bug?

Popped up windows can move point (e.g., if point is at the bottom of the window, and the new window is created below), that's expected Emacs behaviour, not a bug, I think.

Issue #2970, opened 2017-01-22
Cooperate with outline-minor-mode et al.


tarsius:
Magit needs to cooperate with modes such as `outline-minor-mode`, `outline-mode`, and `org-mode` in order to preserve (and/or properly setting up) outline visibility.  The main issue is that reverting a buffer, so that its content matches that of the visited file, which has changed on disk, messes with section visibility.  After a revert, it is usually necessary to expand everything, and then setup the "narrowing" again from scratch.

A buffer can be reverted automatically using `auto-revert-mode`, or explicitly using the command `revert-buffer`.  When using Magit, then `auto-revert-mode` is enabled by default in buffers visiting tracked files, and situations in which a Magit command triggers a revert include checking out a branch/revision and rebasing.

But this issue is not specific to Magit, outline visibility gets messed up whenever a buffer is reverted, regardless of why it is reverted.  So the proper place to implement this is the built-in `outline.el`, and I intend to do that because (1) I use outline features a lot myself, regardless of whether Magit is involved in the work-flow, and (2) I have plans for some features that could be implement once it is possible to preserve the outline visibility across reverts.

Some of these features are version-control related, which is why I am opening this as a Magit issued.

----

Like the mentioned modes, Magit too has "sections", but they are implemented quite differently.  In Magit a tree of sections (aka "dom") is available at all times, while `outline` based modes use font-lock to give section headings the appropriate look and their section movement and visibility commands always have to determine the context from scratch.

So `outline` lacks a feature to preserve visibility because it lacks a feature to remember structure.  However implementing that shouldn't be too hard and Magit's implementation can be a guide here.

Basically we have to hook into the revert process: first we determine and cache the structure/visibility, then we revert as before, and then we restore the visibility as best as we can.  Some issues that will have to be dealt with in the final step are sections that appear and disappear, but again, that's something that has already been solved for Magit buffers.

----

Once this basic functionality exists (and has hopefully been accepted into Emacs), we can implement the following:

- **1)** Persist the visibility state across Emacs sessions, i.e. save the visibility like `savehist` does for variables.
- **2)** Save multiple, task-specific visibility states (for example in bookmarks) and allow switching between them.

Again, these features have little to do with Magit or version-control in general.  However the following feature is particularly useful when using some version-control system.  (It's not specific to Magit however, so here too I will try to get my changes into Emacs.)

- **3)** Add a command to hide everything, except those sections that contain changes.

[...]

PR #3545, opened 2018-08-08
magit-diff-visit-file: call org-reveal when visiting org-mode files.

Alexander-Miller:
When visiting a change deep within the hierarchy an org-mode file you sometimes see just the top headline and have to reveal the context yourself (it usually happens when the file is already open in emacs). This change makes sure that the correct context is always revealed.

If you don't like packing org-related code directly into magit I can also include this as a default part of `magit-diff-visit-file-hook`.


Alexander-Miller:
reveal-mode is too permanent a solution for my own taste, I would want the reveal to only happen on visit, but if that's not a change you want in magit then I can just use the present hooks.

Just one more thing: the part about reveal-mode is no longer in the FAQ, https://github.com/magit/magit/wiki/FAQ#wiki-jumping-into-a-file-might-ends-up-in-an-invisible-region leads nowhere. I cannot find anything when grepping the code base for "reveal" either. Should it stay that way, or should I reinstate that entry?

tarsius:
> I can just use the present hooks.

Please do that for now. I am actually considering to implement something that goes a bit further than what you have proposed here -- eventually. E.g. when jumping to a file with conflicts that have to resolved, then make sure all the conflicts are visible but everything else is not. #2970

> Just one more thing: the part about reveal-mode is no longer in the FAQ

I've added an entry to the FAQ in the manual.

PR #3546, opened 2018-08-08
Do not disable show-trailing-whitespace globally.

kurnevsky:
It seems `magit-mode` should use `setq-local` to disable `show-trailing-whitespace` in it's buffers only. Instead magit disables it globally.


basil-conto:
> Instead magit disables it globally.

It doesn't, see <kbd>C-h</kbd><kbd>v</kbd>`show-trailing-whitespace`<kbd>RET</kbd>:

    show-trailing-whitespace is a variable defined in ‘xdisp.c’.
    Its value is nil

      Automatically becomes buffer-local when set.

kurnevsky:
ok, my bad. I thought it's not buffer-local for some reason. Sorry for disturbance.

PR #3552, opened 2018-08-11
Add ability to visit a remote

georgek:
This adds the ability to "visit" a remote. In many cases the remote can be online when it is a place like Github, Gitlab etc.

If this feature is welcome I would add a more comprehensive list of functions to `magit-visitable-remotes-list` and maybe update the manual if necessary.

### DIFF
--- a/lisp/magit-refs.el
+++ b/lisp/magit-refs.el
@@ -460,10 +460,35 @@ Branch %s already exists.
                (call-interactively #'magit-show-commit))))
     (call-interactively #'magit-show-commit)))
 
+(defvar magit-visitable-remotes-list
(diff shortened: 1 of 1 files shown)


vermiculus:
Should this be in the forge source files instead?

tarsius:
Yes and I have already implemented it. I am currently refactoring some low-level stuff in`ghub-graphql.el` to avoid tons of boilerplate, but once I am done with that I will return to work on [Forge]( https://github.com/magit/forge) and get it ready for an alpha release.

tarsius:
I am done with that and am now adding at least "visit in a browser" support for a few more forge types.

PR #3559, opened 2018-08-21
use value of magit-log-select-arguments instead of --graph

d1egoaz:
It has huge performance implications in large repos.

I've defined `magit-log-select-arguments '("-n256")` but when using git commit fixup it was always showing up the log graph.

This PR tries to use the value of `magit-log-select-arguments` instead of the hardcoded `--graph`

### DIFF
--- a/lisp/magit-commit.el
+++ b/lisp/magit-commit.el
@@ -296,7 +296,7 @@ depending on the value of option `magit-commit-squash-confirm'."
               "" "true" nil t)))
         (format "Type %%p on a commit to %s into it,"
                 (substring option 2))
-        nil nil (list "--graph"))
+        nil nil magit-log-select-arguments)
       (when magit-commit-show-diff
         (let ((magit-display-buffer-noselect t))
           (apply #'magit-diff-staged nil (magit-diff-arguments)))))))


d1egoaz:
Thanks for the info @kyleam .

That's a very interesting idea to add `--graph` to `magit-log-select-arguments` by default.

I'd like to have the final decision to use additional arguments or no when calling some git commands, that's why it's so nice to have `*-arguments`, as you can see, I've modified in order to get performance:
```
        magit-log-arguments '("-n50")  ;; was: '("-n256" "--graph" "--decorate")
        magit-log-section-arguments  '("-n50") ;; was: ("-n256" "--decorate")
        magit-log-select-arguments '("-n50")  ;; was: '("-n256" "--decorate")
```

I understand it helps some users to identify merges, in my case, I'd love to have an option to remove the `--graph` and take the risk.

Another option would be to add a flag `magit-commit-squash-show-graph t` in order to just use the `magit-log-select-arguments`

kyleam:
> In the simple case where there is no merge involved all lines would simply contain a needless * and when there is a merge, then seeing them might actually be beneficial when using those callers too.
>
> Am I missing something or should I just add --graph to the default value and drop the hard-coded --graph?

The needless graph characters is the worst consequence I could think of as well, so I think it's worth adding to the default value.

> Anyway, I have customized the variable locally to test for a while whether that becomes annoying.

Sounds good.  I'll do that too.

PR #3560, opened 2018-08-23
Edit FAQ wording concerning handling of color

ntdef:
I noticed some typos and ambiguity in the docs concerning color control sequence output. Thought I would clean it up while I had it in front of me.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -8,7 +8,7 @@
 #+TEXINFO_DIR_CATEGORY: Emacs
 #+TEXINFO_DIR_TITLE: Magit: (magit).
 #+TEXINFO_DIR_DESC: Using Git from Emacs with Magit.
-#+SUBTITLE: for version 2.13.0 (2.13.0-182-g38fd25ee6+1)
+#+SUBTITLE: for version 2.13.0 (2.13.0-200-g30ccf86f1+1)
 #+BIND: ox-texinfo+-before-export-hook ox-texinfo+-update-version-strings
 
 #+TEXINFO_DEFFN: t
@@ -23,7 +23,7 @@ directly from within Emacs.  While many fine Git clients exist, only
 Magit and Git itself deserve to be called porcelains.
 
 #+TEXINFO: @noindent
-This manual is for Magit version 2.13.0 (2.13.0-182-g38fd25ee6+1).
+This manual is for Magit version 2.13.0 (2.13.0-200-g30ccf86f1+1).
 
 #+BEGIN_QUOTE
 Copyright (C) 2015-2018 Jonas Bernoulli <jonas@bernoul.li>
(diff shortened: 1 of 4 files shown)


Documentation/magit.org
```diff
@@ -7857,12 +7857,12 @@ tag signing", then you'll also need to synchronize ~$GPG_AGENT_INFO~.
...
 and ~color.diff~ are ~false~, ~true~ and ~always~, and the default is ~true~.
-You should leave it that because then you get colorful output in
-terminals but git's output is consumed by something else, then no
-colors are used.
+You should leave it that way because then you get colorful output in
+terminals by default but when git's output is consumed by something else, 
+then no color control sequences are used.
 
-If you actually use some other tool which expects that requires that
-you force git to output control sequences (which is highly unlikely),
```
npostavs: You have to remove the first "that" as well (i.e., "which expects git to output...").
Or else remove only "expects that" (i.e., "which requires that you force git to output...").

ntdef: oh good catch

tarsius:
Thanks! I have pushed a fixup commit that

1. Rewords the second paragraph separately. With your change we loose information.
2. Fixes the issues that were caused by you use of Github. I just realized that I need to update the pull-request template in order to tell users not to do that.  (Actually some of these issues won't be fixed until I squash before merge.)

   - Commit message contains very long line. (That pissed off Linus a lot, they didn't care.)
   - Commit message does not end with a newline character. (That pisses me off. I told them multiple times and they don't care.)
   - Trailing whitespace.
   - No paragraph filling.

   Because Github makes such defects extremely likely, I do not consider Github to be a decent interface to make changes to a file and create commit messages. Really its mostly their fault not yours.

Documentation/magit.org
```diff
@@ -7858,12 +7858,12 @@ This happens when you configure Git to always color diffs and/or all
...
-terminals by default but when git's output is consumed by something else, 
-then no color control sequences are used.
+terminals by default but when git's output is consumed by something
+else, then no color control sequences are used.
 
-If you actually use some other tool which expects git to output control 
-sequences (which is highly unlikely), then you can override these settings 
-just for Magit by using:
+If you actually use some other tool that requires that git is forced
+to output control sequences (which is highly unlikely), then you can
```
npostavs: Hmm, this doesn't quite parse right to me. What's wrong with the previous version?

tarsius: The distinction is between (1) tools that themselves make sure that the control sequences that they need are actually present, (2) tools that instead rely on something else (e.g. the user setting `color.*.ui=true`) making sure that is so, (3) tool that don't need them or don't care. "that git is forced to output control sequences" refers to (2) setting them apart mainly from (1).

npostavs: I don't think I understand the difference between (1) and (2). Do you mean that (1) would check and print an error message if control sequences are missing whereas (2) would just give the wrong answer? It seems like a fairly academic distinction.

tarsius: (1) call git like `git -c color.diff.ui=true ...`, while
(2) simply hope that the user either has done `git config color.diff.ui true` or does not dare to run the tool outside a terminal.

npostavs: Then perhaps something more explicit, like:

```
If you actually use some other tool which requires setting
~color.diff~ to ~always~ (which is highly unlikely), then you can
override these settings just for Magit by using:
```

tarsius: That's better.

tarsius:
@ntdef Would you mind giving us your real email address, so `AUTHORS.md` isn't poluted with `9503857+ntdef@users.noreply.github.com` (another reason not to use Github for committing).

PR #3562, opened 2018-08-26
Only jump to commits when first creating the status buffer

### LINKED ISSUE #3561: magit-status always jumps to "Recent commits"
Magit 20180820.2042, Git 2.18.0, Emacs 26.1, gnu/linux

I can't say for sure if this is new behavior but it _seems_ new.  It's certainly started bothering me recently but perhaps I have just started getting ornery recently :)

I have `C-c g` bound to `magit-status` and when I use that keybinding for the second and subsequent times in a buffer, it brings up the existing magit: buffer for my Git repository but instead of leaving point where it was the last time I visited the status buffer, it always puts point at the "Recent commits" line in the buffer.

This makes me sad because what I almost always want to look at is NOT my "Recent commits" but rather the staged/unstaged changes in my workspace.  And since I have a lot of stashes (25 right now) that come between staged/unstaged changes and "Recent commits", what I want to see and use is always off the window above and I have to scroll or jump back up to get to it.

I'd prefer to have `magit-status` bring the status buffer to the front without moving point from where it was previously.  Or, failing that, I'd rather have the default selected section *not* be "Recent commits" which may be useful but to my mind is much less commonly what one wants to see when opening up the status buffer than unstaged/staged changes (I realize this section might not always exist...)

Thoughts?

tarsius:
Fixes #3561.

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -250,6 +250,12 @@ then fall back to regular region highlighting."
   :type 'hook
   :options '(magit-section-update-region magit-diff-update-hunk-region))
 
+(defcustom magit-create-buffer-hook nil
+  "Normal hook run after creating a new `magit-mode' buffer."
+  :package-version '(magit . "2.90.0")
+  :group 'magit-refresh
+  :type 'hook)
+
 (defcustom magit-refresh-buffer-hook nil
   "Normal hook for `magit-refresh-buffer' to run after refreshing."
   :package-version '(magit . "2.1.0")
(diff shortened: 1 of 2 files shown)


tarsius:
> Is that intentional?

Yes, but...

> Also, the current order in magit-status-sections-hook lists the unpushed variants first, but magit-status-goto-commits-1 prefers unpulled sections over unpushed sections.

We could make it customizable and I was expecting that that would be requested eventually.

What should we offer? We could use an option of type `repeat` to implement fallback behavior (like it currently goes to the unpulled if that exists, else to the recent commits). And the individual behaviors I can think of are (1) go to section by identity and (2) skip headers, which could be generalized as (3) go to nth section.

tarsius:
Merged since it is a bugfix. Unfortunately we cannot reopen this, so I have reopened #3561 instead.

kyleam:
> What should we offer? We could use an option of type repeat to implement fallback behavior (like it currently goes to the unpulled if that exists, else to the recent commits). And the individual behaviors I can think of are (1) go to section by identity and (2) skip headers, which could be generalized as (3) go to nth section.

Sounds reasonable.  With the combination of the fallback behavior and "go to nth top-level section", I think users should be able to get the behavior they want.   What are you thinking for the default?

tarsius:
> What are you thinking for the default?

Probably "goto 2th".

PR #3570, opened 2018-08-30
magit-log-select-arguments: add --graph to default value

tarsius:
And stop hard-coding `--graph' in `magit-commit-squash-internal'.
This makes it possible for users NOT to use `--graph' if that has
negative effects on performance.  On the other hand it also means
that all uses of `magit-log-select' now show the graph by default,
but the latter might actually be beneficial too.

Re #3559.

### DIFF
--- a/lisp/magit-commit.el
+++ b/lisp/magit-commit.el
@@ -295,8 +295,7 @@ depending on the value of option `magit-commit-squash-confirm'."
                 (list "--autosquash" "--autostash")
               "" "true" nil t)))
         (format "Type %%p on a commit to %s into it,"
-                (substring option 2))
-        nil nil (list "--graph"))
+                (substring option 2)))
       (when magit-commit-show-diff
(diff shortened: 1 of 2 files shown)


kyleam:
Testing this out, I didn't spot anything that made me think using --graph here is a bad idea.

There was only one situation where the new behavior felt odd to me.  Before this change, if I call magit-rebase-subset from a log buffer, the transition from the regular log buffer (with --graph) and the log select buffer (without --graph) was very distinct visually because every line shifts. Now, the lines look the same.  It's still obvious that you're in a select mode buffer because the header changes and point moves to the beginning of the buffer, but it's just a less drastic transition than I've become used to.

d1egoaz:
Thanks folks!

tarsius:
It didn't annoy me, then again I might not have used much any features that use log-select. Lets merge and see if it annoys anyone else.

PR #3582, opened 2018-09-12
Add magit-submodule-remove function

manateelazycat:
magit-submodule-remove prompt user which submodule need remove, then all clean works finish automatically.

Github is an insufficient interface for editing files and creating commits.  Please don't do it when contributing to this Magit.

What you should write here

Please summarize the changes made in the commits.  Explain why you are making these changes, not just what changes you are making.  This also applies to the commit messages.

How to update the manual

If you make changes to the manual, then edit only "magit.org".  Do not manually edit "magit.texi".  The latter has to be generate from the former.  If you want to do that yourself, then follow the instructions at [2].  Otherwise a maintainer will do it and amend that to your commit.

### DIFF
--- a/lisp/magit-submodule.el
+++ b/lisp/magit-submodule.el
@@ -169,6 +169,75 @@ it is nil, then PATH also becomes the name."
     (magit-refresh)))
 
 ;;;###autoload
+(defun magit-submodule-remove (&optional module-name)
+  (interactive)
+  (save-excursion
+    ;; Get current directory for restore after remove submodule.
+    (setq current-directory default-directory)
+    ;; Cd magit toplevel directory make sure `magit-list-module-paths' can work.
+    (cd (magit-toplevel))
+    ;; Get subodule name/path will to remove.
+    (setq submodule-name (or module-name (completing-read "Remove submodule: " (magit-list-module-paths))))
+    (setq submodule-path (concat (magit-toplevel) submodule-name))
+    ;; Delete submodule directory first.
+    (when (file-exists-p submodule-path)
+      (delete-directory submodule-path t))
+    ;; Delete submodule from .gitmodules file.
+    (setq gitmodules-file-path (concat (magit-toplevel) ".gitmodules"))
+    (magit-kill-buffer-match-file gitmodules-file-path)
+    (with-current-buffer (find-file gitmodules-file-path)
+      (goto-char (point-min))
+      (when (search-forward-regexp (format "path\\s-=\\s-%s" submodule-name) nil t)
+        (previous-line)
+        (beginning-of-line)
+        (when (search-forward-regexp "\\[submodule\\s-\"\\(.*\\)\"]" nil t)
+          (setq git-modules-submodule-path (concat (magit-toplevel) ".git/modules/" (match-string 1))))
+        (beginning-of-line)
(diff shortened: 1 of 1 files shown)


manateelazycat:
https://www.emacswiki.org/emacs/magit-extension.el Because i upload it to emacswiki.

Of course, i write it myself.

Why no merge my PR ? You can point out the problem in the code

manateelazycat:
@tarsius This is my emacs homepage: https://www.emacswiki.org/emacs/AndyStewart

I have play emacs many many years, I think i don't need copy other people's code.

vermiculus:
I don't think it was meant as an accusation, but the code *does* look hacked-together and doesn't really follow good development standards:
* free variables exist (the symbols you use aren't let-bound)
* you depend on the structure of `.gitmodules` and `.git/config`/`.gitconfig` instead of asking git for the information using established communication API
* as mentioned, you're using try-catch patterns where simple find-index patterns would be much more robust (and efficient)
* probably other things that would present themselves upon a more thorough review…
This is not a personal attack; they are just facts of the current implementation. It's probably easiest to simply start over with a good understanding of magit plumbing than it is to try and fix the current implementation.

manateelazycat:
@vermiculus 

My personal opinion is, if there is any problem with my implementation, I can modify my code to help everyone, instead of simply rejecting others' help to implement a new version.

I believe that pointing out the code problem can allow other developers (include me) to learn more and better help magit grow.

manateelazycat:
@vermiculus Do you know any magit API parse submode's name and path information ?

I believe i can write better code that use API, and re-send new PR.

thanks.

manateelazycat:
IMO, We can use API code (maybe git command) instead search-forward-regexp stuff, but we still need change content of .gitmodules, .git/config and delete subdirectory from .git/modules at last.

manateelazycat:
@vermiculus I will write new version, and post here let you and @tarsius review. 
Thanks for your suggesion.

manateelazycat:
> I hate to say it, but teaching others how to do development and helping a project grow are not often goals in agreement. You've already helped magit grow by demonstrating the need for this functionality and giving credibility to its sincerity by providing an implementation (essentially 'proving' you _did_ need it). Take the opportunity to learn from the new implementation.
> 
> I suspect every reasonable edit to a `.git*/*` files is possible from git and will not require modifying the files directly. The implementation will be significantly different.
> 
> Still, if you want to augment this same PR with a new implementation (just push to the same branch again), you are free to do so. That is no guarantee the implementation will be used, however, but you will still learn from the process if that's your goal:
> 
> * Check out the functions in `magit-git.el`.
> * Grep the sources for `config`.
> * Read the git documentation about how to retrieve submodule information properly.
> * Learn patterns to find information other than try-catch. (The dash.el package documentation may be helpful here.)
> * Read up on good Lisp style and consider the impact of free variables in a global namespace.
> 
> To be absolutely clear, this is a critique of the implementation, not you personally. You can't learn this stuff overnight. The hard truth is that it will take longer to learn all of this (or even do a very targeted back-and-forth) than it would be to simply provide a new implementation that will benefit users sooner.

@vermiculus 
I have rewrite new version at: https://github.com/magit/magit/pull/3583

Please review.

Issue #3584, opened 2018-09-13
It possible make magit-submodule-add asynchronization ?


manateelazycat:
Emacs will freeze if when we add submodule that have huge commits (such as magit).

So it's possible make magit-submodule-add asynchronization?

We still can use emacs even magit is cloning huge git repository.

Thanks

tarsius:
Yes. Implemented in #3585.

PR #3583, opened 2018-09-12
Add magit-submodule-remove

manateelazycat:
Use magit-submodule-remove remove submodule automatically.

PR #3585, opened 2018-09-13
New and improved module commands

tarsius:
I haven't tested any of this yet but am publishing it now to prevent others from implementing and reviewing things that won't be merged.

* `magit-submodule-add: Run asynchronously` in response to #3584.

  This is a recurring pattern and should be abstracted, but there is no hurry to do so.

* `magit-submodule-remove: New command`

  The major differences to the implementations suggested in #3582 and #3583 are 
  * It refuses to do anything if there are uncommitted changes.
  * It does *not* remove the git directory of the submodule. If the gitdir is located in the working tree, then it is first moved into the gitdir of the super-repository. Removing a gitdir is very dangerous.
    * An additional command `magit-submodule-purge-gitdir` could be added which does remove a gitdir that is no longer connected to a module.
    * `magit-submodule-add` should be taught to ask the user how to deal with a gitdir that conflicts with the gitdir that is about to be added.

My `borg` package manager already does these things, but it can only deal with modules in `~/.emacs.d/lib`. But for that limited use-case this approach has worked well.

### DIFF
--- a/lisp/magit-submodule.el
+++ b/lisp/magit-submodule.el
@@ -177,6 +177,7 @@ it is nil, then PATH also becomes the name."
              (magit-submodule-read-name-for-path path)
              (magit-submodule-filtered-arguments "--force")))))
   (magit-with-toplevel
+    (magit-submodule--maybe-reuse-gitdir name path)
     (magit-run-git-async "submodule" "add"
                          (and name (list "--name" name))
                          args "--" url path)
@@ -596,6 +597,33 @@ These sections can be expanded to show the respective commits."
 (defun magit-modulelist-column-path (path)
   "Insert the relative path of the submodule."
   path)
+
+;;; Utilities
+
+(defun magit-submodule--maybe-reuse-gitdir (name path)
+  (let ((gitdir
+         (magit-git-dir (convert-standard-filename (concat "modules/" name)))))
+    (when (and (file-exists-p gitdir)
+               (not (file-exists-p path)))
+      (pcase (read-char-choice
+              (concat
+               gitdir " already exists.\n"
+               "Type [u] to use the existing gitdir and create the working tree\n"
+               "     [r] to rename the existing gitdir and clone again\n"
+               "     [t] to trash the existing gitdir and clone again\n"
+               "   [C-g] to abort ")
+              '(?u ?r ?t))
+        (?u (magit-submodule--restore-worktree (expand-file-name path) gitdir))
+        (?r (rename-file gitdir (concat gitdir "-"
+                                        (format-time-string "%F-%T"))))
+        (?t (delete-directory gitdir t t))))))
+
+(defun magit-submodule--restore-worktree (worktree gitdir)
+  (make-directory worktree t)
+  (with-temp-file (expand-file-name ".git" worktree)
+    (insert "gitdir: " (file-relative-name gitdir worktree) "\n"))
+  (let ((default-directory worktree))
+    (magit-call-git "reset" "--hard" "HEAD")))
 
 (provide 'magit-submodule)
 ;;; magit-submodule.el ends here
--- a/lisp/magit-submodule.el
+++ b/lisp/magit-submodule.el
@@ -123,6 +123,7 @@ an alist that supports the keys `:right-align' and `:pad-right'."
         magit-submodule-synchronize)
     (?d "Unpopulate     git submodule deinit [--force]"
         magit-submodule-unpopulate)
+    (?k "Remove" magit-submodule-remove)
     nil
     (?l "List all modules"  magit-list-submodules)
     (?f "Fetch all modules" magit-fetch-modules))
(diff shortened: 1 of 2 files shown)


manateelazycat:
@tarsius 

My implementation of magit-submodule-remove https://github.com/magit/magit/pull/3583 will do four things:
1. Remove the submodule entry from .git/config
2. Delete the submodule entry from .gitmodules file.
3. Delete submodule directory.
4. Delete submodule under .git/modules/ directory.

Your version just do 2 and 3.

I mean, can you provide a option make user can choose clean from .git/config and .git/modules ?
It's not handy if user still need do 1 and 4 manually after execute magit-submodule-remove.

Let user choose, if user know what he did, clean from .git/config and .git/modules.

tarsius:
I didn't realize it didn't remove from `.git/config`. I'll add that.

I am considering using the prefix argument to allow removing the gitdir.

manateelazycat:
@tarsius 

Your implementation of magit-submodule-remove have two bug.

```Elisp
(dolist (module modules)
          ;; Disregard if `magit-delete-by-moving-to-trash'
          ;; is nil.  Not doing so would be to dangerous.
          (delete-directory module nil t))
```

1. In above code module is submodule name under git root, not the submodule path under .git/modules.
2. We should make third argument of delete-directory with t, and not nil, otherwise delete-directory can't delete directory recursively.

Below is my version to delete .git/modules/submode:

```Elisp
(delete-directory (concat default-directory
                                    (file-name-as-directory ".git")
                                    (file-name-as-directory "modules")
                                    module)
                            t t)
```

BTW, can you provide a customize option that i don't need press C-u before call `magit-submodule-remove' ?

Thanks for your great work!

 -- Andy

tarsius:
Fixes. (I still haven't test this at all.)

> BTW, can you provide a customize option that i don't need press C-u before call `magit-submodule-remove' ?

What do others thing?

manateelazycat:
> Fixes. (I still haven't test this at all.)
> 
> > BTW, can you provide a customize option that i don't need press C-u before call `magit-submodule-remove' ?
> 
> What do others thing?

@tarsius

Last version still have bug.

```
(delete-directory
             (magit-git-dir (convert-standard-filename
                             (concat "modules/" (cdr (assoc module alist)))))
             t t)
```

Should change to:
```
(delete-directory
             (magit-git-dir (convert-standard-filename
                             (concat "modules/" module)))
             t t)
```

Because (cdr (assoc module alist)) is nil, then `magit-submodule-remove' will remove .git/modules directory, include all submodules in current git repository!

tarsius:
> What do others thing?

I meant "What do others think?". Asking for opinions of others on whether to add an option or not; not how to do it.

tarsius:
> Should change to:

No. `module` is a path not a name. We need a name here.

> Because (cdr (assoc module alist)) is nil,

Which should never be the case, but see the TODO related to `alist`'s creation.

manateelazycat:
> > Should change to:
> 
> No. `module` is a path not a name. We need a name here.
> 
> > Because (cdr (assoc module alist)) is nil,
> 
> Which should never be the case, but see the TODO related to `alist`'s creation.

Do you have test it?

I use magit version, (cdr (assoc module alist)) is nil.

tarsius:
> Do you have test it?

Not yet.

manateelazycat:
> > Do you have test it?
> 
> Not yet.

Delete gitdir path is still not right, you should test it.

I hope finish this will merge those two feature in master, then i can use those two commands everyday.

tarsius:
Also that bug was closely related to the TODO. It turned out that I could not use the null character as separator and so I added the TODO, but I did not replace it with a newline.

I am going to let this rest for a few days now.

vermiculus:
> Fixes. (I still haven't test this at all.)
> 
> > BTW, can you provide a customize option that i don't need press C-u before call `magit-submodule-remove' ?
> 
> What do others think?

My two cents would be that introduces unnecessary complication. There are other ways of achieving this at the user level.

manateelazycat:
@tarsius 

magit-submodule-add also have bug, below is Backtrace from emacs:
```
Debugger entered--Lisp error: (void-variable path)
  (magit-call-git "submodule" "absorbgitdirs" path)
  (if (version< (magit-git-version) "2.12.0") nil (magit-call-git "submodule" "absorbgitdirs" path))
  (if (> (process-exit-status process) 0) (magit-process-sentinel process event) (process-put process 'inhibit-refresh t) (magit-process-sentinel process event) (if (version< (magit-git-version) "2.12.0") nil (magit-call-git "submodule" "absorbgitdirs" path)))
  (progn (if (> (process-exit-status process) 0) (magit-process-sentinel process event) (process-put process 'inhibit-refresh t) (magit-process-sentinel process event) (if (version< (magit-git-version) "2.12.0") nil (magit-call-git "submodule" "absorbgitdirs" path))))
  (if (memq (process-status process) '(exit signal)) (progn (if (> (process-exit-status process) 0) (magit-process-sentinel process event) (process-put process 'inhibit-refresh t) (magit-process-sentinel process event) (if (version< (magit-git-version) "2.12.0") nil (magit-call-git "submodule" "absorbgitdirs" path)))))
  (lambda (process event) (if (memq (process-status process) '(exit signal)) (progn (if (> (process-exit-status process) 0) (magit-process-sentinel process event) (process-put process 'inhibit-refresh t) (magit-process-sentinel process event) (if (version< (magit-git-version) "2.12.0") nil (magit-call-git "submodule" "absorbgitdirs" path))))))(#<process git> "finished\n")
```

tarsius:
> magit-submodule-add also have bug,

No, `path` is captured in a closure.

> below is Backtrace from emacs:

Did you copy the code into a buffer where `lexical-binding` is `nil` before evaluating it? In that case no closure is created and `path` is indeed unbound.

manateelazycat:
> > magit-submodule-add also have bug,
> 
> No, `path` is captured in a closure.
> 
> > below is Backtrace from emacs:
> 
> Did you copy the code into a buffer where `lexical-binding` is `nil` before evaluating it? In that case no closure is created and `path` is indeed unbound.

I just use magit-submodule-add as M-x command, everything is fine, but last expression of magit-submodue-add will failed:
```
(unless (version< (magit-git-version) "2.12.0")
             (magit-call-git "submodule" "absorbgitdirs" path))
```

Submodule is add sucess though there have (void-variable path) error.

What's is (magit-call-git "submodule" "absorbgitdirs" path) do ? It's necessary for add submodule?

tarsius:
> (void-variable path) error.

I can only think of one reason why that could happen, which is why I asked:

> Did you copy the code into a buffer where `lexical-binding` is `nil` before evaluating it?

manateelazycat:
> > (void-variable path) error.
> 
> I can only think of only one reason why that could happen, which is why I asked:
> 
> > Did you copy the code into a buffer where `lexical-binding` is `nil` before evaluating it?

Can you please fix magit-submodule-remove and merge this PR first? 

magit-submodule-add not a big problem for me. ;)

I can help test after those two commands merge in magit master. ;)

tarsius:
> > > Did you copy the code into a buffer where `lexical-binding` is `nil` before evaluating it?

Still no answer.

manateelazycat:
> > > > Did you copy the code into a buffer where `lexical-binding` is `nil` before evaluating it?
> 
> Still no answer.

I have function
```
(defun magit-submodule-add+ (url)
  (interactive "sURL: ")
  (magit-submodule-add
   url
   (concat (file-name-as-directory lazycat-emacs-extension-dir) (file-name-base url))
   (file-name-base url)))
```

It wrap magit-submodule-add

tarsius:
> > > > > Did you copy the code into a buffer where `lexical-binding` is `nil` before evaluating it?
> > 
> > Still no answer.
>
> I have a function

That is not an answer to my question.

Here is how to answer a question:

> It wrap magit-submodule-add, is cause by this?

No, I do not think that this could possibly cause the observed behavior.

manateelazycat:
> > > > > > Did you copy the code into a buffer where `lexical-binding` is `nil` before evaluating it?
> > > 
> > > 
> > > Still no answer.
> > 
> > 
> > I have a function
> 
> That is not an answer to my question.
> 
> Here is how to answer a question:
> 
> > It wrap magit-submodule-add, is cause by this?
> 
> No, I do not think that this could possibly cause the observed behavior.

I don't know why path is void, haven't idea now. :(

lisp/magit-submodule.el
```diff
@@ -270,6 +295,80 @@ single module from the user."
...
+          (cl-find-if (lambda (module)
+                        (let ((default-directory (file-name-as-directory
+                                                  (expand-file-name module))))
+                          (magit-anything-modified-p)))
+                      modules)))
+      (if (member "--force" args)
+          (if (magit-confirm 'remove-dirty-modules
+                "Remove dirty module %s"
+                "Remove %i dirty modules"
+                t modules)
```
kyleam: Should `modules` be `modified`?

PR #3587, opened 2018-09-14
magit--github-url-p: fix a bug in github enterprise handling

### LINKED ISSUE #3572: Impossible to use github pull-request commands with github enterprise
version: Magit 20180815.2257, Git 2.18.0, Emacs 25.1.1, gnu/linux

I tried to configure magit to work with github enterprise today. The
goal is to get functions like `magit-branch-pull-request` to work. The
configuration I wrote is based on those two pages:

- https://magit.vc/manual/ghub/Getting-Started.html#Getting-Started
- https://mrdias.com/2018/04/01/using-magit-with-github-enterprise.html

The instance of github enterprise is on a domain like git.foo.com.

My git configuration is the following:

```
[github]
user = louis
host = "git.foo.com"
[github "git.foo.com"]
user = louis
```

My `~/.authinfo` info file:

```
machine git.foo.com/api/v3 login louis^magit password ...
machine git.foo.com login louis^magit password ...
machine api.github.com login khady^magit password ...
```

I open a file in the git repository and then execute M-x
magit-branch-pull-request or b Y from the magit status buffer. It
prints a long html error in messages. The interesting parts are probably:

```
mapcar: Wrong type argument: listp, "<!DOCTYPE html>
[...]
    You can't perform that action at this time.
[...]
```

`toggle-debug-on-error` tells me it comes from `magit-read-pull-request`.

In the blog post above, the person has `/api/v3` at the end of the
host. So I tried this too.

```
[github]
user = louis
host = "git.foo.com/api/v3"
[github "git.foo.com/api/v3"]
user = louis
```

This time `magit-branch-pull-request` tells me `Currently only Github
is supported` and `b Y` tells me `magit-upstream-repository: Currently
only Github is supported`.

I traced this message to `magit--github-url-p`.

If I change this function to `(defun magit--github-url-p (url) t)`,
then `magit-branch-pull-request` works.

The url parameter of this function is something like
`git@git.foo.com:foo/project.git`. The host is `git.foo.com`. Which
doesn't match `github.com` or `(ghub--host)` (that is
`"git.foo.com/api/v3"`).

Is it considered a bug or I did something wrong with my configuration?

I'm sorry if this bug report is not written in a very formal way and doesn't 
respect the expected format. Thanks for all your work on magit.

Khady:
The `separator' argument of `split-string' should be in second
position.

Fixes #3572

### DIFF
--- a/lisp/magit-collab.el
+++ b/lisp/magit-collab.el
@@ -134,7 +134,7 @@ exist, then raise an error."
            ;; in false-positives, but that's rather unlikely.  #3392
            (and (or (string-match-p (regexp-quote "github.com") host)
                     (string-match-p
-                     (regexp-quote (car (split-string "/" (ghub--host))))
+                     (regexp-quote (car (split-string (ghub--host) "/")))
                      host))
                 host)))))
 


tarsius:
Thanks!

PR #3595, opened 2018-09-21
magit-display-buffer: Don't reselect the same frame

### LINKED ISSUE #3330: Possible bug with mouse-autoselect-window and magit-log-current
With

- Magit 2.11.0, Git 2.15.1, Emacs 25.2.2, gnu/linux (Debian elpa-magit 2.11.0-1)
- Emacs 25 (Debian emacs25 and emacs25-lucid 25.2+1-6)
- i3 4.14.1 (Debian 4.14.1-1)

and

    (setq mouse-autoselect-window t)

if I split the Emacs frame via `C-x 3` and then run `M-x magit-status` from the right window, so that the status shows up in the left window, and then hit `l l` in the status window, the pointer jumps to the top-right corner of the status window, and loses focus.  That's of course awkward since subsequent movement commands, etc. will be sent to the wrong window.

Someone else (with a very similar configuration) was able to reproduce the same thing after setting the autoselect variable.

Please let me know if I can help diagnose the problem, and thanks much
for Magit.

npostavs:
Fixes #3330 (see commit message).

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -619,9 +619,11 @@ and `magit-post-display-buffer-hook'."
     (run-hooks 'magit-pre-display-buffer-hook))
   (let ((window (funcall magit-display-buffer-function buffer)))
     (unless magit-display-buffer-noselect
-      (select-frame-set-input-focus
-       (window-frame
-        (select-window window)))))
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks!

PR #3609, opened 2018-10-04
magit-mode: Set the repository-local mode-line-process value

phil-s:
Previously only pre-existing magit buffers for the repository would indicate the current Git process; new magit buffers created while the process was running would not display that information.

`mode-line-process` is `nil` by default, and `(magit-repository-local-get 'mode-line-process)` returns `nil` if there is no such value, or if `magit-process-unset-mode-line` has explicitly set the value to `nil`.

It's my guess that `magit-mode` is the sensible place to do this.


phil-s:
I should *possibly* be calling `force-mode-line-update` after doing that, as in https://github.com/magit/magit/pull/3314/commits/0c85788ff5c61d28d775017a76febd08a02d0ee9 , but I'm not sure that's *actually* needed here?

tarsius:
> It's my guess that magit-mode is the sensible place to do this.

Sound good.

> I should possibly be calling force-mode-line-update after doing that, as in 0c85788 , but I'm not sure that's actually needed here?

Starting a long running process and then creating a new buffer should tell you whether it is necessary or not.

phil-s:
It certainly all looked fine in my tests (n.b. initiating a commit is probably the easiest way to create a long-running Git process for testing).

tarsius:
Thanks!

Issue #3616, opened 2018-10-10
Can't compile magit 


sidhartharya10:
I am using debian linux 4.9 and i am not able to compile magit from git repository. 
It seems to be asking for graphql???
make error:
```
magit-status.el:30:1:Error: Cannot open load file: No such file or directory, graphql
```
`$ make emacs-Q`
```
Debugger entered--Lisp error: (file-missing "Cannot open load file" #("No such file or directory" 0 25 (charset iso-8859-1)) "graphql")
  require(graphql)
  eval-buffer(#<buffer  *load*-341553> nil "/home/arya/.emacs.d/packages/magit/ghub/ghub-graph..." nil t)  ; Reading at buffer position 870
  load-with-code-conversion("/home/arya/.emacs.d/packages/magit/ghub/ghub-graph..." "/home/arya/.emacs.d/packages/magit/ghub/ghub-graph..." nil t)
  require(ghub-graphql)
  eval-buffer(#<buffer  *load*-217132> nil "/home/arya/.emacs.d/packages/magit/ghub/ghub.el" nil t)  ; Reading at buffer position 40199
  load-with-code-conversion("/home/arya/.emacs.d/packages/magit/ghub/ghub.el" "/home/arya/.emacs.d/packages/magit/ghub/ghub.el" nil t)
  require(ghub)
  eval-buffer(#<buffer  *load*-800398> nil "/home/arya/.emacs.d/packages/magit/magit/lisp/magi..." nil t)  ; Reading at buffer position 1434
  load-with-code-conversion("/home/arya/.emacs.d/packages/magit/magit/lisp/magi..." "/home/arya/.emacs.d/packages/magit/magit/lisp/magi..." nil t)
  require(magit-collab)
  eval-buffer(#<buffer  *load*> nil "/home/arya/.emacs.d/packages/magit/magit/lisp/magi..." nil t)  ; Reading at buffer position 24160
  load-with-code-conversion("/home/arya/.emacs.d/packages/magit/magit/lisp/magi..." "/home/arya/.emacs.d/packages/magit/magit/lisp/magi..." nil t)
  require(magit)
  (progn (setq debug-on-error t) (require 'magit) (global-set-key "\30g" 'magit-status))
  eval((progn (setq debug-on-error t) (require 'magit) (global-set-key "\30g" 'magit-status)))
[...]

PR #3617, opened 2018-10-10
Documentation: graphql and treepy reference.

sidhartharya10:
Added reference to graphql and treepy in the Documentation. #3616


tarsius:
I've pushed 21f786a15405414e913af0a201e4e6e609ee244c instead.

PR #3619, opened 2018-10-11
Rename regular commands, freeing their names for transient commands

tarsius:
Once I replace `magit-popup` with `transient`, I am going to shorten the names of the "popup commands" (going forward known as "transient" or "prefix" commands).

For example I am going to rename `magit-tag-popup` to just `magit-tag`. But there already is a command named `magit-tag`, so that has to be renamed first. In this case the new name is easy to find: `magit-tag-create`. Even if I didn't want to free the old name so that it could be used for the transient command, the new name is better. All the other (suffix) commands that are invoked from the popup are already called `magit-tag-VERB`.

Unfortunately coming up with a new name is not always as easy and that's why I am opening this pull-request. Does anyone have suggestions for better names than those I have chosen? Look especially at the commits prefixed with `[*]`.

I intend to merge this soon, *before* switching to `transient`. That package is still work in progress and I am testing it by porting existing popups on a feature branch. By performing the renames now on `master` I will have to deal less with conflicts when I rebase that feature branch.


kyleam:
> Does anyone have suggestions for better names than those I have chosen? Look especially at the commits prefixed with `[*]`.

`magit-revert-do-commit`: another option is `magit-revert-and-commit`.  I have a slight preference for `-and-commit` because I'd be likely to read the `commit-do` combination as "'commit' is the namespace, do is the 'verb'".

`magit-reset-quick`: perhaps `magit-reset-quickly`, in line with `magit-gitignore-{locally,globally}`.

`magit-merge-plain`: I thought about `magit-merge-and-commit`, but I think `-plain` is better because commit on merge is the normal/default case.

tarsius:
Thanks Kyle.

I've also defined obsolete-function aliases.

Packages such as `evil-magit` should continue to use the old names until 2.90.0 is released. After that they should depend on that release and use the new names. In 2.91.0 or 2.92.0 the aliases will be removed and the old names will be used for the new transient commands.

/cc @justbur

PR #3622, opened 2018-10-15
Use libgit if possible and desired

tarsius:
See the commit messages.

This is so complicated and ugly because we have to do the following.

* Because `libgit` cannot be used when accessing remote repositories using Tramp, we have to keep the old git-base implementations, add new libgit-based implementations, and teach the former to dispatch to the latter if appropriate.
* Because we have to keep the git-based implementation anyway, we give users the option to stick to that even when accessing local repositories. (That makes it easier for users and maintainers alike. Otherwise users would be forged to compile Emacs themselves if the official build for their distribution/os was build without module support. Since `libgit` is optional, we can leave it to users to help each other, without taking anything away from anybody.)
* Because functions defined in `magit-git` need to dispatch to functions that are defined in `magit-libgit`, and functions defined in `magit-libgit` need to call low-level functions defined in `magit-git`, we have a circular dependency. This makes it necessary to declare many functions and variables. It should be possible to reduce the interdependence, by splitting `magit-git` into two libraries, but I strongly doubt that it can be reduced to zero.
* Because building the module may fail, we don't want to load `libgit` unless we know that the module already exists, because loading `libgit` causes `libegit2` to be build. It might be better to rethink `libgit` instead of forcing `magit-libgit` to work around how `libgit` does things.
* ...


luismbo:
How about using two classes to model the backends and generic functions to model each operation. Something along these lines should break the circular dependency.

```elisp
(require 'eieio)

(defclass magit-git-backend () ())
(defclass magit-libgit-backend (magit-git-backend) ())

(defvar magit-git-backend (make-instance 'magit-git-backend))
(defvar magit-libgit-backend (make-instance 'magit-git-backend))

(defmacro defgitfun (name args &optional docstring)
  (let ((gf (intern (format "%s-gf" name))))
    `(progn
       (defgeneric ,gf ,(cons (gensym "backend") 'args) ,docstring)
       (defun ,name ,args
         ,docstring
         (,gf (if (magit-use-libgit-p)
                  magit-libgit-backend
                  magit-git-backend)
              ;; or however we're supposed to parse lambda lists...
              ,@ (cl-set-difference args '(&optional &rest)))))))

(defgitfun magit-bare-repo-p (&optional noerror)
  "Return t if the current repository is bare.
If it is non-bare, then return nil.  If `default-directory'
isn't below a Git repository, then signal an error unless
NOERROR is non-nil, in which case return nil.")

(cl-defmethod magit-bare-repop-gf :around (be &optional noerror)
  (and (magit--assert-default-directory noerror)
       (call-next-method)))

(cl-defmethod magit-bare-repo-p-gf ((be magit-git-backend) &optional noerror)
  (condition-case nil
      (magit-rev-parse-true "--is-bare-repository")
    (magit-invalid-git-boolean
      (and (not noerror)
           (signal 'magit-outside-git-repo default-directory)))))

(cl-defmethod magit-bare-repo-p-gf ((be magit-libgit-backend) &optional noerror)
  (if-let ((repo (magit-libgit-repo)))
      (libgit-repository-bare-p repo)
    (unless noerror
      (signal 'magit-outside-git-repo default-directory))))```

In this scenario magit-libgit-backend methods can give up and fall back to the git backend by calling `cl-call-next-method`, plus the ability to share behaviour like the around method in this example.

There are many various on this theme with more or less macros and more or less eieio.

**Comment by luismbo** [comment]
@tarsius this would also open way for an hypothetical third backend based on some sort of RPC on top of libgit that's being discussed in issue #2959. Not that maintaining three backends would be an ideal situation. 😅

**Comment by tarsius** [comment]
@luismbo That's an interesting idea. It might even be possible use a `&context` specializer, making the additional classes, `defgitfun`, and the `-gf` suffix unnecessary:

```lisp
(cl-defmethod magit-bare-repo-p (&context (magit libgit) &optional noerror)
  ...)

(cl-defmethod magit-bare-repo-p (&context (magit git) &optional noerror)
  ...)
```

To make that possible we would have to write a new generalizer and a new context rewriter based on the value of `magit-use-libgit-p`, similar to how that is implemented for `derived-mode` at the end of `cl-generic.el`.

You you like to have a go at this? using the implementation of the `major-mode` generalizer as an example you should be able to implement this.

luismbo:
@tarsius Oh, that's pretty cool. I didn't know about that EIOIO feature. I'll look into it.

tarsius:
@luismbo I've started working on it.

luismbo:
Interesting. Where did it go? :-)

PR #3627, opened 2018-10-23
Support log --left-right output

Gravemind:
Magit currently doesn't like the output of `--left-right` (as in `git log --graph --oneline --boundary --left-right @...@{u}`).

For example in magit: magit-status log other *`l` `o` `--graph --boundary --left-right a1b8baa99c...e8fcd70e79`*

I think Magit doesn't display the log properly only because left-right changes the `*` graph marker to `<` or `>`.

I found 3 regexps that seemed to match the graph marker (I looked for the `*` and the `o` of `--boundaries`), then and I just added `<` and `>`. I don't know much magit source code, so **I don't know if I missed something**.

(If the feature is approved but this PR inadequate, I'll close it and create a more general Issue feature request)

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -943,7 +943,11 @@ Do not add this to a hook variable."
          (remove "--literal-pathspecs" magit-git-global-arguments)))
     (magit-git-wash (apply-partially #'magit-log-wash-log 'log)
       "log"
-      (format "--format=%%h%s%%x00%s%%x00%%aN%%x00%%at%%x00%%s%s"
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks!

I had to add to add a second commit to support `--left-right` without `--graph`.

Github got really confused when I pushed the rebased commits to `master` before I pushed them to this feature branch. So it doesn't list them anymore.

kyleam:
Nice.  I'm happy to see this since I drop into a terminal occasionally for `--left-right`.

PR #3635, opened 2018-10-27
Add default directory option for magit-clone

kevinjfoley:
Adds option to specify default directory for `magit-clone`.  I usually clone all of my repos to the same base repository and rarely to the directory I'm currently in.  Adding this option allows the user to specify which directory they'd like to serve as the default if any.


tarsius:
I did something similar in 5029811bc005c3add0a582f61f963f3881e7a565.

kevinjfoley:
Thanks! I didn't even think of allowing for a function but can already think of uses for that feature.

Issue #3611, opened 2018-10-06
magit-emacs-Q-command is missing graphql and treepy


CyberShadow:
In a new directory, run `HOME=$PWD emacs -Q`. There, evaluate:

```elisp
(progn
  (require 'package)
  (add-to-list 'package-archives
               '("melpa" . "http://melpa.org/packages/") t)
  (package-initialize)
  (package-refresh-contents)
  (package-install 'magit)

  (magit-emacs-Q-command))
```

Here is what this puts in my kill-ring:

```
/usr/bin/emacs -Q --eval \(setq\ debug-on-error\ t\)
-L /home/vladimir/tmp/2018-10-06b/.emacs.d/elpa/magit-20181004.1127/
-L /home/vladimir/tmp/2018-10-06b/.emacs.d/elpa/magit-popup-20181003.921/
-L /home/vladimir/tmp/2018-10-06b/.emacs.d/elpa/with-editor-20180726.2044/
-L /home/vladimir/tmp/2018-10-06b/.emacs.d/elpa/git-commit-20180912.1012/
-L /home/vladimir/tmp/2018-10-06b/.emacs.d/elpa/dash-20180910.1856/
-L /home/vladimir/tmp/2018-10-06b/.emacs.d/elpa/ghub-20181003.1456/
-l /home/vladimir/tmp/2018-10-06b/.emacs.d/elpa/magit-20181004.1127/magit
```

And here is what's in the newly-created `~/.emacs.d/elpa`:

```
archives
async-20180527.1730
dash-20180910.1856
ghub-20181003.1456
git-commit-20180912.1012
gnupg
graphql-20180912.31
magit-20181004.1127
magit-popup-20181003.921
treepy-20180724.656
with-editor-20180726.2044
```

It looks like there are three packages that were installed alongside magit, but are not present in the output of `magit-emacs-Q-command` (`async`, `graphql`, and `treepy`).

Running the command results in the following error:

```
Debugger entered--Lisp error: (file-missing "Cannot open load file" "No such file or directory" "graphql")
  require(graphql)
  byte-code("\300\301!\210\300\302!\210\300\303!\210\300\304!\210\300\305!\207" [require dash ghub graphql subr-x treepy] 2)
  require(ghub-graphql)
  byte-code("\300\301!\210\302\303!\207" [provide ghub require ghub-graphql] 2)
  require(ghub)
  byte-code("\300\301!\210\300\302!\207" [require magit ghub] 2)
  require(magit-collab)
[...]

PR #3637, opened 2018-10-28
magit-emacs-Q-command: Find deps automatically

npostavs:
This should prevent a problem like #3611 from occuring in the future when one of magit's dependent packages changes its dependencies again.

One drawback, it does include `async` (if it has been `require`d by loading `git-rebase`). We could add a special case to exclude it.

EDIT: it occurs to me that this might not work if called before some packages have been loaded yet.
EDIT2: maybe we should specify magit's immediate dependencies explicitly, load them, and then crawl their dependencies automatically?

### DIFF
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -714,43 +714,49 @@ DEFAULT may be a number or a numeric string."
 
 ;;; Debug Utilities
 
+(declare-function file-requires "loadhist" (file))
+(declare-function feature-file "loadhist" (feature))
+
 ;;;###autoload
 (defun magit-emacs-Q-command ()
   "Show a shell command that runs an uncustomized Emacs with only Magit loaded.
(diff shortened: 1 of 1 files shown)


tarsius:
> EDIT: it occurs to me that this might not work if called before some packages have been loaded yet.
> EDIT2: maybe we should specify magit's immediate dependencies explicitly, load them, and then crawl their dependencies automatically?

If we go down that road, then yes we would have to use a mixture of explicit and auto-detected dependencies. I am a bit skeptical all this would win us much. We could just try to remember, and if we forget then we'll be told and fix it. If on the other hand `magit-emacs-Q-command` itself had a bug, and a change like this increases the odds, then noticing what the issue was would become more difficult I think.

npostavs:
> I am a bit skeptical all this would win us much.

Yeah, the automatic method could miss conditional dependencies in dependent packages as well, so it's not reliable enough to be worth it.

PR #3638, opened 2018-10-29
magit-reverse-files: Extend binary file check

### LINKED ISSUE #3625: Reversing binary files doesn't seem to work
`Magit 20180915.659, Git 2.19.1, Emacs 26.1, gnu/linux`

What I did: 

  - opened the commit log (`l l`)
  - selected a commit (`return`)
  - moved point to a binary file that was deleted by that commit (it was an `.ogg` file)
  - pressed `v`

Output:

```
  1 git … apply --reverse --reject -p0 --ignore-space-change -
Checking patch myfile...
error: cannot apply binary patch to 'myfile' without full index line
error: myfile: patch does not apply
```

kyleam:
Provide a more informative error message when a user tries to reverse binary files from a revision or diff buffer.

Closes #3625.

### DIFF
--- a/lisp/magit-apply.el
+++ b/lisp/magit-apply.el
@@ -644,7 +644,14 @@ so causes the change to be applied to the index as well."
 
 (defun magit-reverse-files (sections args)
   (pcase-let ((`(,binaries ,sections)
-               (let ((bs (magit-binary-files "--cached")))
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks!

PR #3641, opened 2018-10-29
Use libgit if possible and desired (r2)

tarsius:
This replaces #3622.

This uses generic functions to prevent `magit-git.el` and `magit-libgit.el` depending on one another. Thanks @luismbo for the suggestion.

/cc @TheBB

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -54,6 +54,72 @@
 
 (defvar magit-tramp-process-environment nil)
 
+;;; Git implementations
+
+(defconst magit--git-implementations '(git libgit))
+
+(defvar magit-inhibit-libgit nil)
+
+(defvar magit--libgit-available-p eieio-unbound)
+
+(defun magit--libgit-available-p ()
+  (if (eq magit--libgit-available-p eieio-unbound)
+      (setq magit--libgit-available-p
+            (and module-file-suffix
+                 (let ((libgit (locate-library "libgit")))
+                   (and libgit
+                        (or (locate-library "libegit2")
+                            (let ((load-path
+                                   (cons (expand-file-name
+                                          (convert-standard-filename "build")
+                                          (file-name-directory libgit))
+                                         load-path)))
+                              (locate-library "libegit2")))))))
+    magit--libgit-available-p))
+
+(defun magit--git-implementation ()
+  (if (and (not magit-inhibit-libgit)
+           (not (file-remote-p default-directory))
+           (magit--libgit-available-p))
+      'libgit
+    'git))
+
+;; We want to write (&context (gitimpl IMPL)), which is only possible
+;; if a variable `gitimpl' exists.  We don't use it for anything but
+;; satisfying that requirement.  `cl-generic' assumes the use of a
+;; variable, but we use a function to determine the context.
+(with-no-warnings (defvar gitimpl nil))
+(defun cl--generic-gitimpl-tagcode (varsym &rest _)
+  `(progn ,varsym ; silence byte-compiler
+          (let ((val (magit--git-implementation)))
+            (and (memq val magit--git-implementations) val))))
+
+(defun cl--generic-gitimpl-generalizers (tagval &rest _)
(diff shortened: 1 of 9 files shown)


luismbo:
@tarsius this looks great. Thanks for doing the hard work of following up my suggestion.

lisp/magit-git.el
```diff
@@ -48,6 +48,72 @@
...
+    magit--libgit-available-p))
+
+(defun magit--git-implementation ()
+  (if (and (not magit-inhibit-libgit)
+           (not (file-remote-p default-directory))
+           (magit--libgit-available-p))
+      'libgit
+    'git))
+
+;; We want to write (&context (gitimpl IMPL)), which is only possible
```
npostavs: Um, this seems like a lot of work just to avoid writing `(&context ((magit-git-impl) (eql IMPL)))`

tarsius: It's twenty lines...

It's unfortunate that this little hack wtih the `gitimpl` variable is needed but I don't think that's all that bad either. What's wrong is basically the lines 85 and 87.

npostavs: Okay, your call, but it hardly seems worth saving 15 characters per defmethod to me (and presumably similar savings could be had by defining a macro that expands to `cl-defmethod`).

fice-t: I find this part somewhat confusing, myself.

Perhaps I'm misinterpreting your statement, but your comment states "`cl-generic' assumes the use of a variable", which doesn't seem to be the case. [Per the manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Generic-Functions.html):

> Alternatively, the argument specializer can be of the form &context (expr spec), in which case the value of expr must be compatible with the specializer provided by spec; spec can be any of the forms described above. In other words, this form of specializer uses the value of expr instead of arguments for the decision whether the method is applicable.

So it can be an arbitrary expression. As an example:

```elisp
(cl-defgeneric test ())
(cl-defmethod test (&context ((magit--git-implementation) (eql git)))
  (message "Just git"))

(cl-defmethod test (&context ((magit--git-implementation) (eql libgit)))
  (message "libgit!"))

;; Force a 'git result
(cl-letf (((symbol-function 'magit--git-implementation)
	   (lambda () 'git)))
  (test))

;; Force a 'libgit result
(cl-letf (((symbol-function 'magit--git-implementation)
	   (lambda () 'libgit)))
  (test))
```

I apologize if I got it wrong, but I believe the following is a simpler solution. Though it doesn't handle certain edge cases (e.g., using both `&gitimpl` and `&context`). Adjust the macro interface to taste:

```elisp
(defmacro magit-defmethod (name spec-lambda-list &rest body)
  (declare (indent defun))
  ;; &gitimpl impl -> &context ((magit--git-implementation (eql impl)))
  (when-let ((impl-expr (memq '&gitimpl spec-lambda-list)))
    (setf (cl-first  impl-expr) '&context)
    (setf (cl-second impl-expr) `((magit--git-implementation)
				  (eql ,(cl-second impl-expr)))))
  `(cl-defmethod ,name ,spec-lambda-list ,@body))

(magit-defmethod test (&gitimpl git)
  (message "Just git"))

(magit-defmethod test (&gitimpl libgit)
  (message "libgit!"))
```

It also might be nice to put the `cl-defgeneric` into this macro, so that the definition isn't split up into 2 top-level forms. One would have to take care of both the `cl-defgeneric` and `cl-defmethod` docstrings in the interface if having a separate method docstring is desired, though.

tarsius: Yes I was wrong about `cl-generic` expecting a variable, I think its only the `eql` and `head` specializers but I might be wrong about that too.

The macro thingy is something I am considering. I haven't made up my mind yet.

npostavs: > **EDIT**: I'm not sure how I didn't notice that @npostavs already used a procedure as a context

Probably because I was too terse and my exampe too abbreviated. But you've explained exactly what I had mind. :)

PR #3643, opened 2018-10-31
magit-init: New separate-gir-dir optional parameter

SaGagnon:
Hi,

I'd love to have support for the `--separate-git-dir` flag in `magit-init` as I use it very often. I decided to make a patch instead of a feature request because this is a minor addition. By the way, thank you guys for this great package!

Samuel


SaGagnon:
Hi kyleam,

After reading your response, I feel you are right; the use case is too specific. Thank you for taking the time to suggests alternatives, I didn't know you could simply reinitialize a repo, so advising `magit-init` should do the trick.

Samuel

PR #3649, opened 2018-11-04
Require subr-x at compile time

jabranham:
If/when-let are macros that can be expanded at compile time rather
than at run time

### DIFF
--- a/lisp/magit-apply.el
+++ b/lisp/magit-apply.el
@@ -30,6 +30,9 @@
 
 ;;; Code:
 
+(eval-when-compile
+  (require 'subr-x))
+
 (require 'magit-core)
 (require 'magit-diff)
 (require 'magit-wip)
--- a/lisp/magit-blame.el
+++ b/lisp/magit-blame.el
@@ -28,6 +28,9 @@
 
 ;;; Code:
 
+(eval-when-compile
+  (require 'subr-x))
+
 (require 'magit)
 
 ;;; Options
--- a/lisp/magit-bookmark.el
+++ b/lisp/magit-bookmark.el
@@ -27,6 +27,9 @@
 
 ;;; Code:
 
+(eval-when-compile
+  (require 'subr-x))
+
 (require 'magit)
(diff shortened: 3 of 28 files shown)


tarsius:
Yes, but they are also used in more files than just `magit-utils.el`. Run `make clean lisp` -- it's not pretty.

I don't think this is worth it. `(eval-when-compile (require 'subr-x))` would to be added to every library, including the once that don't currently use any of the macros. If we didn't do that, then I would sooner or later start using one of the macros without also adding the `require` and not notice an issue (because I have `subr-x` loaded at run-time, so `auto-compile` would not report an issue).

jabranham:
> Run make clean lisp -- it's not pretty.

Yikes! I should've seen that. 

> I don't think this is worth it. 

It makes sense to me to do this work at compile time rather than runtime since we can. But if you don't want to, that's fine too. If you want to, I can add the requirement to all the other files, just let me know. If not, feel free to just close this PR.

tarsius:
Did you add the `require` to all libraries? If so, then please split this into to commits (a) libraries that need this now and (b) libraries that might need it in the future. Also please update the commit message to the one I used elsewhere.

lisp/git-commit.el
```diff
@@ -120,7 +120,8 @@
 (require 'server)
 (require 'with-editor)
 
-(eval-when-compile (require 'recentf))
+(eval-when-compile (require 'recentf)
```
tarsius: Please insert a newline after `(eval-when-compile` *especially* if more than one form follow.

tarsius:
I have merged the first commit into `master` and have pushed the second commit to `origin/subr-x` for reference.

In the first commit I moved the new require form in most cases. `subr-x` should be required next to `cl-lib` and `dash` and like those libraries it should be required before `magit...`. Those are the conventions I try to follow, though there are already some exceptions to the latter.

jabranham:
Makes sense. I was wondering why you wanted separate commits!

PR #3652, opened 2018-11-08
Clarify and implement other interpretation of "global" in magit-gitignore

CyberShadow:
I was confused that, after updating Magit, I could no longer ignore files using Magit (by asking it to place their paths in `.gitignore`). The only options were to ignore locally (which I knew to mean `info/exclude`), and globally (which I understood to have the same general meaning as `git config --global`, i.e. whatever `core.excludepath` was pointing at).

Here's an attempt to clarify this confusion, plus implement my initial interpretation to further disambiguate by presenting all reasonable options.

I did not rename `magit-gitignore-globally`, to avoid breaking users' configurations.

### DIFF
--- a/lisp/magit-gitignore.el
+++ b/lisp/magit-gitignore.el
@@ -39,7 +39,8 @@
   "Popup console for gitignore commands."
   :man-page "gitignore"
   :actions '((?l "ignore locally"  magit-gitignore-locally)
-             (?g "ignore in .gitignore" magit-gitignore-globally))
+             (?g "ignore in .gitignore" magit-gitignore-globally)
+             (?s "ignore system-wide" magit-gitignore-system))
   :max-action-columns 1)
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


tarsius:
> I did not rename `magit-gitignore-globally`, to avoid breaking users' configurations.

We probably should rename it. Maybe `*-publicly`?

CyberShadow:
Sounds good.

Should I redo this against the `transient` branch?

tarsius:
> Should I redo this against the `transient` branch?

If you are already using that, then yes. Otherwise I would recommend you wait two weeks -- by then this should be merged into `master`.

tarsius:
I have done this on the `transient` branch now in the commits

* Add additional magit-gitignore variants
* Improve magit-gitignore commands

(A lot of rebasing is still going on, so no hashes.)

I'll merge this this month.

PR #3656, opened 2018-11-10
doc typo: magit-revision-show-gravatar→magit-revision-show-gravatars

unhammer:
wrong variable name

(needs .texi; I get `Cannot open load file: No such file or directory, org-man` when I try to make it)


tarsius:
Thanks!

PR #3666, opened 2018-11-17
magit-diff-type: Guard rev-eq call to fix range edge case

kyleam:
```
As of 296f8dc7 (magit-diff-type: Fix case when the HEAD is used as
RANGE, 2018-09-13), we use `magit-rev-eq' to check whether RANGE
resolves to HEAD to avoid classifying `git diff HEAD'-equivalent
output as `committed'.

That fix introduced another edge case.  `magit-rev-eq' calls
`magit-rev-verify' on each argument, which returns the first line of
`git rev-parse --verify'.  In the case of a range A..B, the standard
output spans multiple lines, with first line being B's hash and the
second being "^A-hash".  This means that `(magit-rev-eq "A..B"
"HEAD")', where B resolves to "HEAD", will return true, and
`magit-diff-type' will incorrectly classify the diff as either
`unstaged' or `staged'.

Avoid this misclassification by confirming that RANGE isn't actually a
range before passing it to `magit-rev-eq'.  An alternative would be to
update `magit-rev-eq' so that it doesn't get confused by a range, but
`magit-diff-type' is the only caller that appears to pass it a range.
```

---

I think the fix is obviously correct.  I'm opening a PR in case there are objections about the level at which this should be fixed (e.g., within `magit-rev-eq` itself or even deeper).

### DIFF
--- a/Documentation/RelNotes/2.91.0.org
+++ b/Documentation/RelNotes/2.91.0.org
@@ -101,3 +101,7 @@
   ~magit-commit-instant-fixup~ point was no longer placed on the same
   commit as was at point in the buffer from which the command was
   invoked.  #3674
+
+- ~magit-diff-type~ falsely concluded that a diff buffer showed
+  unstaged changes when diffing a range where the right side resolves
+  to the same commit as ~HEAD~.  #3666
--- a/lisp/magit-branch.el
+++ b/lisp/magit-branch.el
@@ -286,7 +286,7 @@ does."
   (interactive
    (let ((arg (magit-read-other-branch-or-commit "Checkout")))
      (list arg
-           (and (not (magit-rev-verify-commit arg))
+           (and (not (magit-commit-p arg))
                 (magit-read-starting-point "Create and checkout branch" arg)))))
   (when (string-match "\\`heads/\\(.+\\)" arg)
     (setq arg (match-string 1 arg)))
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -1090,7 +1090,7 @@ for a revision."
     (when module
       (setq default-directory
             (expand-file-name (file-name-as-directory module))))
-    (unless (magit-rev-verify-commit rev)
+    (unless (magit-commit-p rev)
       (user-error "%s is not a commit" rev))
     (magit-mode-setup #'magit-revision-mode rev nil args files)))
 
@@ -2122,12 +2122,12 @@ or a ref which is not a branch, then it inserts nothing."
                             (`quicker  ; false negatives (number-less hashes)
                              (and (>= (length text) 7)
                                   (string-match-p "[0-9]" text)
-                                  (magit-rev-verify-commit text)))
+                                  (magit-commit-p text)))
                             (`quick    ; false negatives (short hashes)
                              (and (>= (length text) 7)
-                                  (magit-rev-verify-commit text)))
+                                  (magit-commit-p text)))
                             (`slow
-                             (magit-rev-verify-commit text)))
+                             (magit-commit-p text)))
                       (put-text-property beg (point) 'face 'magit-hash)
                       (let ((end (point)))
                         (goto-char beg)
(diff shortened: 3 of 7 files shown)


tarsius:
I think doing it were you did is correct, but we should also improve `magit-rev-eq`. I have pushed an additional commit to that effect; please have a look.

kyleam:
> `magit-rev-eq' should only return t if A and B refer to the same
> commit and a range never refers to any single commit, so we should
> return nil in that case.  Do that by using `magit-rev-verify-commit'
> instead.

Hmm, it doesn't look like `magit-rev-verify-commit` can ensure that REV isn't a range.

```shell
$ git rev-parse --verify 901ff^..901ff
901ff0bd68422a5c5b728d5c9f497ccfe1c269c5
^262fb1dc2be89f0cd21dce27050bfaf01118c1b2
fatal: Needed a single revision
$ git rev-parse --verify "901ff^..901ff^{commit}"
901ff0bd68422a5c5b728d5c9f497ccfe1c269c5
^262fb1dc2be89f0cd21dce27050bfaf01118c1b2
fatal: Needed a single revision
```

```elisp
(magit-rev-verify-commit "901ff^..901ff")  ; => "901ff0bd68422a5c5b728d5c9f497ccfe1c269c5"
```

My guess is that peeling operators have precedence over range dots, so appending a `{commit}` suffix to a range just dereferences right-hand side.

I hadn't thought of this yesterday, but there's a range that will sneak by the `magit-range-re` guard: Git's newer `REV^-<n>` syntax.  And using `magit-rev-verify-commit` does catch that:


```shell
$ git rev-parse --verify "901ff^-"
901ff0bd68422a5c5b728d5c9f497ccfe1c269c5
^262fb1dc2be89f0cd21dce27050bfaf01118c1b2
fatal: Needed a single revision
$ git rev-parse --verify "901ff^-^{commit}"
fatal: Needed a single revision
```

```elisp
(magit-rev-verify-commit "901ff^-")  ; => nil
```

So, it looks like using `magit-rev-verify-commit` only protects against the most obscure range format :/

> It also return with exit status 128, but `magit-rev-verify' ignores that.

When thinking about this yesterday, checking the status seemed like the right thing to do, but I didn't pursue that because it'd be a deeper change and all callers except `magit-diff-type` seem to respect `magit-rev-verify` expectation of a single revision arguments.  But given that `verify-commit` doesn't seem to guard against standard ranges and that my proposed `magit-diff-type` change doesn't detect `rev^-` ranges, perhaps it's worth reconsidering.

tarsius:
> When thinking about this yesterday, checking the status seemed like the right thing to do, but I didn't pursue that because it'd be a deeper change and all callers except `magit-diff-type` seem to respect `magit-rev-verify` expectation of a single revision arguments.  But given that `verify-commit` doesn't seem to guard against standard ranges and that my proposed `magit-diff-type` change doesn't detect `rev^-` ranges, perhaps it's worth reconsidering.

I was thinking the same. Lets do it.

tarsius:
What the hell is Github doing to this pull-request? Are all my nasty force-pushes being preserved for all time now? :scream:

Anyway... I've added two new functions `magit-git-string-{p,ng}` as potential replacements for `magit-git-str{,ing}`. But those will need more though before being used more widely -- for now I am only using `-ng` to fix this issue.

kyleam:
> What the hell is Github doing to this pull-request? Are all my nasty force-pushes being preserved for all time now?

I put the base at maint when I opened the PR, since the fix was relevant there as well.

kyleam:
> Anyway... I've added two new functions `magit-git-string-{p,ng}` as potential replacements for `magit-git-str{,ing}`. But those will need more though before being used more widely -- for now I am only using `-ng` to fix this issue.

Ah, thanks, I hadn't realize you were going to work on this, but I also hadn't revisited it yet.  If you want another set of eyes, I'll try to take a look over the changes today/tomorrow.

tarsius:
> I'll try to take a look over the changes today/tomorrow.

That would be great, thanks!

PR #3672, opened 2018-11-21
--keep-empty commits when autosquashing

### LINKED ISSUE #3670: Git commit instant fixup/squash: allow to specify --keep-empty
When performing an instant fixup or squash on a previous commit, in case there is an empty commit between the target commit and the current head, the empty commit is implicitly removed as result of the command `git rebase -i --autosquash --autostash`.
This behaviour can modified to keep empty commits by adding the `--keep-empty` option to `git rebase`.

It would be nice to add an option to `magit-commit-instant-fixup/squash`, so that empty commits can possibly be preserved.

tarsius:
Closes #3670.

### DIFF
--- a/Documentation/RelNotes/2.91.0.org
+++ b/Documentation/RelNotes/2.91.0.org
@@ -49,6 +49,13 @@
   shown in the status buffer was the closest approximation readily
   available when this feature was first added.  #3671
 
+- The commands ~magit-commit-instant-fixup~ and
+  ~magit-rebase-instant-squash~ now unconditionally preserve empty
+  commits that were already present before the autosquash rebase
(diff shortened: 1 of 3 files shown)


Documentation/RelNotes/2.91.0.org
```diff
@@ -49,6 +49,13 @@
   shown in the status buffer was the closest approximation readily
   available when this feature was first added.  #3671
 
+- The commands ~magit-commit-instant-fixup~ and
+  ~magit-rebase-instant-squash~ now unconditionally preserve commits
```
kyleam: s/commits/empty commits/

PR #3675, opened 2018-11-22
On instant squash/fixup, keep commit cursor is at

### LINKED ISSUE #3674: Instant squash/fixup no longer jumps to the right place in *magit-log-select: *
To reproduce:
1 `git stage` something
2 Open your project's main `*magit:` buffer
3 Move cursor to any commit visible in the short log at the bottom
4 Type `c S`

Expected behavior: `magit-log-select:` buffer opens and point jumps to commit with the same commit ID we selected in step 3 above

Actual behavior: point stays at the first line 

M-x magit-version → `Magit v2.90.1-20-g01240e9b, Git 2.16.2, Emacs 26.1, darwin`

Note that `make emacs-Q` didn't *quite* work for me, as I had to make a couple of symlinks into my `load-path` for the dependencies:

```
Mac (master):~/Library/emacs$ ls -l dash magit-popup treepy with-editor graphql ghub
lrwxr-xr-x  1 dom  staff  28 22 nov 18:43 dash -> elpa-26.1/dash-20180910.1856
lrwxr-xr-x  1 dom  staff  28 22 nov 18:45 ghub -> elpa-26.1/ghub-20181112.1755
lrwxr-xr-x  1 dom  staff  29 22 nov 18:45 graphql -> elpa-26.1/graphql-20180912.31
lrwxr-xr-x  1 dom  staff  34 22 nov 18:44 magit-popup -> elpa-26.1/magit-popup-20181003.921
lrwxr-xr-x  1 dom  staff  29 22 nov 18:45 treepy -> elpa-26.1/treepy-20180724.656
lrwxr-xr-x  1 dom  staff  35 22 nov 18:44 with-editor -> elpa-26.1/with-editor-20181113.1845
```

Once I got these symlinks in order, I could indeed reproduce this here issue with `make emacs-Q`.

I'm pretty sure this Used To Work Before™, with some evidence that "before" could be along the lines of `elpa-25.3.1/magit-20180116.921/` (still on Mac OS X)

domq:
Fix  #3674 by explicitly plumbing down the last optional argument to `magit-log-select`.

TL;DR: there could be a better way to do this but I couldn't find it.

I did read the code of `magit-log-goto-same-commit`, and its fallback logic did sound a little bit like what we want to use to me. However, it appears to work based on the `magit-previous-section` variable, which is only called when setting up a new buffer afaict.

### DIFF
--- a/Documentation/RelNotes/2.91.0.org
+++ b/Documentation/RelNotes/2.91.0.org
@@ -77,3 +77,8 @@
 
 - The Custom type definition of ~magit-diff-adjust-tab-width~ was
   broken.  #3671
+
+- In the log-select buffer shown by ~magit-commit-instant-squash~ and
+  ~magit-commit-instant-fixup~ point was no longer placed on the same
+  commit as was at point in the buffer from which the command was
+  invoked.  #3674
--- a/lisp/magit-commit.el
+++ b/lisp/magit-commit.el
@@ -317,7 +317,8 @@ depending on the value of option `magit-commit-squash-confirm'."
                 (list "--autosquash" "--autostash" "--keep-empty")
               "" "true" nil t)))
         (format "Type %%p on a commit to %s into it,"
-                (substring option 2)))
+                (substring option 2))
+        nil nil nil commit)
       (when magit-commit-show-diff
         (let ((magit-display-buffer-noselect t))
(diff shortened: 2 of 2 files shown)


tarsius:
> TL;DR: there could be a better way to do this but I couldn't find it.

Looks perfect.

PR #3679, opened 2018-11-26
Visibility indicators

tarsius:
I am undecided whether to enable this by default and if so what style
to default to.  Enable it for now on the development branch, so that
users notice and get a chance to comment here before the next
release.

### DIFF
--- a/Documentation/RelNotes/2.91.0.org
+++ b/Documentation/RelNotes/2.91.0.org
@@ -12,6 +12,15 @@
 
 ** Changes since v2.90.0
 
+- Added new option ~magit-section-visibility-indicator~.
+
+  TODO:
+
+  Decide whether to enable this by default and if so what style to
+  default to.  Enable it for now on the development branch, so that
+  users notice and get a chance to comment on #3679 before the next
+  release.
+
 - Added ~bug-reference-mode~ to ~git-commit-setup-hook~.  e19e8fb1e
 
 - The "Version" column in ~magit-submodule-list-mode~ and
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -1095,6 +1095,29 @@ certain sections can also be overwritten using the hook
   then the buffer is being refreshed and these functions should
   immediately return ~nil~.
 
+- User Option: magit-section-visibility-indicator
+
+  This option controls whether and how to indicate that a section can
+  be expanded/collapsed.
+
+  If nil, then no visibility indicators are shown.  Otherwise the
+  value has to have one of these two forms:
+
+  - ~(EXPANDABLE-BITMAP . COLLAPSIBLE-BITMAP)~
+
+    Both values have to by variables whose values are fringe
+    bitmaps.  In this case every section that can be expanded
+    or collapsed gets an indicator in the left fringe.
+
+  - ~(STRING . BOOLEAN)~
+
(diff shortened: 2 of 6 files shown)


wedens:
image

I'm not sure where to add this suggestion, but could you add a space between the arrows and the text? Currently it's a bit tight.

tarsius:
> could you add a space between the arrows and the text? Currently it's a bit tight.

This should help:

```lisp
(add-hook 'magit-mode-hook
          (lambda ()
            (setq left-fringe-width 16)
            (set-window-buffer (selected-window) (current-buffer))))
```

> Also, looks like it doesn't scale with a font size.

I actually switched to a "4k" on the same day as I implemented this, so I know what you mean...

I tried using a 16x16 bitmap instead of a 8x8 one, but that didn't work. The result looked like the top-left corner of the bitmap I specified with every second line being "empty". I would appreciate help from someone who has previously fought the bitmap and won.

NicolasPetton:
I do not mind this addition and having it turned on by default, but I'm not sure if it is useful or not.

If you added this to help newcomers, I don't think it will help that much, because I don't think it makes the interface much more discoverable. What about mouse actions? 

The indicators makes the status buffer look like a tree, but expanding sections result in a flat list (no indentation), which could be also be confusing.

NicolasPetton:
If this is for newcomers, what about a minor mode that would display "..." at the end of expandable sections as a button. `RET` on mouse click would expand the section.

It might be too intrusive in the status buffer, but advanced users could turn this off. This could also be part of a broader "getting started with Magit" kind of option that could guide the user in various aspects of Magit.

tarsius:
> If this is for newcomers, what about a minor mode that would display "..." at the end of expandable sections as a button. `RET` or mouse click would expand the section.


Yes, this is mostly for the benefit of new users. (And long-time users of the less adventurous sort who never bothered to try what <kbd>TAB</kbd> does. Again and again someone gets excited then being told that they can use that to show diffs directly in the status buffer. It's really weird.)

> If this is for newcomers, what about a minor mode that would display "..." at the end of expandable sections as a button.

That's the alternative style, which I would have used by default if it didn't have the potential of decreasing performance. As mentioned above I might go with that style anyway.

> `RET` or mouse click would expand the section.

I plan to do that.

aaronjensen:
The hook is helpful, thanks. I bumped it up to 20px and it feels a little more comfortable. I'd probably want them to be a little closer to the background color so they are less distracting (can the color be configurable?)

tarsius:
> I bumped it up to 20px and it feels a little more comfortable.

What did you bump to 20px and how? (As you can see above I was unable to increase the size of the fringe bitmaps myself, so some help with that would be appreciated.)

> (can the color be configurable?)

Yes, using the `fringe` face.

aaronjensen:
> What did you bump to 20px and how?

Sorry, I meant the fringe width via the hook you suggested: `(setq left-fringe-width 20)`

> Yes, using the `fringe` face.

Interesting, I would have expected there to be a way to set colors for specific fringe items.

I ended up disabling the indicators last night anyway, they were too noisy for me. I think the problem for me is that too many things in magit are expandable so I'd see those fringe markers all over the place, plus I already know most of what is expandable and use it. 

I do think it or something like it is a good feature for discoverability, however.

PR #3683, opened 2018-12-02
magit-clone: execute magit-credential-hook

marsam:
The git clone command need credentials in order to perform on private repositories.


tarsius:
Thanks!

PR #3690, opened 2018-12-07
Fix spelling error in magit-status.el

aeggenberger:
Buried was misspelled.

Github is an insufficient interface for editing files and creating commits.  Please don't do it when contributing to Magit.

What you should write here

Please summarize the changes made in the commits.  Explain why you are making these changes, not just what changes you are making.  This also applies to the commit messages.

How to update the manual

If you make changes to the manual, then edit only "magit.org".  Do not manually edit "magit.texi".  The latter has to be generated from the former.  If you want to do that yourself, then follow the instructions at [2].  Otherwise a maintainer will do it and amend that to your commit.

### DIFF
--- a/lisp/magit-status.el
+++ b/lisp/magit-status.el
@@ -96,7 +96,7 @@ all."
   "The section point is placed on when a status buffer is created.
 
 When such a buffer is merely being refreshed or being shown again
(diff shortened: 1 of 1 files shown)


kyleam:
thanks

PR #3698, opened 2018-12-20
magit-diff-inside-hunk-body-p: fix scroll in sections with no content

kidd:
In diff sections where the change is a file mode change, there's no
content in the section (only title) but scrolling through it is
allowed now.


tarsius:
Thanks!

I fixed this a bit differently.  Instead of pretending that those special hunks have a body, I just prevent the code from running that would error if it was run even though there is no body.

kidd:
cool.

Thanks!

PR #3702, opened 2018-12-25
Fix: (magit-describe-section-briefly) Display ident interactively

alphapapa:
The interactive form and args did not match up, so using a prefix interactively did not cause the ident to be displayed, but rather the vector (which is not useful for magit-status-initial-section).


tarsius:
Thanks!

Issue #3668, opened 2018-11-18
Support for --rebase-merges


kyleam:
In Git 2.18.0, `rebase` gained a `--rebase-merges` flag. Supporting it fully requires updates to git-rebase.el and magit-sequence.el. The git-rebase.el changes are needed regardless of whether Magit supports using `--rebase-merges` itself because users initiating a rebase from the command line can be placed in a `git-rebase-mode` buffer depending on their editor settings.

I plan to look into this within the next couple of weeks, focusing on the git-rebase.el support first.

PR #3711, opened 2019-01-11
Add support for `git rebase --rebase-merges'

agrn:
``git rebase --rebase-merges`` is a new mode to rebase merge commits properly, even with interactive rebase.  It adds three new commands, ``label``, ``reset``, and ``merge``.  The goal is to make it as easy and convenient to handle todo files generated by ``git rebase -ir`` as more traditional todo files without ``--rebase-merges`` (i.e. what exists today in magit).

This mode first came out in git 2.18, I hope it’s not a problem to advertise or support such recent features in magit.

### DIFF
--- a/lisp/git-rebase.el
+++ b/lisp/git-rebase.el
@@ -56,6 +56,9 @@
 ;;   x        Add a script to be run with the commit at point
 ;;            being checked out.
 ;;   z        Add noop action at point.
+;;   l        Add a label at point.
+;;   t        Add a reset action after the commit at point.
+;;   M        Add a merge action at point.
 ;;
 ;;   SPC      Show the commit at point in another buffer.
 ;;   RET      Show the commit at point in another buffer and
--- a/lisp/git-rebase.el
+++ b/lisp/git-rebase.el
@@ -157,6 +157,7 @@
     (define-key map (kbd "e") 'git-rebase-edit)
     (define-key map (kbd "l") 'git-rebase-label)
     (define-key map (kbd "m") 'git-rebase-edit)
+    (define-key map (kbd "M") 'git-rebase-merge)
     (define-key map (kbd "f") 'git-rebase-fixup)
     (define-key map (kbd "q") 'undefined)
     (define-key map (kbd "r") 'git-rebase-reword)
@@ -423,6 +424,19 @@ input remove the command on the current line, if any."
   (git-rebase-insert-at-point-with-args
    "\\(t\\|reset\\) \\(.*\\)" "reset"
    (lambda (initial) (read-from-minibuffer "Reset to label: ")) nil arg))
+
+(defun git-rebase-merge (rev)
+  "Add a merge command to be inserted before the proceeding commit.
+
+REV is the commit to be merged."
+  (interactive (list (magit-read-branch-or-commit "Insert revision")))
+  (goto-char (line-beginning-position))
+  (let ((inhibit-read-only t)
+        (label (read-from-minibuffer "Label: "))
+        (commit-message (read-from-minibuffer "Message for merge commit: ")))
+    (insert (concat "merge " (if (y-or-n-p "Reword the commit message? ")
+                                 "-c "
+                               "-C ") rev " # " commit-message "\n"))))
 
 (defun git-rebase-noop (&optional arg)
   "Add noop action at point.
--- a/lisp/git-rebase.el
+++ b/lisp/git-rebase.el
@@ -350,35 +350,53 @@ current line."
         (insert "pick " it ?\n))
     (user-error "Unknown revision")))
 
-(defun git-rebase-exec (arg)
-  "Insert a shell command to be run after the proceeding commit.
+(defun git-rebase-insert-at-point-with-args (regex insn read-fun at-point arg)
+  "Insert a command with a parameter in the buffer.
 
-If there already is such a command on the current line, then edit
-that instead.  With a prefix argument insert a new command even
-when there already is one on the current line.  With empty input
-remove the command on the current line, if any."
-  (interactive "P")
-  (let ((inhibit-read-only t) initial command)
+The command to insert is set by INSN, and its argument is
+prompted from the user with the function READ-FUN.  It takes one
+argument, the initial value.  If AT-POINT is not nil, the command
+will be inserted on the current line instead of the next line.
+
+If there already is such a command on the current
+line (determined by REGEX), then edit that instead.  With a
+prefix argument (ARG) insert a new command even when there
+already is one on the current line.  With empty input remove the
+command on the current line, if any."
+  (let ((inhibit-read-only t) initial value)
     (unless arg
       (goto-char (line-beginning-position))
       (when (looking-at (concat git-rebase-comment-re "?"
-                                "\\(e\\|exec\\) \\(.*\\)"))
+                                regex))
         (setq initial (match-string-no-properties 2))))
-    (setq command (read-shell-command "Execute: " initial))
-    (pcase (list command initial)
+    (setq value (funcall read-fun initial))
+    (pcase (list value initial)
       (`("" nil) (ding))
       (`(""  ,_)
        (delete-region (match-beginning 0) (1+ (match-end 0))))
       (`(,_ nil)
-       (forward-line)
-       (insert (concat "exec " command "\n"))
+       (unless at-point
+         (forward-line))
+       (insert (concat insn " " value "\n"))
        (unless git-rebase-auto-advance
          (forward-line -1)))
       (_
-       (replace-match (concat "exec " command) t t)
+       (replace-match (concat insn " " value) t t)
        (if git-rebase-auto-advance
            (forward-line)
          (goto-char (line-beginning-position)))))))
+
+(defun git-rebase-exec (arg)
+  "Insert a shell command to be run after the proceeding commit.
+
+If there already is such a command on the current line, then edit
+that instead.  With a prefix argument (ARG) insert a new command
+even when there already is one on the current line.  With empty
+input remove the command on the current line, if any."
+  (interactive "P")
+  (git-rebase-insert-at-point-with-args
+   "\\(e\\|exec\\) \\(.*\\)" "exec"
+   (lambda (initial) (read-shell-command "Execute: " initial)) nil arg))
 
 (defun git-rebase-noop (&optional arg)
   "Add noop action at point.
--- a/lisp/git-rebase.el
+++ b/lisp/git-rebase.el
@@ -470,7 +470,7 @@ running 'man git-rebase' at the command line) for details."
   (setq git-rebase-comment-re (concat "^" (regexp-quote comment-start)))
   (setq git-rebase-line
         (concat "^\\(" (regexp-quote comment-start) "? *"
-                "\\(?:[fprse]\\|pick\\|reword\\|edit\\|squash\\|fixup\\|exec\\|noop\\)\\) "
+                "\\(?:[fprsetlm]\\|pick\\|reword\\|edit\\|squash\\|fixup\\|exec\\|noop\\|reset\\|label\\|merge\\)\\) "
                 "\\(?:\\([^ \n]+\\) \\(.*\\)\\)?"))
   (setq font-lock-defaults (list (git-rebase-mode-font-lock-keywords) t t))
   (unless git-rebase-show-instructions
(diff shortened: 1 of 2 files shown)


kyleam:
Thanks for the PR!  I should have been quicker with my #3668 :]  I'm excited to review this and give it a spin.

Just for reference, I'm pushing my mostly working but unpolished/finished git-rebase.el changes to [a branch](https://github.com/magit/magit/compare/scratch/km/rebase-merges).  I may point to it when reviewing if there's something that was addressed there but not here.

lisp/git-rebase.el
```diff
@@ -350,36 +353,72 @@ current line."
...
+  "Add a merge commadn to be inserted before the proceeding commit."
+  (interactive (list (magit-read-branch-or-commit "Insert revision")))
+  (goto-char (line-beginning-position))
+  (let ((inhibit-read-only t)
+        (label (read-from-minibuffer "Label: "))
+        (commit-message (read-from-minibuffer "Message for merge commit: "))
+        (edit-message (if (y-or-n-p "Reword the commit message? ")
+                          "-c "
+                        "-C ")))
+    (insert (concat "merge " edit-message rev " " label " # " commit-message "\n"))))
```
kyleam: I'm not sure what the best design for the merge command is.  I don't think we want to always read a revision or the message, since the user just might want to do `merge <label>` and be popped into the editor to change the message.

And it seems useful to be able to modify the merge command at point, to alternate between -C and -c at least, but I'm not sure if that's  best to have as a separate command.

lisp/magit-sequence.el
```diff
@@ -467,6 +467,7 @@ This discards all changes made since the sequence started."
               (?A "Autostash"                "--autostash")
               (?i "Interactive"              "--interactive")
               (?h "Disable hooks"            "--no-verify"))
+  :options  '((?r "Rebase merges" "--rebase-merges=" magit-rebase-merges-select-mode))
```
kyleam: I'm undecided on whether --rebase-merges should be in the popup by default.

lisp/magit-sequence.el
```diff
@@ -564,7 +570,11 @@ START has to be selected from a list of recent commits."
           (setq commit (concat commit "^"))
         (setq args (cons "--root" args)))))
   (when (and commit (not noassert))
-    (setq commit (magit-rebase-interactive-assert commit delay-edit-confirm)))
+    (let ((rebase-merges nil))
+      (dolist (elt args rebase-merges)
+        (when (string-match-p "^--rebase-merges" elt)
+          (setq rebase-merges t)))
```
kyleam: Could be simplified to

    (--some (string-prefix-p "--rebase-merges" it)
            args)

and passed directly as the argument.

lisp/magit-sequence.el
```diff
@@ -580,7 +590,7 @@ START has to be selected from a list of recent commits."
 (defvar magit--rebase-published-symbol nil)
 (defvar magit--rebase-public-edit-confirmed nil)
 
-(defun magit-rebase-interactive-assert (since &optional delay-edit-confirm)
+(defun magit-rebase-interactive-assert (since &optional delay-edit-confirm rebase-merges)
```
kyleam: nit: break long line

lisp/magit-sequence.el
```diff
@@ -605,7 +615,7 @@ START has to be selected from a list of recent commits."
           (concat m1 "%i public branches" m2)
           nil branches))
       (push (magit-toplevel) magit--rebase-public-edit-confirmed)))
-  (if (magit-git-lines "rev-list" "--merges" (concat since "..HEAD"))
+  (if (and (magit-git-lines "rev-list" "--merges" (concat since "..HEAD")) (not rebase-merges))
```
kyleam: nit: move `(not rebase...` to next line.

lisp/git-rebase.el
```diff
@@ -350,36 +353,72 @@ current line."
...
-       (insert (concat "exec " command "\n"))
+       (unless at-point
+         (forward-line))
+       (insert (concat insn " " value "\n"))
        (unless git-rebase-auto-advance
          (forward-line -1)))
       (_
-       (replace-match (concat "exec " command) t t)
+       (replace-match (concat insn " " value) t t)
        (if git-rebase-auto-advance
```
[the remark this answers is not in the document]
agrn: I don’t think so.  ``at-point`` defines whether the command should be inserted on the current line or not, not whether the cursor should move to the next line after the fact.

For instance, when I put my cursor on a commit and insert a label, the label will be inserted to the current line instead of the next line, unlike with execs and resets.  But if I want to replace a label (or a command or a reset), I think it should go to the next line in any case (except if the user set ``git-rebase-auto-advance`` to ``nil``).

Now that I think about it two weeks later, I’m not sure it’s the right behaviour for inserting ``label`` commands.  Perhaps the user means “I want this commit to be part of the label” instead of “I want the label before this commit”.  Perhaps I should drop ``at-point`` altogether.

lisp/git-rebase.el
```diff
@@ -350,36 +353,72 @@ current line."
...
+        (regex "\\(l\\|label\\) \\(.*\\)")
+        (fn '(lambda (initial) (read-from-minibuffer "Label: " initial))))
+    (git-rebase-insert-at-point-with-args regex "label" fn t arg)))
+
+(defun git-rebase-reset (arg)
+  "Add a reset command to be inserted after the proceeding commit."
+  (interactive "P")
+  (let ((insn "reset")
+        (regex "\\(t\\|reset\\) \\(.*\\)")
+        (fn '(lambda (initial) (read-from-minibuffer "Reset to label: " initial))))
```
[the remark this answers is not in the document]
agrn: > Also, do you think it's worth considering the collection of labels currently in the buffer ([example](https://github.com/magit/magit/blob/6c18e1045878d64aaf8c211f45424c8ed5d97fee/lisp/git-rebase.el#L483-L489))?

I think it’s a good idea, yes.

lisp/git-rebase.el
```diff
@@ -350,36 +353,72 @@ current line."
...
+    (git-rebase-insert-at-point-with-args regex insn fn nil arg)))
+
+(defun git-rebase-merge (rev)
+  "Add a merge commadn to be inserted before the proceeding commit."
+  (interactive (list (magit-read-branch-or-commit "Insert revision")))
+  (goto-char (line-beginning-position))
+  (let ((inhibit-read-only t)
+        (label (read-from-minibuffer "Label: "))
```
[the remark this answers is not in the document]
agrn: Same answer: I think it’s a good idea.

lisp/git-rebase.el
```diff
@@ -350,36 +353,72 @@ current line."
...
+  (let ((insn "reset")
+        (regex "\\(t\\|reset\\) \\(.*\\)")
+        (fn '(lambda (initial) (read-from-minibuffer "Reset to label: " initial))))
+    (git-rebase-insert-at-point-with-args regex insn fn nil arg)))
+
+(defun git-rebase-merge (rev)
+  "Add a merge commadn to be inserted before the proceeding commit."
+  (interactive (list (magit-read-branch-or-commit "Insert revision")))
+  (goto-char (line-beginning-position))
+  (let ((inhibit-read-only t)
```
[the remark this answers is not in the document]
agrn: I think it’s needed here to ask the question in the right order (``edit-message`` is prompted last, but it’s inserted first in ``(concat)``).  The best I can do is to drop ``edit-message``.  Or maybe you have a better solution?

lisp/magit-sequence.el
agrn: ``rebase -p`` will be deprecated and replaced by ``--rebase-merges`` in “[a couple of major Git versions](https://colabti.org/irclogger/irclogger_log/git-devel?date=2019-01-22#l12)”, so I think it should eventually be part of the popup.

lisp/magit-sequence.el
agrn: Ah, I wanted to write something like that, but I couldn’t find ``(--some)``.  Thank you for pointing this out!

agrn:
Hi, sorry for the late answer.

I skimmed through ``scratch/km/rebase-merges``, and perhaps your approach is better than mine regarding ``git-rebase.el``.  I fixed (most of) the issues you pointed out in this file, but I’m not sure if we should keep them.

lisp/git-rebase.el
```diff
@@ -350,36 +353,72 @@ current line."
         (insert "pick " it ?\n))
     (user-error "Unknown revision")))
 
-(defun git-rebase-exec (arg)
-  "Insert a shell command to be run after the proceeding commit.
+(defun git-rebase-insert-at-point-with-args (regex insn read-fun at-point arg)
+  "Insert a command with a parameter in the buffer.
```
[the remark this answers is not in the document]
agrn: I rewrote the doc, perhaps the wording is not the best.

lisp/git-rebase.el
```diff
@@ -350,36 +356,91 @@ current line."
...
+  "Add a reset command to be inserted after the proceeding commit.
+
+If there already is such a command on the current line, then edit
+that instead.  With a prefix argument (ARG) insert a new command
+even when there already is one on the current line.  With empty
+input remove the command on the current line, if any."
+  (interactive "P")
+  (git-rebase-insert-at-point-with-args
+   "\\(t\\|reset\\) \\(.*\\)" "reset"
+   (lambda (initial) (read-from-minibuffer "Reset to label: ")) nil arg))
```
kyleam: Looks like you accidentally dropped `initial` from the `read-from-minibuffer` call.

lisp/git-rebase.el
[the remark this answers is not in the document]
kyleam: > I think it’s needed here to ask the question in the right order

OK.


> Or maybe you have a better solution?

As I mentioned in my initial review, I'm not quite sure what the best design is, but I think we might want to just support the simple `merge <label>` form and let the user be popped into a buffer to edit the message.  There could be another command that toggles -c/-C.

The merge command from the latest update looks broken, since it is missing the label field.  When I try to rebase with a line of that form, the execution fails with

```
error: nothing to merge: '# Merge pull request #3132 from kyleam/stale-comments'
hint: Could not execute the todo command
```

lisp/git-rebase.el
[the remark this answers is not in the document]
kyleam: >> Should this git-rebase-auto-advance logic consider at-point?
>
> I don’t think so [...]

I see.  Thanks.

> Now that I think about it two weeks later, I’m not sure it’s the right behaviour for inserting `label` commands. Perhaps the user means “I want this commit to be part of the label” instead of “I want the label before this commit”. Perhaps I should drop `at-point` altogether.

Yeah, I'm not sure either, though if neither behavior is clearly better, I lean towards being consistent across commands.  It is easy enough for the user to follow up with a M-p.

lisp/magit-sequence.el
kyleam: Fair enough.

lisp/git-rebase.el
```diff
@@ -350,36 +356,91 @@ current line."
...
+
+(defun git-rebase-merge (rev)
+  "Add a merge command to be inserted before the proceeding commit.
+
+REV is the commit to be merged."
```
kyleam: Isn't REV actually specifying a commit whose message should be used?

tarsius:
@kyleam How's your implementation coming along?

kyleam:
> @kyleam How's your implementation coming along?

I haven't revisited since this PR came in to avoid duplicating effort.  @agrn, do you have an interest in following up on the blockers I mentioned [here] or would you prefer that I proceed with my implementation?

[here]: https://github.com/magit/magit/pull/3711#pullrequestreview-198418370

kyleam:
I've revisited my changes and pushed the WIP state to the [km/rebase-merges][0] branch.  Aside from polish/further thought, the main to-do is the sequencer.

[0]: https://github.com/magit/magit/compare/km/rebase-merges

tarsius:
You should probably open a pr. The "Create Pull Request" button now has a drop-down that lets you mark it as a "draft".

Looks very good overall. (I haven't actually tested it yet though.)

`git-rebase-move-line-down: Fix max bound definition` probably has to be improved--it doesn't take internationalization into account.

kyleam:
I've opened a PR.

> `git-rebase-move-line-down: Fix max bound definition` probably has to be improved--it doesn't take internationalization into account.

Thanks for pointing that out.  That should have dawned on me given my 8e762d34 (git-rebase-mode-show-keybindings: detect translated output, 2017-09-19).

agrn:
Hi,

I’m not really in favour of keeping my changes in `git-rebase.el`, as I said 
earlier.

I was going to tell you to feel free to cherry-pick the last two commits, but 
I see you have already done so.  :)

tarsius:
Closing in favor of #3762.

PR #3720, opened 2019-01-27
Make magit-blame-echo actually echo the commit summary

peniblec:
Hi!

`magit-blame-echo` never quite worked for me: I did get thin lines
separating chunks belonging to different commits, but no message in
the echo area.

I poked around `magit-blame.el` and found a few things that look like
impedance mismatches to me:

- `m-b-maybe-show-message` checks the current style for the
  `show-message` key, but `magit-blame-echo-style` is set to the
  `lines` visualization style, which only contains `show-lines`.

- `m-b-maybe-show-message` looks for the `'heading` symbol in the
  commit alist, but the keys returned by `m-b--commit-alist` are
  strings, the first of which is `"summary"`.

- `assq` uses `eq` to browse alists, which tests "Lisp
  object"-equality; to match on a string such as `"summary"`, one must
  use `assoc`, which uses `equal`.

This commit is pretty much "works-for-me", I didn't take the time to
consider the following issues:

- I don't know if it makes sense to shoehorn `(show-message . t)` in
  the `lines` style or if a new style should be created.

- I don't know if the alist should use symbols or strings.

- I might have misread the documentation and missed an obvious
  explanation as to why `magit-blame-echo` does not work
  out-of-the-box on my setups (Emacs 25.1, 26.1 and 27.0).

Maybe this commit is not usable as-is, in which case I should have
filed an issue instead.  On the off-chance that the commit is sound
enough, I figured a pull-request could work just as well…

Thank you for your time.

### DIFF
--- a/lisp/magit-blame.el
+++ b/lisp/magit-blame.el
@@ -705,8 +705,7 @@ modes is toggled, then this mode also gets toggled automatically.
                                (gethash (oref (magit-current-blame-chunk)
                                               orig-rev)
                                         magit-blame-cache)))))
-          (progn (setq msg (substring msg 0 -1))
-                 (set-text-properties 0 (length msg) nil msg)
+          (progn (set-text-properties 0 (length msg) nil msg)
                  (message msg))
         (message "Commit data not available yet.  Still blaming.")))))
 
--- a/lisp/magit-blame.el
+++ b/lisp/magit-blame.el
@@ -51,7 +51,8 @@
     (highlight
      (highlight-face   . magit-blame-highlight))
     (lines
-     (show-lines       . t)))
+     (show-lines       . t)
+     (show-message     . t)))
   "List of styles used to visualize blame information.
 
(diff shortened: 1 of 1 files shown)


tarsius:
Perfect!

I think this bug was caused by some last minute refactoring and failure to test again afterwards.  There probably was a `show-lines` style at some point, then I replaced that with a `show-lines` "style property" and renamed the style itself to just `lines`.  Likewise for the alist keys; I probably used symbols as keys for a while and then decided that the strings returned by code used to determine these values were good enough, without making all the necessary adjustments.

(Obviously I haven't use the "echo" mode myself.  I'm guessing that's because it is not available from the relevant popups and because once in the mode I could not figure out how to exit (`C-c C-q`).  I'll try do do something about that when I merge the `transient` branch.

peniblec:
Glad I could help!

Good luck on your work with `transient` :)

PR #3725, opened 2019-02-06
enable signoff during merge

MHBauer:
- git supports signoff during merge
 - signoff description copied from magit-commit.el
 - aligned the strings in the switches block
 - manually tested locally

The `strategy` key in merge is `s` and the `signoff` key in `commit` is `s`, so I used `o` from `signOff` in `merge`. This probably merits some discussion.

Is there a test somewhere I need to add to?
Is there a way to check that a specifically supported version of git is installed before showing this particular switch?
Is there a contributor agreement I should be aware of?

### DIFF
--- a/lisp/magit-merge.el
+++ b/lisp/magit-merge.el
@@ -40,8 +40,9 @@
 (magit-define-popup magit-merge-popup
   "Popup console for merge commands."
   :man-page "git-merge"
-  :switches '((?f "Fast-forward only" "--ff-only")
-              (?n "No fast-forward"   "--no-ff"))
+  :switches '((?f "Fast-forward only"      "--ff-only")
+              (?n "No fast-forward"        "--no-ff")
+              (?o "Add Signed-off-by line" "--signoff"))
   :options  '((?s "Strategy" "--strategy="))
   :actions  '((?m "Merge"                  magit-merge-plain)
               (?p "Preview merge"          magit-merge-preview)


tarsius:
For quite some time I did not merge such pull-requests for the reasons given at https://github.com/magit/magit/wiki/Additional-proposed-infix-arguments-and-suffix-commands.

That is no longer the case (I still have to update the pull-request template) because now Magit uses `transient`.  That makes it possible to place arguments on "levels".  By placing an argument on level 5 for example it is hidden from most users, but those who want it can enable it as described at https://magit.vc/manual/transient/Enabling-and-Disabling-Suffixes.html.

Feel free to open a new pr. You should place this on level 5. To make this really useful you should add this to all transients for which it makes sense. Otherwise I will get around to this eventually, but probably not any time soon.

MHBauer:
Thanks for the feedback, I'll take a look at the levels.

MHBauer:
Wish I had seen that wiki page ahead of time. Maybe need to add a link to it and others like it in a PR_TEMPLATE or the CONTRIBUTING guidelines. I would not have opened this PR given the guidance in the document.

tarsius:
It was mentioned in https://github.com/magit/magit/blob/master/.github/ISSUE_TEMPLATE/feature_request.md but not the pull-request template. Anyway it is not accurate anymore, so I will actually *remove* it from the feature-request template.

Issue #3276, opened 2017-11-22
Magit doesn't handle corrupt repositories with "fatal: bad object HEAD"


hrehfeld:
Version: Magit 20171029.612, Git 2.15.0, Emacs 25.3.1, gnu/linux

Opening magit for a file in a repository where `git status` reports:

```
fatal: bad object HEAD
```

magit crashes with:

```
magit-insert-head-branch-header: Wrong type argument: stringp, nil
```

I would have expected it to show me this or some other error message.

tarsius:
I think it is okay if Magit doesn't function properly in corrupted repositories.

PR #3727, opened 2019-02-13
magit-insert-head-branch-header: Error on failed magit-rev-format

### LINKED ISSUE #3444: Return a meaningful error when the repository is corrupted
Magit 20180509.913 Git 2.16.1 Emacs 24.5.1 gnu/linux

When trying to deal with broken repository, magit returns an error (Wrong type argument: stringp, nil). The error doesn't tell user what is going on and suggests a bug in the magit implementation (tbh I would treat it as such if the error I'm reporting can be reproduced).

As an example, I include a gzippted tar archive that contains the artificial corrupted repository. Here you can see the output of `git status`:

```
[testre.tar.gz](https://github.com/magit/magit/files/1990879/testre.tar.gz)

error: object file .git/objects/e6/8ea0525a4f3d996578db995fe9ce6e76a9ca9d is empty
error: object file .git/objects/e6/8ea0525a4f3d996578db995fe9ce6e76a9ca9d is empty
fatal: loose object e68ea0525a4f3d996578db995fe9ce6e76a9ca9d (stored in .git/objects/e6/8ea0525a4f3d996578db995fe9ce6e76a9ca9d) is corrupt
```
I think it would be useful to report this error to the user (I have to admit that I spent some time 'fixing' my emacs, thinking that my setup was at fault instead of suspecting a broken repository). At the very least, propagating the error message could indicate that something is wrong. 

Please let me know if I can provide you with any additional information.

nickdrozd:
Fixes #3444 and #3276 (kinda)

If a repo is corrupt, magit-rev-format will return nil, and
string-match will raise a type error. This is confusing, making it
sound like something went wrong with Magit, when the true issue is
more dire. The error message in this commit is vague, but it at least
hints that something has gone wrong at the Git level. I don't think
it's possible to get any more specific while remaining unobtrusive.

### DIFF
--- a/lisp/magit-status.el
+++ b/lisp/magit-status.el
@@ -449,6 +449,8 @@ the status buffer causes this section to disappear again."
 When BRANCH is nil, use the current branch or, if none, the
 detached `HEAD'."
   (let ((output (magit-rev-format "%h %s" (or branch "HEAD"))))
+    (unless output
+      (error "Magit: Failed to get header line"))
     (string-match "^\\([^ ]+\\) \\(.*\\)" output)
     (magit-bind-match-strings (commit summary) output
       (when (equal summary "")


nickdrozd:
My personal stake in this is that my computer is a piece of junk that freezes all the time, and if it freezes during a Git operation, then whoops, my repo is busted. So this issue ends up affecting me several times a year.

tarsius:
I think this came up before. My take is that it would be to costly to make sure Magit gives meaningful errors when the file-system or git metadata is corrupted. It would be costly to implement and it would also add an overhead that would always have to be payed and only have a benefit in those very rare cases when there actually is some corruption. IMO it is okay if Magit assumes that does things are not corrupted.

As for this particular change, it is cheap enough, but the error message does not really suggest to the user that they should check whether there is some corruption. And simply changing the message to say that explicitly probably also isn't the right thing to do. There might be other reasons for this to fail and then claiming that it is due to corruption would be misleading.

So I think we should just leave it as is.

nickdrozd:
Fair enough. For anyone who's interested, the following seems to work:

```emacs-lisp
  (defun magit-verify-rev-format (magit-rev-format &rest args)
    (if-let ((output (apply magit-rev-format args)))
        output
      (error "Magit: Failed to get header line")))

  (advice-add 'magit-rev-format :around #'magit-verify-rev-format)
```

nickdrozd:
Actually that doesn't work. Ignore!

PR #3739, opened 2019-02-16
Unconditionally initialize fringe bitmaps

dzhu:
This allows the bitmaps to be shown when emacs is started with `emacs --daemon` and a graphical emacsclient frame is later created.

(I realize this may not be the right solution, but figure it's simple enough that I might as well start by putting it out there.)

### DIFF
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -1123,7 +1123,7 @@ the %s(1) manpage.
 
 ;;; Bitmaps
 
-(when (window-system)
+(when (fboundp 'define-fringe-bitmap)
   (define-fringe-bitmap 'magit-fringe-bitmap+
     [#b00000000
      #b00011000


tarsius:
Looks like I was expecting that that function would not be available if Emacs was *build* without support for a graphical system. Maybe the commit message mentions something. Else we have to double check.

dzhu:
Ah, excellent point. As of [2018-10-21](https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=08192e40093bdbc8e6be6b283935b51c12d66eca), it is defined even in a non-GUI build, but I imagine that is too recent to rely on. I've instead changed the guard to `(fboundp 'define-fringe-bitmap)`.

sunqingyao:
Not sure if that's related, but I'm getting the following warnings when installing the latest version of Magit on MEPLA (`20190216.2144`) for the latest Homebrew Emacs that doesn't support graphical systems.

```
In magit-diff--dwim:
magit-diff.el:916:8:Warning: Unknown slot ‘base-ref’

In end of data:
magit-utils.el:1204:1:Warning: the function ‘define-fringe-bitmap’ is not
    known to be defined.
```

tarsius:
The second is, the first is not.

tarsius:
You email address is empty. That cause issues which I would rather not be forced to deal with. Please correct the author information to make the email address non-empty.

dzhu:
Fixed; sorry for making this take way longer than doing it yourself would've.

(I'm curious: what issues would the blank address cause?)

tarsius:
Np.

> (I'm curious: what issues would the blank address cause?)

I haven't noticed, but it wouldn't surprise me if somewhere in Magit (or some other tools) something assumes that the address isn't blank, e.g. a regexp containing `"<\\(.+\\)>"`.

PR #3748, opened 2019-02-21
Add missing require for magit-gitignore

DamienCassou:
Without it, pressing "i" or "I" in magit-status shows an error message:

    Wrong type argument: commandp, magit-gitignore


tarsius:
Provided the autloads file has been loaded, this should not happen, so I think this should be added inside the  `(unless (load "magit-autoloads" t t) ...)` below.

DamienCassou:
I confirm that regenerating the autoloads fixes the problem. I've updated my commit.

PR #3756, opened 2019-02-23
Document how to set the fringe width for the visibility indicator

### LINKED ISSUE #3726: Set left-fringe-width to 20 by default when fringe markers are enabled?
When `magit-section-visibility-indicator` is enabled, the lack of padding around the indicator looks rather awkward.

As noted in https://github.com/magit/magit/pull/3679#issuecomment-443435062 a value of 20 for `left-fringe-width` looks much better.

Perhaps this could be set as a default?

Failing that, the documentation could mention that users can set `left-fringe-width` in `magit-mode-hook` to provide such padding.

phil-s:
Closes #3726.

I've omitted the call to `set-window-buffer` which was suggested in https://github.com/magit/magit/pull/3679#issuecomment-441747653 on the basis that, so far as I can see experimentally, it's unnecessary in this case -- `magit-mode` and its derivatives are called before the buffer is displayed, so I believe there is no need to forcibly re-display the buffer at this point.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -1108,9 +1108,17 @@ certain sections can also be overwritten using the hook
 
   - ~(EXPANDABLE-BITMAP . COLLAPSIBLE-BITMAP)~
 
-    Both values have to by variables whose values are fringe
+    Both values have to be variables whose values are fringe
     bitmaps.  In this case every section that can be expanded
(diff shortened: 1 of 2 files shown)


phil-s:
I opted to leave the example `add-hook` code out of the docstring copy, but am happy to re-do with that included.

Issue #3765, opened 2019-02-25
Auto-complete HEAD, FETCH_HEAD, ORIG_HEAD


benma:
Magit 20190215.1903, Git 2.17.1, Emacs 25.2.2, gnu/linux

When switching branches, or doing a hard reset, or anywhere where there is autocompletion for a ref, it would be nice if those three would also be auto completed in the minibuffer.

:heart:

tarsius:
Some commands do that now.

benma:
When I checkout a branch, I still can't autocomplete FETCH_HEAD. To this day I manually type it out every time 😅 

Could you please re-open this issue?

tarsius:
I've opened #5331 for this.

PR #3766, opened 2019-02-26
Offer rev pointers as completion candidates

tarsius:
Requested in #3765.  Not sure this is the way to go.

We don't want to offer any pointers that don't actually exist, so we have to check whether the respective files exist.  There are more such pointers; anyone know of a complete list?

`FETCH_HEAD` and `ORIG_HEAD` actually aren't symbolic refs, which is why I used the term "pointer".  Any better names (including for the new function)?

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1513,8 +1513,20 @@ SORTBY is a key or list of keys to pass to the `--sort' flag of
                         name)))
             (magit-git-lines "show-ref"))))
 
-(defun magit-list-refnames (&optional namespaces)
-  (magit-list-refs namespaces "%(refname:short)"))
+(defun magit-list-refnames (&optional namespaces include-special)
+  (nconc (magit-list-refs namespaces "%(refname:short)")
+         (and include-special
+              (magit-list-special-refnames))))
+
+(defvar magit-special-refnames
+  '("HEAD" "ORIG_HEAD" "FETCH_HEAD" "MERGE_HEAD" "CHERRY_PICK_HEAD"))
+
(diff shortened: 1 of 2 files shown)


cichli:
According to [this answer](https://stackoverflow.com/a/17596689) some are documented by [`git help revisions`](http://git-scm.com/docs/gitrevisions), but there are others referred to only in the git source. 

Why not list all files in the git dir whose names end in `_HEAD`? There might be other scripts/extensions that use the same format.

On the other hand: it might not actually be useful to offer completions for any but those currently hardcoded in this PR.

>Any better names (including for the new function)?

`magit-list-special-refs` or `magit-list-special-refnames`? Even if they're not technically symbolic refs, `refname` is used in the git docs above.

lisp/magit-git.el
```diff
@@ -1513,9 +1513,19 @@ SORTBY is a key or list of keys to pass to the `--sort' flag of
                         name)))
             (magit-git-lines "show-ref"))))
 
-(defun magit-list-refnames (&optional namespaces)
+(defun magit-list-refnames (&optional namespaces) ; TODO? ... heads ats)
   (magit-list-refs namespaces "%(refname:short)"))
 
+;; TODO? (defvar magit-at-refnames '("@{upstream}" ...))
```
kyleam: I certainly use @{upstream} (or @{u}, I suppose, if I'm manually entering it) and @{push} more than any of the _HEAD refs, though I do use our existing support for FETCH_HEAD (`magit-log-read-revs`) and ORIG_HEAD (`magit-read-reuse-message`) a good amount.

PR #3770, opened 2019-02-27
magit-branch-delete: Don't abort when preserving PR remotes

cichli:
Currently if you delete the only local branch from a pull request remote, you are prompted to also delete the remote corresponding to that PR. However if you press `n` at this prompt, deletion of the branch is aborted. Solution: pass `t` as the value of the `noabort` parameter to `magit-confirm` for the prompt in question.

As an aside: at this point I've been using Magit daily for over five years. Thank you for the continued dedication and hard work put into this project.


tarsius:
Thanks!

PR #3773, opened 2019-02-28
Fix harmless typo in documentation: "turn of" => "turn off"

fritzgrabo:
Fix harmless typo in documentation (magit.org): "turn of" => "turn off".


tarsius:
Thanks!

PR #3775, opened 2019-03-05
Log run time of section and header hooks

aaptel:
I often work with large repos that make the status buffer take 6-8 seconds to render and despite reading the FAQ I couldn't figure out what was taking time.

The verbose mode only showed that refreshing the status buffer was taking time, which I already knew.

I measure run time of various parts of the refresh func and finally figured out that some of the default section hooks were taking a lot of time. When I wanted to edit those on a per-repo basis it turns out the hook lists are not local-safe.

This patchset adds:

* A macro help to convieniently measure run time of forms
* Replaces a couple of time measurements with the macro
* Makes the section/header hook lists local-safe
* Adds a little paragraph about it in the Performance section of the docs

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -6978,6 +6978,10 @@ current buffer.
   (setq magit-refresh-status-buffer nil)
 #+END_SRC
 
+If showing the status buffer is slow, some of the information
+displayed can be disabled by removing hooks from
+~magit-status-headers-hook~ and ~magit-status-sections-hook~.
+
 You should also check whether any third-party packages have added
 anything to ~magit-refresh-buffer-hook~, ~magit-status-refresh-hook~,
 ~magit-pre-refresh-hook~, and ~magit-post-refresh-hook~.  If so, then
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -1061,39 +1061,34 @@ Run hooks `magit-pre-refresh-hook' and `magit-post-refresh-hook'."
                                  (substring (symbol-name major-mode) 0 -5))))
         (magit--refresh-cache (or magit--refresh-cache (list (cons 0 0)))))
     (when (functionp refresh)
-      (when magit-refresh-verbose
-        (message "Refreshing buffer `%s'..." (buffer-name)))
-      (let* ((buffer (current-buffer))
-             (windows
-              (--mapcat (with-selected-window it
-                          (with-current-buffer buffer
-                            (when-let ((section (magit-current-section)))
-                              (list
-                               (nconc (list it section)
-                                      (magit-refresh-get-relative-position))))))
-                        (or (get-buffer-window-list buffer nil t)
-                            (list (selected-window))))))
-        (deactivate-mark)
-        (setq magit-section-highlight-overlays nil)
-        (setq magit-section-highlighted-section nil)
-        (setq magit-section-highlighted-sections nil)
-        (setq magit-section-unhighlight-sections nil)
-        (magit-process-unset-mode-line-error-status)
-        (let ((inhibit-read-only t))
-          (erase-buffer)
-          (save-excursion
-            (apply refresh magit-refresh-args)))
-        (dolist (window windows)
-          (with-selected-window (car window)
-            (with-current-buffer buffer
-              (apply #'magit-section-goto-successor (cdr window)))))
-        (run-hooks 'magit-refresh-buffer-hook)
-        (magit-section-update-highlight)
-        (set-buffer-modified-p nil))
-      (when magit-refresh-verbose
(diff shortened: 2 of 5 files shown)


aaptel:
FYI here's the breakdown of the hook runtimes for me:

```

magit hook magit-insert-error-header                         ... 0.000s
magit hook magit-insert-diff-filter-header                   ... 0.000s
magit hook magit-insert-head-branch-header                   ... 0.006s
magit hook magit-insert-upstream-branch-header               ... 0.008s
magit hook magit-insert-push-branch-header                   ... 0.000s
magit hook magit-insert-tags-header                          ... 2.299s
magit hook magit-insert-status-headers                       ... 2.016s
magit hook magit-insert-merge-log                            ... 0.003s
magit hook magit-insert-rebase-sequence                      ... 0.000s
magit hook magit-insert-am-sequence                          ... 0.000s
magit hook magit-insert-sequencer-sequence                   ... 0.000s
magit hook magit-insert-bisect-output                        ... 0.000s
magit hook magit-insert-bisect-rest                          ... 0.000s
magit hook magit-insert-bisect-log                           ... 0.000s
magit hook magit-insert-untracked-files                      ... 0.083s
magit hook magit-insert-unstaged-changes                     ... 0.047s
magit hook magit-insert-staged-changes                       ... 0.019s
magit hook magit-insert-stashes                              ... 0.015s
magit hook magit-insert-unpushed-to-pushremote               ... 0.000s
magit hook magit-insert-unpushed-to-upstream-or-recent       ... 1.707s
magit hook magit-insert-unpulled-from-pushremote             ... 0.000s
magit hook magit-insert-unpulled-from-upstream               ... 1.709s
```

tarsius:
> Makes the section/header hook lists local-safe

Running arbitrary code is inherently unsafe. We must not claim it is safe.

aaptel:
How is it arbitrary?

tarsius:
You can add whatever function you want to a hook.

aaptel:
You cannot add a lambda (see :safe definition in my commit), therefore you're limited to existing 0-arg functions. Which I'm ok with.

We can restrict it even further if define a list of valid hook values and made sure every item is a member of it. Or check for symbol starting with "magit-" or whatever.

vermiculus:
```elisp
(defun magit-insert-status-headers () (shell-command-to-string "rm -rf /")) ;muahhahaha
```

aaptel:
You would have to evalute this somehow before hand, and if you can do that you might as well run shell-command-to-string directly. Could you describe your scenario in more details?

if you try to set the var to something not validated by the :safe predicate I put (which is a list of symbols) you will be prompted for confirmation like usual for unsafe vars.

And again if this is not enough we can check that each symbol is one of the standard hooks.

vermiculus:
The point is that I can redefine any symbol to mean anything. If I enter a directory that somehow evaluates this form and then sets the hook value, then entering that directory could wipe my machine. Hooks are generally not safe for this kind of on-the-fly customization.

I wouldn't choose this hill to die on. Your other patches are useful.

aaptel:
Please read again. We can check that the symbols belong to the standard magit hook list. I can add a commit to do that, just let me know. This commit would define a magit-standard-hooks var defined to this list:

- magit-insert-error-header
- magit-insert-diff-filter-header
- magit-insert-head-branch-header
- magit-insert-upstream-branch-header
- magit-insert-push-branch-header
- magit-insert-tags-header
- magit-insert-status-headers
- magit-insert-merge-log
- magit-insert-rebase-sequence
- magit-insert-am-sequence
- magit-insert-sequencer-sequence
- magit-insert-bisect-output
- magit-insert-bisect-rest
- magit-insert-bisect-log
- magit-insert-untracked-files
- magit-insert-unstaged-changes
- magit-insert-staged-changes
- magit-insert-stashes
- magit-insert-unpushed-to-pushremote
- magit-insert-unpushed-to-upstream-or-recent
- magit-insert-unpulled-from-pushremote
- magit-insert-unpulled-from-upstream

And the safe predicate would check against this list.

This doesn't *add* any hole, because `magit-status-headers-hook` and
`magit-status-sections-hook` **already contains those by default**.
They are "vulnerable" to the same (non-)problem you described to begin
with. I hope this is clear now.

tarsius:
> Please read again. [...] I hope this is clear now.

I have reservations not only about the `:safe` thing, but about the other changes as well. Contentiously implying that we fail to understand makes me not want to deal with this at all. I have better things to do, e.g. *improve* performance. Your patch can come in handy for that, but in its current form it is not something I would want to merge.

So I am closing this and marking it as something to deal with later.

aaptel:
>Contentiously implying that we fail to understand makes me not want to deal with this at all.

Same here... especially when I spend time explaining myself and you and vermiculus don't, beside a one line example with *evil laugh* as the only comment.

Nevermind I guess.

Ciao

PR #3780, opened 2019-03-07
magit-push: Use -t for --follow-tags

xendk:
I believe --follow-tags should have a `-` prefix and not `=` as it doesn't set an argument.

### DIFF
--- a/lisp/magit-push.el
+++ b/lisp/magit-push.el
@@ -45,7 +45,7 @@
    ("-n" "Dry run"          ("-n" "--dry-run"))
    ("-u" "Set upstream"     "--set-upstream"
     :if-nil magit-remote-set-if-missing)
-   (7 "=t" "Follow tags" "--follow-tags")]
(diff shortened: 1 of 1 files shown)


tarsius:
> It's a switch, not an option.

That's no longer relevant. From the transient release announcement:

> While using Magit-Popup, the key bindings for all switches had the form `-<char>`, while the key bindings for all options had the form `=<char>`. That was a hard-coded convention and now that this restriction has been lifted we always prefer a key binding that match the corresponding short argument.

What this doesn't say is that we still use `=` in some cases. But that doesn't depend on the type of the argument and should only be done if the corresponding `-` cannot be used. It cannot be used if another argument already uses that binding. There may be other reasons, but in this case I do not see any, so this probably was an oversight and the change is good. Please adjust the commit message, talking instead about `-t` not being used for something else. Also mention the pr number, in case this turns out to be wrong, making it easier to find back here.

xendk:
Will do.

xendk:
Updated

PR #3783, opened 2019-03-08
magit-dispatch: Add Cherries

CeleritasCelery:
The key `Y` (`magit-cherry`) is bound in the magit status buffer, but not the dispatch transient. It is applicable to both scenarios. I often find my self going `c-x M-g Y` only to have it tell me that it is unbound. Seems it should be made available.


tarsius:
Thanks!

PR #3784, opened 2019-03-09
magit-revision-filter-files-on-follow: New option

### LINKED ISSUE #3680: Toggling diff file filter doesn't work with --follow
Using latest magic, execute `magit-log-buffer-file` in a git-controlled buffer.  The history of commits affecting that file shows, but no longer restricted to that file.  

In the magic-log buffer, select a diff that affects more than that one file.  All of the diff's show.  Hit 'D' to bring up the diff-refresh popup; hit F to toggle file filter: there is no change, the message shown is:
'No diff file filter to toggle'

kyleam:
Let the user configure which ugliness they prefer when showing a
commit from a log that uses --follow: empty diffs or ignoring the
log's file restriction.

Closes #3680.


tarsius:
Thanks!

PR #3795, opened 2019-03-19
Fix typo in `magit-cherry-pick:--mainline`

kini:
The transient library doesn't have a function called `transient-read-natural-number-N+`.  This seems to be a typo for `transient-read-number-N+`, which does exist, and makes `A - m` work as expected rather than suddenly closing the transient.


tarsius:
Thanks!

PR #3810, opened 2019-03-28
Fixing git error while trying to stash

h0t0sh0:
Hello!
I recently started getting errors from git when trying to make stash(Magit shortcut: **z z**):

```
128 git … reset --hard HEAD
fatal: ambiguous argument 'HEAD': both revision and filename
```

My PR is  attempt to fix this.
I'm not a sophisticated git user, so comments and corrections are welcome.


kyleam:
> I recently started getting errors from git when trying to make stash

Your working tree must have a HEAD file, which is presumably unintentional (?).  Anyway, we certainly should disambiguate in these spots.  I've made a slight tweak to your commit message.

Thanks!

PR #3816, opened 2019-04-02
Super small messaging edit - less passive aggresive

patrickeganfoley:
Friendlier magit / less passive aggresive.  👍 

Does not change any code in any way, just the messaging.

### DIFF
--- a/lisp/magit-ediff.el
+++ b/lisp/magit-ediff.el
@@ -329,7 +329,7 @@ mind at all, then it asks the user for a command to run."
        (cond ((not command)
               (call-interactively
                (magit-read-char-case
-                   "Failed to read your mind; do you want to " t
+                   "Do you want to " t
                  (?c "[c]ommit"  'magit-ediff-show-commit)
                  (?r "[r]ange"   'magit-ediff-compare)
                  (?s "[s]tage"   'magit-ediff-stage)


tarsius:
```diff
-Failed to read your mind; do you want to <list of available ediff variants>
+Do you want to <list of available ediff variants>
```

You are not the first to consider this message to be passive-aggressive. However more users mentioned that they find it hilarious.

The current message already is an abbreviation. By removing the "passive-aggressive" part some information is lost. The "full message" is:

> This dwim command uses a set of rules to determine which ediff variant should be used based on the cursor position. None of these rules applied given the current location. Do you want to {list of available ediff variants}

Personally I wouldn't read the current message as

> You provided silly input, so you have to provide the required information explicitly.

but rather

> We take pride in the quality of our dwim commands, we are terribly sorry, but this sadly failed in this case.

----

This is not the hill I want to die on, but I also don't want to just click "merge" to make it go away. The current message hopefully is not offensive to anyone. The worst that can happen is that it is interpreted as passive-aggressive; which has happened a few times now. Magit has other Easter-eggs, so even if we remove this one, it won't be without any humor build in.

So...

* What do others think?
* Any suggestions for an alternative message, which doesn't lose any meaning and is not considerably longer?

I can totally see how this can be interpreted as passive-aggressive. I won't even deny that it is possible that I actually was frustrated when I wrote it. But I am ambivalent about self-censorship for the sake of avoiding any possible offense. I mean what's the worst that can happen? Someone thinking that **I** am a bit of an ass. Maybe that means that that user won't donate. Should I really censor myself because of that?

Honest question. I think in the past I initially resist the removal of the word "insane", but later agreed that it should be avoided. I feel the issue at hand is different though.

tarsius:
> While we take pride in the quality of our dwim command, we are terribly sorry to inform you that this sadly failed in this case.

Actually that's rather funny I think. But too long.

npostavs:
> I can totally see how this can be interpreted as passive-aggressive.

Yeah, I can sort of see it, though a computer trying to be rude always reads as humorous to me.

> * Any suggestions for an alternative message, which doesn't lose any meaning and is not considerably longer?

How about this (slightly shorter even, though not quite as colourful):

```diff
-Failed to read your mind; do you want to <list of available ediff variants>
+Unsure what you mean; do you want to <list of available ediff variants>
```

Acronym based humour can be even shorter still, but probably not clear enough:
```
DWIM failed, try SWIM instead: <list of available ediff variants>
```

patrickeganfoley:
I kind of like ```Could not determine which ediff variant should be used based on the cursor position.  Do you want to {list of available ediff variants}``` or just ```Could not determine which ediff variant to use.  Do you want to {list of available ediff variants}``` if that's a bit too long. 

I think for me this might stem from a few other Emacs packages having sort of rude/passive-aggressive tones (e.g. ["might as well have a documentation string"](https://github.com/flycheck/flycheck/blob/master/test/flycheck-test.el#L3184) ) and it just kind of sticks out when a program momentarily adopts human tone and sounds like a person complaining at you.  

Anyway thanks for all the work you do maintaining magit!  Love using it!

tarsius:
>> I can totally see how this can be interpreted as passive-aggressive.

> Yeah, I can sort of see it, though a computer trying to be rude always reads as humorous to me.

Right?! I feel the same way.

> I think in the past I initially resist the removal of the word "insane", but later agreed that it should be avoided. I feel the issue at hand is different though.

The difference is that the use of the word "insane" could make someone with health issues uncomfortable and I don't want that. Everyone is welcome.

A prompt that might be considered rude is way different from using language that could make someone feel excluded. I think this particular message is funny. If others disagree that's reasonable too, but this does not put anyone down, not even unintentionally.

I don't want to remove that message, it feels wrong to refrain from trying to be funny once in a while just because someone else might think its not all that funny.

PR #3817, opened 2019-04-02
Fix magit-status "Merge:" header commit+ref order

jguenther:
To be consistent with the "Head:" ordering when `magit-status-show-hashes-in-headers` is enabled (seems to have happened in 86c095ab)

Before:
image

After:
image

Haven't looked into adding a regression test yet but I can if necessary.


tarsius:
Thanks!

PR #3819, opened 2019-04-04
Various auto-revert fixes

basil-conto:
From oldest to newest, the three commits are in order of decreasing importance and increasing controversiality:
1. Fix a logic error in `auto-revert-buffers--buffer-list-filter` which used `auto-revert-buffer-list-filter` as a variable, rather than a function.
2. Audit auto-revert-related docs for typos, grammar, and inconsistencies (such as `auto-revert-mode` being the only mode labelled a Command rather than a User Option).
3. Minor `magit-autorevert.el` code touch-ups.

I'm happy to reword/amend commits, but feel free to also make any changes you see fit yourselves. Note that someone else will have to update the Texinfo manual for me.

Thanks.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -720,7 +720,7 @@ refresh explicitly.
 By default Magit automatically reverts buffers that are visiting files
 that are being tracked in a Git repository, after they have changed on
 disk.  When using Magit one often changes files on disk by running
-git, i.e. "outside Emacs", making this a rather important feature.
+Git, i.e. "outside Emacs", making this a rather important feature.
 
 For example, if you discard a change in the status buffer, then that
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks!

I change the last commit to continue to require `cl-lib` at runtime. It's gonna be loaded by other libraries anyway and when we did the same `subr-x` that lead to issues and I don't want to risk similar issues just in order to "do the right thing".

basil-conto:
> I change the last commit to continue to require `cl-lib` at runtime. It's gonna be loaded by other libraries anyway and when we did the same `subr-x` that lead to issues and I don't want to risk similar issues just in order to "do the right thing".

Fair enough, thanks!

Issue #3833, opened 2019-04-14
status buffer not automatically refreshed on file changed inside emacs


zakora:
## Steps to reproduce
1. Do `magit-status` in a git folder, displays:
```
Head:     master initial commit

Recent commits
0aa0c82 master initial commit
```
2. Change a versioned file from this git repository.
3. Save that file within emacs

## Expected behavior
The `magit-status` buffer should show:
```
Head:     master initial commit

Unstaged changes (1)
modified   README

Recent commits
0aa0c82 master initial commit
```

## Actual behavior
Magit status buffer is not updated, showing the same:
```
Head:     master initial commit

Recent commits
0aa0c82 master initial commit
```
Typing `g` in this buffer actually refreshes to the current status.

## Notes
The [screenshots page](https://magit.vc/screenshots/#status-is-automatically-updated) says:

> If you use Emacs to edit and save a file which is located inside the repository or if you use Magit to perform some action such as staging a change, then the status buffer is automatically updated.

But maybe my interpretation is wrong, since somewhere else in the [docs](https://magit.vc/manual/magit/Automatic-Refreshing-of-Magit-Buffers.html) it is less clear that saving a file would always trigger an automatic update of magit status:

>  The status buffer may optionally be automatically refreshed whenever a buffer is saved to a file inside the respective repository. 

### `magit-version`
```
Magit 20190413.1201, Git 2.17.1, Emacs 25.2.2, gnu/linux
```
from MELPA.

### debug mode

Tested the reproducibility steps above with:
```
/usr/bin/emacs25 -Q --eval \(setq\ debug-on-error\ t\) -L /home/zakora/.emacs.d/elpa/dash-20190413.1058/ -L /home/zakora/.emacs.d/elpa/transient-20190319.1928/ -L /home/zakora/.emacs.d/elpa/with-editor-20190224.931/ -L /home/zakora/.emacs.d/elpa/magit-20190413.1201/ -L /home/zakora/.emacs.d/elpa/git-commit-20190413.1201/ -l /home/zakora/.emacs.d/elpa/magit-20190413.1201/magit
```

tarsius:
You have to enable that feature using:

```lisp
(add-hook 'after-save-hook 'magit-after-save-refresh-status t)
```

zakora:
Thank you for your answer. It does solve my issue!

I made an attempt to clarify the documentation  about this with PR magit/magit.vc#11 and #3834 .
Feel free to integrate or discard them.

PR #3834, opened 2019-04-15
update documentation on automatic buffer refresh

zakora:
Github: magit/magit#3833

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -8,7 +8,7 @@
 #+TEXINFO_DIR_CATEGORY: Emacs
 #+TEXINFO_DIR_TITLE: Magit: (magit).
 #+TEXINFO_DIR_DESC: Using Git from Emacs with Magit.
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks!

I was considering just enabling it by default, but lets leave that for another time.

You made the edit on Github, which I am forced to consider an unsuitable interface for editing code and creating commits.  Its the little things.  E.g. when using `git`, then every commit message ends with a newline character.  When using Github, then they don't unless you explicitly put it there; which nobody does.

We would like to properly attribute changes to people.  It may be possible to configure Github to do that, but the way you have configured it the commit is attribute to your Github alias and some dummy email.  We would like to use your real name and email; but won't force you to do so.

Please put an empty line between the paragraph and the code block.  Also

``` diff
-#+begin_src elisp
-#+BEGIN_SRC emacs-lisp
```

zakora:
Thank you for taking the time to look at this PR and explaining the details.

I updated the commit with magit so it should not contain any Github specific things now, and also changed the paragraph to match your suggestion.

Hope it's good now.

PR #3835, opened 2019-04-16
magit-worktree-branch: Fix bug where start-point is ignored

CeleritasCelery:
If `magit-worktree-branch` is called interactively, it will read a `start-point` but ignore it. Instead it will use the current branch. This PR fixes that.


kyleam:
Thanks!  (I tweaked your commit message to add a bit more information.)

PR #3836, opened 2019-04-17
Fix and improve handling of infix arguments and buffer-local argument variables

### LINKED ISSUE #3750: Saved transient values are not restored for magit-{log*,diff*,show-refs}
>  (1) what behavior you expected

When `~/.emacs.d/transient/values.el` has the following contents,
```
((magit-log "-n256" "--graph" "--color" "--decorate"))
```

In magit-log, `--color` option is selected as well as `-n256`, `--graph` and `--decorate`.

>  (2) what behavior you observed

`--color` option is not selected, but other (default) options are selected.

>  (3) and how we can reproduce the issue.

1. Execute `magit-status`.
2. Type <kbd>l</kbd>, <kbd>-c</kbd>, and then <kbd>C-x C-s</kbd>
3. Restart emacs.
4. Execute `magit-status`.
5. Type <kbd>l</kbd>.

>  M-x magit-version RET

`Magit 20190220.2315, Git 2.20.1, Emacs 26.1, gnu/linux`

tarsius:
1. Most importantly this fixes setting and saving the diff, log and refs arguments.  That's a feature that broke with the switch to Transient.  Fixes #3750.  For details see:

   *  log: Fix and refactor argument handling
   *  diff: Fix and refactor argument handling
   *  refs: Sanitize argument handling

   Originally I tried to do this without all the other changes, but that proofed to be too difficult.  To much effort would have been required to not change things I intended to change anyway (but at a later time).
  
2. Deprecate the `magit-refresh-args` variable.  That variable was a central part of the abstraction to map arguments as specified in popups/transients to arguments that are in effect in buffers.  The "refresh" in it was due to these arguments having to remain in effect when the buffer was refreshed.

   * Define more buffer-local variables
   * magit-refresh-args: Deprecate

   The abstraction is still there, but the variable is not.  A central part of the abstraction is that for every mode derived from `magit-mode` there should be a function `magit-*-refresh-buffer`, which is responsible for refreshing the buffer while ensuring that the arguments remain in effect.  That function wasn't implemented for every mode; now it is.  
   
   Previously this function received the refresh arguments as function arguments.  `magit-mode-setup` was responsible for calling this function while passing it `magit-refresh-arguments` as arguments.  Now `magit-*-refresh-buffer` functions are called without any arguments and they have to retrieve the arguments themselves.
   
   The problem with `magit-refresh-args` was that a **lot** of `(nth N magit-refresh-args)` was going on.  Now we use many different variables named `magit-buffer-MEANINGFUL` which is more convenient because `MEANINFUL` is more informative than `N`.
   
   This change is split across many commits.  It is done for each mode individually and for each mode it usually involved more than one commit.  I tried to ensure that intermediate commits did not break anything, but there might be instances where that failed.  Splitting across multiple commits certainly helped detecting omissions that I would not have been able to spot in one humongous commit.
   
   Third-party code that still relies on `magit-refresh-args` should continue to work.
   
3. Along with the deprecation of `magit-refresh-args` the "infix argument to effective arguments" abstraction has seen some other important changes and deprecations.

   1. Most importantly like there now is a `magit-*-refresh-buffer` function for every mode there now is a `magit-*-setup-buffer` function.  Magit commands that run Git commands to display the information in a dedicated buffer now do so by running the respective `magit-*-setup-buffer` function.  That function sets the buffer-local arguments values using the `magit-setup-buffer` macro.
   
      For some modes such a function existed before but neither the names nor interfaces  were consistent.  In many cases no such function existed and each respective command did the setup individually, which lead to even more inconsistency.
	 
	  Now we use `magit-*-setup-buffer` even if there is just a single command and when it might look like overkill when looked at in isolation.
   
   2. The `magit-setup-buffer` marco replaces the `magit-mode-setup` buffer.  The mean difference is that the former sets many variables, while the latter sets `magit-refresh-args`.
   
      * magit-setup-buffer: New macro
	  * magit-mode-setup{,-internal}: Deprecate
	  
      It is possible to use `magit-setup-buffer` to set `magit-refresh-args`, i.e. to only adapt partially to the new interface, but for compatibility reasons `magit-mode-setup` is preserved for the time being.
	  
	  Like its predecessor `magit-setup-buffer` calls `magit-refresh-buffer` (see what I did here when I deprecated the old implementation ;-), which then calls the appropriate `magit-*-refresh-buffer` function.
	  
  Lets summarize how the abstraction works:
  
  1. A suffix command receives the infix arguments of the respective transient prefix command (or it retrieves the arguments through other means, see below) and calls `magit-*-setup-buffer`.
  
  2. `magit-*-setup-buffer` maps the received arguments to the names of buffer-local variables and hands that alist of to `magit-setup-buffer`.
  
  3. `magit-setup-buffer` sets the buffer-local values of the variables according to the received alist.

  4. `magit-setup-buffer` calls `magit-refresh-buffer`, which re-generates the buffer contents using the values it gets from the buffer-local variables.  How exactly this happens is not standardized.  It may call `git` directly, handing it the values of the variables.  Or it may call `magit-*-buffer-hook`, and the functions on that hook do their thing while using the values of the variable.
  
  The above is what happens when the user invokes a command in order to display **something else**.  The user may also wish to display the same thing but change **how** it is presented.  For the most part that is limited to how diffs and logs are presented.  For diffs and logs "refresh" prefix commands, `magit-diff-refresh` and `magit-log-refresh`, exist.  The allow the user to change how the diff(s)/log(s) in the current buffer are presented, while `magit-diff` and `magit-log` allow the user to change what diff/log is shown (and optionally also change how).
  
  1. The suffixes of the "refresh" prefixes only change the variables that control how the information is presented.  These commands do not call `magit-*-setup-buffer` (and therefore `magit-setup-buffer` isn't called either).
  
  2. Instead they update the appropriate (i.e. the "how") buffer-local variables and then call `magit-refresh`, which calls `magit-refresh-buffer`, which calls `magit-*refresh-buffer`.
  
  This is how `magit-setup-buffer` relates to `magit-refresh-buffer` and `magit-*-setup-buffer` relates to `magit-*-refresh-buffer`.  "Setup" means "setup plus 'initial refresh'".  "Setup" involves things that do not have to be done again during subsequent "refresh", such as displaying the buffer in some window.
  
3. (aka **4**) Magit supported "sticky" diff and log (and refs) arguments for quite some time now.  Sticky means that the arguments that are effective in the appropriate buffer are used when invoking a command that is going to display something else in that buffer.  In other words the stickiness prevents a change in the arguments that control **how** the diff/log is displayed when changing **what** is being displayed.

   * Change stickiness of infix arguments

   That is not universally desirable though.  If you for example change the diff arguments while looking at some particularly difficult to understand diff, then you might not want those settings to stick around.  If you then go on to do different things, which cause the diff buffer to be buried and later show another diff, then it might be an unwelcome surprise for those arguments to have stuck around.
   
   Therefore this pr makes the buffer-local arguments less sticky.  The buffer-local variables are only used now if the respective buffer is shown in a window of a current frame, the reasoning being that if you can see what is going to happen, then you are not going to be surprised when it does happen.
   
   There are two variables that control this now, `magit-prefix-use-buffer-arguments` and `magit-direct-use-buffer-arguments`, previously there was only one.  One of them controls what the initial infix argument are going to be.  The other controls what happens if you invoke an argument-using command directly, i.e. if you do not get a change to see the arguments that are going to be used before they are being put into effect.
   
   I recommend you don't change the stickiness.  I think we have finally arrived at a reasonable default behavior.
  
4. The diff transient features some suffixes that do not update the `magit-diff-mode` buffer.  These commands `magit-show-commit` and `magit-stash-show` are usually invoked by pressing `RET` on the commit in a lot or list of stashes.  In that case they used to get the arguments from the `magit-diff-buffer` anyway.  Now they get it from the appropriate buffer instead (subject to `magit-direct-use-buffer-arguments`).

   * log: Fix and refactor argument handling
   * diff: Fix and refactor argument handling

5. Support for `bookmark` has been re-implemented from scratch.  There are some small differences and `magit-submodule-list-mode` isn't supported anymore because it does not derive from `magit-mode` and only that is supported now.

   * Restore bookmark support

6. Lots of other small changes and bugfixes.

Piuuu....


npostavs:
> If the buffer is not being displayed in a buffer

(In the first commit message) I think you mean, "in a window".

tarsius:
It would be nice if Markdown did just use the numbers I used for the numbered list items instead of going "Ha!! You got the intention wrong dufus".

tarsius:
I still need to debug some issue.

PR #3839, opened 2019-04-19
[WIP] Add `magit-clone-transient`

cireu:
A new transient command for `magit-clone`

图片

Now I meet some problems

1. Some options only significant when some other options were turned on. (`--share` or `--no-hardlink` must use with `-l`),  I can't find a way to control user which options are disable with current transient.

2. Some options should be used like `--option value` or even `--option valA --option valB`. I don't have good idea to deal with these options too.

### DIFF
--- a/lisp/magit-clone.el
+++ b/lisp/magit-clone.el
@@ -62,6 +62,24 @@ with the remote url as only argument and use the returned value."
 
 ;;; Commands
 
+;;;###autoload (autoload 'magit-clone-transient "magit-clone" nil t)
+(define-transient-command magit-clone-transient ()
+  "Clone a repo from remote."
+  :man-page "git-clone"
+  :dependant '(("--\\(?:no-hardlinks\\|shared\\)" . "--local"))
+  ["Local Clone"
+   ("-l" "Clone from local remote" "--local")
+   (magit-clone:=T)]
+  ["Shallow Clone"
+   (magit-clone:--depth)
+   (magit-clone:--shallow-since)
+   ("=t" "Don't clone any tags" "--no-tags")]
+  ["Submodules"
+   ("=R" "Recursive Clone" "--recursive")
+   ("=M" "Shallow Clone Submodules" "--shallow-submodules")]
+  ["Actions"
+   ("RET" "Continue" magit-clone)])
+
 ;; Local Clone
 
 (define-infix-argument magit-clone:=T ()
@@ -87,7 +105,7 @@ with the remote url as only argument and use the returned value."
   :class 'transient-option
   :key "=S"
   :argument "--shallow-since="
-  :reader 'magit-read-date)
+  :reader 'transient-read-date)
 
 ;; (define-infix-argument magit-clone:--shallow-exclude ())
 
@@ -99,32 +117,33 @@ with the remote url as only argument and use the returned value."
 ;;   :argument "-j"
 ;;   :reader 'transient-read-number-N+)
 
-
-(defun magit-read-date (&rest _)
-  "Read a date from minibuffer, format it to \"YYYY-MM-DD\"."
-  (->> (calendar-read-date)
-       (nreverse)
-       (apply #'format "%04d-%02d-%02d")))
-
-;;;###autoload (autoload 'magit-clone-transient "magit-clone" nil t)
-(define-transient-command magit-clone-transient ()
-  "Clone a repo from remote."
-  :man-page "git-clone"
-  ["Local Clone"
-   ("-l" "Clone from local remote" "--local")
-   (magit-clone:=T)]
-  ["Shallow Clone"
-   (magit-clone:--depth)
-   (magit-clone:--shallow-since)
-   ("=t" "Don't clone any tags" "--no-tags")]
-  ["Submodules"
-   ("=R" "Recursive Clone" "--recursive")
-   ("=M" "Shallow Clone Submodules" "--shallow-submodules")]
-  ["Actions"
-   ("RET" "Continue" magit-clone)])
-
-
 (defun magit-clone--arguments ()
+  (let* ((transient-arg (magit-clone--arguments-1))
+         (localp (member "--local" transient-arg))
+         (source (if localp
+                     (read-directory-name "Clone repository: ")
+                   (magit-read-string-ns "Clone repository"))))
+    (if localp
+        (list source
+              (read-directory-name
+               "Clone to: "
+               (if (functionp magit-clone-default-directory)
+                   (funcall magit-clone-default-directory source)
+                 magit-clone-default-directory)
+               nil nil nil)
+              transient-arg)
+      (list source
+            (read-directory-name
+             "Clone to: "
+             (if (functionp magit-clone-default-directory)
+                 (funcall magit-clone-default-directory source)
+               magit-clone-default-directory)
+             nil nil
+             (and (string-match "\\([^/:]+?\\)\\(/?\\.git\\)?$" source)
+                  (match-string 1 source)))
+            transient-arg))))
+
+(defun magit-clone--arguments-1 ()
   "Get Arguments from `magit-clone-transient'."
   (-when-let (args (transient-args (and (eq current-transient-prefix
                                             'magit-clone-transient)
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks!

I have to admit that `git clone` is one of the very few things I still do on the command-line, which probably explains why `magit-commit` isn't fancier yet.  I also never use any of these arguments, so I will have to read up on those to be able to better answer your questions.

Helping you with the issues you are running into will probably amount to more work for me than if I just did this by myself, but if you are willing for this to take a while, then we can do that.

> I can't find a way to control user which options are disable with current transient.

I cannot parse that sentence.  What is it that you cannot figure out how to do?

> or even `--option valA --option valB`.

There isn't currently any support for specifying the same argument twice.

> Refactor: Extract `magit-clone` to `magit-clone--internal`

Good idea to do that in a separate commit, but actually this does quite a bit more.

> Add `magit-clone-transient`

The way we do this now is that we name the transient prefix command `magit-clone`.  There are some other prefixes that only have a single suffix (though whether that should be done here is to be seen).  In such cases the prefix and suffix can actually be the same command, see `magit-patch-apply` for an example.

cireu:
> I cannot parse that sentence. What is it that you cannot figure out how to do?

Sorry for my bad english :sweat_smile: . As I referred, `--no-hardlinks` only significant when `-l` was turned on. But now user can still turn on `--no-hardlinks` without `-l`, we have to do extra work to clean these useless arguement(You can see the comment in `magit-clone--arguments`)

I hope we can make some options depend on other options. When their dependencies were turned off, their value will be ignored and their keys won't be avaible.

It sounds like `:if` `:if-not` predicator, but seems that `:if` `:if-not` can't do this job interactively.

> Good idea to do that in a separate commit, but actually this does quite a bit more.

I can do a rebase on it if neccessary.

> There isn't currently any support for specifying the same argument twice.

Some git commands need this feature, You can check the man page of `git-clone`, and you will see `--recuse-submodules` can be given multiple times to specific more than one value.

Maybe we should extend transient to do this


> The way we do this now is that we name the transient prefix command magit-clone. There are some other prefixes that only have a single suffix (though whether that should be done here is to be seen). In such cases the prefix and suffix can actually be the same command, see magit-patch-apply for an example.

Thanks for your notice, but I'm afraid that we will break the behavior of `magit-clone`,especially for who using `magit-clone` in lisp to do batch process.

tarsius:
> I hope we can make some options depend on other options. When their dependencies were turned off, their own value will be ignored and their keys won't be avaible.

I implemented something similar yesterday. Support for disabling other incompatible arguments when turning on an argument.

I have now done something similar for automatically enabling required arguments. Add this to the prefix definition:

```emacs-lisp
:dependant '(("--\\(?:no-hardlinks\\|shared\\)" . "--local"))
```

now enabling `--no-hardlinks` or `--shared` enables `--local`, but disabling `--local` does not yet disable the former too. You will need to use `transient`'s `dependant` branch for this.

----

Please rebase your branch on top of `master`.

lisp/magit-clone.el
```diff
@@ -61,24 +62,120 @@ with the remote url as only argument and use the returned value."
...
+
+
+(defun magit-read-date (&rest _)
+  "Read a date from minibuffer, format it to \"YYYY-MM-DD\"."
+  (->> (calendar-read-date)
+       (nreverse)
+       (apply #'format "%04d-%02d-%02d")))
+
+;;;###autoload (autoload 'magit-clone-transient "magit-clone" nil t)
+(define-transient-command magit-clone-transient ()
```
tarsius: The prefix command should be the first command.

lisp/magit-clone.el
```diff
@@ -61,24 +62,120 @@ with the remote url as only argument and use the returned value."
...
+;; Submodules
+
+;; (define-infix-argument magit-clone:--jobs ()
+;;   :description "Specific the how many jobs to clone submodules."
+;;   :class 'transient-option
+;;   :argument "-j"
+;;   :reader 'transient-read-number-N+)
+
+
+(defun magit-read-date (&rest _)
```
tarsius: Probably use `transient-read-date` instead.

lisp/magit-clone.el
```diff
@@ -61,24 +62,120 @@ with the remote url as only argument and use the returned value."
...
 Then show the status buffer for the new repository."
   (interactive
-   (let  ((url (magit-read-string-ns "Clone repository")))
-     (list url (read-directory-name
+   (let* ((transient-arg (magit-clone--arguments))
+          (localp (member "--local" transient-arg))
+          (source (if localp
+                      (read-directory-name "Clone repository: ")
+                    (magit-read-string-ns "Clone repository"))))
+     (if localp
```
tarsius: The THEN and ELSE parts are nearly identical. Instead keep only the ELSE part and move `localp` its `and`.

lisp/magit-clone.el
```diff
@@ -61,24 +62,120 @@ with the remote url as only argument and use the returned value."
...
+             transient-arg))))
+  (magit-clone--internal repository directory args))
+
+(defun magit-clone--internal (repository directory args)
+  "Do the real job for `magit-clone'"
   (run-hooks 'magit-credential-hook)
   (setq directory (file-name-as-directory (expand-file-name directory)))
-  (magit-run-git-async "clone" repository
-                       (magit-convert-filename-for-git directory))
+  (apply #'magit-run-git-async "clone"
```
tarsius: `apply` is unnecessary here. `magit-run-git-async` automatically flattens its arguments.

lisp/magit-clone.el
```diff
@@ -61,24 +62,120 @@ with the remote url as only argument and use the returned value."
...
+         `(,@args "--" ,repository ,(magit-convert-filename-for-git directory)))
```
tarsius: Are you sure `--` should come before `repository`?

cireu: Probably sure, see https://www.git-scm.com/docs/git-clone

```
git clone [--template=<template_directory>]
	  [-l] [-s] [--no-hardlinks] [-q] [-n] [--bare] [--mirror]
	  [-o <name>] [-b <name>] [-u <upload-pack>] [--reference <repository>]
	  [--dissociate] [--separate-git-dir <git dir>]
	  [--depth <depth>] [--[no-]single-branch] [--no-tags]
	  [--recurse-submodules[=<pathspec>]] [--[no-]shallow-submodules]
	  [--jobs <n>] [--] <repository> [<directory>]
```

cireu:
> Edit: Actually I don't see a need to split that out into a separate function. It would probably be best if this pr was a single commit, at least for the time being.

I extract it becasue `magit-clone` have a long and winding interactive form. It seems that a better idea is extract interactive form to another function.

lisp/magit-clone.el
cireu: I don' t consider they are identical, anyway, using `read-directory-name` while doing a local clone is more convenient.

lisp/magit-clone.el
tarsius: You are right.

lisp/magit-clone.el
tarsius: But they are nearly identical:

```diff
diff --git a/then.el b/else.el
index c83cbf5..d29f103 100644
--- a/then.el
+++ b/else.el
@@ -4,5 +4,7 @@ (list source
        (if (functionp magit-clone-default-directory)
            (funcall magit-clone-default-directory source)
          magit-clone-default-directory)
-       nil nil nil)
+       nil nil
+       (and (string-match "\\([^/:]+?\\)\\(/?\\.git\\)?$" source)
+            (match-string 1 source)))
       transient-arg)
```

My suggestion is that you remove the `if` and the THEN part and instead make this change to the ELSE part:

```diff
diff --git a/else.el b/shared.el
index d29f103..afe10f0 100644
--- a/else.el
+++ b/shared.el
@@ -5,6 +5,7 @@ (list source
            (funcall magit-clone-default-directory source)
          magit-clone-default-directory)
        nil nil
-       (and (string-match "\\([^/:]+?\\)\\(/?\\.git\\)?$" source)
+       (and (not localp)
+            (string-match "\\([^/:]+?\\)\\(/?\\.git\\)?$" source)
             (match-string 1 source)))
       transient-arg)

```

tarsius:
I wanted to add a cloning transient/popup for a long time, but never got around to it. This pr finally gave me the necessary push. Thanks for that.

My implementation takes a different approach than yours. Most importantly I added several suffix commands that implement different "kinds of cloning" (e.g. regular, bare and sparse), instead of adding the respective arguments to the transient and only a single suffix command. (Note that you have to invoke `magit-clone` with a prefix argument, otherwise it just acts like `magit-clone-regular`, i.e. it invokes `git clone` directly without showing a transient popup first.)

----

Above we discussed how enabling/disabling certain arguments can enable/disable other arguments. After reading `git-man(1)` I have concluded that is not necessary here.

Most importantly it appears that you have misunderstood what `--local` does. It controls *how* the cloning is done if and only if the repository that is being cloned is a local repository. If that argument is used when cloning a remote repository, then it is simply ignored. `--share` and `--no-hardlinks` can be used with or without `--local` and these arguments too are ignored when cloning a remote repository.

Issue #2821, opened 2016-10-11
whitespace-mode kills colors


lazywithclass:
Thanks for the efforts coding this!

I reproduced this bug with plain Emacs with magit, then `M-x magit-status`, then open latest commit and notice the presence of colors.
`whitespace-mode` and all the colors are gone from the buffer.

Tested with Magit 20161011.744, Git 2.8.4 (Apple Git-73), Emacs 25.1.1, darwin

This was originally reported in Spacemacs https://github.com/syl20bnr/spacemacs/issues/4940

tarsius:
`whitespace-mode` uses font-lock to do its thing. Magit does not use font-lock to to add faces/colors to its buffers. That makes them incompatible because turning on font-lock removes all faces keeping only those that it adds itself.

The only thing we could do is to somehow prevent `whitespace-mode` from being turned on in magit buffers. And because this issue comes up ever so often, I have done that now. For an alternative see `magit-diff-paint-whitespace`.

lazywithclass:
Thanks a lot for the prompt answer.

Issue #3394, opened 2018-03-15
Magit conflicts with rainbow mode. No color shown.


junhaowww:
If I set rainbow mode to a global condition, no color except gray and white is applied to the magit mode. I found it strange and took for a while to locate the issue. Thanks a lot!

macOS 10.13.2
Emacs 25.3 

Here is my config.
``` elisp
(require 'rainbow-mode)
;; conflict with magit
(define-globalized-minor-mode my-global-rainbow-mode rainbow-mode
(lambda () (rainbow-mode 1)))
(my-global-rainbow-mode 1)
```

[rainbow-mode.el](https://github.com/emacsmirror/rainbow-mode)

tarsius:
The problem is probably the same as for `whitespace-mode`.

> Because `whitespace-mode' uses font-lock and Magit does not, they are not compatible.  Therefore you cannot turn on that minor-mode in Magit buffers.  If you try to enable it anyway, then this advice prevents that.

tarsius:
These other modes are to blame - they assume that every buffer uses Font-Lock mode and that it is therefore safe for them to enable themselves everywhere.

junhaowww:
I see. Thank you!

Issue #2942, opened 2017-01-09
Diffs should optionally support syntax highlighting


tarsius:
After adding this feature, the line-prefix (`+`, `-`, or ` `) continues to be visible and shows whether a given line is an added, removed, or context line. And as before the background color (red/green/gray) continues to be used for the same purpose, but the foreground color.

(When there is a hunk-internal region, then the parts outside of the region currently lose the background color, so that's a first complication to think about.)

Roughtly speaking this feature requires that for each modified file both versions (`A..B`) have to be visited in a buffer in which font-lock is enabled.  After font-lock is done, the text properties are extracted and re-applied in the buffer showing the diff.

This obviously is costly and therefore has to be done asynchronously.  For large diffs it is quite possible that the user initially sees a diff with no syntax highlighting and that that only appears after a delay.

---

This is only one of the many improvements I want to make to diffs and it will be a while until I get to it.  It's also not one of the most important improvements, improved performance very clearly has to be the priority.  Also the changes that I intend to make in order to achieve better performance are actually prerequisits for this feature.  Not only in the sense that we "cannot make diff parsing even slower just to get some additional eye-candy", but also for technical reasons.  Asynchronous and selective diff parsing in a separate buffer is a most before the feature proposed here can be implemented.)

tarsius:
This could also be used to highlight whitespace errors. In order to do that `whitespace-mode` would have to be enabled in the two file-visiting buffers and the text-properties set by that would have to be copied along with the text-properties set by font-lock.

dsvensson:
How does this relate to the new Emacs 27 `diff-font-lock-syntax`?

https://github.com/emacs-mirror/emacs/commit/6973b1489b24ca4190d24be9e5f887aef2cc9eff

tarsius:
`diff-mode` beat `magit-diff-mode` to it.

tarsius:
I still intend to implement this eventually but since I (also still) plan to give the whole diffing functionality a major overhaul, I will delay doing the former until it becomes part of the latter.

rirze:
Is this a feature that could benefit from tree-sitter?

Pardon me if I'm misunderstanding the feature.

jperras:
@rirze fwiw you can get diff highlights via [`magit-delta`](https://github.com/dandavison/magit-delta) and something like the following in your `config.el`:

```elisp
(use-package! magit-delta
  :after magit
  :config
  (setq
    magit-delta-default-dark-theme "Monokai Extended"
[...]

PR #3840, opened 2019-04-19
make magit-mode compatible with whitespace-mode

### LINKED ISSUE #3885: magit loses faces when num3-mode active
### Context

Magit v2.90.1-423-g0abc761f, Git 2.21.0, Emacs 26.2, darwin

I have num3-mode globally enabled, but magit status modes and num3-mode interact poorly.

### Expected

magit faces in all their glory even using num3-mode:

Screenshot 2019-05-23 at 10 27 44

### Observed

Most of the faces were missing. See attached images.

Screenshot 2019-05-23 at 10 31 05

3. I can reproduce using the following command-line, inspired by M-x magit-emacs-Q-command:

```
/Applications/Emacs.app/Contents/MacOS/Emacs-x86_64-10_14 -Q --eval \(setq\ debug-on-error\ t\) -L /Users/andstrib/.emacs.d/elpa/dash-2.15.0/ -L /Users/andstrib/.emacs.d/straight/repos/transient/lisp/ -L /Users/andstrib/.emacs.d/elpa/with-editor-2.8.1/ -L /Users/andstrib/.emacs.d/straight/repos/magit/lisp/ -L  /Users/andstrib/.emacs.d/straight/repos/num3-mode/ -l /Users/andstrib/.emacs.d/straight/repos/magit/lisp/magit --eval \(load-library\ \"num3-mode\"\) --eval \(global-num3-mode\)
```

Then go to a git repo and run `M-x magit`.

The reason this feels like a Magit bug is that num3-mode works fine with (eg) programming language and `org-mode` major modes.

npostavs:
https://github.com/magit/magit/issues/2821#issuecomment-253082085 says

> `whitespace-mode` uses font-lock to do its thing. Magit does not use font-lock to to add faces/colors to its buffers. That makes them incompatible because turning on font-lock removes all faces keeping only those that it adds itself.
> 
> The only thing we could do is to somehow prevent `whitespace-mode` from being turned on in magit buffers.

But it looks like all that's needed to make magit compatible with `whitespace-mode` is `s/'face/'font-lock-face/` (I've only done a quick search&replace without reformatting, if this approach is okay I can go back and reformat lines which got longer as needed). The Elisp manual sort of mentions this in [(elisp) Precalculated Fontification](https://www.gnu.org/software/emacs/manual/html_node/elisp/Precalculated-Fontification.html); I'm think about adding some more explicit text about this there:

```diff
modified   doc/lispref/modes.texi
@@ -3201,7 +3201,12 @@ is disabled, @code{font-lock-face} has no effect on the display.
   It is ok for a mode to use @code{font-lock-face} for some text and
 also use the normal Font Lock machinery.  But if the mode does not use
 the normal Font Lock machinery, it should not set the variable
-@code{font-lock-defaults}.
+@code{font-lock-defaults}.  In this case the @code{face} property will
+not be overriden, so using the @code{face} property could work too.
+However, using @code{font-lock-face} is generally preferable as it
+allows the user to control the fontification by toggling
+@code{font-lock-mode}, and lets the code work regardless of whether
+the mode uses Font Lock machinery or not.
```

Also fixes #3885.

### DIFF
--- a/Documentation/RelNotes/2.91.0.org
+++ b/Documentation/RelNotes/2.91.0.org
@@ -249,6 +249,10 @@
   now also kills all other Magit buffers associated with the current
   repository.  #3863
 
+- Magit buffers are now compatible with ~whitespace-mode~ (and other
+  modes which use font lock).  #3840
+
+
 ** Fixes since v2.90.0
 
 - Bumped the minimal required version of ~git-commit~ to the correct
--- a/lisp/magit-bisect.el
+++ b/lisp/magit-bisect.el
@@ -187,7 +187,7 @@ bisect run'."
                              (and bad-line (match-string 1 bad-line)))
         (magit-insert-heading
           (propertize (or bad-line (pop lines))
-                      'face 'magit-section-heading))
+                      'font-lock-face 'magit-section-heading))
         (dolist (line lines)
           (insert line "\n"))))
     (insert "\n")))
@@ -220,7 +220,8 @@ bisect run'."
           (narrow-to-region beg (point))
           (goto-char (point-min))
           (magit-insert-section (bisect-item heading t)
-            (insert (propertize heading 'face 'magit-section-secondary-heading))
+            (insert (propertize heading 'font-lock-face
+                                'magit-section-secondary-heading))
             (magit-insert-heading)
             (magit-wash-sequence
              (apply-partially 'magit-log-wash-rev 'bisect-log
--- a/lisp/magit-blame.el
+++ b/lisp/magit-blame.el
@@ -622,7 +622,7 @@ modes is toggled, then this mode also gets toggled automatically.
           magit-blame-separator))))
 
 (defun magit-blame--update-highlight-overlay (ov)
-  (overlay-put ov 'face (magit-blame--style-get 'highlight-face)))
+  (overlay-put ov 'font-lock-face (magit-blame--style-get 'highlight-face)))
 
 (defun magit-blame--format-string (ov format face)
   (let* ((chunk   (overlay-get ov 'magit-blame-chunk))
@@ -643,15 +643,15 @@ modes is toggled, then this mode also gets toggled automatically.
              (propertize (concat (if (string-prefix-p "\s" format) "\s" "")
                                  "Not Yet Committed"
                                  (if (string-suffix-p "\n" format) "\n" ""))
-                         'face face)
+                         'font-lock-face face)
            (magit--format-spec
-            (propertize format 'face face)
+            (propertize format 'font-lock-face face)
             (cl-flet* ((p0 (s f)
-                           (propertize s 'face (if face
-                                                   (if (listp face)
-                                                       face
-                                                     (list f face))
-                                                 f)))
+                           (propertize s 'font-lock-face (if face
+                                                             (if (listp face)
+                                                                 face
+                                                               (list f face))
+                                                           f)))
                        (p1 (k f)
                            (p0 (cdr (assoc k revinfo)) f))
                        (p2 (k1 k2 f)
@@ -670,15 +670,16 @@ modes is toggled, then this mode also gets toggled automatically.
                          (magit-blame--style-get 'margin-width))))
         (concat str
                 (propertize (make-string (max 0 (- width (length str))) ?\s)
-                            'face face))
+                            'font-lock-face face))
       str)))
 
 (defun magit-blame--format-separator ()
   (propertize
    (concat (propertize "\s" 'display '(space :height (2)))
            (propertize "\n" 'line-height t))
-   'face (list :background
-               (face-attribute 'magit-blame-heading :background nil t))))
+   'font-lock-face (list :background
+                         (face-attribute 'magit-blame-heading :background
+                                         nil t))))
 
 (defun magit-blame--format-time-string (time tz)
   (let* ((time-format (or (magit-blame--style-get 'time-format)
--- a/lisp/magit-branch.el
+++ b/lisp/magit-branch.el
@@ -759,8 +759,9 @@ and also rename the respective reflog file."
   [:description
    (lambda ()
      (concat
-      (propertize "Configure " 'face 'transient-heading)
-      (propertize (oref transient--prefix scope) 'face 'magit-branch-local)))
+      (propertize "Configure " 'font-lock-face 'transient-heading)
+      (propertize (oref transient--prefix scope)
+                  'font-lock-face 'magit-branch-local)))
    ("d"   magit-branch.<branch>.description)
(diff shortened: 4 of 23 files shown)


kyleam:
I don't spot any obvious downsides to using `font-lock-face`.  The only thing that came to mind was whether there could be a noticeable overhead due to the font-lock-mode-managed mapping from `font-lock-face` to `face`, but I doubt that's a reasonable concern.

With this change, I found it oddly satisfying to pointlessly toggle fontification in Magit buffers with `M-x font-lock-mode` :)

tarsius:
I would rather not do that.

* Magit already implements some `whitespace-mode`-like functionality and while it might not be as good, it should cover the main use-cases.
* Does "highlight wrong whitespace used for indentation" actually work after applying this? I would expect that the leading `[-+\s]` poses a problem.
* The rules about what is the right indentation may be different for different file types, but they may be displayed in the same buffer.
* One might want the "show spaces as dots" style, but only in diffs. But `whitespace-mode` is a buffer-wide mode, so the dots would also appear in commit messages and everywhere else.

And finally I still plan to support syntax highlighting in diffs (see #2942). This would copy syntax faces from file-visiting buffers and other fontification would be copied at the same time.

So not only do I have doubts about the benefit of this in the short-run, I also think that eventually we would end up reverting it.

npostavs:
> * Does "highlight wrong whitespace used for indentation" actually work after applying this? I would expect that the leading `[-+\s]` poses a problem.
> * The rules about what is the right indentation may be different for different file types, but they may be displayed in the same buffer.
> 
> * One might want the "show spaces as dots" style, but only in diffs. But `whitespace-mode` is a buffer-wide mode, so the dots would also appear in commit messages and everywhere else

You're right about all these. But it's not just `whitespace-mode` that has a problem with the current use of `face`, but any other mode which uses font lock search based highlighting will have the same problem, e.g., rainbow mode #3394. And https://github.com/jlr/rainbow-delimiters/issues/31#issuecomment-54159846

> And finally I still plan to support syntax highlighting in diffs (see #2942). This would copy syntax faces from file-visiting buffers and other fontification would be copied at the same time.
> 
> So not only do I have doubts about the benefit of this in the short-run, I also think that eventually we would end up reverting it.

Wouldn't the solution for #2942 also copy `font-lock-face`?

tarsius:
Lets leave this open for the time being. The other two options that I can choose from at this point are "nah, sorry" and "if you insist but then it's your fault". :stuck_out_tongue: Seems better to wait until I can make a more informed choice (though it will take a while for me to get there).

npostavs:
> Seems better to wait until I can make a more informed choice (though it will take a while for me to get there).

Sure, this isn't really urgent.

> > Wouldn't the solution for #2942 also copy `font-lock-face`?
> 
> Probably but wouldn't that make it _more_ problematic to merge this?

I don't see why? If everything uses `font-lock-face` then they're all compatible and everything is fine, no?

tarsius:
@npostavs I am okay with you finishing and merging this now. *Maybe* that will cause issues eventually, but more likely not and reverting wouldn't be that horrible. Meanwhile we no longer have to deal with font-lock-dependant modes that enable themselves in non-font-locked buffers, so that's a win.

kyleam:
I can't take a closer look at the moment, but running with this locally I noticed that the upstream and push branches are no longer fontified in the push transient (I'd guess there are similar issues in other transients).

npostavs:
Oh, it's because `font-lock-mode` doesn't get turned on in the buffers transient creates, e.g., the following advices "fixes" it:

```elisp
(defun turn-on-font-lock-in-transient (&rest _)
  (with-selected-window transient--window
    (font-lock-mode +1)))
(advice-add 'transient--show :after #'turn-on-font-lock-in-transient)
```

@tarsius would you rather turn on font lock for transient buffers, or revert the part of this commit that affects strings inserted into transient buffers?

tarsius:
I have *removed* the commit while it was still the last commit.

tarsius:
> would you rather turn on font lock for transient buffers,

I would rather not do that because in those buffers turning on font-lock-using modes is definitely wrong.

> or revert the part of this commit that affects strings inserted into transient buffers?

What would the consequences of that be?

tarsius:
Also this is not just about `transient`. All functions that return a propertized string (more or less the functions that have "get" or "format" in their name) should use `face` because any third-party package may use those.

npostavs:
> I have _removed_ the commit while it was still the last commit. Okay now we can talk about it in peace...

Yeah, I was a bit hasty in merging this.

> > or revert the part of this commit that affects strings inserted into transient buffers?
> 
> What would the consequences of that be?

Hopefully make it work as before, but I forgot about...

> Also this is not just about `transient`. All functions that return a propertized string (more or less the functions that have "get" or "format" in their name) should use `face` because any third-party package may use those.

...this. So I think these should use both `face` and `font-lock-face`, because a 3rd party package might insert the string into a buffer without font lock (like transient, where we want `face`), or a buffer with font lock searching enabled (where we would want `font-lock-face`).

stribb:
I notice that ``speedbar.el` also has a problem with font-lock mode, so it has the following code to tell font-lock-mode to keep its hands off:


;; Fix a font lock problem for some versions of Emacs
(and (boundp 'font-lock-global-modes)
     font-lock-global-modes
     (if (eq font-lock-global-modes t)
	 (setq font-lock-global-modes '(not speedbar-mode))
       (if (eq (car font-lock-global-modes) 'not)
	   (add-to-list 'font-lock-global-modes 'speedbar-mode t)
	 (setq font-lock-global-modes (delq 'speedbar-mode
					    font-lock-global-modes)))))


Obviously, the correct place for a general font-lock mode exclusion function is in font-lock code itself, but that's an Emacs iteration away. So maybe:

```elisp
(defun magit-disable-font-lock-mode (mode)
  "Disable font-lock mode for MODE."
  (and (boundp 'font-lock-global-modes)
     font-lock-global-modes
     (if (eq font-lock-global-modes t)
	 (setq font-lock-global-modes (list 'not mode))
       (if (eq (car font-lock-global-modes) 'not)
	   (add-to-list 'font-lock-global-modes mode t)
	 (setq font-lock-global-modes (delq mode
					    font-lock-global-modes)))))
```

then liberally scatter `magit-disable-font-lock-mode` invocations throughout the Magit codebase?

npostavs:
> hen liberally scatter `magit-disable-font-lock-mode` invocations throughout the Magit codebase?

This would stop `whitespace-mode`, `rainbow-delimiters`, `num3-mode` and others from breaking magit fontification; but would still prevent them from actually working in magit buffers.

----

> I have _removed_ the commit while it was still the last commit. Okay now we can talk about it in peace...

Since github doesn't follow these kind of tricks, I've opened a new pull request: #3886.

Issue #2959, opened 2017-01-12
Implement an Elisp binding for libgit2


tarsius:
*This description was taken from #2956. I intend to replace it with a more in-depth description at a later time.*

Magit is slow and part of fixing that involves the use of [`libgit2`](https://libgit2.github.com/), "a portable, pure C implementation of the Git core methods provided as a re-entrant linkable library with a solid API, allowing you to write native speed custom Git applications in any language which supports C bindings."  Unfortunately nobody has written that for Elisp yet and since improving performance is a top priority now, I'll to it.

This will be named just `libgit.el` (or `libgit2.el`) and be pretty basic, i.e. just expose the functions provided by `libgit2` to Elisp.

---

Older discussions: #2539, https://github.com/magit/magit/issues/2442#issuecomment-165093660, https://github.com/magit/magit/issues/1327#issuecomment-39678570. (Yes, this goes back a while, but note that doing this is only even possible since Emacs v25.1, which was released in September 2016.)

---

Some resources:

- In depth documentation https://phst.github.io/emacs-modules
- Another list of resources https://github.com/emacs-pe/emacs-modules
- A simple module, also with links http://kitchingroup.cheme.cmu.edu/blog/2017/07/08/Adding-numerical-methods-to-emacs-with-dynamic-modules
- Emacs, Dynamic Modules, and Joysticks http://nullprogram.com/blog/2016/11/05/

And of course...

- libgit2 API https://libgit2.github.com/libgit2/#HEAD

jwiegley:
@tarsius If you have any `libgit2` related questions, feel free to ask me. I've used it extensively, both for personal projects and in production, and I maintain the Haskell bindings to `libgit2`.

tarsius:
That's awesome to hear! I suspect you also have some experience with the new module support. I intend to get started with this soon, but could definitely need some help.

jwiegley:
I haven't yet looked into the module support, but this would be a great way to learn it. Count me in!

vermiculus:
@jwiegley I'm not current on emacs-devel, but have we thought at all about how modules will be distributed? Is package.el planning support for them?

mgalgs:
> part of fixing that involves the use of libgit2

I would love to see some data to back up that claim. It _sounds_ right to me, but it would be a shame for you to spend precious time on an optimization that might not bear fruit as anticipated. Perhaps we could start with just a minimal implementation of the libgit2 bindings and do some benchmarks to prove the concept.

[...]

PR #3841, opened 2019-04-19
Basic libgit support

tarsius:
Description coming soon. (Need to coordinate changes to three repositories.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ env:
   global:
     - CURL="curl -fsSkL --retry 9 --retry-delay 9"
     - GHRAW="https://raw.githubusercontent.com"
+    - BUILD_MAGIT_LIBGIT="false"
   matrix:
     - EMACS_VERSION=25.1
     # 25.2 is identical to 25.3 except for a critical security bug in
--- a/Makefile
+++ b/Makefile
@@ -197,6 +197,14 @@ define set_package_requires
 `((emacs ,emacs-version) ;`
   (dash ,dash-version)
   (with-editor ,with-editor-version)))))
+(with-temp-file "lisp/magit-libgit.el"
+  (insert-file-contents "lisp/magit-libgit.el")
+  (re-search-forward "^;; Package-Requires: ")
+  (delete-region (point) (line-end-position))
+  (insert (format "%S"
+`((emacs "$(LIBGIT_EMACS_VERSION)") ;`
+  (magit "$(LIBGIT_MAGIT_VERSION)")
+  (libgit ,libgit-version)))))
 (with-temp-file "lisp/magit-pkg.el"
   (insert (pp-to-string
 `(define-package "magit" "$(VERSION)" ;`
@@ -222,6 +230,7 @@ bump-versions-1:
         (async-version \"$(ASYNC_VERSION)\")\
         (dash-version \"$(DASH_VERSION)\")\
         (git-commit-version \"$(GIT_COMMIT_VERSION)\")\
+        (libgit-version \"$(LIBGIT_VERSION)\")\
         (transient-version \"$(TRANSIENT_VERSION)\")\
         (with-editor-version \"$(WITH_EDITOR_VERSION)\"))\
         $$set_package_requires)"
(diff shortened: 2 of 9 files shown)


tarsius:
I am deleting the earlier iterations now but just in case I am recording the old &context magic here:

``` emacs-lisp
;; We want to write (&context (gitimpl IMPL)), which is only possible
;; if a variable `gitimpl' exists.  We don't use it for anything but
;; satisfying that requirement.  `cl-generic' assumes the use of a
;; variable, but we use a function to determine the context.
(with-no-warnings (defvar gitimpl nil))
(defun cl--generic-gitimpl-tagcode (varsym &rest _)
  `(progn ,varsym ; silence byte-compiler
          (let ((val (magit--git-implementation)))
            (and (memq val magit--git-implementations) val))))

(defun cl--generic-gitimpl-generalizers (tagval &rest _)
  `((gitimpl ,tagval)))

(cl-generic-define-generalizer cl--generic-gitimpl-generalizer
  80
  #'cl--generic-gitimpl-tagcode
  #'cl--generic-gitimpl-generalizers)

(cl-defmethod cl-generic-generalizers ((_specializer (head gitimpl)))
  (list cl--generic-gitimpl-generalizer))

(cl-generic-define-context-rewriter gitimpl (impl &rest impls)
  `(gitimpl ,(if (consp impl)
                 (progn (cl-assert (null impls)) impl)
               `(gitimpl ,impl . ,impls))))
```

ethomson:
Excited to see this!  I'm sorry that I didn't see this earlier, because I would want to point out that libgit2 is actually _libgit2_, which is the proper name of the library.  (It's not "libgit", version "2"; which is a common misunderstanding for obvious reasons.)

libgit itself is the shared library that is part of git, the command-line application.  So there's some ambiguity calling it simply "libgit".

(Actually, libgit2 _was_ originally intended to replace that libgit that's in git, while providing a linkable library for everybody else, too.  But that's not really what happened, I guess, and here we are.)

Anyway.  I just wanted to mention this.  I don't expect you to necessarily change this if it's a) obviously too late, or b) you don't care.  But if this is something that you care about then I wanted to mention it before it was _way_ too late.  😉 

Excited again to see this.  Please let me know if the libgit2 developers can help with anything!

slyrus:
I second the notion that calling the libegit2 directory libgit is user-unfriendly and confusing. It would be great if this could just work with that directory being named libegit2.

tarsius:
Thanks for the feedback.

If there is a last chance to change the name, then it is now.

I don't think we talked much about the name in the past, and the extend of the "conversation" is:

1. (1) of https://github.com/magit/magit/issues/2959#issuecomment-383221426 and
2. the reply to that in https://github.com/magit/magit/issues/2959#issuecomment-383276092.

One difficulty is that Emacs Lisp doesn't do namespaces. Different packages avoid trampling on each others foot by using dedicated symbol prefixes. These prefixes should match the name of the package. The name of the package must match the name of the primary lisp library of the package.

So if we change the name of the package, then we don't only have to rename the library to `libgit2.el`, we *should* also change the symbol prefixes to `libgit2-` and rename all the symbols.

I would rather not do that. It is a unwieldy; its a bit long and makes touch typing harder.

I guess the reason we use the `libgit-` prefix instead of `libgit2-` is the same as why libgit2 itself uses `git-` instead of `libgit2-`. We just couldn't go as far because of the lack of namespacing. `libgit2` isn't "git", so the symbol prefix that it uses is "not correct". But you concluded that this doesn't matter because once you actually use these function you already know how `libgit2` relates to `git` and how they are not the same thing.

In our case the choice of symbol prefix leaks into the name of the library and the package; at least by convention. While the names of the library and the package *must* match, the prefix only *should* match.

There aren't that many C modules yet, so the conventions haven't been challenged much yet when it comes to this special-case. We do have the option to name the package and library `libgit2` (or `libegit2`, `egit`...) and keep the `libgit-` symbol prefix. Lets summon someone who does a lot of convention enforcement.

@purcell we call to you to bless us with your  guidance. :wink: 

Also @TheBB is who will ultimately decide, so :wave:, please join in too.

basil-conto:
FWIW, there is slightly related precedent with, for example, Emacs linking to `libxml2` and exposing it via `libxml-*` functions.

tarsius:
@basil-conto Thanks for the information. I know favor doing the same here.

PR #3842, opened 2019-04-21
Fix pull/push push remote strings

danielmartin:
The description strings of `magit-pull-from-pushremote` and `magit-push-current-to-pushremote` were mixed: `magit-pull-from-pushremote` ended the sentence with "push there", and `magit-push-current-to-pushremote` ended the sentence with "pull from there".


kyleam:
Thanks!

PR #3849, opened 2019-04-27
magit-process-filter: Strip \r *before* checking prompts

### LINKED ISSUE #3843: Cannot enter SSH passphrase during fetch or push with openssh-8.0
## Summary
After upgrading OpenSSH from 7.9 to 8.0 the SSH passphrase during git fetch or push can no longer be entered in minibuffer.

## Software versions
- Distro: Arch Linux, x86_64
- OpenSSH: OpenSSH_8.0p1, OpenSSL 1.1.1b  26 Feb 2019
- Emacs: GNU Emacs 26.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.8) of 2019-04-12
- Magit (from MELPA): Magit 20190422.1347, Git 2.21.0, Emacs 26.2, gnu/linux

## Steps to reproduce
1. Have a local clone of a remote git repository (e.g. from Github) over SSH with keys set up
1. Open a version controlled file in Emacs
1. `M-x magit-pull`
1. `u` (default remote branch)

### Expected result
- The minibuffer shows "`Enter passphrase for key '/home/user/.ssh/id_rsa':`", the passphrase can be entered and the action completes successfully

### Actual result
- The minibuffer shows "running git pull origin refs/heads/master"
- The passphrase query can be seen in magit-process buffer, but cannot enter any input there
- The process hangs in waiting for the key input

## Notes
- Downgrading the OpenSSH to version 7.9 fixed the problem, so the change that has broken the behaviour seems to come from there. Decided to report here, because Magit will probably need to react and adapt anyway.
- Fetching and pushing over HTTPS still works, the problem is only with interacting with the remote repository over SSH.
- Git fetch and push works fine from command line (I use bash 5.0.3)
- Link to the OpenSSH release notes for quick reference: https://www.openssh.com/txt/release-8.0
- Upgraded Magit and dependencies, restarted Emacs, used `M-x magit-emacs-Q-command` and started Emacs with the provided command, but problem persisted in all the cases.
- `M-x toggle-debug-on-error` produced no stack trace, since nothing actually crashes

npostavs:
As I said in https://github.com/magit/magit/issues/3843#issuecomment-485944696

> We already strip `\r` in `magit-process-filter`; the correct fix would
> be to shuffle the code around so that it happens before password
> prompt checking instead of after.

Fixes #3843.

### DIFF
--- a/Documentation/RelNotes/2.91.0.org
+++ b/Documentation/RelNotes/2.91.0.org
@@ -220,3 +220,6 @@ c47913461 * Rename magit-{push-current-set-remote => remote-set}-if-missing
 - When deleting a remote branch failed, the logic for deciding whether
   to prune the local remote-tracking ref was too loose, leading to
   false positives.  #3650
+
+- Fix handling of passphrase prompts which are output with leading
+  carriage return, such as those produced by Openssh 8.0.  #3843
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -684,22 +684,19 @@ Magit status buffer."
   "Default filter used by `magit-start-process'."
   (with-current-buffer (process-buffer proc)
     (let ((inhibit-read-only t))
-      (magit-process-yes-or-no-prompt proc string)
-      (magit-process-username-prompt  proc string)
-      (magit-process-password-prompt  proc string)
       (goto-char (process-mark proc))
-      (setq string (propertize string 'magit-section
-                               (process-get proc 'section)))
       ;; Find last ^M in string.  If one was found, ignore
       ;; everything before it and delete the current line.
-      (let ((ret-pos (length string)))
-        (while (and (>= (cl-decf ret-pos) 0)
-                    (/= ?\r (aref string ret-pos))))
-        (if (< ret-pos 0)
-            (insert string)
-          (delete-region (line-beginning-position) (point))
-          (insert (substring string (1+ ret-pos)))))
-      (set-marker (process-mark proc) (point)))))
+      (when-let ((ret-pos (cl-position ?\r string :from-end t)))
+        (cl-callf substring string (1+ ret-pos))
+        (delete-region (line-beginning-position) (point)))
+      (insert (propertize string 'magit-section
+                          (process-get proc 'section)))
+      (set-marker (process-mark proc) (point))
+      ;; Make sure prompts are matched after removing ^M.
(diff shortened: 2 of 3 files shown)


edwargix:
Getting this error when trying to fetch:
```
error in process filter: substring: Symbol’s value as variable is void: ret-post
error in process filter: Symbol’s value as variable is void: ret-post
```

npostavs:
@edwargix: oops, that was a silly typo.  I force-pushed a fix, please try again.

edwargix:
No worries. Just tried the new fix and I still can't enter my password when I try to fetch.

npostavs:
Hmm, can you try `(trace-function 'magit-process-filter)` and show the
`*trace-output*` buffer contents?

edwargix:
```
======================================================================
1 -> (magit-process-filter #<process git<1>> "
Enter passphrase for key '/home/edwargix/.ssh/id_rsa': ")
1 <- magit-process-filter: #<marker at 158 in magit-process: mozzarella>
```

edwargix:
The magit process buffer stalls at
```
run git … fetch origin
Enter passphrase for key '/home/edwargix/.ssh/id_rsa': 
```

npostavs:
Hmm, so you have a newline at the front, not a carriage return?  Seems
like that should work fine (even without the patch).  I tested with
the following in `*scratch*` and I do get a prompt:

``` emacs-lisp
(setq test-proc (start-process "test-cat" (current-buffer) "cat"))

(magit-process-filter test-proc "
Enter passphrase for key '/home/edwargix/.ssh/id_rsa': ")
```

----

> After `f p` from magit status buffer
> ```
> ======================================================================
> 1 -> (magit-process-filter #<process git<1>> "
> Enter passphrase for key '/home/edwargix/.ssh/id_rsa': ")
> 1 <- magit-process-filter: #<marker at 158 in magit-process: mozzarella>
> ```

And I'm a bit confused because in your `*trace-output*` it looks like
the function returned, so it shouldn't have stalled.  Can you step
through in the debugger to see what's happening?

edwargix:
It seems to be working now

juristi:
Worked for me as well. The prompt appeared in the minibuffer and the passphrase could entered successfully.

Cloned the magit repo earlier today and tried to play around locally, but ended up only with much clumsier solutions :)

npostavs:
Okay thanks, I think we're all good then.  Merged.

Issue #3855, opened 2019-05-01
magit-tag-release hangs when using loopback pinentry


darkfeline:
Trying to use the new `magit-tag-release` with loopback pinentry hangs.  Changing the `magit-run-git` call to `magit-run-git-async` allows the command to proceed smoothly.

It's possible to configure Emacs to serve GnuPG pinentry requests (eliding the details, see `epa-pinentry-mode`).  However, this doesn't work with `magit-tag-release` because it uses the synchronous `magit-run-git`.  Since this is synchronous, Emacs is unable to handle the pinentry request that comes back, causing deadlock.

    M-x magit-version RET

Magit 20190426.1457, Git 2.21.0, Emacs 26.2, gnu/linux

PR #3856, opened 2019-05-01
magit-tag-release: Unblock Emacs to handle pinentry

darkfeline:
Fixes https://github.com/magit/magit/issues/3855

### DIFF
--- a/lisp/magit-tag.el
+++ b/lisp/magit-tag.el
@@ -155,7 +155,10 @@ review the result."
                                   (directory-file-name (magit-toplevel))))
                                 ver)))))
        (list tag (read-string (format "Message for %S: " tag) msg)))))
-  (magit-run-git "tag" "--annotate" "--sign" "-m" msg tag)
+  (let ((proc (magit-run-git-async "tag" "--annotate" "--sign" "-m" msg tag)))
+    ;; Allow Emacs to handle pinentry for signing.
+    (while (eq (process-status proc) 'run)
+      (accept-process-output proc)))
   (magit-refs-setup-buffer nil (magit-show-refs-arguments)))
 
 (defun magit--list-releases ()


tarsius:
This should use a sentinel instead. `magit-merge-absorb` looks like a decent example for how to do that.

lisp/magit-tag.el
```diff
@@ -155,7 +155,10 @@ review the result."
                                   (directory-file-name (magit-toplevel))))
                                 ver)))))
        (list tag (read-string (format "Message for %S: " tag) msg)))))
-  (magit-run-git "tag" "--annotate" "--sign" "-m" msg tag)
+  (let ((proc (magit-run-git-async "tag" "--annotate" "--sign" "-m" msg tag)))
+    ;; Allow Emacs to handle pinentry for signing.
+    (while (eq (process-status proc) 'run)
+      (accept-process-output proc)))
```
basil-conto: There is a race condition here. Quoting `(elisp) Accepting Output` from Emacs 27:
```
   If a connection from a process contains buffered data,
‘accept-process-output’ can return non-‘nil’ even after the process has
exited.  Therefore, although the following loop:

     ;; This loop contains a bug.
     (while (process-live-p process)
       (accept-process-output process))

will often read all output from PROCESS, it has a race condition and can
miss some output if ‘process-live-p’ returns ‘nil’ while the connection
still contains data.  Better is to write the loop like this:

     (while (accept-process-output process))
```

darkfeline:
@tarsius I don't think using a sentinel here is right (or I misunderstand). `magit-tag-release` still has to be "blocking" before it calls `magit-refs-setup-buffer`, and as far as I can tell, the only way to wait for a process is `(while (accept-process-output process))` as @basil-conto suggested.

Also, `magit--merge-absort` appears to be typoed.

lisp/magit-tag.el
darkfeline: Makes sense, will fix it after confirming @tarsius 's suggestion.

Although technically I don't think it matters here since we mostly care about waiting for the process to exit and not really about processing all of the output before popping up the refs buffer.  It doesn't matter of the final bit of the output from `git tag` doesn't get processed until after this command returns.

basil-conto: Still, it's more correct and sets a better precedent.

basil-conto:
> magit-tag-release still has to be "blocking" before it calls magit-refs-setup-buffer, and as far as I can tell, the only way to wait for a process is (while (accept-process-output process))

I'm not familiar with the code in question, but in general, reading the entire output of a process is not the only way to ensure the process has completed before calling the next function. The alternative is to conditionally call the next function from the process sentinel. I assume this is what @tarsius is suggesting here, but again, I haven't been following the code or discussion closely.

darkfeline:
> The alternative is to conditionally call the next function from the process sentinel.

That was my interpretation too, but I think that is wrong, because that would make opening the refs buffer asynchronous.  For example, if `git tag` takes 10 seconds for the sake of example, then after the user runs `magit-tag-release` which runs asynchronously and goes to edit some other file, the refs buffer will pop up after 10 seconds while the user is typing away and this is a Bad Thing.

tarsius:
> if `git tag` takes 10 seconds for the sake of example

In practice it takes a few milliseconds though and if it takes longer then you have bigger problems than a buffer randomly popping up.

More importantly though not waiting until the tag is ready would completely defeat the purpose of why why this buffer is shown in the first place. It lets you quickly verify that the new tag indeed confirms to the tagging conventions of the current repository.

So I went with the sentinel.

darkfeline:
> More importantly though not waiting until the tag is ready would completely defeat the purpose of why why this buffer is shown in the first place. It lets you quickly verify that the new tag indeed confirms to the tagging conventions of the current repository.

Er, that's what my commit (improved by @basil-conto) does?  This is just a question of whether the command exits and returns to the Emacs command loop before or after waiting for git tag to finish.

This PR:

1. User runs magit-tag-release
2. git tag runs
3. wait for git tag to finish
4. Show refs buffer
5. Return to command loop

Landed commit:

1. User runs magit-tag-release
2. git tag runs
3. Return to command loop
4. wait for git tag to finish
5. Show refs buffer

I mean, I guess it's fine, but it seems like a weird design choice to me.

PR #3858, opened 2019-05-01
Print current hash along with the tags in the status buffer

georgek:
I use `git describe` quite a lot and it's nice to be able to see the same information in the magit status buffer. The only thing the tags section doesn't currently contain is the current hash. This is only displayed if we're not currently on the tag.

### DIFF
--- a/lisp/magit-status.el
+++ b/lisp/magit-status.el
@@ -561,7 +561,8 @@ arguments are for internal use only."
          (next-cnt (cadr next-tag))
          (this-tag (car this-tag))
          (next-tag (car next-tag))
-         (both-tags (and this-tag next-tag t)))
+         (both-tags (and this-tag next-tag t))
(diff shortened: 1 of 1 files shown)


kyleam:
Are you aware of the `magit-status-show-hashes-in-headers` option?

tarsius:
That line is about tags, not about `HEAD`. A tag may point at the same commit as `HEAD` but more often than not that is not the case. It would therefore be better to show the hash on the `Head: ...` line. If `magit-status-show-hashes-in-headers` is non-nil, then that is what happens.

georgek:
It is about HEAD because you show the distance from the tags. It shows similar information to `git describe` except it doesn't show the hash.

Maybe I should just add a section that has the output of `git describe` instead.

I don't find `magit-status-show-hashes-in-headers` to be very nice. It detracts form the elegance of the status buffer.

Issue #3862, opened 2019-05-03
Kill all associated buffers when killing a project's status buffer


tsdh:
I work with many repositories throughout a day.  Most of the I only need temporarily, i.e., I open a Magit status buffer for one of them, update the project, maybe view the log, add a commit, and then kill the status buffer when I'm done and move on.

It would be nice if killing a project's status buffer would (maybe optionally) also kill all associated Magit buffers like the `magit-process` buffer, all `magit-log`, `magit-revision` and `magit-diff` buffers for that project, etc.

Right now, those buffers just stick around until you kill them manually.  Because of their uniform naming scheme (common prefix), they can easily make buffer switching more cumbersome.

If you think that's a good idea in general, feel free to comment how you'd like to have it implemented and I'll do it and add a PR.

tarsius:
I would merge that.

The interface seems simple enough:

```emacs-lisp
(defun magit-mode-bury-buffer (&optional kill-buffer)
  "Bury the current buffer.
With a prefix argument, kill the buffer instead.
This is done using `magit-bury-buffer-function'."
  (interactive "P")
  (funcall magit-bury-buffer-function kill-buffer))
```

I would suggest adding a `magit-bury-buffer-function`-implementation, which takes two prefix arguments to mean "... and also kill all the other Magit buffers for this repository".

tsdh:
Will do!

tsdh:
@tarsius That was easier than I've thought.  And my first encounter with Forge which is really awesome!

PR #3863, opened 2019-05-03
Kill associated Magit buffers with C-u C-u q

tsdh:
As discussed in #3862, with this pull request, a double prefix argument given
to `magit-mode-bury-buffer` will also kill all other Magit buffers associated
with the current project.

I've chosen not to add a new `magit-bury-buffer-function` implementation which
would have that feature (as you've suggested) but just do it directly in
`magit-mode-bury-buffer` so that it will work no matter which
`magit-bury-buffer-function` people are using.

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -1071,8 +1071,16 @@ latter is displayed in its place."
 (defun magit-mode-bury-buffer (&optional kill-buffer)
   "Bury the current buffer.
 With a prefix argument, kill the buffer instead.
+With two prefix arguments, also kill all Magit buffers associated
+with this repository.
 This is done using `magit-bury-buffer-function'."
   (interactive "P")
+  ;; Kill all associated Magit buffers when a double prefix arg is given.
+  (when (>= (prefix-numeric-value kill-buffer) 16)
+    (let ((current (current-buffer)))
+      (dolist (buf (magit-mode-get-buffers))
+        (unless (eq buf current)
+          (kill-buffer buf)))))
   (funcall magit-bury-buffer-function kill-buffer))
 
 (defun magit-mode-quit-window (kill-buffer)
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -8,7 +8,7 @@
 #+TEXINFO_DIR_CATEGORY: Emacs
 #+TEXINFO_DIR_TITLE: Magit: (magit).
 #+TEXINFO_DIR_DESC: Using Git from Emacs with Magit.
-#+SUBTITLE: for version 2.90.1 (v2.90.1-628-gabc75923d+1)
+#+SUBTITLE: for version 2.90.1 (v2.90.1-652-gab4811f54+1)
 
 #+TEXINFO_DEFFN: t
 #+OPTIONS: H:4 num:3 toc:2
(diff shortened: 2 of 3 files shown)


lisp/magit-mode.el
```diff
@@ -1071,8 +1071,13 @@ latter is displayed in its place."
 (defun magit-mode-bury-buffer (&optional kill-buffer)
   "Bury the current buffer.
 With a prefix argument, kill the buffer instead.
+With two prefix arguments, also kill all Magit buffers associated
+with this repository.
 This is done using `magit-bury-buffer-function'."
   (interactive "P")
+  (when (>= (prefix-numeric-value kill-buffer) 16)
```
basil-conto: The `>=` comparison precludes a future triple prefix argument. ;) (I'm not being serious.)

tsdh: @basil-conto That's intended so that an accidental triple or quadruple prefix works like a double one as long as we're not having a special treatment for more prefixes.  You know, more is more.

tsdh:
A side question: the PR in Forge just lists the first 2 commits in the PR but not the last fix 8d439d8.  Is that a bug or is the API's point of view just a bit behind the Git/website view?  (I'll check again later to see if it appears in my Forge PR...)

lisp/magit-mode.el
```diff
@@ -1071,8 +1071,16 @@ latter is displayed in its place."
...
+With two prefix arguments, also kill all Magit buffers associated
+with this repository.
 This is done using `magit-bury-buffer-function'."
   (interactive "P")
+  ;; Kill all associated Magit buffers when a double prefix arg is given.
+  (when (>= (prefix-numeric-value kill-buffer) 16)
+    (let ((buf (current-buffer)))
+      (mapc #'kill-buffer
+            (--filter (not (eq it buf))
+                      (magit-mode-get-buffers)))))
```
tarsius: I would use

```emacs-lisp
(dolist (buf (magit-mode-get-buffers))
  (unless (eq buf current)
    (kill-buffer buf)))
```

tsdh: Ok, whatever you prefer. :-)

lisp/magit-mode.el
basil-conto: @tsdh I know, I was joking about the fact that, in the unlikely event that a third prefix is added in the future, it would constitute a breaking change. ;)

tsdh:
@tarsius Ok, I've changed it according to your suggestion and rebased the PR into two commits, one with the code, and one for the docs.

tarsius:
Thanks!

manuel-uberti:
@tsdh @tarsius will this restore the window configuration as well? I have this in my configuration:
```
(defun mu-magit-kill-buffers ()
              "Restore window configuration and kill all Magit buffers."
              (interactive)
              (let ((buffers (magit-mode-get-buffers)))
                (magit-restore-window-configuration)
                (mapc #'kill-buffer buffers)))
```
And I was wondering if I can ditch in favour of this new binding.

tsdh:
@manuel-uberti It depends on `magit-bury-buffer-function` where the current default value is `magit-restore-window-configuration`, so I had assumed yes. :-)

PR #3876, opened 2019-05-14
magit-branch: Allow it to specify a branch when deleting the current

yashi:
While working on a topic branch, it's not the master branch I want to
go back to when I delete the current branch, but the topic branch I'm
working on.  This often happens when the current branch is just a
throw away branch for a build test at the middle of the topic branch.

This commit add another option to magit-branch-delete so that we can
specify a branch to checkout when deleting the current branch.

### DIFF
--- a/lisp/magit-branch.el
+++ b/lisp/magit-branch.el
@@ -580,6 +580,7 @@ defaulting to the branch at point."
                    (magit-read-char-case prompt nil
                      (?d "[d]etach HEAD & delete"     'detach)
                      (?c "[c]heckout master & delete" 'master)
+                     (?o "checkout [o]ther & delete"  'other)
                      (?a "[a]bort"                    'abort)))
             (`detach (unless (or (equal force '(4))
                                  (member branch force)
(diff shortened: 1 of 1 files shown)


tarsius:
You have probably noticed the comment at the beginning of this function. I would like to avoid stuffing even more functionality into this command.

I don't think this command should be used to perform the actions "delete current branch" and "make another branch the current branch". Rarely I use this command like this when I *accidentally* try to delete the current branch thinking that I have already switched to another branch. I use that feature because it already exists; I strongly doubt I would implement it myself if that weren't the case.

Why do you use this command like this? Or more precisely in what situations to you use it like this?

The only situation that I can think of where I feel doing so makes sense is when you do some experiment, it turns out what you tried doesn't work, and now you want to throw it away. In that case your extension might be useful.

yashi:
Let's say I created a PR with commits.  A maintainer noticed something and wanted me to fix here and there.  To do that I'll go back each commit by creating and checking out a throwaway branch and build and test.  Once you are done fixing the commit, I rebase the topic branch on top of the throwaway branch.

If you noticed that you don't need to do fix anything after testing (with code modification) at the throwaway branch, you want to go back to the topic branch and kill the current throwaway branch.

As you see while working on a topic branch, I rarely go back to the master because the local master branch usually tracks the upstream master branch.  But the current commands only has options I don't use.

tarsius:
How about just performing these two tasks in two separate steps? First check out the feature branch and only then delete the experimental branch.

The current variants "checkout out something else (`HEAD` or `master`) to make it possible to carry out the requested action (deleting the current branch)".

Your proposed variant goes a step further. It optimizes one particular workflow, making it possible to carry out two steps in one go. But `magit-branch-delete` is Magit's only command for deleting branches, and it should not be overloaded with too much only marginally related dwim behavior. It is *already* terribly overloaded. I also think it is weird if the command that deletes a branch prompts for a branch to check out.

tarsius:
I recommend you create a *separate* command for your use-case.

PR #3881, opened 2019-05-19
Give require-match the correct meaning in magit-ido-completing-read

DarwinAwardWinner:
Since the REQUIRE-MATCH argument to `magit-completing-read` means that an empty string is not a valid completion (unlike in `completing-read`), I've modified the call to `ido-completing-read+` to prevent it from accepting an empty string from the user when REQUIRE-MATCH is non-nil. This means that magit's ido completion will no longer offer the empty string as the first completion option, and pressing RET without typing anything will accept the first choice.

<strike>If there actually *is* a case where the empty string is desired as the default choice, then the only way to specify that now is to explicitly pass the empty string (not nil) as the default. This might be desired if the calling function *really* wants to make sure the user is intentionally selecting an option instead of just hitting RET to pick the first thing on the list.</strike> Edit: Actually this won't really work because it will kind of mess up the prompt. I think there might not be any good way to specify the empty string as the default when a match is required. This is probably fine.

This pull request also includes a 2nd commit that fixes a typo in the docstring relating to `magit-ido-completing-read`.

### DIFF
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -69,7 +69,7 @@
 If you have enabled `ivy-mode' or `helm-mode', then you don't
 have to customize this option; `magit-builtin-completing-read'
 will work just fine.  However, if you use Ido completion, then
-you do have to use `magit-ido-completion-read', because Ido is
+you do have to use `magit-ido-completing-read', because Ido is
 less well behaved than the former, more modern alternatives.
 
 If you would like to use Ivy or Helm completion with Magit but
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -570,7 +570,8 @@ drop-in replacement for `completing-read', instead we use
 same name."
   (if (require 'ido-completing-read+ nil t)
       (ido-completing-read+ prompt choices predicate require-match
-                            initial-input hist def)
+                            initial-input hist
+                            (or def (and require-match (car choices))))
     (display-warning 'magit "ido-completing-read+ is not installed
 
 To use Ido completion with Magit you need to install the


DarwinAwardWinner:
Note that you could probably do more or less the same trick in `magit-builtin-completing-read`, which would prevent the user from even selecting the empty string.

tarsius:
If we make this change, then we should do it in `magit-completing-read` and update this part of the doc-string "If REQUIRE-MATCH is non-nil and the users exits without a choice, an user-error is raised."

The purpose of this change is to prevent this from happening, but I am unsure this is really necessary. Do users often exit without making a choice? I don't know whether I would do that because I use `ivy` and that *always* makes the first choice the default if there is no explicit default.

tarsius:
> Fix a typo in the docstring for magit-completing-read-function

Does the manual have the same typo?

DarwinAwardWinner:
The reason for the change is that when there is no default, `ido-completing-read+` adds the empty string to the front of the list of completions in order to maintain compatibility with `completing-read-default`. However, this compatibility is not needed for `magit-completing-read`, since the empty string is defined in the documentation to be an invalid completion, so if anything, adding the empty string to the collection is counterproductive. This patch ensures that won't happen, which brings ido completion more or less in line with ivy: the first choice becomes the default.

The possible argument that I see against this change is that maybe when there's no default, you *want* to require the user to explicitly make a choice by typing at least one letter or using the arrow keys to select an item instead of just hitting RET without thinking. But given that ivy doesn't work that way, I don't get the feeling that's the intent.

> Does the manual have the same typo?

Not the same typo, but there are some occurrences of "completion-read" in the manual and elsewhere:

```
$ grep -r completion-read .
./Documentation/magit.texi:This function performs completion using the built-in @code{completion-read}
./Documentation/RelNotes/2.90.0.org:- ~magit-completion-read~ did not protect the value of ~this-command~ even
./Documentation/magit.org:  This function performs completion using the built-in ~completion-read~
```

tarsius:
Makes a  lot more sense now. This originally threw me of quite a bit:

> Note that you could probably do more or less the same trick in `magit-builtin-completing-read`, which would prevent the user from even selecting the empty string.

Only later you explained why we wouldn't want that:

> However, this compatibility is not needed for `magit-completing-read`, since the empty string is defined in the documentation to be an invalid completion, so if anything, adding the empty string to the collection is counterproductive.

tarsius:
Thanks!

DarwinAwardWinner:
Well, it's true, you *could* do the same thing in `magit-builtin-completing-read`, for the same reason: if the empty string is always an invalid completion for `magit-completing-read`, it's reasonable for magit completion functions to prevent the user from selecting the empty string entirely. If you call `completing-read-default` with `REQUIRE-MATCH` non-nil and `DEF` a non-empty string, then it won't let the user exit completion with an empty string -- it will *require* a valid completion to be entered.

On a related note, this pull request assumes that COLLECTION is *always* a list and never a function, and looking at the code for `magit-completing-read`, it also makes the same assumption. If nothing else, you should probably document the difference, e.g. saying only list collections are allowed.

PR #3884, opened 2019-05-22
Hooks after workspace create and delete

happy-barney:
Allow user to react on workspace operations, for example
alter projectile's list of known projects, using common pattern - hooks.

Both `magit-worktree-create-hook` and `magit-worktree-delete-hook`
expect function(s) with one argument, workspace directory.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -6687,6 +6687,16 @@ Also see [[man:git-worktree]]
   If the worktree at point is the one whose status is already being
   displayed in the current buffer, then show it in Dired instead.
 
+- User Option: magit-workspace-create-hook
+
+  Hook called right after new workspace has been created with workspace
+  directory as an argument.
+
+- User Option: magit-workspace-delete-hook
+
+  Hook called right after new workspace has been delete with workspace
+  directory as an argument.
+
 ** Common Commands
 
 These are some of the commands that can be used in all buffers whose
--- a/lisp/magit-worktree.el
+++ b/lisp/magit-worktree.el
@@ -40,6 +40,22 @@ Used by `magit-worktree-checkout' and `magit-worktree-branch'."
   :group 'magit-commands
   :type 'function)
 
+(defcustom magit-worktree-create-hook nil
+  "Hook executed after new worktree is created.
+It is called with one argument, worktree directory."
+  :package-version '(magit . "2.91.0")
+  :group 'magit-commands
+  :options '(projectile-add-known-project)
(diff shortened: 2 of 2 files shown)


tarsius:
I think this goes to far. We cannot add a hook to every command that runs git and these commands don't sound like they need a hook more than the others. You can achieve the same using advices.

happy-barney:
These hooks should be called at every place where new checkout is created or where existing checkout is discarded - ie every time where entity accessible by external module, treated differently than in `magit`, is created (thus `projectile` examples)

`magit-worktree-` is not best best naming, I just don't have any better yet :-(

tarsius:
> These hooks should be called at every place where new checkout is created or where existing checkout is discarded

The same would be the case if you advised these functions.

tarsius:
https://www.gnu.org/software/emacs/manual/html_node/elisp/Advising-Functions.html

happy-barney:
yes, difference is, that in case of hooks you don't need to know `magit` internals (and you need to advise lot of functions instead of adding one hook) and that argument is usually not known - so imho only think you can reliably advise is `magit-run-git`

tarsius:
> (and you need to advise lot of functions instead of adding one hook)

How so? If only function `foo` calls `foo-hook`, then advising `foo` to call some code has the same effect as adding a function, which calls the same code, to `foo-hook`.

> and that argument is usually not known

The directory? I didn't mention this before because I don't intend to merge this, but iff these hooks were added they would have to be changed. If a hook takes arguments, then it should be named `*-functions` not `*-hook`. However the argument is unnecessary because the directory is available as the value of `default-directory`.

happy-barney:
ad one function-one hook - yes

for example, `magit-workspace-create-hook` is called on two places and properly named hook should be called from `magit-clone-internal` as well (though I don't speek elisp fluently enough to do it right know - apparently if at all :-) )

in case of `magit-workspace-delete`, `default-directory` contains dirname of buffer which executed command, for example let's have checkout `magit@master` and `magit@hpb-dev-workspace-hooks`

`%` `k` in 'magit@master`, providing parameter `magit@hpb-dev-workspace-hooks`  => `default-directory` is `magit@master` (though hook should be called with `magit@hpb-dev-workspace-hooks`

happy-barney:
yes, difference is, that in case of hooks you don't need to know `magit` internals (and you need to advise lot of functions instead of adding one hook) and that argument is usually not known - so imho only think you can reliably advise is `magit-run-git`

PR #3886, opened 2019-05-24
make magit-mode compatible with whitespace-mode and other font lock using modes (v2)

npostavs:
This is a continuation of #3840 which was prematurely merged (and then unmerged, but github doesn't realize that part).

> So I think [all functions that return a propertized string] should use both `face` and `font-lock-face`, because a 3rd party package might insert the string into a buffer without font lock (like transient, where we want `face`), or a buffer with font lock searching enabled (where we would want `font-lock-face`).

I've added a `squash!` commit which does that. Though it occurs that it might be simpler for future maintenance to just always use both `face` and `font-lock-face` regardless of circumstances.

### DIFF
--- a/Documentation/RelNotes/2.91.0.org
+++ b/Documentation/RelNotes/2.91.0.org
@@ -267,6 +267,9 @@
   ~magit-jump-to-skip-worktree~ and ~magit-jump-to-assume-unchanged~, and
   section inserter ~magit-insert-assume-unchanged-files~.  2d0de8614 ff
 
+- Magit buffers are now compatible with ~whitespace-mode~ (and other
+  modes which use font-lock).  #3840
+
 ** Fixes since v2.90.0
 
 - Bumped the minimal required version of ~git-commit~ to the correct
--- a/lisp/magit-bisect.el
+++ b/lisp/magit-bisect.el
@@ -187,7 +187,7 @@ bisect run'."
                              (and bad-line (match-string 1 bad-line)))
         (magit-insert-heading
           (propertize (or bad-line (pop lines))
-                      'face 'magit-section-heading))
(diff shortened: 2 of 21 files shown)


npostavs:
Hmm, not sure why the the CI is failing, I get no errors locally.

tarsius:
> Hmm, not sure why the the CI is failing, I get no errors locally.

Because it didn't include faa0e52a6b67280e5114bd87b8958071dbac1af6 and travis uses the latest transient.

tarsius:
This broke `bug-reference-mode` in Magit buffers. I addressed it using:

```emacs-lisp
(put 'bug-reference 'font-lock-face 'link)
```

@npostavs Do you understand why this is necessary here but not in file-visiting buffers whose major-mode used font-lock?

npostavs:
Um, no, and it doesn't seem to be broken here?

tarsius:
Hm. I though I discovered an issue while working on #3933 but it appears I *introduced* it while doing so. False-alarm.

tarsius:
This broke all our uses of `add-face-text-property`, which continues to operate on `face` and cannot be told to to use `font-lock-face` instead. @npostavs could you please take care of this? I fear this might involve having to reimplement `add_text_properties_1` in lisp because that cannot be called from lisp and `add-text-properties` lacks the `set_type` argument.

npostavs:
Oh darn, I searched only for `'face` so I completely missed these calls.

> could you please take care of this?

I'm on it.

> I fear this might involve having to reimplement `add_text_properties_1` in lisp because that cannot be called from lisp and `add-text-properties` lacks the `set_type` argument.

Yeah, I suspect you're correct.

PR #3892, opened 2019-06-04
manual: Add example of how to set magit-status-margin

sankalp-khare:
noted the instructions. I think I've covered my bases. 🤞 

feel free to edit as you like, or refuse if it seems unwarranted.


sankalp-khare:
I pushed a fix for a typo that I let through. Please squash. Thanks!

tarsius:
Thanks for the effort but this does indeed seem unnecessary to me.

sankalp-khare:
np. thanks for taking a look.

PR #3902, opened 2019-06-18
Add autoload cookies for `magit--handle-bookmark`

cireu:
Make it possible to trigger loading of magit when visiting a magit-status bookmark


tarsius:
Thanks!

PR #3911, opened 2019-07-05
git-commit-setup-changelog-support: Fix filling

basil-conto:
Change log entries are customarily unindented[1].  In `git-commit-mode`, however, `adaptive-fill-regexp` picks up leading asterisks `"* "` in change log entries, which are subsequently interpreted as a fill prefix of 2 spaces.  Thus, entries like the following:

```
* file: very long description...
```

are "indented" to column 2 when filled:

```
* file: very long
  description...
```

Instead of messing with `adaptive-fill-mode`, the simplest way to avoid this `fill-prefix` by default (but still allow users to enforce it by manually indenting the paragraph's 2nd line) is to enable `fill-indent-according-to-mode`, as was recently done in Emacs' `python-mode` for bug#36056[2].

[1]: `(info "(standards) Style of Change Logs")`
https://www.gnu.org/prep/standards/html_node/Style-of-Change-Logs.html
[2]: https://debbugs.gnu.org/36056

### DIFF
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -574,7 +574,8 @@ Don't use it directly, instead enable `global-git-commit-mode'."
 (put 'git-commit-mode 'permanent-local t)
 
 (defun git-commit-setup-changelog-support ()
-  "Treat ChangeLog entries as paragraphs."
+  "Treat ChangeLog entries as unindented paragraphs."
+  (setq-local fill-indent-according-to-mode t)
   (setq-local paragraph-start (concat paragraph-start "\\|\\*\\|(")))
 
 (defun git-commit-turn-on-auto-fill ()


basil-conto:
The reason I am marking this PR as a draft is because of the recent work of @npostavs on Emacs' built-in log entry filling as part of [bug#16301#11](https://debbugs.gnu.org/16301#11).

@npostavs Will this PR still be relevant if Magit can reuse the functionality you are working on?

Thanks.

npostavs:
> @npostavs Will this PR still be relevant if Magit can reuse the functionality you are working on?

That functionality will only be in Emacs 27, so as long as Magit supports earlier versions it won't help.

If we assume that functionality is available, then the filling problem you describe might be fixable with `(setq fill-paragraph-function #'log-edit-fill-entry)`, but as I've mentioned the code there is "not entirely baked" so I'm pretty sure it won't work properly yet. So far I've only tested filling of the function name list part. It's possible that we will find adding `(setq-local fill-indent-according-to-mode t)` to `log-edit-mode` is the right thing (to be honest, the docstring of this variable is so vague that I find it hard to say what setting it really means).

basil-conto:
> That functionality will only be in Emacs 27, so as long as Magit supports earlier versions it won't help.

Of course. I just wanted to check whether this PR could be doing something differently in the meantime, or in order to cover both bases.

> It's possible that we will find adding `(setq-local fill-indent-according-to-mode t)` to `log-edit-mode` is the right thing (to be honest, the docstring of this variable is so vague that I find it hard to say what setting it really means).

Indeed, my only understanding of the variable's effect/purpose comes from stepping through the relevant part of `fill-region-as-paragraph`:

```el
    ;; Don't let Adaptive Fill mode alter the fill prefix permanently.
    (let ((fill-prefix fill-prefix))
      ;; Figure out how this paragraph is indented, if desired.
      (when (and adaptive-fill-mode
                 (or (null fill-prefix) (string= fill-prefix "")))
        (setq fill-prefix (fill-context-prefix from to))
        ;; Ignore a white-space only fill-prefix
        ;; if we indent-according-to-mode.
        (when (and fill-prefix fill-indent-according-to-mode
                   (string-match "\\`[ \t]*\\'" fill-prefix))
          (setq fill-prefix nil)))
```

Here, `fill-context-prefix` returns a string of 2 spaces, and the simplest way to disregard it as a fill prefix is by enabling `fill-indent-according-to-mode`, so that `fill-prefix` is then set to `nil` again. The more complicated way would be to mess with the adaptive fill settings so that `fill-context-prefix` does not interpret leading asterisks `"* "` as a fill prefix, but this doesn't seem kosher to me, given that all adaptive fill variables are user options.

So I'll mark this PR as ready for review.

tarsius:
> Change log entries are customarily unindented

Well... I wish they were more readable instead. But my preference doesn't really matter here.

> So I'll mark this PR as ready for review.

I'll leave it to @npostavs to review this.

npostavs:
> the filling problem you describe might be fixable with `(setq fill-paragraph-function #'log-edit-fill-entry)`, but as I've mentioned the code there is "not entirely baked" so I'm pretty sure it won't work properly yet. So far I've only tested filling of the function name list part. It's possible that we will find adding `(setq-local fill-indent-according-to-mode t)` to `log-edit-mode` is the right thing

I've made some progress on this, and I've ended up let-binding `fill-indent-according-to-mode` inside `log-edit-fill-entry`: https://debbugs.gnu.org/16301#14. I think setting `fill-indent-according-to-mode` in `git-commit-setup-changelog-support` makes sense if `log-edit-fill-entry` is unavailable, so I'll go ahead and merge this now. Later we can make it conditional on the Bug#16301 stuff.

PR #3914, opened 2019-07-07
Fix typo magit-revision-headerS-format

topimiettinen:
Option magit-revision-header-format is actually called
magit-revision-headers-format.


tarsius:
Thanks!

PR #3915, opened 2019-07-08
Remove conditions for Emacs version lower than 25

ynilu:
Since Magit no longer supports Emacs version lower than 25, these conditions should be removed.


tarsius:
Thanks!

Issue #531, opened 2013-01-12
support temporarily ignoring files using git update-index --assume-unchanged


jixiuf:
```
[alias]
    ignore = !git update-index --assume-unchanged
    unignore = !git update-index --no-assume-unchanged
    # ignored = !git ls-files -v | grep ^[a-z]
         # ignored = !git ls-files -v | grep "^[[:lower:]]" | awk '{print $2}'
    ignored = !git ls-files -v | grep "^[[:lower:]]"
```

I can use this command in my shell ,

```
     git ignore filename 
     git unignore filename 
     gilt ignored  
```

and I hope magit can support this feature.

kuanyui:
Just want to say that I indeed encountered some situation which need this function...

During development, I want to modify a file temporarily for debug / testing; however this modification:

1. Don't want to be committed nor pushed because it's quite large and not needed by others.
2. The modification is quite large, which cause fold/expand function of Magit crash.
3. Even if the fold/expand function is available, that's easily be committed & pushed carelessly.

zilongshanren:
@jixiuf 

I have write some elisp code to implement this feature  and it works perfect for assume-unchanged:

```
(defun magit-skip-assume-unchanged-files ()
  (--keep (and (and (= (aref it 0) ?h)
                    (substring it 2)))
          (magit-git-items "ls-files"
                           (string-trim
                            (car (magit-git-items "rev-parse" "--show-toplevel")))
                           "-v" "--full-name" "-z")))

(defun magit-insert-assume-unchanged-files ()
  "Insert a tree of assume unchanged files.

    If the first element of `magit-buffer-diff-files' is a
    directory, then limit the list to files below that.  The value
    of that variable can be set using \"D -- DIRECTORY RET g\"."
  (when-let ((files (magit-skip-assume-unchanged-files)))
    (let* ((base (car magit-buffer-diff-files))
           (base (and base (file-directory-p base) base)))
      (magit-insert-section (assume-unchanged nil t)
        (magit-insert-heading "Assume-unchanged files:")
        (magit-insert-files files base)
        (insert ?\n)))))

(defun magit-list-all-files ()
    (magit-with-toplevel
      (with-temp-buffer
        (apply #'magit-git-insert '("ls-files" "-z" "--full-name"))
        (split-string (buffer-string) "\0" t))))

(defun magit-assume-unchanged (file)
  "Call \"git update-index --assume-unchanged FILE\"."
[...]

PR #3925, opened 2019-07-24
add assume unchanged support. It also improves the skip worktree.

zilongshanren:
resolve issue https://github.com/magit/magit/issues/531

I have added two keybindings, it might not be optimal, helps arewanted.

thanks.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -828,6 +828,13 @@ tracked file."
 (defun magit-file-tracked-p (file)
   (magit-git-success "ls-files" "--error-unmatch" file))
 
+ 
+(defun magit-list-all-files ()
+  (magit-with-toplevel
+    (with-temp-buffer
+      (apply #'magit-git-insert '("ls-files" "-z" "--full-name"))
+      (split-string (buffer-string) "\0" t))))
+
 (defun magit-list-files (&rest args)
   (apply #'magit-git-items "ls-files" "-z" "--full-name" args))
 
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -868,10 +868,21 @@ tracked file."
   (magit-git-items "ls-files" "-z" "--others" "--ignored"
                    "--exclude-standard" "--directory"))
 
+(defun magit-assume-unchanged-files ()
+  (--keep (and (and (= (aref it 0) ?h)
+                    (substring it 2)))
+          (magit-git-items "ls-files"
+                           (string-trim
+                            (car (magit-git-items "rev-parse" "--show-toplevel")))
+                           "-v" "--full-name" "-z")))
+
 (defun magit-skip-worktree-files ()
   (--keep (and (and (= (aref it 0) ?S)
                     (substring it 2)))
-          (magit-git-items "ls-files" "-z" "-t")))
+          (magit-git-items "ls-files"
+                           (string-trim
+                            (car (magit-git-items "rev-parse" "--show-toplevel")))
+                           "-z" "--full-name" "-t")))
 
 (defun magit-revision-files (rev)
   (magit-with-toplevel
--- a/lisp/magit-gitignore.el
+++ b/lisp/magit-gitignore.el
@@ -130,6 +130,26 @@ Rules that are defined in that file affect all local repositories."
     (magit-completing-read "File or pattern to ignore"
                            choices nil nil nil nil default)))
 
+;;; Assume unchanged Commands
+;;;###autoload
+(defun magit-assume-unchanged (file)
+  "Call \"git update-index --assume-unchanged FILE\"."
+  (interactive (list (magit-read-file-choice "Assume unchanged for: "
+                                             (cl-set-difference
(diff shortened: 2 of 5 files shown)


lisp/magit-git.el
```diff
@@ -828,6 +828,13 @@ tracked file."
 (defun magit-file-tracked-p (file)
   (magit-git-success "ls-files" "--error-unmatch" file))
 
+ 
+(defun magit-list-all-files ()
```
tarsius: How is this different from `magit-list-files` except for the use of `magit-with-toplevel` and lacking the ARGS argument?

IMO it is the callers that should use `magit-with-toplevel` if appropriate. The difference in the names of these functions, i.e. whether `all-` appears in it or not, doesn't hint at what the differences between the functions are.

Is the `magit-with-toplevel` really necessary in

lisp/magit-git.el
```diff
@@ -861,10 +868,21 @@ tracked file."
...
+                           "-v" "--full-name" "-z")))
+
 (defun magit-skip-worktree-files ()
   (--keep (and (and (= (aref it 0) ?S)
                     (substring it 2)))
-          (magit-git-items "ls-files" "-z" "-t")))
+          (magit-git-items "ls-files"
+                           (string-trim
+                            (car (magit-git-items "rev-parse" "--show-toplevel")))
+                           "-z" "--full-name" "-t")))
```
tarsius: Do this in a separate commit and explain what it does.

lisp/magit-gitignore.el
```diff
@@ -130,22 +130,42 @@ Rules that are defined in that file affect all local repositories."
...
+                                              (magit-list-all-files)
                                               (magit-skip-worktree-files)))))
   (magit-with-toplevel
     (magit-run-git "update-index" "--skip-worktree" "--" file)))
 
 ;;;###autoload
 (defun magit-no-skip-worktree (file)
   "Call \"git update-index --no-skip-worktree FILE\"."
-  (interactive (list (magit-read-file-choice "Do not skip worktree for: "
+  (interactive (list (magit-read-file-choice "Do not skip worktree for"
```
tarsius: Do this and the other removal of `: ` in a separate commit.

lisp/magit-status.el
```diff
@@ -685,6 +687,22 @@ of that variable can be set using \"D -- DIRECTORY RET g\"."
         (magit-insert-files files base)
         (insert ?\n)))))
 
+(defun magit-insert-assume-unchanged-files ()
+  "Insert a tree of assume unchanged files.
+
+    If the first element of `magit-buffer-diff-files' is a
+    directory, then limit the list to files below that.  The value
+    of that variable can be set using \"D -- DIRECTORY RET g\"."
```
tarsius: Do not indent doc-strings. (Look at this using `describe-variable` to see why.)

lisp/magit-git.el
zilongshanren: `magit-list-all-files` will list all the files in the git repository regardless where you run the function , while the `magit-list-files` only list files under current buffer directory.

With the `magit-with-toplevel` wrapper, the git commands could be executed in the root of git repository.

lisp/magit-git.el
zilongshanren: @tarsius Sure. I could split the commit into two speparate commits.

It means run `git ls-files` in the root dir of a git repository.

lisp/magit-status.el
zilongshanren: @tarsius Ok, I don't notice the difference here and just bindly re-indent the doc string here.
I will take a look at what it behaviors using `describe-variable`

tarsius:
The obvious part of this is to copy the relevant skip-worktree code and doing `s/skip-worktree/assume-unchanged`.  There are some additional details you overlooked though. You also noticed that there was an issue with paths, regarding sub-directories.  I do not agree with the way you addressed those issue.

I decided it would be easier to make the changes myself instead of guiding you to do so.  I did however try to do it in a away that allows you to see what different decisions I came to and why.

I did not add the section inserters to the the hook (you added it to the wrong hook in your commit).

zilongshanren:
@tarsius Alright.

phil-s:
Thanks for the work on this handy new feature @zilongshanren.

@tarsius, I didn't see key bindings for the new commands in the commits, but I feel that they would logically fit under the "ignore" menu?

tarsius:
> logically fit under the "ignore" menu?

Done.

zilongshanren:
should we mention adding the ignored section in magit status buffer? I think it's more obvious when people mark some files as unchanged/skip worktree and they can see these files in the status buffer.

```      
  (magit-add-section-hook 'magit-status-sections-hook
                                'magit-insert-assume-unchanged-files nil t)

        ;; insert the hidden files section in the magit status buffer.
        (magit-add-section-hook 'magit-status-sections-hook
                                'magit-insert-skip-worktree-files nil t)
```

tarsius:
Done that too.

Issue #3924, opened 2019-07-23
Staging with wrong line number


svenssonaxel:
Version info: Magit 2.90.1, Git 2.20.1, Emacs 26.1, gnu/linux

* Steps to reproduce:
1. Create a file with the following contents (including a newline after the last line):
line1
line2
line3
2. Commit the file
3. Edit the file to have the following contents:
line1
add1
line2
add2
line3
4. Run magit-status
5. Run magit-diff-less-context repeatedly until diffs show with no context.
6. Stage the following diff using magit-stage:
@@ -2,0 +4 @@ line2
+add2

* Expected:
7. The magit-status buffer contains only one unstaged change; the one including add1.
8. The output of ‘git diff --cached’ shows the following staged change:
@@ -1,3 +1,4 @@
 line1
 line2
+add2
 line3

* Actual:
7. The magit-status buffer contains three unstaged changes.
8. The output of ‘git diff --cached’ shows the following staged change:
@@ -1,3 +1,4 @@
 line1
 line2
 line3
+add2

kyleam:
When there are no context lines, the function responsible for staging the hunk, `magit-apply-patch`, calls `git apply` with `--unidiff-zero`:

```
--unidiff-zero
    By default, git apply expects that the patch being applied is a
    unified diff with at least one line of context. This provides good
    safety measures, but breaks down when applying a diff generated
    with --unified=0. To bypass these checks use --unidiff-zero.

    Note, for the reasons stated above usage of context-free patches
    is discouraged.
```

I suspect the weirdness you're describing (confirmed on my end) is something that we'll have a hard time avoiding when applying context-free patches.  Note that you can keep the context that `apply` needs but still stage just the "add2" line by selecting it before calling `magit-stage`.

tarsius:
I'll file this under "not great but expected".

svenssonaxel:
I agree that it's not great, but I don't think it's expected. Let me show why I think so:

1. Create a file with the following contents (13 lines containing "line", including a newline after the last line):
   line
   line
   line
   line
   line
   line
   line
   line
   line
   line
   line
   line
   line

2. Commit the file
3. Edit the file to have the following contents:

   line
   line
   line
   ADD1
   line
   line
   line
   line
   line
   line
   line
   ADD2
   line
   line
   line

4. Run magit-status
5. Set the number of context lines to 3
6. Stage the diff containint "ADD2".

This also exhibits the "not great" behaviour, but **git CLI does not**:
[...]

PR #3927, opened 2019-07-29
apply: Adjust hunk line positions for partial application

kyleam:
When applying a hunk or a region of a hunk that is not the first hunk
from the file, the line number in the header for the new start
position will often be incorrect because it accounts for content of
the ignored hunk(s) from above.  This usually isn't problematic
because there is enough context for 'git apply' to locate the
appropriate spot, but, in some cases where the context is too
repetitive to identify the location, it can lead to incorrect staging.

Note that the problem here is simpler than the one that 'git add -p' needs
to solve because, due to Magit's restrictions on what makes a valid
selection, we know that the hunks are contiguous and from the same
file.

Re: #3924

### DIFF
--- a/Documentation/RelNotes/2.91.0.org
+++ b/Documentation/RelNotes/2.91.0.org
@@ -371,6 +371,10 @@
   paths relative to the current directory instead of the top-level of
   the repository.  6d3f12d58
 
+- Staging a hunk applied the change to the wrong location in rare
+  cases where repetitive diff context prevented ~git apply~ from
+  finding the correct location.  #3924
+
 - Various bug fixes to
   ~magit-branch-delete~ (3e73ff19d),
   ~magit--{upstream,pushbranch}-suffix-predicate~ (0ce7fbbc2),
--- a/lisp/magit-apply.el
+++ b/lisp/magit-apply.el
@@ -161,21 +161,56 @@ so causes the change to be applied to the index as well."
                      (concat (magit-diff-file-header section)
                              (magit-apply--section-content section))))
 
+(defun magit-apply--adjust-hunk-new-starts (hunks)
+  "Adjust new line numbers in headers of HUNKS for partial application.
+HUNKS should be a list of ordered, contiguous hunks to be applied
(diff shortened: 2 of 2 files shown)


tarsius:
Thanks for this!

Please also add a changelog entry.

kyleam:
> Please also add a changelog entry.

Updated.

<details>
<summary>range-diff</summary>

```
1:  fae5f57b8 ! 1:  b3f7a08a1 apply: Adjust hunk line positions for partial application
    @@ -10,13 +10,28 @@
         appropriate spot, but, in some cases where the context is too
         repetitive to identify the location, it can lead to incorrect staging.
     
    -    Note that problem here is simpler than the one that 'git add -p' needs
    -    to solve because, due to Magit's restrictions on what makes a valid
    -    selection, we know that the hunks are contiguous and from the same
    -    file.
    +    Note that the problem here is simpler than the one that 'git add -p'
    +    needs to solve because, due to Magit's restrictions on what makes a
    +    valid selection, we know that the hunks are contiguous and from the
    +    same file.
     
         Re: #3924
     
    + diff --git a/Documentation/RelNotes/2.91.0.org b/Documentation/RelNotes/2.91.0.org
    + --- a/Documentation/RelNotes/2.91.0.org
    + +++ b/Documentation/RelNotes/2.91.0.org
    +@@
    + - The default value for ~magit-view-git-manual-method~ was treated as
    +   invalid.  #3873
    + 
    ++- Staging a hunk applied the change to the wrong location in rare
    ++  cases where repetitive diff context prevented ~git apply~ from
    ++  finding the correct location.  #3924
    ++
    + - Various bug fixes to ~magit-branch-delete~ (3e73ff19d),
    +   ~magit--{upstream,pushbranch}-suffix-predicate~ (0ce7fbbc2),
    +   ~magit-patch-create~ (470272a0f), ~magit--cherry-spinoff-read-args~
    +
      diff --git a/lisp/magit-apply.el b/lisp/magit-apply.el
      --- a/lisp/magit-apply.el
      +++ b/lisp/magit-apply.el
```

</details>

Issue #2931, opened 2017-01-07
Pre-populate commit messages with a list of modified symbols


tarsius:
The library `add-log`, which is part of Emacs allows semi-automatically creating ChangeLog files.  I think this library addresses a very real need, but does so in a highly inefficient and restrictive way.

A few years back there was a period where many users requested that similar features be implemented in Magit and my position was that, this should be done by improving `add-log` and Magit would then use that improved `add-log`.  (I wasn't the only one who thought that this would be a good idea, see http://debbugs.gnu.org/cgi/bugreport.cgi?bug=16301.)

Unfortunately nobody stepped up to improve that library, so I am now entertaining to do it myself or to write an alternative from scratch if I come to the conclusion that is more efficient and or that my improvements are likely to be rejected by the Emacs maintainers on grounds of them being to invasive.

---

Currently Magit *does* wrap `add-log` to provide the following commands:

- `magit-commit-add-log` Add a stub for the current change into the commit message buffer. If no commit is in progress, then initiate  it.
- `magit-add-change-log-entry` Find change log file and add date entry and item for current change. This differs from `add-change-log-entry` (which see) in that it acts on the current hunk in a Magit buffer instead of on a position in a file-visiting buffer.
- `magit-add-change-log-entry-other-window` Find change log file in other window and add entry and item.

I don't think the latter two commands make much sense in the age of Git.  ChangeLog are now used for high-level descriptions of the changes and the extensive log can be found in the commit messages.  But they will remain, just in case someone actually uses them.

But `magit-commit-add-log` isn't really all that useful either.  It does not generate and insert a list of changed changes.  Instead the user has to go to each change in the diff and say "and for that change too, add a stub to the commit message".

---

The functionality should be split like so:

1. Generate a list of all changed symbols.
2. Format the list according to some preference.
3. Insert the formatted list somewhere.

Currently `add-log` does not clearly separate these tasks and, if I remember correctly it doesn't do (1) at all.  But that's the most essential part.

---

Once (1) exists (in `add-log.el` or some other library), Magit will provide one function that can be added to `git-commit-setup-hook` to automatically insert the list of modified symbols.  Users will be able to select from multiple functions that format the list using different styles or even write their own versions.

---

(And once (1) exists as a separate function it will also be possible to use it for other purposes.  For example it could be used to retroactively add notes (`git notes`) to existing commits.  That information could then be used to very efficiently get a list of commits that touched a certain symbol.  I'll add a separate issue about that.)

[...]

PR #3928, opened 2019-07-29
Use Emacs 27 ChangeLog generating function if available

npostavs:
This is an initial implementation for #2931, which uses the new ChangeLog generating functions I've written in [Emacs Bug#16301](https://debbugs.gnu.org/16301).

I'm not too sure about the setting of `rev1` and `rev2` in `magit-generate-changelog`. I tried to follow `magit-diff-while-committing` but I don't know if I really understood the logic there.

### DIFF
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -579,25 +579,36 @@ Don't use it directly, instead enable `global-git-commit-mode'."
 (declare-function change-log-insert-entries "add-log" (changelogs))
 (declare-function magit-commit-message-buffer "magit-commit" ())
 
-(defun magit-generate-changelog ()
+(defun magit-generate-changelog (&optional amending)
   "Insert ChangeLog entries into the current buffer.
-The entries are generated from the diff being committed."
-  (interactive)
+The entries are generated from the diff being committed.  If
+prefix argument, AMENDING, is non-nil, include changes in HEAD as
+well as staged changes in the diff to check."
+  (interactive "P")
   (unless (magit-commit-message-buffer)
     (user-error "No commit in progress"))
   (require 'diff-mode) ;; `diff-add-log-current-defuns'.
   (require 'vc-git)    ;; `vc-git-diff'.
-  (let ((rev1 nil)
+  (require 'add-log)   ;; `change-log-insert-entries'.
+  (unless (and (fboundp 'change-log-insert-entries)
+               (fboundp 'diff-add-log-current-defuns))
+    (user-error "`magit-generate-changelog' requires Emacs 27 or better"))
+  (setq default-directory
+        (if (and (file-regular-p "gitdir")
+                 (not (magit-git-true "rev-parse" "--is-inside-work-tree"))
+                 (magit-git-true "rev-parse" "--is-inside-git-dir"))
+            (file-name-directory (magit-file-line "gitdir"))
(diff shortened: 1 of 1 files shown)


tarsius:
I've actually also worked a bit on this a while back. Don't remember if it is any good but you might want to take a look at the `modified-defuns` branch.

tarsius:
> I'm not too sure about the setting of `rev1` and `rev2` in `magit-generate-changelog`. 

Yeah. My version contains some code to deal with this but it is not actually being used and I don't think I considered it ready. I'll look into that.

npostavs:
> take a look at the `modified-defuns` branch.

The `magit-diff--modified-lines` part is interesting, I didn't think of gathering the line numbers in a list; I used some local functions instead and it turned out more convoluted than I expected (I'm speaking of `diff-add-log-current-defuns` in the Emacs patch).

> Yeah. My version contains some code to deal with this but it is not actually being used and I don't think I considered it ready. I'll look into that.

After looking at this, I'm afraid I'm actually more confused than before. So I hope you can figure out what's needed here...

tarsius:
Have you pushed the change to Emacs to some branch?

npostavs:
I just pushed it to `master`: https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=01661f33c11654d1fe5fe1013332db2500b7f449

dickmao:
I'd like to submit https://github.com/dickmao/magit-patch-changelog as an alternative.

npostavs:
> After looking at this, I'm afraid I'm actually more confused than before.

So I've taken another look at this, I believe the reason for my confusion was that I was trying to solve an impossible problem: there's no good way of deducing whether or not we're amending after the fact. I've made it controllable by prefix arg for now.

Comparing to @dickmao's `magit-patch-changelog`  solution:

- `magit-patch-changelog` avoids the problem mentioned above by restricting the command flow, so the user can't call the changelog generating function from the standard magit-commit command (i.e., `--amend` simply isn't an option).
- My solution is shorter because it relies on code I've added in Emacs 27, whereas `magit-patch-changelog` is longer and works in earlier Emacs versions.
- I also didn't think to implement anything like `magit-patch-changelog-xref`, that does seem quite nifty.
- `magit-patch-changelog` provides some fine-grained `magit-patch-changelog-agg-down` and `magit-patch-changelog-agg-up` commands, whereas my code expects the user to just write up the changelog and use `M-q` to aggregrate any adjacent empty entries all at once. I could see preferences going either way on this.

dickmao:
I'm gratified someone, anyone, even looked at magit-patch-changelog.  I'm fairly sure no one uses it.

My feeling is the Changelog is a procedural relic of the past similar to TPS reports in form and dubious function, but so long as the emacs leadership continues to insist on it I suppose something ought to be done to make Changelog entries less like doing one's taxes.

tarsius:
I've merged this. Who would have though.

I made some minor changes, which I did not push to the branch, so that they could be inspected.

Thanks @npostavs. And for your patience too. :grinning:

npostavs:
To be honest, it was less patience specifically, and more that I haven't had much time to look at Emacs related stuff over the past months/year. I'm just starting to dig out now. :)

tarsius:
Welcome back!

PR #3936, opened 2019-08-06
magit-repolist-column-dirty: Change order to staged, unstaged, untracked

stepnem:
Use the order of urgency ("WIPness"): uncommitted files are generally
much more interesting than untracked files.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -8,7 +8,7 @@
 #+TEXINFO_DIR_CATEGORY: Emacs
 #+TEXINFO_DIR_TITLE: Magit: (magit).
 #+TEXINFO_DIR_DESC: Using Git from Emacs with Magit.
-#+SUBTITLE: for version 2.90.1 (v2.90.1-750-g7e6b544aa+1)
+#+SUBTITLE: for version 2.90.1 (v2.90.1-751-gc7077c593+1)
 
 #+TEXINFO_DEFFN: t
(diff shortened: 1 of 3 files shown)


tarsius:
The old order also sorts by WIPness. By decreasing WIPness instead of the increasing WIPness you prefer.  I prefer the old order and disagree on which changes are more "urgent".

When I look at the list of repositories in order to figure out which require my attention, then I want to take care of those with untracked files first. These repositories either lack the appropriate ignore rules, which is usually a quick fix. Or they contain new files that should be tracked but are not, in which case it is urgent to track them to ensure they are backed up. Next I take care of random changes that are laying around, for example leftover debug statements and accidentally saved typos (both of which are usually unstaged). And finally I do the real but unfinished work of polishing and then committing intentional changes. I do that last, after doing the urgent and the cleanup work.

stepnem:
So it _is_ more subjective than I thought, after all... :-)

Thank you for your rationale. I on the other hand just about never leave
things in the index, so staged files are definitely something that I want
to be alerted about.

Would you accept a patch adding an option to make the order
customizable?

-- 
Štěpán

tarsius:
> Would you accept a patch adding an option to make the order customizable?

Yes, but please make sure it doesn't need an inappropriate amount of code and/or documentation.

stepnem:
Thanks, please have a look at #3937.

PR #3937, opened 2019-08-06
magit-repolist-column-flag: Replacement for magit-repolist-column-dirty

stepnem:
Allow specifying arbitrary flags and predicates in custom order.

With the default value of magit-repolist-column-flag-alist, this is
fully backward compatible with the original magit-repolist-column-dirty,
which is also preserved as an obsolete function alias.

[This is a continuation of #3936]


tarsius:
That's a nice approach.  Thanks!

I made some mostly cosmetic changes before merging.

PR #3942, opened 2019-08-12
Fix keybinding typo in 2.91.0 release notes

danielmartin:
(magit-status-here) is in C-c M-g g; C-c M-g s is (magit-stage-file).


kyleam:
Thanks!

PR #3947, opened 2019-08-16
Fix ...-maybe-update-... functions.

bastianbeischer:
These functions did not show the correct objects when point was moving quickly
through the list. The object to show was stored when the command to go to the
next item was issued for the first time and then displayed later after a timer
elapsed.

This is inadequate when moving through the list quickly: By the time the objects
are shown point is actually further ahead so there is a discrepancy between the
shown object and the place where point is.

This commit also fixes a small bug in magit--maybe-update-blob-buffer which was
looking at the wrong variable.


tarsius:
Thanks!

PR #3949, opened 2019-08-16
Fix magit-read-file-trace argument error / undefined behaviour

jguenther:
In bcb501476, validation via `string-match` was removed, but `match-string` is still being called. This PR restores that call to `string-match` but doesn't do anything with the returned value.

* Error: "Wrong type argument: (or eieio-object class), nil, obj"

* Or sometimes no error, but the "-L" argument returned by this function would contain unrelated string-match data. Good ol' emacs global mutable state

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -557,6 +557,8 @@ the upstream isn't ahead of the current branch) show."
 (defun magit-read-file-trace (&rest _ignored)
   (let ((file  (magit-read-file-from-rev "HEAD" "File"))
         (trace (magit-read-string "Trace")))
+    (string-match "^\\(/.+/\\|:[^:]+\\|[0-9]+,[-+]?[0-9]+\\)\\(:\\)?$"
+                  trace)
     (concat trace (or (match-string 2 trace) ":") file)))
 
 ;;;; Setup Commands


lisp/magit-log.el
```diff
@@ -557,6 +557,8 @@ the upstream isn't ahead of the current branch) show."
 (defun magit-read-file-trace (&rest _ignored)
   (let ((file  (magit-read-file-from-rev "HEAD" "File"))
         (trace (magit-read-string "Trace")))
+    (string-match "^\\(/.+/\\|:[^:]+\\|[0-9]+,[-+]?[0-9]+\\)\\(:\\)?$"
+                  trace)
     (concat trace (or (match-string 2 trace) ":") file)))
```
npostavs: The value of `(match-string 2 trace)` is still undefined if the `string-match` fails; is it guaranteed to succeed?

kyleam: Also, when there is a match, I'm not seeing the point of calling `(or (match-string 2 trace) ":")` because the second match would be ":" or nil, so why not reduce it to ":"?

tarsius:
The proper fix is

```diff
 (defun magit-read-file-trace (&rest _ignored)
   (let ((file  (magit-read-file-from-rev "HEAD" "File"))
         (trace (magit-read-string "Trace")))
-    (concat trace (or (match-string 2 trace) ":") file)))
+    (concat trace ":" file)))
```

right?

lisp/magit-log.el
jguenther: ah good point :+1:

PR #3953, opened 2019-08-23
magit--add-face-text-property

npostavs:
For https://github.com/magit/magit/pull/3886#issuecomment-523511418

I think this should work, but it's getting late enough that I might
have missed something silly.


tarsius:
You look at the `face` property in order to modify that *and also `font-lock-face`*. Shouldn't you instead look at the `font-lock-face` to modify that and only that?

npostavs:
> Shouldn't you instead look at the `font-lock-face` to modify that and only that?

Yeah, I was confused.

PR #3956, opened 2019-08-28
Bump transient dependency

### LINKED ISSUE #3945: Symbol's function definition is void: transient-get-value when viewing logs
I'm getting an error when trying to view git logs.

Steps to reproduce:

1. Open a file that is under source control.
2. Open magit transient window.
3. Type `l l`, `l b` or `l L` to open logs.
4. Receive error message `Symbol's function definition is void: transient-get-value`

I expect to see the logs.
I actually see an error message.

Debug Output:
```
Debugger entered--Lisp error: (void-function transient-get-value)
  transient-get-value()
  magit-log-arguments()
  byte-code("\300\301!\302 B\207" [magit-log-read-revs t magit-log-arguments] 2)
  call-interactively(magit-log-current nil nil)
  command-execute(magit-log-current)
```

I have tried uninstalling magit and reinstalling and I'm still seeing the issue.  I recently updated magit this morning and this started happening.

Emacs Version: GNU Emacs 26.2
OS: Mac OS 10.15 Beta

dickmao:
Closes #3945


Silex:
I think the commit message should be "Bump transient dependency" :)

tarsius:
Thanks!

I redid the work in my own commit. IMO it's a trivial change that doesn't really have to be attributed to the author and when we do attribute contributions then we prefer if the author information is complete, which wasn't the case here. I wanted to avoid having to ask you for the missing information because that would delay this trivial change being "merged".

Issue #3957, opened 2019-08-29
Allow passing --branches to magit log


DamienCassou:
I would like to be able to pass `--branches=<some pattern>` to magit log. This would let me filter the result to some branch pattern. My use case is to get a log of our recent releases (which are tags in git). I can't just log the `master` branch, because some releases have cherry-picked commits and are thus not in master's ancestry.

Because it seems `--branches` (without argument) is already used for Magit's different log actions, I'm not sure how to proceed.

PR #3958, opened 2019-08-29
Support passing --branches=<pattern> to git-log [#3957]

DamienCassou:
Support passing --branches=<pattern> to git-log.

Tentative fix for #3957. I'm not sure if having an action is the best way to do what I did. I think maybe an infix argument would be better but then it may conflict with other usages of `--branches`.


tarsius:
Thanks!

I have changed the placement of the suffix.

PR #3972, opened 2019-09-29
magit-emacs-Q-command: Save to system clipboard instead kill-ring.

cireu:
Kill-ring is only accessible in Emacs, this allow user use generated command in terminal(or any other places)

### DIFF
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -763,9 +763,11 @@ See info node `(magit)Debugging Tools' for more information."
                 ;; Avoid Emacs bug#16406 by using full path.
                 "-l" ,(file-name-sans-extension (locate-library "magit")))
               " ")))
-    (message "Uncustomized Magit command saved to kill-ring, %s"
+    (message "Uncustomized Magit command saved to clipboard, %s"
              "please run it in a terminal.")
-    (kill-new cmd)))
+    (with-temp-buffer
+      (insert cmd)
+      (clipboard-kill-ring-save (point-min) (point-max)))))
 
 ;;; Text Utilities
 


tarsius:
I am not familiar with how the kill-ring works; especially integration with the system clipboard.

Don't we have to use *both* functions so that the string becomes available outside as well as inside Emacs? Also isn't there a function that does not require that the text is inserted into a temporary?

Ah, there is: `clipboard-kill-ring-save`. It also appears that looking at the code and following the trails could shed some light on my other question.

tarsius:
That's not what I meant. Doesn't that function make it unnecessary to use `with-temp-buffer` and `insert`? Also you are adding some random whitespace errors in some unrelated code.

cireu:
Use `clipboard-kill-ring-save` now.

> Don't we have to use both functions so that the string becomes available outside as well as inside Emacs? Also isn't there a function that does not require that the text is inserted into a temporary?

Personally, I can use `yank` to get content killed by `clipboard-kill-region`. But I think using `clipboard-kill-ring-save` maybe safer and that's mentiomed by manual.

cireu:
> Doesn't that function make it unnecessary to use with-temp-buffer and insert?

Can't find one.

tarsius:
> Can't find one.

Indeed there is no `clipboard-kill-new`. However look at the definitions of `clipboard-{yank,kill-ring-save,kill-region}`. You will notice that each of these functions is a simple wrapper around the respective prefix-less function and that the only difference is that `select-enable-clipboard` is let-bound to `t`. This suggests that let-binding that variable around `kill-new` would have the same effect as for the other functions.

But before we do that, lets have a look at `select-enable-clipboard` itself. Turns out the default global value is `t` too! So unless the user actually changed the value, our effort to bind it to `t` doesn't actually have any effect.

And if the user *has* set `select-enable-clipboard` to `nil` shouldn't we then respect that here too? If we decide to ensure the clipboard is used shouldn't we also ensure the primary selection is used by binding `select-enable-clipboard`?

```elisp
(let ((select-enable-primary   t)
      (select-enable-clipboard t))
  (kill-new "testibus"))
```

My answer to both tends to be *no*. If the user decides to disable these features, then so be it--even if they would be useful in this case at least. But cc @npostavs.

cireu:
It seems to be the issue of my config that disable `select-enable-clipboard` accidentally.

PR #3978, opened 2019-10-03
Add magit-worktree-move

J3RN:
I noticed that there exists a [`git worktree move` command](https://git-scm.com/docs/git-worktree#Documentation/git-worktree.txt-move) in the git worktree documentation, and figured it would make a good addition to the repertoire of existing magit worktree commands.

Since having implemented it, I've actually integrated it into my workflow—largely as a mechanism to move worktrees that I have accidentally created where they ought not to be.

### DIFF
--- a/lisp/magit-worktree.el
+++ b/lisp/magit-worktree.el
@@ -50,6 +50,7 @@ Used by `magit-worktree-checkout' and `magit-worktree-branch'."
     ("b" "worktree"              magit-worktree-checkout)
     ("c" "branch and worktree"   magit-worktree-branch)]
    ["Commands"
+    ("m" "Move worktree"         magit-worktree-move)
     ("k" "Delete worktree"       magit-worktree-delete)
     ("g" "Visit worktree"        magit-worktree-status)]])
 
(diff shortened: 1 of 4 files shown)


lisp/magit-worktree.el
```diff
@@ -76,6 +77,22 @@ Used by `magit-worktree-checkout' and `magit-worktree-branch'."
...
+   (list (magit-completing-read "Move worktree"
+                                (cdr (magit-list-worktrees))
+                                nil t nil nil
+                                (magit-section-value-if 'worktree))
+         (funcall magit-worktree-read-directory-name-function
+                  "Move worktree to: ")))
+  (let ((inhibit-magit-refresh t))
+    (magit-run-git "worktree" "move" worktree (expand-file-name path)))
+  (when (and (derived-mode-p 'magit-status-mode)
+             (string= (concat worktree "/") magit--default-directory))
```
npostavs: This comparison looks a bit dubious, maybe use `file-equal-p` instead?

J3RN: Dubious indeed! And Window-incompatible, I might add. I was actually looking for `file-equals-p` when writing this but Google wasn't being helpful. I'm new to elisp. :smile:

I moved this comparison to the `let` before the actual move occurs, because, as the docs say:
>If FILE1 or FILE2 does not exist, the return value is unspecified.

J3RN:
@kyleam Thanks for your feedback! I've made the following alterations:

- Added the same "main worktree" check from `magit-worktree-delete`.
- Removed setting `inhibit-magit-refresh`. After rewriting this, I tried running it without setting that variable and it seems to behave well even when you're moving the worktree you're currently viewing. I can reinstate this in a more thoughtful way if it is necessary.
- Used `magit-toplevel` to check whether you're viewing the worktree you're moving.
- Only call `magit-diff-visit-directory` if git command to move the worktree exits successfully (by way of using `magit-call-git` instead of `magit-run-git`).
- Solved the issue of existing directories by doing the path joining ourselves. This approach is somewhat questionable; I'm open to feedback on it.
- Added release note.
- Squashed my commits.

I'd be happy to add a call to `kill-buffer` if you decide that's necessary. I can see the argument either way.

J3RN:
Thanks again for the feedback @kyleam! Here are the changes I've made:
- Switched back from backquote to `list`. I was just trying to be fancy, really :)
- Moved path concatenation to after the move has been performed, and we're visiting the worktree's new directory. Since we had to detect whether the target directory existed _before_ the move, I assigned a boolean var in a `let` for this purpose before the move occurs.
- Switched to using `(file-exists-p default-directory)` over path comparison to determine if we're viewing the moved worktree. This follows `magit-worktree-delete`. Does it seems odd not to use `file-directory-p` to this end? It's inconsequential, I'm just curious.
- Use `zerop` over `(= ... 0)`. Nifty, I didn't know this existed!
- Consolidate down to one `magit-call-git` call over a `magit-call-git` and `magit-run-git`.

And squash, of course :smile:

kyleam:
Nice, thanks.

I'm going to push an update to your branch.  Aside from rebasing on the current master, it has a fixup commit with a regenerated *.texi file, as well as an additional fixup commit that adds in a magit-refresh call (along with a couple of cosmetics tweaks).

This is good to go from my POV.  I'll let it sit for another day before merging to see if any other comments come in.

>  * Switched to using `(file-exists-p default-directory)` over path comparison to determine if we're viewing the moved worktree. This follows `magit-worktree-delete`. Does it seems odd not to use `file-directory-p` to this end? It's inconsequential, I'm just curious.

Hmm, I dunno.  I hadn't thought about doing that way before seeing that -worktree-delete did, but once I saw it I thought it was a nice way to go about it.

Issue #3087, opened 2017-05-24
Magit needlessly causes server contact for unsaved disconnected Tramp buffers


jdtsmith:
`Magit 2.10.3, Git 2.11.0 (Apple Git-81), Emacs 25.2.1, darwin`

Issue: when you have _unsaved_ Tramp buffers that are no longer contactable (offline, or in another network environment), MAGIT still inquires for the `file-truename` of those files, which causes Tramp to try to contact the uncontactable server, leading to timeouts and interrupts, during e.g., commits.  This is easily reproduced by creating (but not saving) a file on a remote server via Tramp, then disconnecting the network, and performing a commit on an unrelated repository.

Borrowing a bit from the new autorevert code which attempts to prevent this behavior, and adding it to the predicate in `magit-save-repository-buffers` seems to solve this:

```
;; Avoid needlessly connecting to unrelated remotes.
(equal (file-remote-p buffer-file-name)
       (file-remote-p topdir))
```

tarsius:
The existing code already *tries* to accomplish this, but it makes sense that the use of `file-truename` counteracts that effort. However simply drooping that, does not seem to be the proper fix. It is there for a reason; symlinks may be involved, and when that is so, then a file may be inside a repository even when the repository path isn't a prefix of the file path (in one case symlinks might already have been followed, but not in the other).

tarsius:
One possible approach could be to not follow symlinks for *remote* files only.

jdtsmith:
Good point.  But symlinks can't span remotes.  What's wrong with first testing if both `topdir` and the buffer-file in question are on the same remote before worrying about their `file-truename`'s?  I took this idea from [your autorevert fix](https://github.com/magit/magit/commit/f4dc04c20e77756a8386777ce5c43d1e25745263) dealing with Tramp issues and specifically the new `magit-auto-revert-repository-buffer-p`.

```
(defun magit-save-repository-buffers (&optional arg)
  "Save file-visiting buffers belonging to the current repository.
After any buffer where `buffer-save-without-query' is non-nil
is saved without asking, the user is asked about each modified
buffer which visits a file in the current repository.  Optional
argument (the prefix) non-nil means save all with no questions."
  (interactive "P")
  (-when-let (topdir (magit-rev-parse-safe "--show-toplevel"))
    (save-some-buffers
     arg (-partial (lambda (topdir)
                     (and buffer-file-name
                          ;; Avoid needlessly connecting to unrelated remotes.
[...]

Issue #1472, opened 2014-08-24
`magit-status` on local repo connects to all open TRAMP servers


DarwinAwardWinner:
Calling `magit-status` on any repo will attempt to run a command on all TRAMP servers for currently open files. If any of those servers is unavailable, Emacs will hang forever until you do C-g, and it will be impossible to to `magit-status` without first killing the buffers for all inaccessible files.

Essentially this is because `magit-save-buffers-predicate-tree-only` calls (indirectly) `file-directory-p` in every file-visiting buffer, which needs to run a TRAMP command on each remote server (relevant parts of backtrace included below. I think this could be fixed by first testing whether the buffer file is in a subdirectory of the magit top dir in question, and only then testing whether it has the same magit top dir. I'll submit a pull request shortly.

```
  tramp-maybe-open-connection(["ssh" nil "s14" "/home/ryan/Projects/proton-globin" nil])
  tramp-send-command(["ssh" nil "s14" "/home/ryan/Projects/proton-globin" nil] "test 0 2>/dev/null; echo tramp_exit_status $?")
  ... [More TRAMP stuff] ...
  tramp-run-test("-d" "/ssh:s14:/home/ryan/Projects/proton-globin")
  ... [More TRAMP stuff] ...
  tramp-file-name-handler(file-directory-p "/ssh:s14:/home/ryan/Projects/proton-globin")
  file-directory-p("/ssh:s14:/home/ryan/Projects/proton-globin/")
  magit-get-top-dir("/ssh:s14:/home/ryan/Projects/proton-globin/")
  magit-save-buffers-predicate-tree-only()
  #[(buffer) "\305!\205V\306!\205V\307!?\205V\310!\204+ \205Vq\210\n\205V\311 \312V\205V\313    !\203;rq\210   )\205V\203A\314\207\314\310!\203P\315\316\310!\"\207\315\317\320!\"\207" [buffer pred buffer-offer-save arg queried buffer-live-p buffer-modified-p buffer-base-buffer buffer-file-name buffer-size 0 functionp t format "Save file %s? " "Save buffer %s? " buffer-name] 4](#<buffer rnaseq-count-star.R<proton-globin>>)
  map-y-or-n-p(...[LONG LINE OMITTED]...)
  save-some-buffers(nil magit-save-buffers-predicate-tree-only)
  magit-save-some-buffers()
  magit-status("/Users/ryan/Projects/Dist-Zilla-Role-MainPodReader/")
```

PR #3979, opened 2019-10-06
Check if buffer's file is unrelated remote before checking it exists

kevinjfoley:
`file-exists-p' calls tramp functions for remote files which can cause
Emacs to hang if the remote host is no longer accessible.

### Testing
It looks like this bug has been addressed a few times before (#1472 #3087) but subsequent updates have reintroduced it.  It would be nice to add a test to prevent this but I'm not sure the best way to test this. 

I'm happy to try adding a test if there are any suggestions on a good approach.  I'll also put more thought into it later and try to come up with something for discussion.

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -1337,10 +1337,13 @@ argument (the prefix) non-nil means save all with no questions."
        arg (lambda ()
              (and (not magit-inhibit-refresh-save)
                   buffer-file-name
-                  (file-exists-p (file-name-directory buffer-file-name))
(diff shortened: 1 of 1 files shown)


tarsius:
I think a comment would also get the job done. It should state which functions are problematic and why, and why the chosen order in which the various checks are performed is the way to go. That explanation should also be part of the commit message.

tarsius:
I have merged after adding such a comment.

kevinjfoley:
Sorry I missed the reply on this, I agree a comment should suffice, thanks for adding one and merging!

PR #3981, opened 2019-10-08
async-bytecomp-allowed-packages value can now be

thierryvolpiatto:
the symbol 'all instead of '(all) previously
'(all) is still supported though, but if a user set it to 'all, magit
will fail to load.

### DIFF
--- a/lisp/git-rebase.el
+++ b/lisp/git-rebase.el
@@ -81,7 +81,10 @@
 (require 'magit)
 
 (and (require 'async-bytecomp nil t)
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks! I pushed a somewhat improved version. Please take a look in case I did something stupid.

thierryvolpiatto:

> Thanks! I pushed a somewhat improved version. Please take a look in
> case I did something stupid.

Looks good, thanks.

-- 
Thierry

Get my Gnupg key:

PR #3985, opened 2019-10-15
Allow the directory at point as gitignore rule

krobelus:
In the magit status buffer, move the point over an untracked directory, that
is, a directory with only (and at least one) untracked files. Pressing "ip"
to add an ignore rule should probably suggest that directory.  This commit
makes it so.


tarsius:
I think this just works around defects of the current code. Using these debug statements might help to instead *fix* the current issues:

```elisp
(defun magit-gitignore-read-pattern ()
  (let* ((default (magit-current-file))
         (choices
          (delete-dups
           (--mapcat
            (cons (concat "/" it)
                  (when-let ((ext (file-name-extension it)))
                    (list (concat "/" (file-name-directory "foo") "*." ext)
                          (concat "*." ext))))
            (magit-untracked-files)))))
    (message "default(1) %S" default)
    (when default
      (setq default (concat "/" default))
      (unless (member default choices)
        (setq default (concat "*." (file-name-extension default)))
        (message "default(2) %S" default)
        (unless (member default choices)
          (setq default nil))))
    (message "default(3) %S, choices %S" default choices)
    (magit-completing-read "File or pattern to ignore"
                           choices nil nil nil nil default)))
```

krobelus:
Thanks for your guidance! I don't know how I could miss that spurious `"foo"` :sweat_smile: 
I have pushed a different approach now that not only suggests untracked files but also the untracked files or directories which are shown in the status buffer.

Honoring `magit-buffer-diff-files` probably rarely makes a difference but why not I guess.

lisp/magit-gitignore.el
```diff
@@ -116,14 +116,19 @@ Rules that are defined in that file affect all local repositories."
...
             (cons (concat "/" it)
-                  (when-let ((ext (file-name-extension it)))
-                    (list (concat "/" (file-name-directory "foo") "*." ext)
-                          (concat "*." ext))))
-            (magit-untracked-files)))))
+              (when-let ((ext (file-name-extension it)))
+                (list
+                  (concat "/" (file-name-directory it) "*." ext)
+                  (concat "*." ext))))
+            (-concat
```
tarsius: `nconc` works just as well here.

krobelus: done

lisp/magit-status.el
```diff
@@ -700,19 +700,20 @@ value of that variable can be set using \"D -- DIRECTORY RET g\"."
...
             (dolist (file files)
               (magit-insert-section (file file)
                 (insert (propertize file 'font-lock-face 'magit-filename) ?\n)))
             (insert ?\n)))))))
 
+(defun magit-status--untracked-files (base)
+  (--mapcat (and (eq (aref it 0) ??)
+                 (list (substring it 3)))
+            (magit-git-items "status" "-z" "--porcelain"
+                             (magit-ignore-submodules-p) "--" base)))
```
kyleam: Co-opting this for use in `magit-gitignore-read-pattern` seems problematic to me.  Its behavior depends on how the user configures `status.showUntrackedFiles`.  That's fine in the context of the status buffer.  But, in the context of adding directories to the choices in `magit-gitignore-read-pattern`, this means that users that set `status.showUntrackedFiles=all` won't be offered directories as candidates.  I don't think that's good because their preference for how files are shown via `status` doesn't imply they won't be intereseted in ignoring an entire directory.

Perhaps it's better to leave this be and work with the return value of `git ls-files -z --other --directory --exclude-standard --no-empty-directory`, either by extending `magit-untracked-files` or just with a direct `magit-git-items` call in `magit-gitignore-read-pattern`.

tarsius:
I am not sure why the code from `magit-insert-untracked-files` that you have turned into a new function `magit-status--untracked-files` is necessary in the first place. Wouldn't `magit-untracked-files` do the trick.

Since you now use both functions at the same time I assume you have realized something that escapes me for now. Please explain. Given the values I have feed them they return the same list. Do you have an example which causes them to return different values?

tarsius:
I have factored the "foo"-fix into a separate commit and already pushed that to master.

lisp/magit-status.el
tarsius: Could you add a code comment about `status.showUntrackedFiles`. That wasn't obvious to me as can be seen here https://github.com/magit/magit/pull/3985#issuecomment-544195626.

kyleam:
> I have factored the "foo"-fix into a separate commit and already pushed that to master.

Thanks :] 

> Wouldn't `magit-untracked-files` do the trick.

This isn't directed at me, but I'll try because I was just touched on the same topic with my inline comment.  The output of `magit-untracked-files` includes individual files in a directory even when all the files in the directory are untracked.  So the desired output, I think, is `git ls-files -z --other --exclude-standard --directory --no-empty-directory`; the current `magit-untracked-files` lacks `--directory --no-empty-directory`.

krobelus:
> Since you now use both functions at the same time I assume you have realized something that escapes me for now. Please explain. Given the values I have feed them they return the same list. Do you have an example which causes them to return different values?

My motivation was that `magit-untracked-files` didn't list untracked directories - which was ill-advised. I've realized now that there is a `--directory` option to `git ls-files` which will make everything simpler.

> So the desired output, I think, is `git ls-files -z --other --exclude-standard --directory --no-empty-directory`; the current `magit-untracked-files` lacks `--directory --no-empty-directory`.

Thanks, I'll do that

krobelus:
I have looked at the other uses of `magit-untracked-files` and I think the only one that might be able to benefit from `--directory --no-empty-directory` is `magit-find-file`, and even there it's probably not very useful. Hence I have only added the options for this call inside `magit-gitignore-read-patterns` instead of creating a function `magit-untracked-files-and-directories`.

lisp/magit-gitignore.el
```diff
@@ -116,14 +116,17 @@ Rules that are defined in that file affect all local repositories."
...
          (choices
           (delete-dups
            (--mapcat
             (cons (concat "/" it)
                   (when-let ((ext (file-name-extension it)))
                     (list (concat "/" (file-name-directory it) "*." ext)
                           (concat "*." ext))))
-            (magit-untracked-files)))))
+            (magit-list-files "--other" "--exclude-standard" "--directory"
+                              "--no-empty-directory" "--" base)))))
```
kyleam: Dropping the `magit-untracked-files` call drops the candidates for individual files in an untracked directory.  Say I have a directory `d/` with no tracked files and an untracked file `foo.el`.  On master (618c90f4) this would give the following candidates:

    ("/d/foo.el" "/d/*.el" "*.el")

With this change, it'd be

    ("/d/")

while, as I understand it, the goal is

    ("/d/foo.el" "/d/*.el" "*.el" "/d/")

So, similar to the last iteration of this PR, I think we'd need to take the union of `magit-untracked-files` and this `magit-list-files` call.

If we wanted to avoid the second git call, we could instead stick with only `magit-untracked-files` and add `(file-name-directory it)` to the list.  That has the disadvantage of adding the directory even when it already has tracked files, and it also would fail to offer the top-most untracked directory for nested paths (e.g. `<untracked 1>/<untracked 2>/foo.el`).

krobelus: Well spotted, I completely missed that. I added another call to `magit-untracked-files`.

kyleam:
Sorry for being slow to get back to this.  I'll have another look tonight (est).

tarsius:
> I've pushed an update that [... non-trivial changes ...]

We should probably get into the habit of adding a `Modified-by` pseudo-header in such cases.

kyleam:
> We should probably get into the habit of adding a `Modified-by` pseudo-header in such cases.

Good point.  Given this is still open, I've updated the commit message.

PR #3988, opened 2019-10-22
Introduce a new face to display commit message summaries.

bastianbeischer:
I think it would be nice if users were able to configure the face for the commit message summaries in magit. For example, I am using a very dim face for diff hunk headings, but I would like to use something more prominent for the commit message summary line in magit-revision buffers.


tarsius:
Thanks!

PR #3991, opened 2019-10-28
Add new option, `stage` for magit-commit-ask-to-stage

conao3:
Hi!

I usually commit with some changes.
Adding all changes without any confirming helps me to easy commit.
So I added 'stage new option to stage anything without confirm.


tarsius:
> I every check what I commit before commitng by magit showing diff.

Well that makes sense.
Merged with minor rewordings.

conao3:
Good! Thank you for maintain this package for many years!

PR #3992, opened 2019-10-28
[WIP] use remote name as default value for input remote URL

conao3:
Hi!
I want to create dwim interface for `magit-clone`.
My clone and pull-request flow is showed as below.
1. `M-x magit-clone`
2. Press `u` to select `Clone from URL`
3. Input `magit/magit` ; for example
4. Press `Enter`
  because I set `magit-clone-default-directory` to "~/dev/forks/"
  (Currently, remote is only `origin` to git@github.com:magit/magit.git)
5. Create branch and Commit something
6. Create fork via `M-x magit-status` and press `! !`, input `git fork`.
  (As my configure, `git` is alias to `hub`)
7. Add my fork url named `conao3` to remote
  via
  - `M a`
  - input `conao3`
  - browse my fork via browser
  - click `Clone or download`
  - click `Copy` button to copy `git@github.com:conao3/magit.git`
  - paste `git@github.com:conao3/magit.git` to mini-buffer
8. Push local branch to my fork
  via `P e` input `conao3/<feature>`
9. Fetch all things via `' f f` using `forge`
10. Create pull-request via `' c p`

Well, I want to fix input fork URL as remote named `conao3` step.
That work is almost done, but I couldn't find a way to get the SSH URL dynamically.
Is this interface good? And can you give me advice.

Also, since I first started using Magit in earnest, it has only been a week.
Please let me know if there is an efficient flow to create pull-request.

### DIFF
--- a/lisp/magit-remote.el
+++ b/lisp/magit-remote.el
@@ -98,9 +98,16 @@ has to be used to view and change remote related variables."
 ;;;###autoload
 (defun magit-remote-add (remote url &optional args)
   "Add a remote named REMOTE and fetch it."
-  (interactive (list (magit-read-string-ns "Remote name")
-                     (magit-read-url "Remote url")
-                     (transient-args 'magit-remote)))
+  (interactive (let* ((remote-name (magit-read-string-ns "Remote name"))
+                      (url (magit-get "remote.origin.url"))
+                      (repo-name (and (string-match "\\`git@github\\.com:\\([^/]+\\)/\\(.+\\)\\.git\\'" url)
+                                      (match-string 2 url))))
+                 (list remote-name
+                       (magit-read-url
+                        "Remote url"
+                        (and repo-name
+                             (format "git@github.com:%s/%s.git" remote-name repo-name)))
+                       (transient-args 'magit-remote))))
   (if (pcase (list magit-remote-add-set-remote.pushDefault
                    (magit-get "remote.pushDefault"))
         (`(,(pred stringp) ,_) t)
--- a/lisp/magit-remote.el
+++ b/lisp/magit-remote.el
@@ -61,6 +61,12 @@ has to be used to view and change remote related variables."
   :group 'magit-commands
   :type 'boolean)
 
+(defcustom magit-remote-guess-url nil
+  "Whether to guess remote URL using remote name."
+  :package-version '(magit . "2.91.0")
+  :group 'magit-commands
+  :type 'boolean)
+
 ;;; Commands
 
 ;;;###autoload (autoload 'magit-remote "magit-remote" nil t)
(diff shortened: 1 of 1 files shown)


tarsius:
> Please let me know if there is an efficient flow to create pull-request.

Install `forge` and then `' c p`.

conao3:
Yes, but we needed to push current branch to my fork isn't it? Or I don't need to do it?

tarsius:
Ah sorry, I overlooked you are already using forge.

I don't quite understand. Of course you have to push your changes somewhere to make it possible for others to pull them. Are you asking whether the push and the "please merge this" actions can be carried out using a single command as opposed to two distinct commands?

At (8) you should us `P p` instead.

tarsius:
I do intend to add a `magit-fork` command, which would create the fork on Github *and* create and fetch the corresponding remote. But I haven't gotten around to doing that yet. I believe there are third-party packages that implement that functionality though.

conao3:
Thanks for your advice!
But `P p` atempts to push `origin`; your `magit/magit`, that command fails...
So I need to use `P e` and input `conao3/<feature>`.

conao3:
Or if magit has `push my fork` feature via `P k`, my pull-request flow will more easy.
Or do you expect `magit/magit` remote to be saved as `upstream` and `origin` to register my fork?
In this case, it is convenient if the function of registering old `origin` as `upstream` and registering fork URL as `origin` is implemented because `magit/magit` is registered as `origin` when cloning.

tarsius:
> Or if magit has `push my fork` feature via `P k`, my pull-request flow will more easy.

Which of the numbered steps above would that command perform?

> Or do you expect `magit/magit` remote to be saved as `upstream` and `origin` to register my fork?

Most certainly not. This is what I recommend: [The Two Remotes](https://magit.vc/manual/magit/The-Two-Remotes.html).

conao3:
If magit has `P k` to `push my fork` feature, I can replace the whole `7` step into one keystroke.

Your `Two remotes` page is interesting. I read and try it.

conao3:
Now, back to the first question. How can I get remote SSH URL via magit api?

tarsius:
> Now, back to the first question. How can I get remote SSH URL via magit api?

I am not sure what you are expecting beyond `(format "git@github.com:%s/%s.git" owner name)`.

conao3:
Now I can change my proposal like this. But how to get `owner`?

tarsius:
Isn't "conao3" the owner in that case? Why not just `(setq who-am-i "conao3")`?

Or would the expected value be "magit"? Then:

```emacs-lisp
(let ((url (magit-get "remote.origin.url")))
  (and (string-match "\\`git@github\\.com:\\([^/]+\\)/\\(.+\\)\\.git\\'" url))
       (match-string 1 url))
```

conao3:
Sorry, I misunderstand it. We have `remote-name` with users input, so we need `repo-name` from magit api.
This interface works fine for me, please try.

tarsius:
I added `forge-fork` a while ago and have now added a key binding `' c f` and updated the documentation to say that it can also be used to add an existing fork as a remote.  I think that command is better than what we have come up here.

conao3:
OK, I try it.

PR #3993, opened 2019-10-31
Add libgit2 implementation for magit-get-current-branch

cedarbaum:
This change implements `magit-get-current-branch` using libgit2. This is to help make progress on #2959.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1313,7 +1313,7 @@ to, or to some other symbolic-ref that points to the same ref."
                (funcall predicate module))
            module))))
 
-(defun magit-get-current-branch ()
+(cl-defgeneric magit-get-current-branch ()
   "Return the refname of the currently checked out branch.
 Return nil if no branch is currently checked out."
   (magit-git-string "symbolic-ref" "--short" "HEAD"))
--- a/lisp/magit-libgit.el
+++ b/lisp/magit-libgit.el
@@ -69,6 +69,13 @@ If optional DIRECTORY is nil, then use `default-directory'."
          (unless noerror
            (signal 'magit-outside-git-repo default-directory)))))
 
+(cl-defmethod magit-get-current-branch
+  (&context ((magit-gitimpl) (eql libgit)))
+  (when-let ((repo (magit-libgit-repo))
+             (ref (libgit-reference-dwim repo "HEAD")))
+    (when (libgit-reference-branch-p ref)
+      (libgit-reference-shorthand ref))))
+
 ;;; _
 (provide 'magit-libgit)
 ;;; magit-libgit.el ends here


tarsius:
Thanks!

I'll delay merging this until after I have released 3.0; which should happen this year.

MatthewRock:
@tarsius any progress on that? I love using magit, but it gets really slow with bigger/more active repos. I'd love to take a look at helping with #2959 but it sounds like this needs to be pulled in first.

tarsius:
The plan now is "early February". It actually looks like I am going to make it.

cedarbaum:
Hi @tarsius - I am wondering if there are any updates regarding the plan/timeline to address #2959. Hope you are doing well!

tarsius:
I'm doing well, thanks!

Take a grain... a sack of salt and then: 3.0 in august or so, then git-handler and libgit2 during the rest of the year.

tarsius:
... but if *you* want to work on this now; I won't keep you from doing that!

tarsius:
So I was actually gonna merge this now but then I took a closer look and this is actually more costly than the non-libgit implementation, so I am closing it instead.

cedarbaum:
Hey @tarsius,

I found some time recently to experiment more with libgit2 and I have a few thoughts/questions - would greatly appreciate your input when you have some time 😄 

1.) For benchmarking, do you have a standard setup you use and could share? I have my own ideas of how to benchmark these functions, but it would probably be better to use the same methodology.

2.) For some functions, I can imagine performance with libgit2 being better than native Git on Windows (due to process overhead) and worse on other Unix-like systems. I don't know if this is the case for `magit-get-current-branch`, though I can find out. In such cases, would you be open to either (1) casing on the OS for which implementation to use per-function or (2) having predefined subsets of libgit2 functions to use (e.g., `libgit-full`, `libgit-hot-path-only`, etc.) which the user could then pick from?

3.) For other functions, such as `rev-parse`, I see from this comment that there was good evidence of significant speedup: https://github.com/magit/magit/issues/2959#issuecomment-353705489. If such functions still show significant performance gains, are you open to merging their libgit2 implementations now?

I appreciate any input you have on above and hope you're doing well!

tarsius:
1. I don't have any hidden benchmark infrastructure, but that certainly would be useful. :grin: 
2. Lets not do that for now it would complicate things and maybe/probably it is not necessary at all. Let's do the benchmarks first.
3. Well, the libgit2 implementation already is merged isn't it? The problem is more that it implements hardly anything, so it is kinda pointless to do so now. That's what's needed next, many implemented methods....

cedarbaum:
> Well, the libgit2 implementation already is merged isn't it? The problem is more that it implements hardly anything, so it is kinda pointless to do so now. That's what's needed next, many implemented methods....

Agreed, I phrased this poorly. What I meant to ask about was merging further _usages_ of libgit2 functions in magit-libgit.el, such as for `rev-parse` (which I've started working on).

tarsius:
I would merge that too but I am hoping you could do a few dozen methods. One by one does take a bit too long.

Gaeric:
I did some benchmarks under my Windows machine,
``` elisp
(defun gaeric/log-time (msg)
  (format "%s: %s" msg (format-time-string "%s.%6N")))

(defun gaeric/profiler-process (func)
  (insert (format ";; %s start\n" (gaeric/log-time (identity func))))
  (let ((time (benchmark-elapse (funcall func))))
    (insert (format ";; elapse %s seconds\n" time)))
  (insert (format ";; %s end\n" (gaeric/log-time (identity func)))))

(gaeric/profiler-process
 (lambda () (magit-process-git nil)))
;; (lambda nil (magit-process-git nil)): 1670136068.505386 start
;; elapse 0.034687 seconds
;; (lambda nil (magit-process-git nil)): 1670136068.540562 end

(gaeric/profiler-process
 (lambda () (magit-gitdir nil)))
;; (lambda nil (magit-gitdir nil)): 1670136079.620565 start
;; elapse 0.037991 seconds
;; (lambda nil (magit-gitdir nil)): 1670136079.659049 end

(gaeric/profiler-process
 (lambda () (libgit-repository-discover)))
;; (lambda nil (libgit-repository-discover)): 1670136216.256266 start
;; elapse 0.000176 seconds
;; (lambda nil (libgit-repository-discover)): 1670136216.256655 end
```
Actually, I ran the same benchmark on another Windows machine.
libgit2 takes about 2.5ms, git subprocess takes more than 120ms...
Can these data show that using libgit2 will improve performance under windows?
Or I need to do another more rigorous benchmarking?

If this data is reasonable, I think I should write some code to make magit work well on windows

PR #3996, opened 2019-11-07
Fix git-commit-save-message

h0t0sh0:
Hi!

Please consider the changes to fix the bug after the last update:
git-commit-save-message: Invalid function: (message (git-commit-buffer-message))


tarsius:
I don't think that's the issued. Emacs lisp is a lisp 2.

My fix is a2168f9ed92d90888f43cd9e88ff9d46673120dd.

PR #3998, opened 2019-11-09
magit-merge: Add an argument to select diff algorithm

honza:
Selecting a different diff algorithm when merging can prevent merge conflicts. This adds an argument to the merge popup which lets the user select which algorithm to use. The same selector is used as for magit-diff for consistency. 

More context can be found in [this post](https://blog.jcoglan.com/2017/06/19/why-merges-fail-and-what-can-be-done-about-it/).


tarsius:
Thanks for the suggestion. I went with an approach that avoids code duplication. I put it on level 5 so you'll have to enable it explicitly.

PR #4000, opened 2019-11-16
Fix `git-commit-co-authored' typo in git-commit-mode-menu

cpitclaudel:
Clicking the Co-authored-by entry in the Commit menu currently produces an error: `apply: Wrong type argument: commandp, git-commit-co-authored`.

This commit fixes that error by referring to the right function name.

Renaming the function would have made its name more consistent with other like functions, but it might have caused breakage? I couldn't find any use of the function on GitHub.

### DIFF
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -743,7 +743,7 @@ With a numeric prefix ARG, go forward ARG comments."
   (interactive (git-commit-read-ident))
   (git-commit-insert-header "Suggested-by" name mail))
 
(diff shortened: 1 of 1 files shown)


kyleam:
Thanks.

> Renaming the function would have made its name more consistent with other like functions, but it might have caused breakage?

That's true.  Given that the command was introduced quite recently (3add5310b, 2019-09-28), I'd be in favor of dropping the "-by" for consistency.

cpitclaudel:
Sounds good, updated.

Issue #3986, opened 2019-10-15
Diff highlighting broken in emacs 27.0.50


sdilts:
Emacs 27.0.50 changes the way line highlighting works, which causes cosmetic issues in diff buffers. The highlight now only extends to the end of the text, not to the end of the line. [Here is a Reddit thread that has pictures.](https://www.reddit.com/r/emacs/comments/diahh1/emacs_27_update_changed_how_highlighted_lines/) The relevant paragraph from `emacs/NEWS` reads:

> New face attribute `:extend` to control face extension at EOL. The new face attribute `:extend` controls whether to use the face for displaying the empty space beyond end of 
line (EOL) till the edge of the window. By default, this attribute is non-nil only for `region` and  `hl-line` faces; any other face that crosses end of line will not affect the display of the empty space at EOL.

The change was merged into emacs master at commit https://github.com/emacs-mirror/emacs/commit/3d6075e3ee8c447f8974b37007a1b1ae1af8917c.

tarsius:
Was this change discussed anywhere?

I for one won't rush to update all my packages to account for this and will instead wait until the dust has settled. Locally I am resetting to before this was merged for the time being.

cireu:
I encounter same issue and just submit a bug here.

https://lists.gnu.org/archive/html/bug-gnu-emacs/2019-10/msg01809.html

I hope more people can discuss this change and let the maintainers of Emacs hear our voice.

thierryvolpiatto:
Zhu Zihao <notifications@github.com> writes:

> I encounter same issue and just submit a bug here.
>
> https://lists.gnu.org/archive/html/bug-gnu-emacs/2019-10/msg01809.html
>
> I hope more people can discuss this change and let the maintainers of Emacs hear our voice.
>
> cc @thierryvolpiatto , this change also break helm's selection overlay

It doesn't break helm-selection, it just change it, overlay ends at eol
instead of at window-width, is it really a problem?

> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.*


-- 
Thierry

Get my Gnupg key:
gpg --keyserver pgp.mit.edu --recv-keys 59F29997

cireu:
> Zhu Zihao <notifications@github.com> writes:
[...]

PR #4002, opened 2019-11-19
Fix face-attribute 'region :extend on Emacs 27

adamkruszewski:
Add ignore-errors around face-attribute 'region :extend as mentioned in my comment on https://github.com/magit/magit/issues/3986#issuecomment-555391760

It caused byte-compiling errors on latest Emacs 27. Just slapped ignore-errors there as advised
by Henrik of doom-emacs at https://github.com/hlissner/doom-emacs/issues/2070

_I'm sending this pull-request - more as a convenience - not really sure if you would like to fix the problem exactly this way. But it fixed the issue I was experiencing._

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -551,7 +551,7 @@ If you prefer the old behaviors, then set this to t."
 (defface magit-diff-hunk-region
   `((t :inherit bold
        ,@(and (>= emacs-major-version 27)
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks! Merged after fixing the commit message. (This is *not* relevant "on latest Emacs 27"--it's only relevant on *non-recent* Emacs 27.)

adamkruszewski:
Thanks!

PR #4006, opened 2019-11-27
magit-log-propertize-keywords: Fix regexp search

basil-conto:
#### TL;DR

The `magit-keyword` face is currently applied to keywords in log messages greedily, e.g. the entire message `"[EMACS] Rebind C-]"` receives the `magit-keyword` face due to the trailing `]`.

#### ChangeLog

- Make unused grouping construct shy.
- Avoid applying `magit-keyword-squash` face to whitespace.
- Hoist `magit-log-highlight-keywords` guard out of loop.
- Make keyword regexp non-greedy so as to handle cases like `"[EMACS] Rebind C-]"`.
- Avoid meaningless quoting of `?\]` inside regexp, as per advice in [`(elisp) Regexp Special`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Regexp-Special.html).  This is because backslash quoting does not remove special meaning of `?-` and `?\]` in regexps.

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -1261,15 +1261,15 @@ Do not add this to a hook variable."
   t)
 
 (defun magit-log-propertize-keywords (_rev msg)
-  (let ((start 0))
-    (when (string-match "^\\(?:squash\\|fixup\\)! " msg start)
-      (setq start (match-end 0))
-      (magit--put-face (match-beginning 0) (1- start)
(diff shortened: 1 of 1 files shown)


tarsius:
`\\[` ... `]` *feels* unbalanced. Merging anyway because technically you are right.

basil-conto:
Thanks!

PR #4014, opened 2019-12-12
magit-extras: New option: magit-revision-format

magnusmalm:
A modest attempt to contribute something small to this great package. 

Any and all comments, critique, and tips on improvements very much welcomed.

### DIFF
--- a/lisp/magit-extras.el
+++ b/lisp/magit-extras.el
@@ -463,6 +463,12 @@ that)."
                (choice (regexp :tag "Find index regexp")
                        (const :tag "Don't number entries" nil))))
 
+(defcustom magit-copy-revision-abbreviated nil
+  "Whether to save abbreviated revision to `kill-ring' and `magit-revision-stack'."
+  :package-version '(magit . "2.91.0")
+  :group 'magit-miscellaneous
+  :type 'boolean)
+
 ;;;###autoload
 (defun magit-pop-revision-stack (rev toplevel)
   "Insert a representation of a revision into the current buffer.
@@ -552,6 +558,10 @@ the current section is a commit, branch, or tag section, push
 the (referenced) revision to the `magit-revision-stack' for use
 with `magit-pop-revision-stack'.
 
+When `magit-copy-revision-abbreviated' is non-nil, save the
+abbreviated revision to the `kill-ring' and the
+`magit-revision-stack'.
+
 When the current section is a branch or a tag, and a prefix
 argument is used, then save the revision at its tip to the
 `kill-ring' instead of the reference name.
@@ -586,7 +596,9 @@ hunk, strip the outer diff marker column."
                     (file-name-as-directory
                      (expand-file-name (magit-section-parent-value section)
                                        (magit-toplevel))))))
-           (setq value (magit-rev-parse value))
+           (setq value (magit-rev-parse
+                        (and magit-copy-revision-abbreviated "--short")
+                        value))
            (push (list value default-directory) magit-revision-stack)
            (kill-new (message "%s" (or (and current-prefix-arg ref)
                                        value)))))
@@ -613,7 +625,11 @@ the current section instead, using `magit-copy-section-value'.
 
 When the region is active, then save that to the `kill-ring',
 like `kill-ring-save' would, instead of behaving as described
-above."
+above.
+
+When `magit-copy-revision-abbreviated' is non-nil, save the
+abbreviated revision to the `kill-ring' and the
+`magit-revision-stack'."
   (interactive)
   (if (use-region-p)
       (call-interactively #'copy-region-as-kill)
@@ -626,7 +642,9 @@ above."
                              magit-buffer-range))
                           (magit-status-mode "HEAD")))))
       (when (magit-commit-p rev)
-        (setq rev (magit-rev-parse rev))
+        (setq rev (magit-rev-parse
+                   (and magit-copy-revision-abbreviated "--short")
+                   rev))
         (push (list rev default-directory) magit-revision-stack)
         (kill-new (message "%s" rev))))))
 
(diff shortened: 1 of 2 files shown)


lisp/magit-extras.el
```diff
@@ -463,6 +463,19 @@ that)."
                (choice (regexp :tag "Find index regexp")
                        (const :tag "Don't number entries" nil))))
 
+(defcustom magit-revision-format 'full
```
kyleam: This option name is pretty general.  I think it should be renamed to make it clearer that this about what gets added to the kill ring.

lisp/magit-extras.el
```diff
@@ -463,6 +463,19 @@ that)."
...
+  "Length of rev hash saved to 'kill-ring'.
+
+'short makes the rev hash short enough to be unique (hopefully).
```
kyleam: Nit: Might as well spell out "revision" here and elsewhere in the docstring.

The convention in elisp docstrings is to quote symbols between \`' (i.e. \`kill-ring' and \`short').

lisp/magit-extras.el
```diff
@@ -463,6 +463,19 @@ that)."
...
+An integer sets the rev hash length to that value
+
+'full (default) means full rev hash."
+  :package-version '(magit . "2.90.1")
```
kyleam: That points to an already released version.

lisp/magit-extras.el
```diff
@@ -463,6 +463,19 @@ that)."
...
+  :group 'magit-miscellaneous
+  :type '(list (choice (integer :tag "Length of rev hash"))
+               (choice (const :tag "Length of rev hash short enough to be unique (hopefully)" short))
+               (choice (const :tag "Full rev hash length (default)." full))))
```
kyleam: Nit: Inconsistent use of trailing periods.

lisp/magit-extras.el
```diff
@@ -626,10 +639,20 @@ above."
                              magit-buffer-range))
                           (magit-status-mode "HEAD")))))
       (when (magit-commit-p rev)
-        (setq rev (magit-rev-parse rev))
+        (if (not (eql magit-revision-format 'full))
```
kyleam: `eq` will do here and is more consistent with this code base.  Same applies below.

lisp/magit-extras.el
```diff
@@ -626,10 +639,20 @@ above."
...
+            (setq rev (magit-rev-parse (revision-format) rev))
+          (setq rev (magit-rev-parse rev)))
         (push (list rev default-directory) magit-revision-stack)
         (kill-new (message "%s" rev))))))
 
+(defun revision-format ()
```
kyleam: This name is problematic given elisp's lack of namespaces.  It should have a `magit-` prefix.

However, I think this code is simple enough to use inline, even more so if the option is reduced to a boolean value, as suggested above.

lisp/magit-extras.el
```diff
@@ -626,10 +639,20 @@ above."
...
+  "Return a rev-parse option string according to 'magit-revision-format'."
+  (cond ((numberp magit-revision-format)
+         (format "--short=%d" magit-revision-format))
+        ((eql 'short magit-revision-format)
+         (format "--short" magit-revision-format))
+        (t "")))
```
kyleam: [ A not-so-relevant note given that above I said I think the option should be reduced to a boolean and this function removed. ]

If the fall-through clause returned nil instead of "", you could have simplified the two `rev` assignments above to a single

```elisp
(setq rev (magit-rev-parse (revision-format) rev))
```

lisp/magit-extras.el
magnusmalm: What is the convention here? Since I do not know which version this code might be included in, should I omit package-version all together or use some placeholder value? I grepped through the magit.git repo but could not find any mentions regarding this.

Apologies if info on this is somewhere obvious that I missed. :)

magnusmalm:
Maybe I should have squashed and force pushed this? I am a bit nervous about force push in general but it might have been safe this time.

Anyway, thanks a lot for all the feedback! I hope I at least have improved the patch a bit and that I interpreted the comments correctly. :)

I also hope it was ok that I used your suggested commit message verbatim as it was really spot on as-is. :)

lisp/magit-extras.el
kyleam: Sorry for not being more explicit.  I think you should go with 2.91.0, which is what the draft release notes for the next version are marked for.

lisp/magit-extras.el
```diff
@@ -626,9 +634,12 @@ above."
...
                           (magit-status-mode "HEAD")))))
       (when (magit-commit-p rev)
-        (setq rev (magit-rev-parse rev))
-        (push (list rev default-directory) magit-revision-stack)
-        (kill-new (message "%s" rev))))))
+        (let ((rev (magit-rev-parse rev)))
+          (if magit-copy-buffer-revision-abbreviated
+              (let ((abbreviated-rev (magit-rev-parse "--short" rev)))
+                (kill-new (message "%s" abbreviated-rev)))
+            (kill-new (message "%s" rev)))
```
kyleam: You could simplify this to something like (untested)

```elisp
(setq rev (magit-rev-parse
           (and magit-copy-buffer-revision-abbreviated "--short")
           rev))
(push (list rev default-directory) magit-revision-stack)
(kill-new (message "%s" rev))
```

That is, limit the change to just the `rev` assignment.  That results in an abbreviated hash being pushed to the revision stack, but I think that is fine (and indeed preferable to an additional rev-parse call).

kyleam:
> Maybe I should have squashed and force pushed this? I am a bit nervous about force push in general but it might have been safe this time.

Yes, please force push in this case (working a topic branch).  The convention in this project is to polish/re-roll series rather than putting fixup commits on top.

tarsius:
No objects to the idea. Needs the latest suggested simplification.

magnusmalm:
I hope I didn't miss anything this time around. Not sure how elaborate the Rel Notes entries should be so I kept it real short. :)

magnusmalm:
Great! And thanks for the excellent feedback! :)

PR #4018, opened 2019-12-17
Add option to push with tags

aragaer:
While there is an option to push tags separately, my preferred workflow is to perform single push command with `--tags` option added.


tarsius:
IMO `--follow-tags` is better. It also already is supported by `magit-push`; the user just has to enable it once (try `C-x l` when the prefix is active).

aragaer:
Interesting. I didn't know about that option and never used it before.

tarsius:
Let me know if after using this argument for a while you still think the other one should be added as well.

PR #4031, opened 2020-01-18
magit-pull.el: Add autostash option

balajisivaraman:
Hey,

Long-time user of Magit. It's one of the main reasons why I keep returning to Emacs after trying out other editors.

There are occasions in my workflow where I will have made a commit and need to pull changes from remote before pushing. But I will also have some local changes that need to be stashed. Git provides the great `--autostash` option for this, which to the best of my knowledge wasn't exposed when pulling through Magit.

This PR attempts to add `--autostash` during Magit pulls. I've tested this locally, and it works.

This is my first PR to Magit, so kindly review and suggest if I've missed something.

Cheers.

### DIFF
--- a/lisp/magit-pull.el
+++ b/lisp/magit-pull.el
@@ -45,7 +45,8 @@
   :man-page "git-pull"
   [:description
    (lambda () (if magit-pull-or-fetch "Pull arguments" "Arguments"))
-   ("-r" "Rebase local commits" ("-r" "--rebase"))]
+   ("-r" "Rebase local commits" ("-r" "--rebase"))
+   ("-a" "Autostash local changes and pop after pull" ("-a" "--autostash"))]
   [:description
    (lambda ()
      (if-let ((branch (magit-get-current-branch)))


tarsius:
I would encourage you to split the pulling step into two separate steps fetching and rebasing/merging. If you do that, then you an be certain about what is going to happen at the second step. If you do both steps in one go, then you cannot know what you will get and that is particular dangerous when using `--autostash`.

Maybe not so much for you and me, who could figure things out if one of those hairy special cases occurred but for someone less experienced this could be a major issue and that's why I don't want to add this argument here, at least not enabled by default.

Lets wait for others to comment on whether we should not add it at all or only "on level 7".

kyleam:
> Lets wait for others to comment on whether we should not add it at all or only "on level 7".

While I agree with your preference for two steps, I'm ok with putting --autostash on level 7 in the pull transient.

lisp/magit-pull.el
```diff
@@ -45,7 +45,8 @@
   :man-page "git-pull"
   [:description
    (lambda () (if magit-pull-or-fetch "Pull arguments" "Arguments"))
-   ("-r" "Rebase local commits" ("-r" "--rebase"))]
+   ("-r" "Rebase local commits" ("-r" "--rebase"))
+   ("-a" "Autostash local changes and pop after pull" ("-a" "--autostash"))]
```
kyleam: * Should the key be `-A` for consistency with the rebase transient?

* I think the second `-a` is incorrect.  My understanding is that the value, when given, should match the actual CLI short argument.  `--autostash` doesn't have one, and `git pull -a` is short for `git pull --append`.

balajisivaraman: - Agreed on the `-A`.
- On the second comment, it's my lack of understanding of how the transient works. I'll relook based on your comments and make the correct change. Sorry.

PR #4039, opened 2020-01-31
Ensure that git-commit-read-ident-history variable is defined

### LINKED ISSUE #4035: git-commit-co-authored fails because of wrong argument
I'm not really sure if it's an issue or the problem of my own configuration (sorry for bothering in this case), but trying to call `git-commit-co-authored` fails with following message:

```
In ‘git-commit-co-authored History’ source: ‘#[0 \304\302\300\305\303$\211\205\306\307\310\301""\207 [nil nil git-commit-read-ident-history t helm-comp-read-get-candidates nil delete  helm-cr-default] 6]’ 
 (wrong-type-argument listp git-commit-read-ident-history)
```

At the begining I was pretty sure that `git-commit-read-ident-history` was not loaded in my doom-emacs config, but looking at https://github.com/magit/magit/blob/master/lisp/git-commit.el I can't find any other occurence of this symbol. Where does it actually come from?

Steps to reproduce:
`C-c tab C-a` when editing commit message.

magit-version:
Magit 20200126.1750, Git 2.25.0, Emacs 26.3, darwin

mbuczko:
Fixes #4035. Ensures that `git-commit-read-ident-history` variable is defined which allows to use `git-commit-read-ident` function without errors.

### DIFF
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -131,6 +131,7 @@
 (defvar flyspell-generic-check-word-predicate)
 (defvar font-lock-beg)
 (defvar font-lock-end)
+(defvar git-commit-read-ident-history nil)
 
 (declare-function magit-completing-read "magit-utils"
                   (prompt collection &optional predicate require-match


tarsius:
Thanks! I ended up doing this myself because I wanted to say a bit more about it in the commit message and place the definition elsewhere.

Issue #4051, opened 2020-02-17
xfuncname for Clojure


dotemacs:
According to https://github.com/magit/magit/issues/3706 this can be configured for Emacs Lisp which is great, but it doesn't quite work for Clojure.

In **.gitconfig**:

```config
[diff "clojure"]
        xfuncname = (^\\(.*|\\s*\\(def|defn|defrecord|defmulti|defmacro|deftest|ns).*
```

In **.gitattributes**:

```config
*.clj  diff=clojure
*.cljc diff=clojure
*.cljs diff=clojure
*.cljx diff=clojure
```

When the pointer is on the line where I have `defn` declared, that is under git, and I run: <kbd>M-x magit-log-trace-definition RET</kbd>, it works as expected: it narrows to the defined `defn` and I can see the git log just for that sexp.

Where it fails is when I try to narrow it to a sexp where I have `defmethod` defined, which will look something like:

```clojure
(defmethod process :foo:bar
...
```

If I were to run this via:

```shell
git log -L'/process :foo:bar/':src/app/foo/bar.clj
```
I do get the desired results back.

(NOTE: You'll note that I've edited the syntax above and instead of using just the colons around the desired string like so: **:function-name:**, I wrapped the desired string to search in single quotes, which I wrapped in forward slashes: **'/function-name /'**. Which works on the command line, since I want to match Clojure's `defmethod` declarations.)

But trying to tweak `magit-log-trace-definition` and changing:

```
   (cons (format ;;"-L:%s%s:%s"
                 "-L'/%s%s/':%s"
                 (regexp-quote fn)
...
```
(NOTE: I'm talking about this: https://github.com/magit/magit/blob/master/lisp/magit-log.el#L729-L730)

Then I chord: <kbd>C-x C-e</kbd>

And running <kbd>M-x magit-log-trace-definition RET</kbd> on a `defmethod` returns a buffer without any log lines. But if I run the command in the shell:

```shell
git log -L'/process :foo:bar/':src/app/foo/bar.clj
```

I do get the results.

What needs to be tweaked to enable the above setup to work on Clojure's `defmetho` definitions? 

Thanks

dotemacs:
Would it help to create a sample repo that would give you example defmethods, so that this can be tried out, rather than just following my write up?

tarsius:
[...]

PR #4055, opened 2020-02-21
Escape function name for magit-log-trace-definition

dotemacs:
This is useful when the xfuncname is defined for Clojure's defmethod.

Here is the setup:

In .gitconfig:

[diff "clojure"]
        xfuncname = (^\\(.*|\\s*\\(def|defn|defrecord|defmulti|defmacro|deftest|ns).*

In .gitattributes:

*.clj  diff=clojure
*.cljc diff=clojure
*.cljs diff=clojure
*.cljx diff=clojure

When the pointer is on the line where I have defn declared, that is
under git, and I run: M-x magit-log-trace-definition RET, it works as
expected: it narrows to the defined defn and I can see the git log
just for that sexp.

Where it fails is when I try to narrow it to a sexp where I have
defmethod defined, which will look something like:

(defmethod process :foo:bar
...

If I were to run this via:

git log -L'/process :foo:bar/':src/app/foo/bar.clj

I do get the desired results back.

With the previous setup the magit-log-trace-definition didn't work on
defmethods from Magit.

Reference: https://github.com/magit/magit/issues/4051

All credit to Kyle Meyer

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -727,7 +727,7 @@ restrict the log to the lines that the region touches."
   (magit-log-setup-buffer
    (list rev)
    (cons (format "-L:%s%s:%s"
-                 (regexp-quote fn)
+                 (replace-regexp-in-string ":" "\\:" (regexp-quote fn) nil t)
                  (if (derived-mode-p 'lisp-mode 'emacs-lisp-mode)
                      ;; Git doesn't treat "-" the same way as
                      ;; "_", leading to false-positives such as


kyleam:
@dotemacs Thanks for putting my suggestion from #4051 into a PR.  From my POV your proposed commit message is a bit off the mark.  For example, the setup that takes up most of the message isn't particularly relevant, and the core issue---that if a function name has a colon, it has to be escaped---isn't directly stated.  I've wrapped up my patch with a different message and add an entry to the release notes.  Thanks again for the report and PR.

dotemacs:
Thank you for for improving magit!

Issue #3808, opened 2019-03-27
magit-status takes several minutes to load the magit buffer


c02y:
(1) what behavior you expected
  `magit-statsu` or other commands in magit buffer will response quickly.
  (2) what behavior you observed
  `magit-status` will freeze Emacs and take several minutes to load the `magit: .emacs.d` buffer
  (3) and how we can reproduce the issue.
  
  Note that this problem occured to me when I firstly upgraded my magit to v2.1 several years ago, and it is still annoying me and never gets better. The old issue page: https://github.com/magit/magit/issues/2099

And I noticed that the magit buffer contains all the untracked/modified/new/deleted files and **their diffs**, in the end, the magit buffer is huge, I believe that's why it is slow. 

Is it able to not show the diffs and just load the files list, and show the diffs only when I execute a command (TAB/Return on a file or other commands)?

I really tried google/read official document/searched the issue pages from time to time about how to improve the performance of magit ever since I posted the old issue, but all of them doesn't help at all.

Here are the details:
  
  My .emacs.d repo currently contains a lot of untracked/modified/new/deleted files (most of them are pacakge files), but a simple `M-x magit-status` for .emacs.d will take several minutes (**around 5 minutes**) and the buffer `magit: .emacs.d` which the `matit-status` loaded is **49MB, 1218648 lines**, after I execute `magit-status` for .emacs.d directory, my Emacs will stuck until the `magit: .emacs.d` is loaded.

  And the CPU usage of Emacs process keeps being over 80% even after the `magit: .emacs.d` is loaded, and moving my cursor or executing other commands are extremely slow.

-----------------------------------
Also post the output of

  M-x magit-version RET
  
~~~
Magit 20190314.1221, Git 2.7.4, Emacs 26.1, gnu/linux
~~~

The result of `M-x profiler-start/report` of CPU for .emacs.d is:
~~~
- command-execute                                                                            329623  99%
 - call-interactively                                                                        329616  99%
  - funcall-interactively                                                                    328021  99%
   - magit-status                                                                            328018  99%
    - magit-status-internal                                                                  328018  99%
     - magit-mode-setup                                                                      328018  99%
      - magit-mode-setup-internal                                                            328018  99%
       - magit-refresh-buffer                                                                328012  99%
        - apply                                                                              328012  99%
         - magit-status-refresh-buffer                                                       327486  98%
          - magit-run-section-hook                                                           326585  98%
           - apply                                                                           326585  98%
[...]

Issue #2982, opened 2017-02-01
Improve performance


tarsius:
Magit's performance isn't so good to say the least.  I would go as far as to say that mediocre and in certain situations outright bad performance currently is Magit's biggest problem.

We have made some improvements in the past, like e.g. caching calls to Git during a buffer refresh, but these improvements are to a large extend outweighed by the addition of new features.  Already there exist useful features that are disabled by default because they can be to costly in some cases.

One major difference between Git and Magit is that the latter shows all kinds of useful information up-front.  The major disadvantage of Magit's approach is that this information has to be always up-to-date (else it couldn't be relied on), and that doing so costs time.

To achieve good performance the cost of keeping (primarily, but not only) the status buffer up-to-date has to be reduced.  And in order to do so, some fundamental changes have to be made.  Doing that will take a while because it makes it necessary to replace some core abstractions.  The good news is that improving or even replacing those abstractions will have other benefits beside improved performance.

---

**A** These are the primary causes of bad performance:

1. After the user has performed some action, the complete status buffer is refreshed.  And that is done by recreating its content from scratch.  Every time.  All of it.

2. The required information is collected by calling `git` many times.  Until recently, there was no alternative, but now Emacs has a foreign function interface and so we can start using `libgit2` now.

3. Expensive parsing (primarily diffs and logs) is done synchronously, even when we don't need the complete result immediately (because most of it is not visible until the user scrolls).

4. The results of expensive parsing are not being cached.  (The result of calls to `git` are cached, but that mitigates the effect of (2), not (1).)

5. Many sections are created by inserting the output directly into the user-visible buffer, where it is then manipulated to look and feel they way we want to.  Without changing that, it would be very hard to address (3) and (4).

---

**B** There are two primary paths for improving performance:

1. #2959 Start using `libgit2` instead of calling `git` over and over again.

2. {TODO create issue} Completely change how sections are created (using separate parsing-buffers, which likely will be preserved as a caching mechanism) and make it possible to update sections asynchronously and independently of the containing buffer.

Luckily these two paths can be tackled in parallel, because in most cases a given section is **either** created by inserting and then "washing" `git` output (A5) **or** by processing a list of items (or a single value) using Elisp and then inserting that as the sections content (A2).

So aside from this overview, these two changes will be discussed independently, in dedicated issues.

jkdufair:
I'd love to help contribute on #2. I've developed some decent familiarity with the section management functions. I'll keep an eye out for the new issue.

[...]

PR #4056, opened 2020-02-23
Ability to hide patches of deleted files, significantly speeding up `magit-status` (#2982)

bramadams:
Hi,

Before this patch, simply updating a package like org-plus-contrib, then running `magit-status` on my .emacs.d git repo would take 20 seconds, followed by similar delays upon (un)staging of files due to refreshing the magit buffer. The main reason was that deleted files have their entire contents shown in the patch subsections of the (un)staged sections of `magit-status`. This issue is similar to for example https://github.com/magit/magit/issues/3808, and is also related to https://github.com/magit/magit/issues/2982.

With this patch, `magit-status` and subsequent refreshes only take a couple of seconds.

The only thing the patch is doing is to invoke "git diff" with the "-D" (--irreversible-delete) option for the (un)staged file sections, see https://git-scm.com/docs/diff-options/2.14.2#diff-options---irreversible-delete. This option still shows the deleted file in the summary of the diff, but excludes the actual contents of the file. As such the buffer of `magit-status` is much smaller, which speeds up `magit-status-refresh-buffer`, while one can still (un)stage deleted files as before. The only thing that has become impossible is opening the deleted file by hitting enter.

Since some people might not like this idea, the patch also introduces a defcustom named `magit-diff-hide-deleted-file-patches` to toggle whether or not deleted files should have their patch shown by `magit-status` for the (un)staged change sections. Default value is `nil`, i.e., the current (slow) behaviour.

Unfortunately, "git diff" does not have an equivalent argument for new files, hence large added files could still slow down `magit-status-refresh-buffer`.

In practice, though, this might not be as bad as it sounds: in case of renamed files
or, for example, updated MELPA packages, if one first stages the old, deleted version of a file or folder (which prompts the now faster refresh), then the new (untracked) version, git's rename detection will kick in, which automatically hides the contents of the renamed files and hence significantly speeds up the second refresh.

TL;DR Significantly faster magit refresh by hiding deleted files' patches during (un)staging; in case of renaming, first stage the deleted file/folder.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -2713,7 +2713,7 @@ It the SECTION has a different type, then do nothing."
   (magit-insert-section (unstaged)
     (magit-insert-heading "Unstaged changes:")
     (magit--insert-diff
-      "diff" magit-buffer-diff-args "--no-prefix"
+      "diff" (if magit-diff-hide-deleted-file-patches (append magit-buffer-diff-args '("-D")) magit-buffer-diff-args) "--no-prefix"
       "--" magit-buffer-diff-files)))
 
 (defvar magit-staged-section-map
@@ -2735,7 +2735,7 @@ It the SECTION has a different type, then do nothing."
     (magit-insert-section (staged)
       (magit-insert-heading "Staged changes:")
       (magit--insert-diff
-        "diff" "--cached" magit-buffer-diff-args "--no-prefix"
+        "diff" "--cached" (if magit-diff-hide-deleted-file-patches (append magit-buffer-diff-args '("-D")) magit-buffer-diff-args) "--no-prefix"
         "--" magit-buffer-diff-files))))
 
 ;;; Diff Type
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -268,6 +268,12 @@ that many spaces.  Otherwise, highlight neither."
                                (integer :tag "Spaces" :value ,tab-width)
                                (const :tag "Neither" nil)))))
 
+(defcustom magit-diff-hide-deleted-file-patches nil
+  "Whether to hide deleted file patches (for performance reasons)."
+  :package-version '(magit . "3.0.0")
+  :group 'magit-diff
+  :type 'boolean)
+
 (defcustom magit-diff-hide-trailing-cr-characters
   (and (memq system-type '(ms-dos windows-nt)) t)
   "Whether to hide ^M characters at the end of a line in diffs."


tarsius:
I don't think a new Custom option is necessary. We can just add this argument to the two relevant transients.

You will have to reveal, enable and save the argument in both transients (on `D` and `d`).

tarsius:

tarsius:
The infix arguments used by a certain transient popups can be [set/saved](https://magit.vc/manual/transient/Saving-Values.html) for future invocations.  TL;DR Enter the transient as usual, enable and/or disable arguments, and finally type <kbd>C-x s</kbd> or <kbd>C-x C-s</kbd>.

bramadams:
Hi,

Works great, thanks for the improved fix!

BA

PS: Perhaps it's worth adding this functionality (+ the "stage deleted first" trick) to https://magit.vc/manual/magit/Performance.html, since it makes a big difference.

tarsius:
Please feel free to suggest a paragraph. For me files being deleted isn't a regular experience, so it doesn't matter much for me. Which reminds me, I wanted to ask:

> Before this patch, simply updating a package like org-plus-contrib,

Why? Why would "updating" (`git pull`?) cause files to disappear from just the working tree but not `HEAD` (or the other way around I suppose)?

bramadams:
> > Before this patch, simply updating a package like org-plus-contrib,
> 
> Why? Why would "updating" (`git pull`?) cause files to disappear from just the working tree but not `HEAD` (or the other way around I suppose)?

Sorry for the confusion: I meant "updating" in the emacs package manager sense, which boils down to deleting the old version of, say org-plus-contrib, and installing the new version. Since I version-control the installed packages in my emacs config, to commit the package update the Magit buffer for my repo shows the deleted version in the unstaged section (by default with full contents shown, hence slow) and the new version in the untracked section (just folder name, hence quick). With the -D tweak, the Magit buffer now refreshes almost instantly for this use case.

> Please feel free to suggest a paragraph. For me files being deleted isn't a regular experience, so it doesn't matter much for me. 

OK, here is the suggested text to replace the start of the 3rd paragraph of 9.2.2 on https://magit.vc/manual/magit/Performance.html#Performance:
***
<p>Meanwhile, there are a number of things that can be done to speed up the Magit buffer refresh. If you often deal with deleted files (folders), or with renamed files (folders), you can disable the preview of the contents of the deleted/renamed files (folders) by enabling the <code>--irreversible-delete</code> option of <code>git diff</code> (possible from version 3.0 of Magit on). The only downside of this performance optimization is that hitting enter on (un)staged deleted/renamed file will no longer open that file.<p>

<p>To enable <code>--irreversible-delete</code>, open the Magit status buffer, then open the <code>magit-diff</code> transient using <code>d</code>. In this transient, type <code>C-x l</code>, followed by <code>-D</code>, then input <code>1</code>, hit enter, then type again <code>-D</code>. Now you just need to make this option permanent for future sessions by typing <code>C-x C-s</code>. If desired, you can hide the (enabled) option again using <code>C-x l</code>, followed by <code>-D</code>, then input <code>5</code>, hit enter).</p>

<p>Unfortunately, there is no equivalent option for newly added files. However, in case of renamed files (folders; typical example is when committing an upgraded emacs package in your personal configuration repo), you can still speed up the Magit buffer refresh by making a habit of first staging the deleted file (folder) before staging the new file (folder). Git's internal rename detection will automatically hide the contents of the renamed file (folder), making the Magit buffer refresh much snappier.</p>

<p>If this does not work well in your case, a more extreme performance optimization is to only automatically refresh the current Magit buffer

PR #4061, opened 2020-02-29
Expand magit-patch-create transient

### LINKED ISSUE #4028: `format-patch`: new option for accessing to non-listed options
Dear,

Thank you for this awesome tool!

Currently, Magit provides `format-patch` and several options are controlled by `C-m <key>` (the main ones). However, if I understand correctly, it is not possible to use the other options of `format-patch` which are not listed. I am not convinced that all the options have to be accessible with `C-m <key>` but it appears to me really useful to add (let say) `C-m !` to authorize to type any arguments.

To be concrete, when I am on a branch, using the CLI, I run:
```
git format-patch --cover-letter --base=$(git merge-base --fork-point master)
```
to get `base-commit: <hash>` at the bottom of the cover letter where `<hash>` is the commit where the patches of the branch can straightforwardly be applied.

Using Magit, I would like to hit: `W c C-m l C-m !` and type `--base=$(git merge-base --fork-point master)` then end  `c`.

Obviously, sometime I need to use the option `--interdiff` or any other.

Let me know if it is already possible and/or I miss something.

What do you think?

All the best,
simon

kyleam:
This PR was prompted by gh-4028.  It reorganizes the `magit-patch-create` transient a bit (in terms of levels, sections, and one binding change) and adds the following options: `--base`, `--interdiff`, `--range-diff`, `--rfc`, `--cover-from-description`, and `--notes`.

I've lazily decided not to add release notes because there's already a catch-all entry for the Transient switch that says "many popups gained new arguments and/or commands, most of which are not mentioned explicitly in these release notes".

I'd be interested to hear alternative suggestions for levels and bindings.

Closes #4028.

### DIFF
--- a/lisp/magit-patch.el
+++ b/lisp/magit-patch.el
@@ -132,7 +132,13 @@ which creates patches for all commits that are reachable from
   :description "Thread style"
   :class 'transient-option
   :key "C-m s  "
-  :argument "--thread=")
+  :argument "--thread="
+  :reader #'magit-format-patch-select-thread-style)
+
+(defun magit-format-patch-select-thread-style (&rest _ignore)
+  (magit-read-char-case "Thread style " t
+    (?d "[d]eep" "deep")
+    (?s "[s]hallow" "shallow")))
 
 (define-infix-argument magit-format-patch:--reroll-count ()
   :description "Reroll count"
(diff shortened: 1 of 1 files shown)


tarsius:
I can't really contribute much to the conversation.  The key bindings are not great, but they derive from what I originally came up with (I think), so I can't complain about that.  This looks all fairly straight forward, so I would say: merge, we can improve the bindings later if the need arises or inspiration strikes.

kyleam:
> so I would say: merge, we can improve the bindings later if the need arises or inspiration strikes

Okay, sounds good to me.  Thanks.

tarsius:
Thanks you!

Issue #3879, opened 2019-05-16
Update documentation concerning `magit-log-section-arguments`


fredefox:
`magit-log-section-arguments` was removed in 4641aa07e0b1f4654fceb139a36b160d594d7846, however they are still mentioned in the user manual.

For my use-case I'm currently trying to figure out how to get rid of the `--graph` argument.

tarsius:
> For my use-case I'm currently trying to figure out how to get rid of the `--graph` argument.

https://magit.vc/manual/transient/Saving-Values.html

PR #4068, opened 2020-03-11
Remove documentation of removed variable magit-log-section-args

dandavison:
The variable `magit-log-section-args` is documented but no longer exists. See also #3879.


tarsius:
Thanks!

Issue #2909, opened 2016-12-28
Performance issues on OSX


atomontage:
*Edit by @tarsius: The cause of this has been found and been fixed on Emacs `master` branch. See [this](https://magit.vc/manual/magit/MacOS-Performance.html) for more information, including how to get a backport of that fix.*

----

"2.9.0-34-gfee9c17c"
GIT: 2.11.0

magit-stage of a single file with a 5-line diff is perceivably slow, around 1 second.
magit-unstage also suffers from the same perceived slowdown.

Some profiling data:

* Stage
```
magit-stage                                       1           0.8872        0.8872
magit-stage-1                                     1           0.887063      0.887063
magit-run-git                                     1           0.885151      0.885151
magit-process-file                                39          0.855452      0.0219346666
magit-refresh                                     1           0.669937      0.669937
magit-refresh-buffer                              1           0.668759      0.668759
magit-status-refresh-buffer                       1           0.666563      0.666563
magit-git-insert                                  14          0.491143      0.0350816428
magit-git-str                                     30          0.3144180000  0.0104806000
magit-rev-parse-safe                              28          0.2852569999  0.0101877499
magit-git-wash                                    4           0.225156      0.056289
magit-call-git                                    1           0.215206      0.215206
magit-toplevel                                    16          0.2069670000  0.0129354375
magit-call-process                                1           0.195608      0.195608
magit-git-items                                   2           0.1731359999  0.0865679999
```
* Unstage
```
magit-unstage                                     1           0.943516      0.943516
magit-unstage-1                                   1           0.943387      0.943387
magit-run-git                                     1           0.924936      0.924936
magit-process-file                                44          0.912173      0.0207312045
magit-refresh                                     1           0.666068      0.666068
magit-refresh-buffer                              1           0.665038      0.665038
magit-status-refresh-buffer                       1           0.662804      0.662804
magit-git-insert                                  14          0.485324      0.0346659999
magit-git-str                                     35          0.3729509999  0.0106557428
magit-rev-parse-safe                              33          0.3445889999  0.0104420909
magit-call-git                                    1           0.258859      0.258859
magit-toplevel                                    17          0.2507959999  0.0147527058
magit-call-process                                1           0.243106      0.243106
magit-git-wash                                    4           0.207833      0.05195825
magit-git-items                                   2           0.165855      0.0829275
magit-insert-unstaged-changes                     1           0.16204       0.16204
magit-insert-status-headers                       1           0.157472      0.157472
[...]

PR #4069, opened 2020-03-13
Add documentation on using absolute git path in Catalina

ymherklotz:
Add documentation on using absolute `git` path in Darwin in `magit-git-executable`. Using the default results in a 20 times slowdown for some reason.

This was reported in #2982 and seems to have been confirmed by a few people [on reddit](https://www.reddit.com/r/emacs/comments/ffwxc9/performance_problem_with_magit_on_macos/).

Let me know if I should change or reword anything.

### DIFF
--- a/Documentation/magit-section.org
+++ b/Documentation/magit-section.org
@@ -8,7 +8,7 @@
 #+TEXINFO_DIR_CATEGORY: Emacs
 #+TEXINFO_DIR_TITLE: Magit-Section: (magit-section).
 #+TEXINFO_DIR_DESC: Use Magit sections in your own packages.
-#+SUBTITLE: for version 2.90.1 (v2.90.1-944-g1916e83aa+1)
+#+SUBTITLE: for version 2.90.1 (v2.90.1-945-g99d09ff72+1)
 
 #+TEXINFO_DEFFN: t
 #+OPTIONS: H:4 num:3 toc:2
@@ -26,7 +26,7 @@ user options see [[info:magit#Sections]].  This manual documents how you
 can use sections in your own packages.
 
 #+TEXINFO: @noindent
-This manual is for Magit-Section version 2.90.1 (v2.90.1-944-g1916e83aa+1).
+This manual is for Magit-Section version 2.90.1 (v2.90.1-945-g99d09ff72+1).
 
 #+BEGIN_QUOTE
 Copyright (C) 2015-2020 Jonas Bernoulli <jonas@bernoul.li>
--- a/Documentation/magit-section.texi
+++ b/Documentation/magit-section.texi
@@ -31,7 +31,7 @@ General Public License for more details.
 @finalout
 @titlepage
 @title Magit-Section Developer Manual
-@subtitle for version 2.90.1 (v2.90.1-944-g1916e83aa+1)
+@subtitle for version 2.90.1 (v2.90.1-945-g99d09ff72+1)
 @author Jonas Bernoulli
 @page
 @vskip 0pt plus 1filll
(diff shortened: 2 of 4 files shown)


tarsius:
Thanks!

I added a note about it potentially also affecting other macOS releases.

Kazark:
I implemented this performance hack in Spacemacs, but now [it is a breaking the use of Magit over TRAMP](https://github.com/syl20bnr/spacemacs/pull/14070#issuecomment-751572477) (per @bcc32):

> FYI, this PR breaks the use of magit over TRAMP when the path to git is different on the local and remote machines. For example, on my mac, magit-git-executable gets set to /usr/local/bin/git (installed via homebrew), but this breaks when I try to access files remotely on my NixOS server using TRAMP, where git is at /run/current-system/sw/bin/git.
>
> Also, as it happens, on macOS Big Sur (11.1), I see no speed difference in magit-status with and without this patch. I think, at the very least, this code should be re-guarded behind an OS check. IMO, it seems weird to lock magit into using whatever git was found in path at startup.

Any advice?

tarsius:
@kyleam @npostavs should we add the following and replace 37 uses of the variable by function calls?

```elisp
(defun magit-git-excecutable ()
  "Return the value of the variable by the same name.

If you use a OS that is abysmal at locating executables *and*
want to connect to other machines using Tramp, then you might
have to set the variable to an absolute value and use another
function to use that only locally:

  (fset \\='magit-git-excecutable
        (lambda ()
          (if (file-remote-p default-directory)
              \"git\"
            magit-git-excecutable)))"
  magit-git-executable)
```

I just wish Apple/Microsoft got their act together or that we would collectively give up fixing their "it just works" OSs.

kyleam:
> should we add the following and replace 37 uses of the variable by function calls?

I suppose the additional overhead of the function call would be swamped by even the simplest git call, so I guess it's not a real concern, though I still dislike it.

The [linked comment above][1] suggests it may no longer be an issue on Big Sur.  I don't know my macOS versions, but quickly searching it seems like Big Sur follows Catalina.  If there hasn't been any confirmation that the workaround helps there, my vote would be to stick with the variable, and update the instructions to note that the workaround doesn't play well with TRAMP.

[1]: https://github.com/syl20bnr/spacemacs/pull/14070#issuecomment-751572477

tarsius:
> I suppose the additional overhead of the function call would be swamped by even the simplest git call, so I guess it's not a real concern, though I still dislike it.

Yes, though the number of call sites doesn't really map to the number of actual calls--the latter is larger. But this was more about "ugh, look how many places have to be adjusted". Not that that is all that important either.

> The linked comment above suggests it may no longer be an issue on Big Sur. I don't know my macOS versions, but quickly searching it seems like Big Sur follows Catalina. If there hasn't been any confirmation that the workaround helps there, my vote would be to stick with the variable, and update the instructions to note that the workaround doesn't play well with TRAMP.

Yes, lets not use a function until we have proof this workaround still helps.

> my macOS versions

Ah ha! :grin:

npostavs:
Wasn't there some change to Emacs that "fixed" forking efficiency on Mac? (that is, the Emacs version is more significant than the macOS one). Something about fork vs vfork, I think.

Looks like the change was in Emacs 26.1, according to #2909.

tarsius:
Yes... there's that. But allegedly there's also this other issue.

From what I have heard I don't think that this is related to forking efficiency. Macos is being slow for some other unknown reason when it looks for executables. Doesn't this information usually get cached? Maybe that cache is broken on macos in some cases. I don't know.

Issue #4074, opened 2020-03-23
magit-release-tag-regexp does not support SemVer pre-release versions


darkfeline:
`magit-release-tag-regexp` does not support [SemVer pre-release versions](https://semver.org/#spec-item-9).  e.g. `v2.0.0-alpha.1`.

This seems like a useful feature to have as I currently cannot use `magit-tag-release` for my pre-release versions.

PR #4075, opened 2020-03-23
Add support for pre-release SemVer versions

darkfeline:
https://github.com/magit/magit/issues/4074

### DIFF
--- a/lisp/magit-tag.el
+++ b/lisp/magit-tag.el
@@ -110,12 +110,27 @@ defaulting to the tag at point.
   (when remote-tags
     (magit-run-git-async "push" remote (--map (concat ":" it) remote-tags))))
 
+(defvar magit-tag-version-regexp-alist
+  '(("^[-._+ ]?alpha\\.?$" . -3)
+    ("^[-._+ ]?beta\\.?$" . -2)
+    ("^[-._+ ]?\\(pre\\|rc\\)\\.?$" . -1))
+  "Value to use for `version-regexp-alist' when parsing and sorting versions.
+The default value matches some common SemVer pre-release formats.
+See also `magit-release-tag-regexp'.")
+
 (defvar magit-release-tag-regexp "\\`\
 \\(?1:\\(?:v\\(?:ersion\\)?\\|r\\(?:elease\\)?\\)?[-_]?\\)?\
-\\(?2:[0-9]+\\(?:\\.[0-9]+\\)*\\)\\'"
+\\(?2:[0-9]+\\(?:\\.[0-9]+\\)*\
+\\(?:-[a-zA-Z0-9-]+\\(?:\\.[a-zA-Z0-9-]+\\)*\\)?\\)\\'"
   "Regexp used to parse release tag names.
 The first submatch must match the prefix, if any.
-The second submatch must match the version string.")
+The second submatch must match the version string.
+The default value matches SemVer version numbers, including
+pre-release versions.
+
+If this will match versions that are not dot separated numbers, you
+also need to set `magit-tag-version-regexp-alist' to recognize them
+and give them a sorting order.")
 
 ;;;###autoload
 (defun magit-tag-release (tag msg)
@@ -179,7 +194,9 @@ a tag qualifies as a release tag."
                        (let ((tag (substring line 0 (match-beginning 0)))
                              (msg (substring line (match-end 0))))
                          (and (string-match magit-release-tag-regexp tag)
-                              (let ((ver (match-string 2 tag)))
+                              (let ((ver (match-string 2 tag))
+                                    (version-regexp-alist
+                                     magit-tag-version-regexp-alist))
                                 (list (list (version-to-list ver)
                                             ver tag msg)))))))
                 ;; Cannot rely on "--sort=-version:refname" because
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -6504,7 +6504,10 @@ Also see [[man:git-tag]]
 
   First it prompts for the name of the new tag using the highest
   existing tag as initial input and leaving it to the user to
-  increment the desired part of the version string.
+  increment the desired part of the version string.  If you use
+  unconventional release tags or version numbers (e.g.,
+  ~v1.2.3-custom.1~), you can set the ~magit-release-tag-regexp~ and
+  ~magit-tag-version-regexp-alist~ variables.
 
   Then it prompts for the message of the new tag.  The proposed tag
   message is based on the message of the highest tag, provided that


darkfeline:
This naive fix is insufficient, because Magit uses Emacs's version functions like `version-to-list`.

tarsius:
> This naive fix is insufficient, because Magit uses Emacs's version functions like `version-to-list`.

It appears you have addressed that?

darkfeline:
> > This naive fix is insufficient, because Magit uses Emacs's version functions like `version-to-list`.
> 
> It appears you have addressed that?

Yes, please take a look.  The fix ended up being more involved than a simple regexp fix I made initially.

lisp/magit-tag.el
```diff
@@ -29,6 +29,20 @@
 
 (require 'magit)
 
+(defgroup git-tag nil
+  "Edit Git tags."
+  :prefix "git-tag-"
+  :link '(info-link "(magit)Tagging")
+  :group 'tools)
+
+(defcustom git-tag-version-regexp-alist
```
kyleam: These symbols should start with `magit-`.  (I haven't reviewed the larger behavioral change and don't use `magit-tag-release` myself, so I probably wouldn't be a good judge.)

lisp/magit-tag.el
```diff
@@ -112,7 +126,7 @@ defaulting to the tag at point.
 
 (defvar magit-release-tag-regexp "\\`\
 \\(?1:\\(?:v\\(?:ersion\\)?\\|r\\(?:elease\\)?\\)?[-_]?\\)?\
-\\(?2:[0-9]+\\(?:\\.[0-9]+\\)*\\)\\'"
+\\(?2:[0-9]+\\(?:\\.[0-9]+\\)*\\(?:-[a-zA-Z0-9-]+\\(?:\\.[a-zA-Z0-9-]+\\)\\)?\\)\\'"
```
kyleam: The [link][0] you gave in gh-4074 has "1.0.0-alpha", so shouldn't the second group you're adding be optional?  (I'd guess the same applies to the periods the `version-regexp-alist` variant above.)

[0]: https://semver.org/#spec-item-9

lisp/magit-tag.el
```diff
@@ -29,6 +29,20 @@
...
+  "Edit Git tags."
+  :prefix "git-tag-"
+  :link '(info-link "(magit)Tagging")
+  :group 'tools)
+
+(defcustom git-tag-version-regexp-alist
+  '(("^[-._+ ]?alpha\\.$" . -3)
+    ("^[-._+ ]?beta\\.$" . -2)
+    ("^[-._+ ]?\\(pre\\|rc\\)\\.$" . -1))
+  "Value to use for `version-regexp-alist' when listing version tags."
```
kyleam: This seems to be motivated by Sematic Versioning, so shouldn't that be mentioned/explained here?

lisp/magit-tag.el
darkfeline: Oversight, fixed

lisp/magit-tag.el
darkfeline: Yes and no.  This isn't strictly exclusive to SemVer; you could use this to support any custom versioning scheme that looks close enough to SemVer.  I added a note about the default value.

lisp/magit-tag.el
```diff
@@ -29,6 +29,21 @@
 
 (require 'magit)
 
+(defgroup magit-tag nil
```
tarsius: A new custom group for a single option seems like a bit much. `magit-miscellaneous` should be fine here.

lisp/magit-tag.el
kyleam: It looks like you adjusted `magit-release-tag-regexp` but not the regexps in `version-regexp-alist`, which still end with a non-optional `\\.`.  Using a tag like `5.2.0-rc`, I can trigger the following error (with a66c6d85 checked out):

```
Debugger entered--Lisp error: (error "Invalid version syntax: ‘5.2.0-rc’")
  signal(error ("Invalid version syntax: ‘5.2.0-rc’"))
  error("Invalid version syntax: `%s'" "5.2.0-rc")
  version-to-list("5.2.0-rc")
  (list (version-to-list ver) ver tag msg)
  [....]
```

Also, it seems problematic that `magit-release-tag-regexp` allows for non-numeric characters following ".", but the `version-regexp-alist` regexps do not account for this.  Cases where `magit-release-tag-regexp` and the regexp in `version-regexp-alist` disagree are susceptible to the above error, I think.

darkfeline: That is unavoidable.  You have to tell Emacs how to recognize and sort non-numeric version.  SemVer does not specify a sorting order.  For example, 2.1-frobble, and 2.1-nibble are both valid SemVer versions, but how are we supposed to know which one is higher?  The default value for `magit-tag-version-regexp-alist` handles what I determined to be "common cases", but will have to be customized by the user.

I made some changes for this, including making `magit-release-tag-regexp` customizable as well.

lisp/magit-tag.el
darkfeline: I have also added `magit-release-tag-regexp`, so I'm assuming two options are sufficient to justify a new group.  If not, just tell me and I will move into `magit-miscellaneous`.

lisp/magit-tag.el
kyleam: > That is unavoidable.

It is not clear what "that" is.  Based on quickly looking at the change you pushed and guessing from what you say next, it looks like you agreed that the first point was an issue and are talking about the second point (what follows "Also...").

> You have to tell Emacs how to recognize and sort non-numeric version. SemVer does not specify a sorting order. For example, 2.1-frobble, and 2.1-nibble are both valid SemVer versions, but how are we supposed to know which one is higher?

My point was more about the mismatch between `magit-release-tag-regexp`, which matches non-numeric bits following the period, and the default value for `magit-tag-version-regexp-alist`, which doesn't.  Why should the default value `magit-release-tag-regexp` match this case if doing so will just lead to an error when `version-to-list` is called?

Anyway, perhaps I'm missing something, but at least I've helped you catch a couple of other issues with this PR.  I'm detaching from this PR now.  As I said, I don't use `magit-tag-release`, and don't feel strongly enough about seeing this land to invest more time in it.

darkfeline: It is not possible in the general case for `magit-tag-version-regexp-alist` to support all values that match `magit-release-tag-regexp`, because you have to tell `magit-tag-version-regexp-alist` what priorities to use.  Let's say you use Greek letters for version names.  You would set `magit-release-tag-regexp` to `"[a-z]+"`, but then you must also set `magit-tag-version-regexp-alist` to `((alpha . 1) (beta . 2) (gamma . 3) ...)` so `version-to-list` recognizes them and `version-list-<` sorts them correctly.  (You could also set `magit-release-tag-regexp` to `"alpha|beta|gamma|..."`, but that's tedious and not semantically correct in the general case).

`magit-release-tag-regexp` controls what we match as a version tag.  Ideally, that's all a user would need to customize.  However, for practical reasons, the user also needs to customize how non-numeric versions sort, as that is something that cannot be deduced from `magit-release-tag-regexp`.  It may also be the case that there is no finite set of non-numeric versions, and the user will have to register new non-numeric versions with a sorting priority as they are encountered.  Magit will recognize the versions correctly since `magit-release-tag-regexp` is set, but will error out when trying to sort them.  The new change I made reports an error specifically for this case.

There are three cases:

1. Magit both matches the release tag with `magit-release-tag-regexp` and knows how to sort it with `magit-tag-version-regexp-alist`.
1. Magit matches the release tag with `magit-release-tag-regexp`, but doesn't know how to sort it.  We raise an error pointing the user to customize `magit-tag-version-regexp-alist`.
1. Magit does not consider the tag to be a release tag.

lisp/magit-tag.el
tarsius: > I have also added `magit-release-tag-regexp`

Very very few people are going to modify that variable and those that do will most likely do it manually instead of using the custom interface, so let's not add another option that just confuses most users.

> so I'm assuming two options are sufficient to justify a new group.

No that still wouldn't be enough to justify a new group. Please just use `magit-miscellaneous`.

lisp/magit-tag.el
```diff
@@ -180,14 +202,22 @@ a tag qualifies as a release tag."
...
                ;; The inverse of this function does not exist.
                #'version-list-< :key #'car)))))
 
+(defun magit--version-to-list (ver)
+  "Convert version string VER into a list of integers.
+This is like `version-to-list', but customizable for Magit."
+  (let ((version-regexp-alist magit-tag-version-regexp-alist))
+    (condition-case err
+        (version-to-list ver)
+      (error (error (format "%s (try changing magit-tag-version-regexp-alist?)" err))))))
```
tarsius, line 219: The `format` is redundant here. Also I think we shouldn't do this at all. Someone who modifies a regexp as complex as `magit-release-tag-regexp` should just test what they have come up with. In case of error the backtrace will also guide the user to the appropriate to the variable, so we don't need this wrapping.

lisp/magit-tag.el
tarsius: Please add the added characters on a separate line (end this line with `\`) for better readability.

tarsius:
I don't see why anything beyond this is needed:

```diff
@@ -112,7 +112,7 @@ (defun magit-tag-prune (tags remote-tags remote)
 
 (defvar magit-release-tag-regexp "\\`\
 \\(?1:\\(?:v\\(?:ersion\\)?\\|r\\(?:elease\\)?\\)?[-_]?\\)?\
-\\(?2:[0-9]+\\(?:\\.[0-9]+\\)*\\)\\'"
+\\(?2:[0-9]+\\(?:\\.[0-9]+\\)*\\(?:[-.]*[a-zA-Z]+[0-9]*\\)?\\)\\'"
   "Regexp used to parse release tag names.
 The first submatch must match the prefix, if any.
 The second submatch must match the version string.")
```

This gives e.g.:

```
(magit--list-releases)
=>
(("3" "3" "x")
 ("2-rc2" "2-rc2" "x")
 ("2rc1" "2rc1" "x")
 ("2beta" "2beta" "x")
 ("2-alpha2" "2-alpha2" "x")
 ("2-alpha" "2-alpha" "x")
 ("1" "1" "x"))
```

darkfeline:
> I don't see why anything beyond this is needed:
> 
> ```diff
> @@ -112,7 +112,7 @@ (defun magit-tag-prune (tags remote-tags remote)
>  
>  (defvar magit-release-tag-regexp "\\`\
>  \\(?1:\\(?:v\\(?:ersion\\)?\\|r\\(?:elease\\)?\\)?[-_]?\\)?\
> -\\(?2:[0-9]+\\(?:\\.[0-9]+\\)*\\)\\'"
> +\\(?2:[0-9]+\\(?:\\.[0-9]+\\)*\\(?:[-.]*[a-zA-Z]+[0-9]*\\)?\\)\\'"
>    "Regexp used to parse release tag names.
>  The first submatch must match the prefix, if any.
>  The second submatch must match the version string.")
> ```
> 
> This gives e.g.:
> 
> ```
> (magit--list-releases)
> =>
> (("3" "3" "x")
>  ("2-rc2" "2-rc2" "x")
>  ("2rc1" "2rc1" "x")
>  ("2beta" "2beta" "x")
>  ("2-alpha2" "2-alpha2" "x")
>  ("2-alpha" "2-alpha" "x")
>  ("1" "1" "x"))
> ```

That doesn't support v1.2.3-alpha.1.2 which is a valid SemVer version

lisp/magit-tag.el
darkfeline: Done

lisp/magit-tag.el
darkfeline: Reverted this to a variable.

lisp/magit-tag.el
darkfeline: done

lisp/magit-tag.el
darkfeline: Based on your comment here, I also made `magit-tag-version-regexp-alist` a regular variable because:

- out of the presumably minority of users using pre-release versions, the default value should hopefully cover almost all of their use cases.
- I think configuring it for a custom version scheme may be non-trivial because it must still conform to the way `version-to-list` works, so such a user can comfortably do it without customize.

lisp/magit-tag.el
```diff
@@ -179,7 +194,9 @@ a tag qualifies as a release tag."
                        (let ((tag (substring line 0 (match-beginning 0)))
                              (msg (substring line (match-end 0))))
                          (and (string-match magit-release-tag-regexp tag)
-                              (let ((ver (match-string 2 tag)))
+                              (let ((ver (match-string 2 tag))
+                                    (version-regexp-alist
```
darkfeline, line 198: I applied your fixup, but note that this causes the dynamic binding to be applied every time the lambda is invoked by map, which I think may perform worse.  (I realize this code is not performance critical, but just FYI).

darkfeline:
Friendly ping

tarsius:
Sorry for the delay.

Issue #4076, opened 2020-03-23
Cannot input GnuPG passphrase through pinentry over TRAMP


Apteryks:
(1) behavior you expected

When using Magit over TRAMP, I'd expect to be able to input my GnuPG passphrase when needed, for example for signing commits.

(2) what behavior you observed

Doing a fetch on an authenticated repository hangs, and I can see in the magit-process buffer ($ key) that it is querying for my passphrase there:

```
─────────────────────────────────────────────────────┐│ Please enter the passphrase to unlock the OpenPGP secret key:  ││ [...]
```

There's no way that I can provide the passphrase there, as input doesn't seem to be read, and keybindings are enabled.

(3) and how we can reproduce the issue.

This is the tricky part.  The following details the *remote* relevant GnuPG configuration. I'm using GnuPG 2.2.19 on the remote, with the config:
```
# PIN entry program
enable-ssh-support
allow-emacs-pinentry
```
I have the regular pinentry package installed (1.1.0), and launch my gpg-agent service with:

```
 gpg-agent --daemon --pinentry-program /abs/path/to/my/pinentry
```
And finally in my `~/.bashrc` I have the following config to use gpg-agent as an ssh-agent as well as redirecting the pinentry to the incoming SSH session:

``` bash
# Make gpg-agent available for any shell invocation.
export GPG_TTY=$(tty)

# We use gpg-agent instead of ssh-agent.
unset SSH_AGENT_PID
if [ "${gnupg_SSH_AUTH_SOCK_by:-0}" -ne $$ ]; then
    export SSH_AUTH_SOCK="$(gpgconf --list-dirs agent-ssh-socket)"
fi

# Allow pinentry to display even for remote SSH connections.
echo UPDATESTARTUPTTY | gpg-connect-agent > /dev/null
```

    M-x magit-version RET

Magit , Git 2.25.2, Emacs 26.3, gnu/linux

The same occurs when launching emacs with the Q switch and loading the magit library manually.

Apteryks:
It seems to me that there needs to be a new process filter to be used in `magit-process-filter', that would detect the pinentry prompt and forward the user's passphrase to the process.

Apteryks:
I added this to my ~/.emacs:
``` lisp
(setq-default magit-process-password-prompt-regexps
  '("^\\(Enter \\)?[Pp]assphrase\\( for \\(RSA \\)?key '.*'\\)?: ?$"
    ;; Match-group 99 is used to identify the "user@host" part.
[...]

PR #4077, opened 2020-03-24
magit-process: Add prompt regexps for GnuPG.

Apteryks:
Fixes <https://github.com/magit/magit/issues/4076>.

When using Magit on a remote Git repository via TRAMP (using SSH), the
gpg-agent of the remote may prompt for a password.  Pinentry displays
the prompt through the terminal of the remote process, which until now
was not being handled by magit-process.

* lisp/magit-process.el (magit-process-password-prompt-regexps): Add
regexps to match the GnuPG prompts when requesting a password for a PGP
or SSH key through pinentry.

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -166,6 +166,8 @@ itself from the hook, to avoid further futile attempts."
   '("^\\(Enter \\)?[Pp]assphrase\\( for \\(RSA \\)?key '.*'\\)?: ?$"
     ;; Match-group 99 is used to identify the "user@host" part.
     "^\\(Enter \\)?[Pp]assword\\( for '?\\(https?://\\)?\\(?99:[^']*\\)'?\\)?: ?$"
+    "Please enter the passphrase for the ssh key"
+    "Please enter the passphrase to unlock the OpenPGP secret key"
     "^.*'s password: ?$"
     "^Yubikey for .*: ?$"
     "^Enter PIN for .*: ?$")


lisp/magit-process.el
```diff
@@ -167,6 +167,8 @@ itself from the hook, to avoid further futile attempts."
     ;; Match-group 99 is used to identify the "user@host" part.
     "^\\(Enter \\)?[Pp]assword\\( for '?\\(https?://\\)?\\(?99:[^']*\\)'?\\)?: ?$"
     "^.*'s password: ?$"
+    "Please enter the passphrase for the ssh key"
+    "Please enter the passphrase to unlock the OpenPGP secret key"
     "^Yubikey for .*: ?$"
     "^Enter PIN for .*: ?$")
```
kyleam: Please bump the `:package-version` on this option.

Apteryks:
Hi! Thanks for having a look. When matching the full line, the prompt shown to the user appears as:
```
Please enter the passphrase to unlock the OpenPGP secret key:  ││ "Your Name <some@email>"       ││ 4096-bit RSA key, ID 182384758382938,  : 
```

To get the prompt, Emacs must be launched is an environment where `echo UPDATESTARTUPTTY | gpg-connect-agent` was run, to allow pinentry to display even for remote SSH connections.

Inputting the OpenPGP key seems to fail; the curses prompt window of pinentry remains active and blocks further actions until gpg-agent is restarted.  So this will need more work it seems :-/.

tarsius:
What's the regexp you used that resulted in that prompt?

tarsius:
Also that sounds like a really strange prompt. Could you post this prompt including surrounding output from when you do this on the command-line?

Anyway, lets go with this then:

    "Please enter the passphrase \\(for the ssh\\|to unlock the OpenPGP secrets\\) key"

tarsius:
I have trimmed down the comment myself.

Apteryks:
@tarsius Hello, and sorry for taking a bit of time to answer. Note that while this manages to prompts the user for the password, as I mentioned in my last comment above, it seems the user's input is not relayed correctly to the remote pinentry, as it hangs.

It's a small improvement (that it at least prompts) but it still doesn't work (I'm not sure why!). Sorry if I didn't stress that out clear enough in my earlier message.

tarsius:
I've reopened #4076 for that.

Apteryks:
@tarsius Thank you!

PR #4078, opened 2020-03-26
Fix Travis CI Build config

tsuu32:
Same as https://github.com/company-mode/company-mode/pull/963.

This PR silences travis Build config validation warnings and corrects `allow_failures:` config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 language: generic
 env:
   global:
@@ -13,6 +12,7 @@ env:
     - EMACS_VERSION=26.1
     - EMACS_VERSION=26     # 26.1.50, emacs-26 branch, built daily
     - EMACS_VERSION=master # 27.0.50, master branch, built daily
+matrix:
   allow_failures:
     - env: EMACS_VERSION=master
 install:


tarsius:
Thanks for looking into this.

These are the current warnings (taken from https://travis-ci.org/github/magit/magit/builds/663924372/config). I added the numbering.

1. env: unknown key allow_failures ([object Object])
2. root: deprecated key sudo (The key `sudo` has no effect anymore.)
3. env: key matrix is an alias for jobs, using jobs
4. root: missing os, using the default linux
5. root: missing dist, using the default xenial 

It appears you are only tackling 1 and 2.

I think 2, 4, 5 could be tackled in a commit. 3 in another commit. And 1 in yet another commit by itself.

2 through 5 are fairly obvious. 1 is not--at least the commit that addresses that should come with a description that covers the question of why we need to make that change. Link to the relevant documentation might help.

tarsius:
@npostavs Could you please look into updating the jobs. (Should 26 be replaced with 27?)

npostavs:
> (Should 26 be replaced with 27?)

Yeah, and maybe replace 26.1 with 26.3 (how much do we care about testing every minor version?)

tsuu32:
> These are the current warnings (taken from https://travis-ci.org/github/magit/magit/builds/663924372/config). I added the numbering.

1 and 2 are warnings, but 3, 4 and 5 are not warnings, just infos. So I only tackled 1 and 2.

> 2 through 5 are fairly obvious. 1 is not--at least the commit that addresses that should come with a description that covers the question of why we need to make that change. Link to the relevant documentation might help.

Travis documentation (https://docs.travis-ci.com/user/build-matrix/#matching-jobs-with-allow_failures) shows `allow_failures` examples which indicate that current magit's `.travis.yml` is not correct.

tsuu32:
This is a question: why magit's travis doesn't run by pull request? Is this intended?

tarsius:
> why magit's travis doesn't run by pull request? Is this intended?

I don't think so.

tsuu32:
> I don't think so.

Please see https://travis-ci.org/github/magit/magit/requests and https://travis-ci.org/github/magit/magit/pull_requests.
magit/magit's travis doesn't run by pull request, which is useful to check if the PR is acceptable. I suspect "Build pushed pull request" setting is turned off.

tarsius:
Travis seems to be broken for me and they no longer help users but refer to "community support" so I might as well describe the issues I am having here instead.

On https://travis-ci.org/account/repositories only a single repository from my user account is listed (its the only one with tests). Not a single repository from my organizations is listen. On the left it says:

> Organizations
> You are not currently a member of any organization.
> 
> Missing an Organization?
> [link]Review and add[/link] your authorized organizations. 

Following the link to https://github.com/settings/connections/applications/[some-hash] I can see that I have granted access to 9 of my 11 organizations, including to `magit`.

---

I did follow the links you provided but, probably due to the above issue, I did not find any nobs at these locations that I could turn to turn on things. @npostavs please check if you can change the settings on the travis pages.

tsuu32:
I have experienced similar situation. In that case signing out and re-log in travis solved it.

tarsius:
That helped... a little. The organizations and there repositories are now listed but I cannot get to the settings due to "insufficient permissions". As the owner...

npostavs:
> I think 2, 4, 5 could be tackled in a commit. 3 in another
> commit. And 1 in yet another commit by itself.

I put 3 together with 2, 4, 5.  See #4080.

> That helped... a little. The organizations and their repositories are
> now listed but I cannot get to the settings due to "insufficient
> permissions". As the owner...

Logging in and out seems to have worked for me.  I've set the CI to
run on pull requests.

tsuu32:
@npostavs Thanks for the following up with detailed commit messages. LGTM.

I'll close this PR after #4080 is marged.

tsuu32:
> Logging in and out seems to have worked for me. I've set the CI to
run on pull requests.

Currently travis seems to run by pull request (https://travis-ci.org/github/magit/magit/pull_requests )😄

Thanks.

tarsius:
And I seem to have gained full access to my repositories again. :grin:

tarsius:
Ah and I have lost access again.

PR #4081, opened 2020-03-30
Optional was misspelled.

EricGebhart:
_Optional_ was misspelled which causes the function _magit-inside-gitdir-p_ to not work.


tarsius:
Thanks!

I changed the commit message.

EricGebhart:
You are welcome.

Issue #4088, opened 2020-04-15
Handle password prompts in Git process


matthewbauer:
Password prompts in magit-process are not handled properly. This happens when `sudo` is needed for sending data to a repo.

My setup works like this:

- run `git remote add origin /etc`
- set `git config remote.origin.receivepack 'sudo git-receive-pack'` in the sending repo
- run `git push`
- observe sudo timeout in magit-process looking like this:

```
Pushing to /etc
[sudo] password for XXX: 
sudo: timed out reading password
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

magit-version:

```
    Magit 20200108.532, Git 2.23.1, Emacs 26.3, gnu/linux
```

matthewbauer:
I think this may be just as simple as adding another entry to magit-process-password-prompt-regexp.

matthewbauer:
Done in https://github.com/magit/magit/pull/4089

tarsius:
See #4089.

PR #4089, opened 2020-04-15
Add sudo password prompt to magit-process-password-prompt-regexps

matthewbauer:
Add sudo password prompt to magit-process-password-prompt-regexps.

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -170,7 +170,8 @@ itself from the hook, to avoid further futile attempts."
     "Please enter the passphrase to unlock the OpenPGP secret key"
     "^.*'s password: ?$"
     "^Yubikey for .*: ?$"
-    "^Enter PIN for .*: ?$")
+    "^Enter PIN for .*: ?$"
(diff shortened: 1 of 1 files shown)


kyleam:
Thanks for the PR.  The change itself looks fine to me, but I wonder if a setup where the user receives this prompt (e.g., what you describe in gh-4088) isn't quite rare.  In which case, it seems acceptable that a user would have to modify a `defcustom` to support their use case.

matthewbauer:
> Thanks for the PR. The change itself looks fine to me, but I wonder if a setup where the user receives this prompt (e.g., what you describe in [gh-4088](https://github.com/magit/magit/issues/4088)) is quite rare. In which case, it seems acceptable that a user would have to modify a `defcustom` to support their use case.

Yeah that's possible. I originally didn't realize something like `magit-process-password-prompt-regexps` exists. But the fact it is a defcustom means it's not too hard to do this in the configuration.

tarsius:
I would guess that this is quite rare too, so I am closing this for the time being.

PR #4090, opened 2020-04-18
Adjust magit-reshelve-since for upstream changes

kyleam:
As of Git v2.24, `magit-reshelve-since` will sit for ten seconds before doing anything while underneath Git warned to not use `git-filter-branch` and is waiting/hoping for users to cancel the call.  We can set an environment variable to get rid of that warning, as the second commit of this series does.  But I wondered how an implementation using `git-fast-{export,import}` would look.

After a couple of minor bug fixes in the surrounding code and some prep commits, the final commit of this series gets to that.  IMO the `fast-{export,import}` approach ends up being pretty nice.  While the `filter-branch` approach is shorter, much of the logic is injected into a shell `case` statement.  The `fast-*` approach, on there other hand, allows for pretty straightforward Emacs code, dumping to a buffer and processing the text.

Thoughts?  I'm open to the argument that this isn't worth the extra code for this obscure function; let's just set the environment variable and be done with it.

### DIFF
--- a/lisp/magit-extras.el
+++ b/lisp/magit-extras.el
@@ -393,31 +393,34 @@ be used on highly rearranged and unpublished history."
                          (float-time
                           (date-to-time
                            (read-string "Date for first commit: "
-                                        time-now 'magit--reshelve-history)))))
-                (process-environment process-environment))
-            (push "FILTER_BRANCH_SQUELCH_WARNING=1" process-environment)
+                                        time-now 'magit--reshelve-history))))))
             (magit-with-toplevel
-              (magit-run-git-async
-               "filter-branch" "--force" "--env-filter"
-               (format "case $GIT_COMMIT in %s\nesac"
-                       (mapconcat (lambda (rev)
-                                    (prog1 (format "%s) \
+              (magit--reshelve-filter-branch tstamp range)
+              (set-process-sentinel
+               magit-this-process
+               (lambda (process event)
+                 (when (memq (process-status process) '(exit signal))
+                   (if (> (process-exit-status process) 0)
+                       (magit-process-sentinel process event)
+                     (process-put process 'inhibit-refresh t)
+                     (magit-process-sentinel process event)
+                     (magit-run-git "update-ref" "-d" backup)))))))))))))
+
+(defun magit--reshelve-filter-branch (tstamp range)
+  (let ((process-environment process-environment))
+    (push "FILTER_BRANCH_SQUELCH_WARNING=1" process-environment)
+    (magit-run-git-async
+     "filter-branch" "--force" "--env-filter"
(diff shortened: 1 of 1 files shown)


tarsius:
> Thoughts? I'm open to the argument that this isn't worth the extra code for this obscure function; let's just set the environment variable and be done with it.

I'm afraid that's the conclusion I arrive at, but don't just discard it, instead keep the feature branch around. And merge the first 4-5 commits of course.

Even if we never merge this it is still nice to have some simple fast-import/export example at hand.

kyleam:
Thanks for the feedback.  I've pushed the changes unrelated to the fast-export rewrite to master.


<details>
<summary>range-diff</summary>

```
1:  fba49d1f = 1:  b96c273f magit-reshelve-since: Fix an un-formatted message
2:  aed90571 ! 2:  3555b06b magit-reshelve-since: Avoid git-filter-branch warning
    @@ Commit message
         alternative).  Each invocation of git-filter-branch has a 10 second
         delay where it shows a warning.
     
    -    Set FILTER_BRANCH_SQUELCH_WARNING to dismiss the warning and avoid the
    +    Based on reading the (long) list of issues in the git-filter-branch
    +    manpage, I don't spot any that I think apply to the restricted fashion
    +    in which we're using filter-branch in magit-reshelve-since.  Set
    +    FILTER_BRANCH_SQUELCH_WARNING to dismiss the warning and avoid the
         delay.
     
    +    If it turns out that we do want to move away from using
    +    git-filter-branch, refs/pull/4090/head provides an implementation that
    +    uses git-fast-{export,import}.
    +
    +    Re: #4090
    +
      ## lisp/magit-extras.el ##
     @@ lisp/magit-extras.el: (defun magit-reshelve-since (rev)
                           (float-time
3:  7d7e7d80 ! 3:  a7f60a7e magit-reshelve-since: Store branch and backup ref as variables
    @@ Metadata
      ## Commit message ##
         magit-reshelve-since: Store branch and backup ref as variables
     
    -    This is arguably an improvement because it avoids some repeated calls,
    -    and it will become more useful when the logic for determining the
    -    backup ref gains a condition.
    +    This is minor, but avoids a few repeated calls.
     
      ## lisp/magit-extras.el ##
     @@ lisp/magit-extras.el: (defun magit-reshelve-since (rev)
4:  4c01538e = 4:  8e28a154 magit-reshelve-since: Abort if head is detached
5:  817ad44f < -:  -------- magit-reshelve-since: Rename a variable for clarity
6:  fe84a590 < -:  -------- magit-reshelve-since: Move filter-branch bits to separate function
7:  9d124099 < -:  -------- magit-reshelve-since: Conditionally use fast-{export,import}
```

</details>

PR #4091, opened 2020-04-19
Customization of ANSI escape sequence processing in diff hunks

dandavison:
This PR adds two new user options:

- `magit-diff-preprocess-git-output-function`
  Function to perform initial processing of git output.

- `magit-diff-preprocess-git-output-always`
  Whether to always call the git output preprocessor function.

These options make it possible to use  a terminal git diff viewer such as [delta](https://github.com/dandavison/delta)<sup>1</sup> to apply 256-color and truecolor (24-bit) ANSI color escape sequences color to the git diff output, and to use [xterm-color](https://github.com/atomontage/xterm-color) instead of `ansi-color` to process the ANSI escape sequences ariving in the magit buffer. Here's a screenshot of Magit using delta (a minor-mode to do this is [here](https://github.com/dandavison/magit-delta)):

<table><tr><td>image</td></tr></table>

The magit diff buffer is fully functional (e.g. for staging hunks and lines) since `delta --color-only` guarantees that, after stripping ANSI escape sequences, the diff output is identical to that produced by git.

While this is a rather niche way to use Magit, it is fun to have the option. The main attraction of using delta with Magit is that one gets color syntax highlighting in the magit diff buffer (somewhat like GitLab, GitHub, etc; the syntax highlighting uses the same color themes as [bat](https://github.com/sharkdp/bat/)). Other possible benefits are

- Responsiblities for within-line string edit inference algorithms (`diff-refine-hunk`) are handed over to a dedicated external application
- Visual consistency between diffs displayed by Magit and diffs viewed in a terminal emulator

A minor-mode toggle for using delta with magit is here: https://github.com/dandavison/magit-delta (I'll submit that to MELPA once this PR is resolved).

I can add documentation of the new variables, but I wanted to check first whether it would be more appropriate to leave them undocumented for now (e.g. pending future work on unifying handling of ANSI escape sequences in contexts other than diffs).

### Testing

- I've checked that `--color-moved` still works (e.g. `-m d u` in the diff transient)
- Incidental magit usage on this branch.
- `make test` output is the same locally on this branch as on master

----------------------------------
<sup>1</sup> I'm the author of delta.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -243,6 +243,33 @@ t         Highlight only in added lines.
                  (const :tag "in added and removed lines" both)
                  (const :tag "in added, removed and context lines" all)))
 
+(defcustom magit-diff-preprocess-git-output-always
+  nil
+  "Whether to always call the git output preprocessor function.
+
+If non-nil, always preprocess git output using
+`magit-diff-preprocess-git-output-function'. If nil, this is done
+only when the current git arguments imply that ANSI escape
+sequences are to be expected in the git output (an example of
+such a git argument is --color-moved)."
+  :group 'magit-diff
+  :type 'boolean)
+
+(defcustom magit-diff-preprocess-git-output-function
+  #'magit-diff-convert-ansi-escape-sequences
+  "Function to perform initial processing of git output.
+
+The function will be called without arguments, with the current
+buffer containing raw git output. At a minimum, it must remove
+any ANSI escape sequences in the buffer and apply Emacs faces to
+the remaining text, creating the same visual effects as were
+implied by the ANSI escape sequences. The faces must be applied
+as overlays, not as text properties. It must leave point at the
+beginning of the buffer."
+  :group 'magit-diff
+  :type '(choice (function-item magit-diff-convert-ansi-escape-sequences)
+                 function))
+
 (defcustom magit-diff-highlight-trailing t
   "Whether to highlight whitespace at the end of a line in diffs.
 Used only when `magit-diff-paint-whitespace' is non-nil."
@@ -1982,6 +2009,10 @@ Staging and applying changes is documented in info node
    ;; oldBold, and newBold.
    ))
 
+(defun magit-diff-convert-ansi-escape-sequences ()
+  (require 'ansi-color)
+  (ansi-color-apply-on-region (point-min) (point-max)))
+
 (defun magit-insert-diff ()
   "Insert the diff into this `magit-diff-mode' buffer."
   (magit--insert-diff
@@ -2015,9 +2046,9 @@ Staging and applying changes is documented in info node
     (magit-git-wash #'magit-diff-wash-diffs args)))
 
 (defun magit-diff-wash-diffs (args &optional limit)
-  (when (--some (string-prefix-p "--color-moved" it) args)
-    (require 'ansi-color)
-    (ansi-color-apply-on-region (point-min) (point-max)))
+  (when (or magit-diff-preprocess-git-output-always
+         (--some (string-prefix-p "--color-moved" it) args))
+    (funcall magit-diff-preprocess-git-output-function))
   (when (member "--show-signature" args)
     (magit-diff-wash-signature))
   (when (member "--stat" args)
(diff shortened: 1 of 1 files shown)


lisp/magit-diff.el
```diff
@@ -2093,9 +2131,6 @@ section or a child thereof."
         (if (looking-at "^$") (forward-line) (insert "\n"))))))
 
 (defun magit-diff-wash-diff (args)
-  (when (cl-member-if (lambda (arg) (string-prefix-p "--color-moved" arg)) args)
-    (require 'ansi-color)
-    (ansi-color-apply-on-region (point-min) (point-max)))
```
dandavison, line 2098: I've moved the ANSI escape sequence processing to slightly earlier in the diff-washing sequence, so that it is called once only on the output received from git. Let me know if this needs further discussion.

lisp/magit-diff.el
```diff
@@ -243,6 +243,37 @@ t         Highlight only in added lines.
...
+
+The function must take two arguments: begin and end, which are
+buffer locations defining the region to operate on. It must
+remove the ANSI escape sequences from the buffer text and apply
+Emacs faces to the remaining text, creating the same visual
+effects as were implied by the ANSI escape sequences. The faces
+must be applied as overlays, not as text properties."
+  :group 'magit-diff
+  :type '(choice (function-item magit-diff-convert-ansi-escape-sequences-using-ansi-color)
+                 function))
```
dandavison, line 275: To switch Magit to use  `xterm-color` instead of  `ansi-color`, one can set this variable to a function like [`magit-delta-convert-ansi-escape-sequences-using-xterm-color`](https://github.com/dandavison/magit-delta/blob/master/magit-delta.el#L107).

lisp/magit-diff.el
tarsius: It should be done in a separate commit at least. Preferably the second.

dandavison:
> It should be done in a separate commit at least. Preferably the second.

Thanks @tarsius. I've altered this branch to have a more logical sequence of commits, and renamed the variables to recognize that they are performing arbitrary preprocessing of raw git output, including but not limited to interpreting and removing ANSI escape sequences:

1. 530794c6cff9ded5d089b7b27023b3b77b788c8f
  Moves the call to `ansi-color-apply-on-region` to be the first thing done in the diff washing sequence. It is called once, on the entire buffer.

2. 2b708107c92b942f3e7629065ae641c372e5a841
  Introduces new user variables `magit-diff-preprocess-git-output-function` and `magit-diff-preprocess-git-output-always`, setting the default value of `magit-diff-preprocess-git-output-function` to be a function that calls `ansi-color-apply-on-region`, as before.

For my current use case, I set `magit-diff-preprocess-git-output-function` to be a [function](https://github.com/dandavison/magit-delta/blob/master/magit-delta.el#L71) that calls delta on the raw git output to add color escape sequences, and then calls `xterm-color-colorize-buffer` to convert these to overlays.

Let me know if you'd prefer different names. The functionality is like a process filter for the synchronous git process. I'm not sure if a name based on "filter" rather than "preprocess" is clearer or not.

lisp/magit-diff.el
```diff
@@ -2015,6 +2015,9 @@ Staging and applying changes is documented in info node
     (magit-git-wash #'magit-diff-wash-diffs args)))
 
 (defun magit-diff-wash-diffs (args &optional limit)
+  (when (--some (string-prefix-p "--color-moved" it) args)
```
tarsius: Continue to use `cl-member-if`.

tarsius:
You make an effort to make this usable for others but I feel that in the process you are over-complicating things.

For now (and really just your package) I am just adding the `magit-diff-wash-diffs-hook`. I am intentionally not documenting or otherwise advertising it. This allows you to do your thing without Magit having to commit to some interface that might not work for others and/or that nobody else needs. (I think `magit-diff-preprocess-git-output-always` is a mistake, that should be done *inside* the hook function.)

A disadvantage is that if `--color-moved` is used, then `ansi-color-apply-on-region` is now called needlessly. That seems acceptable for now. Eventually I might move that onto `magit-diff-wash-diffs-hook`.

dandavison:
>  I feel that in the process you are over-complicating things.

Yes, I think you're right! Using the new hook works perfectly for my use case, thanks very much.

> A disadvantage is that if --color-moved is used, then ansi-color-apply-on-region is now called needlessly. 

Yes, I think I had got distracted by the attempt to avoid that. But it doesn't interfere with face properties added previously by a function on the new hook, so it's absolutely fine for me.

PR #4097, opened 2020-04-23
Add support for git reset --keep

jdelStrother:
I use `git reset --keep` a lot from the command line when I have some uncommitted changes that I realize ought to be applied to a different branch.  How about something like this in magit?

I'd welcome any suggestions that would make the wording clearer.


tarsius:
Maybe just `(HEAD and index, keep uncommitted)`, because it is shorter.

jdelStrother:
OK, updated.

tarsius:
Please also update the manual. (Just the org file. I can update the texi).

tarsius:
Thanks!

I've changed some metadata and whitespace.

Issue #4094, opened 2020-04-19
magit-am-apply-patches: Allow patches from anywhere in the filesystem


aurabindo:
**Use case**: Lots of patch files that needs to be applied but lying around somewhere outside of the git repository.

When `*.patch` is an entry in the `.gitignore`, copying the patches into the git does not make them show up in the magit status area as 'untracked' and hence cannot be conveniently selected from magit to initiate git am.

**Possible solution**: instead of taking one file as input, accept multiple selections

kyleam:
So I guess the main thing here is what the multi-selection interface looks like.  And, then once that's defined, whether it make sense from Magit's perspective to maintain this code in order to support this particular workflow.

One thing worth considering is that Dired seems like a natural interface to use for this, and it doesn't take much to wire that up outside of Magit:

```emacs-lisp
(defun my/dired-apply-patches (repo &optional arg)
  (interactive
   (list
    (and (not (magit-toplevel))
         (magit-read-repository t))
    ;; Or, if you use projectile:
    ;; (and (not (projectile-project-p))
    ;;      (completing-read "Project: "
    ;;                       (projectile-relevant-known-projects)))
    (and current-prefix-arg
         (prefix-numeric-value current-prefix-arg))))
  (let ((files (or (dired-get-marked-files nil arg)
                   (user-error "No marked files"))))
    (magit-status-setup-buffer repo)
    (magit-am-apply-patches files)))
```

tarsius:
I've added that with slight modifications.

PR #4099, opened 2020-04-25
Add magit-patch-apply-buffer

casouri:
This command applies a patch buffer, instead of a file. This is useful when you have opened a patch file (for review, etc) and don't want the hassle to find the file when applying. 

In this command I write the buffer's content to a temp file in `/tmp`. Is this ok to do? I.e., any potential trouble we might encounter?

### DIFF
--- a/lisp/magit-patch.el
+++ b/lisp/magit-patch.el
@@ -249,7 +249,8 @@ which creates patches for all commits that are reachable from
    ("-c" "Only apply to index" "--cached")
    ("-3" "Fall back on 3way merge" ("-3" "--3way"))]
   ["Actions"
-   ("a"  "Apply patch" magit-patch-apply)]
+   ("a"  "Apply patch" magit-patch-apply)
+   (6 "b"  "Apply patch in buffer" magit-patch-apply-buffer)]
   (interactive
    (if (not (eq current-transient-command 'magit-patch-apply))
        (list nil)
@@ -262,6 +263,17 @@ which creates patches for all commits that are reachable from
   (if (not file)
       (transient-setup 'magit-patch-apply)
     (magit-run-git "apply" args "--" (magit-convert-filename-for-git file))))
+
+;;;###autoload
+(defun magit-patch-apply-buffer (buffer &rest args)
+  "Apply the patch buffer BUFFER."
+  (interactive
+   (list (read-buffer "Apply patch in buffer: " nil t)
+         (transient-args 'magit-patch-apply)))
+  (with-temp-buffer
+    (insert-buffer-substring-no-properties buffer)
+    (magit-run-git-with-input "apply" args "-")
+    (magit-refresh)))
 
 ;;;###autoload
 (defun magit-patch-save (file &optional arg)


tarsius:
No temporary file is necessary, see `magit-apply-patch`.

casouri:
Thanks, I don't see any documentation on the section argument, what should one pass to it?

tarsius:
You should only get inspired by this function, you shouldn't actually use it. It is intended for buffers that contain magit sections.

casouri:
I modified the code. Now it doesn't use a temp file anymore.

lisp/magit-patch.el
```diff
@@ -263,6 +264,19 @@ which creates patches for all commits that are reachable from
...
+;;;###autoload (autoload 'magit-patch-apply-buffer "magit-patch" nil t)
+(define-transient-command magit-patch-apply-buffer (buffer &rest args)
+  "Apply the patch buffer BUFFER."
+  (interactive
+   (list (read-buffer "Apply patch: " nil t)
+         (transient-args 'magit-patch-apply)))
+  (with-temp-buffer
+    (insert-buffer-substring-no-properties buffer)
+    (magit-run-git-with-input
+     "apply" args)
```
tarsius, line 276: Use a single line and I think you need the `-` argument.

lisp/magit-patch.el
```diff
@@ -263,6 +264,19 @@ which creates patches for all commits that are reachable from
...
+    (unless inhibit-magit-refresh
```
tarsius, line 277: That's appropriate for `magit-apply-patch` because it is rather low-level. This command on the other hand should refresh unconditionally like most other commands.

lisp/magit-patch.el
```diff
@@ -263,6 +264,19 @@ which creates patches for all commits that are reachable from
       (transient-setup 'magit-patch-apply)
     (magit-run-git "apply" args "--" (magit-convert-filename-for-git file))))
 
+;;;###autoload (autoload 'magit-patch-apply-buffer "magit-patch" nil t)
+(define-transient-command magit-patch-apply-buffer (buffer &rest args)
+  "Apply the patch buffer BUFFER."
+  (interactive
+   (list (read-buffer "Apply patch: " nil t)
```
tarsius, line 271: Wouldn't it make more sense to just unconditionally apply the patch in the current buffer?

lisp/magit-patch.el
casouri: Done. BTW, what does the dash do? I know about "--" but not "-".

lisp/magit-patch.el
casouri: Done.

lisp/magit-patch.el
casouri: I'm not quite sure what do you mean. Maybe I'm not very familiar with the magit workflow you are referring to. If you are in a patch buffer, 1) how do I invoke the magit apply funtion? By M-x? 2) How does magit know which repo this patch should be applied to?

The workflow I have in mind is is a magit status buffer, hit w a b, select a buffer and the patch is applied to the current project.

tarsius: 1) Yes, for example. 2) Based on `default-directory` (same as in buffers with magit sections).

tarsius: > The workflow I have in mind is being in a magit status buffer, hit w a b, select a buffer and the patch is applied to the current project.

Seems more cumbersome than to do it from the buffer that contains the patch. Why do you want to go back to the status buffer first?

casouri: Because the patch buffer could be not in the project's directory. In my case, I sometimes open patch files in emails and they are located in somewhere like 
```
  /Users/yuan/Library/Containers/com.apple.mail/Data/Library/Mail Downloads/602427EA-BD71-458F-87EA-1E1E029FA967
```

I'm not really a magit power user, what would be the "magit way" of doing this? One idea is to do some thing like:

```
(if (and (in diff-mode) (in git repo)) 
    (directly apply current buffer)
  (prompt for buffer))
```

WDYT, is that the magit way?

tarsius: And I would expect that either `diff-mode` or `vc` already provide such a command. I cannot find it though. Please ask elsewhere whether such a thing already exists. I feel it would make more sense to add this elsewhere instead of here.

tarsius: > Because the patch buffer could be not in the project's directory.

Ah that makes sense.

tarsius: Okay we can add this.

* The prompt should mention that it is for a buffer.
* `(magit-run-git-with-input "apply" args "-")` should be on one line.
* Put this command on level 6 in the transient.

tarsius: Like so: `(6 "b"  "Apply patch in buffer" magit-patch-apply-buffer)`. Here's how you then have to activate it:

----

casouri: Okay, this is pretty cool :-) Please see the latest commit. BTW, should I add the condition check to directly apply the current buffer?

tarsius: Please also mention the command in the manual. You only have to edit the org file, I will regenerate the texi file.

tarsius:
Hm, #4094 is about more or less the same thing. I am still unsure whether we should add anything to Magit.

casouri:
Of course, it’s your decision what do do. In the mean time, I’ll just add custom command for myself (thanks to your awesome design).

kyleam:
> Hm, #4094 is about ~more or less the same thing~ something similar. I am still unsure whether we should add anything to Magit _in either case_.

FWIW my view is that new commands that deal with "I have patches somewhere outside a repo, I want to am or apply them" make sense to leave outside of Magit proper.  These involve bringing in various non-Magit bits that depend on the user's setup (Dired, projectile, mail client, and so on).  gh-4094 falls into this category.

The command in this PR seems less clear because it works from a Magit repo, and it feels pretty natural/convenient to say "well, I can apply a file, so why not a buffer?".  On the other hand, I'm not sure if a workflow that would use this is common enough that the added convenience is worth an extra command (or two, as noted at the end).  The motivating usecase seems to be a mailed patch, and I'd guess that most users of Emacs-based mail clients would find it more straightforward to invoke a command from the message buffer, in which case we're back into the category above.

So, I suppose I'm on the fence (with a slight preference for waiting to see if others chime in saying they'd find the command useful).  But if we do add this command, shouldn't `am` get a similar variant?

lisp/magit-patch.el
```diff
@@ -263,6 +264,17 @@ which creates patches for all commits that are reachable from
       (transient-setup 'magit-patch-apply)
     (magit-run-git "apply" args "--" (magit-convert-filename-for-git file))))
 
+;;;###autoload (autoload 'magit-patch-apply-buffer "magit-patch" nil t)
+(define-transient-command magit-patch-apply-buffer (buffer &rest args)
```
kyleam, line 268: Wouldn't a regular `defun` do here?

casouri: It could be, it's just more convenient for me.

casouri:
I agree. I'll wait for your instructions for any further action.

lisp/magit-patch.el
kyleam: If you replaced `define-transient-command` with `defun`, what would be less convenient?

casouri: `M-x magit-apply-patch-buffer` seems more keystrokes than `w a b`. But maybe it doesn't need to be a magit command to be invoked by `w a b`? I don't really know magit that well.

tarsius: It does not have to be defined with `define-transient-command`. Just try it out.

tarsius: I've just pushed this myself.

tarsius:
I've decided against adding this.

We need improvements in this area, but I don't think this goes far enough. (And I am also not sure what I would want to do instead of this.

Ps: You could add this to https://github.com/magit/magit/wiki/Tips-and-Tricks.

tarsius:
> FWIW my view is that new commands that deal with "I have patches somewhere outside a repo, I want to am or apply them" make sense to leave outside of Magit proper. These involve bringing in various non-Magit bits that depend on the user's setup (Dired, projectile, mail client, and so on). gh-4094 falls into this category.

While I generally agree, I think dired is the most standard interface people are going to use if there is no support for doing it directly in their mail client and other third-party packages. So that one (#4094) we can grant.

PR #4100, opened 2020-04-30
Improve the magit-{git,shell}-command-topdir completion

Vifon:
The `magit-shell-command-topdir` and `magit-gitcommand-topdir` completion is now done in the context of the repository top level, which allows completing scripts and other files residing there.  Previously the completion was done from the context of the current `default-directory` which wasn't necessarily the repository top level making the completion of local files nonsensical.

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -350,13 +350,14 @@ is run in the top-level directory of the current working tree."
   (magit-process-buffer))
 
 (defun magit-read-shell-command (&optional toplevel initial-input)
-  (let ((dir (abbreviate-file-name
-              (if (or toplevel current-prefix-arg)
-                  (or (magit-toplevel)
-                      (magit--not-inside-repository-error))
-                default-directory))))
+  (let ((default-directory
+          (if (or toplevel current-prefix-arg)
+              (or (magit-toplevel)
+                  (magit--not-inside-repository-error))
+            default-directory)))
     (read-shell-command (if magit-shell-command-verbose-prompt
-                            (format "Async shell command in %s: " dir)
+                            (format "Async shell command in %s: "
+                                    (abbreviate-file-name default-directory))
                           "Async shell command: ")
                         initial-input 'magit-git-command-history)))
 


lisp/magit.el
```diff
@@ -319,7 +319,9 @@ used as initial input, but can be deleted to run another command.
 
 COMMAND is run in the top-level directory of the current
 working tree."
-  (interactive (list (magit-read-shell-command t "git ")))
+  (interactive (list
+                (let ((default-directory (magit-toplevel)))
+                  (magit-read-shell-command t "git "))))
```
kyleam, line 324: I think the better spot to do this let-binging is `magit-read-shell-command`.  Based on its first argument, that function knows whether it's reading a command for the top-level directory.

Vifon: Done, I've moved it there.

Vifon:
What's the policy regarding history editing over here? Should I squash both commits as the second one essentially reverts the first one?

Vifon:
I did it either way, it just felt dirty to keep it this way.

kyleam:
> What's the policy regarding history editing over here? Should I squash both commits as the second one essentially reverts the first one?

Rewriting/polishing series is encouraged.

lisp/magit.el
```diff
@@ -350,13 +350,14 @@ is run in the top-level directory of the current working tree."
...
-                default-directory))))
+  (let ((default-directory
+          (if (or toplevel current-prefix-arg)
+              (or (magit-toplevel)
+                  (magit--not-inside-repository-error))
+            default-directory)))
     (read-shell-command (if magit-shell-command-verbose-prompt
-                            (format "Async shell command in %s: " dir)
+                            (format "Async shell command in %s: "
+                                    (abbreviate-file-name default-directory))
```
kyleam, line 360: I've pushed a change that moves the `abbreviate-file-name` call down here, as it's only needed for the prompt.  Even if it setting `default-directory` to something with `~/` in it works, I'd rather avoid it.

Vifon: Good call, thanks!

PR #4103, opened 2020-05-02
Scroll down with `S-SPC`

zaeph:
Allow scrolling down with `S-SPC`.

`DEL` already does that, but I often find myself pressing `S-SPC` to do it, as I would in Firefox or elsewhere.

I wanted to mention it in the doc, but I'm not sure how to best proceed when there are _multiple keys for one action_.  Here are the two spots we'd need to modify:
https://github.com/magit/magit/blob/b808d5891492c4e50dc1312cc2bccd33794d1996/Documentation/magit.org#L2620
https://github.com/magit/magit/blob/b808d5891492c4e50dc1312cc2bccd33794d1996/Documentation/magit.org#L3893

-----

I think `scroll-down` and `scroll-up` might have been mixed up:
https://github.com/magit/magit/blob/b808d5891492c4e50dc1312cc2bccd33794d1996/lisp/magit-mode.el#L349-L351
Shouldn't this be the reverse?


tarsius:
Thanks for taking care of the alignment but I think in this case I would rather not do that.

(Maybe I will do it myself at a later time in a commit that does nothing else, once I have added support for `git blame --ignore-revs-file`.)

zaeph:
> Thanks for taking care of the alignment but I think in this case I would rather not do that.
> 
> (Maybe I will do it myself at a later time in a commit that does nothing else, once I have added support for `git blame --ignore-revs-file`.)

Addressed in 526c79716f3b8b76a40488ab68712b40425679d9 & 53088a38d448f14102792ae8e31834da67cb48a7

zaeph:
> Thanks for taking care of the alignment but I think in this case **I would rather not do that**.  [emphasis added]

Just to be clear, you were only talking about the alignment, not the extra key-binding?

tarsius:
Yes.

Thanks!

PR #4104, opened 2020-05-03
manual: Fix default value of magit-branch-read-upstream-first

cpitclaudel:
The default value is `t`, but the manual claimed that it was `nil`.


tarsius:
@kyleam You were about a minute faster than me. :grinning:

kyleam:
Thanks for catching that.  Looks like the code/doc mismatch happened way back in d4992d0e (magit-branch-read-upstream-first: default to t, 2015-07-15).

Issue #4106, opened 2020-05-04
"Cannot determine Magit's version" message on startup


chasecaleb:
I just updated magit from `68b5a13fa1b6b122d4a2547b8c415aadff83d8ca` to the current head of master (`9a388a61d95761b9c748dc186602ba7c691d580f`) and now see this in `*Messages*` on startup:
> Cannot determine Magit’s version (error "/home/caleb/.emacs.d/lib/magit/lisp/magit.el" repo static elpa dirname)

I didn't dig very deeply into what's going on, but the value of `magit-version` variable is `'error'`. I'm using borg (which I love by the way, so thanks!) and use-package for my config. The magit-related portion is:

```emacs-lisp
(use-package magit
  :config
  (setq magit-diff-refine-hunk 'all)
  ;; keep-foreground is a bit hard on the eyes (with spacemarks-dark).
  (setq magit-diff-unmarked-lines-keep-foreground nil)
  (setq magit-branch-prefer-remote-upstream '("master"))
  (setq magit-branch-adjust-remote-upstream-alist '(("origin/master" "master")))
  (setq magit-module-sections-nested nil)
  (setq magit-clone-set-remote.pushDefault t))

(use-package magit-mode
  :config
  (setq magit-display-buffer-function #'magit-display-buffer-same-window-except-diff-v1))

(use-package magit-ediff
  :config
  (setq magit-ediff-dwim-show-on-hunks t))

(use-package magit-wip
  :diminish
  :config
  (magit-wip-mode))
```

However I didn't touch my config between updates (and I made sure to rerun `borg build`), so I think it is a code regression and not config related.

tarsius:
Please either step through `magit-version` or add some debug statements to figure out what is going on. When using `borg`, then the version is supposed to come from `(magit-git-string "describe" "--tags" "--dirty")`. Try to figure out why that isn't called by inspecting the tests on the way there.

chasecaleb:
Alright, I investigated closer and I'll close this issue since it looks like it's working as intended - my setup is just an extreme edge case. Sorry about that and thanks for pointing me in the right direction.

[...]

PR #4113, opened 2020-05-13
Fix regexp matching version from magit directory

janEbert:
I am having the same problem as #4106 (my .emacs.d is also symlinked).
I believe `(magit-version)` fails due to the regexp for matching the version from the magit directory not being up-to-date. The regexp currently tries matching an 8-digit number, followed by a period, followed by any number of digits. I changed it to match a number, followed by more than one number or period, followed by a number.
If desired, I can make the regexp match the SemVer-scheme which it currently does not.

Due to symlinks being an issue in both regards, I initially suspected the usage of `file-chase-links` instead of `file-truename` to be the culprit. However, changing the method did not fix the problem.
Finally, I wonder if there is a better way to test changes than modifying `.emacs.d/elpa/magit-*/magit.el`. Thanks!

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -445,8 +445,7 @@ and Emacs to it."
               (push 'dirname debug)
               (let ((dirname (file-name-nondirectory
                               (directory-file-name topdir))))
-                (when (string-match "\\`magit-\\([0-9]\\{8\\}\\.[0-9]*\\)"
-                                    dirname)
+                (when (string-match "\\`magit-\\([0-9].*\\)" dirname)
                   (setq magit-version (match-string 1 dirname)))))
             ;; If all else fails, just report the commit hash. It's
             ;; better than nothing and we cannot do better in the case


kyleam:
> The regexp currently tries matching an 8-digit number, followed by a period, followed by any number of digits. I changed it to match a number, followed by more than one number or period, followed by a number.

It looks to me like that regexp should be that specific: 4c19758c (magit-version: handle emacs -Q inside magit-YYYYMMDD.N case, 2016-09-20).  If I understand correctly, that case is meant for catching when `emacs -Q` is being called from within the elpa/magit-YYYYMMDD.N directory for debugging.  But for just running as usual and using Magit from MELPA, I think the `(when (featurep 'package) ...)` condition is the one that should be detecting the version. 

What does the value matched by your change look like?

tarsius:
@janEbert Friendly ping!

janEbert:
I'll first lay out what went wrong on my side and – after the horizontal line – what my intention with the PR was.

Thanks @kyleam for pointing me in the right direction. I never wondered why `package-alist` was `nil` (causing `(assoc 'magit package-alist)` to also return `nil`); turns out emacs-27.0.91 does _not_ call `package-initialize` as mentioned in its NEWS file. I replaced the call in init.el with
```emacs-lisp
(when (< emacs-major-version 27)
  (package-initialize))
```
as suggested in the NEWS file, never noticing any problems. As `package-initialize` was never called, `package-alist` was also not set and Magit's version check did not work.

---
I thought the regexp was supposed to check for the Magit directory that is installed in `package-user-dir`, named "magit-2.90.1" in my case. I never saw a directory with the "YYYYMMDD.N" versioning scheme you mentioned, only the SemVer one. Therefore I changed the regexp to match both SemVer as well as the old (?) "YYYYMMDD.N" versioning scheme. My new regexp matches any version that is at least two digits long with any amount of digits and/or periods in between.

As the issue was on my part, this can be closed if desired; if you have use for the more flexible regexp, I can patch whatever you dislike.

kyleam:
> As `package-initialize` was never called, `package-alist` was also not set and Magit's version check did not work.

Hmm, I'm not a package.el user, so I may be getting turned around, but shouldn't `package-initialize` still be called (and `package-alist` still populated)?  I thought the Emacs 27 change was just that the user didn't need to explicitly add it in init.el.  Quickly testing by adding MELPA and installing Magit [*], I see `package-alist` populated at startup (using an Emacs built from origin/emacs-27, 7d323f07c0).

[*] I also moved my real init.el out of the way so that I could run without customization, because using `-q` and friends disables loading packages at startup.

> I thought the regexp was supposed to check for the Magit directory that is installed in `package-user-dir`, named "magit-2.90.1" in my case. I never saw a directory with the "YYYYMMDD.N" versioning scheme you mentioned

That left me scratching my head for a bit, but I think I know what's going on.  Are you using melpa-stable?  The "YYYYMMDD.N" scheme is used for regular melpa.  If others find it useful, I don't see anything problematic about extending that regexp to handle the melpa-stable form.

basil-conto:
Here are the relevant announcements from `etc/NEWS` in Emacs 27:

```
** Emacs can now be configured using an early init file.
The file is called "early-init.el", in 'user-emacs-directory'.  It is
loaded very early in the startup process: before graphical elements
such as the tool bar are initialized, and before the package manager
is initialized.  The primary purpose is to allow customizing how the
package system is initialized given that initialization now happens
before loading the regular init file (see below).

We recommend against putting any customizations in this file that
don't need to be set up before initializing installed add-on packages,
because the early init file is read too early into the startup
process, and some important parts of the Emacs session, such as
'window-system' and other GUI features, are not yet set up, which could
make some customization fail to work.

** Installed packages are now activated *before* loading the init file.
As a result of this change, it is no longer necessary to call
'package-initialize' in your init file.

Previously, a call to 'package-initialize' was automatically inserted
into the init file when Emacs was started.  This call can now safely
be removed.  Alternatively, if you want to ensure that your init file
is still compatible with earlier versions of Emacs, change it to:

(when (< emacs-major-version 27)
  (package-initialize))

However, if your init file changes the values of 'package-load-list'
or 'package-user-dir', or sets 'package-enable-at-startup' to nil then
it won't work right without some adjustment:
- You can move that code to the early init file (see above), so those
  settings apply before Emacs tries to activate the packages.
- You can use the new 'package-quickstart' so activation of packages
  does not need to pay attention to 'package-load-list' or
  'package-user-dir' any more.
```

Note that it does not say that Emacs 27 calls `package-initialize`, because what is actually called is `package-activate-all`, but the effect should be similar. Here's the relevant excerpt from `lisp/startup.el`:

```el
;; If any package directory exists, initialize the package system.
(and user-init-file
     package-enable-at-startup
     (not (bound-and-true-p package--activated))
     (catch 'package-dir-found
       (let (dirs)
         (if (boundp 'package-directory-list)
             (setq dirs package-directory-list)
           (dolist (f load-path)
             (and (stringp f)
                  (equal (file-name-nondirectory f) "site-lisp")
                  (push (expand-file-name "elpa" f) dirs))))
         (push (if (boundp 'package-user-dir)
                   package-user-dir
                 (locate-user-emacs-file "elpa"))
               dirs)
         (dolist (dir dirs)
           (when (file-directory-p dir)
             (dolist (subdir (directory-files dir))
               (when (let ((subdir (expand-file-name subdir dir)))
                       (and (file-directory-p subdir)
                            (file-exists-p
                             (expand-file-name
                              (package--description-file subdir)
                              subdir))))
                 (throw 'package-dir-found t)))))))
     (package-activate-all))
```

If your package manager isn't being initialised properly, it's due to either a misconfiguration on your end or a bug, which would be nice to get reported.

tarsius:
I went with an even more forgiving regexp and added a more verbose commit message.

janEbert:
@tarsius: Thanks for improving the commit!

@kyleam: Thanks for trying to reproduce. My original build was based on the emacs-27.0.91 tag. I just rebuilt on the emacs-27 branch (also commit 7d323f07c0) and had the same problem.

@basil-conto: Thanks for clarifying which function is called. Guess it's time to try and  debug this. At which point would you suggest filing the bug when I can't find out what is wrong myself?

janEbert:
I once again believe the problem is related to my symlinked .emacs.d. However, I am not able to print-debug startup.el; the messages are simply not shown.
I've tried multiple ways to display text (below) but when I enter the \*Messages\* buffer, none of my messages is shown. I did not forget byte-recompiling the file.

Could you help me with that? We can also continue over mail to not pollute this issue further.

```emacs-lisp
;; in `defun command-line`
(let ((msg "trying to activate packages"))
  (error msg)
  (display-warning 'initialization msg)
  (message msg)
  (princ msg)
  (terpri))
```

basil-conto:
Sure, feel free to email me or <kbd>M-x</kbd>`report-emacs-bug` including as much information as you can provide.

PR #4116, opened 2020-05-17
Make magit-version work properly in shallow clone

raxod502:
See https://github.com/raxod502/straight.el/issues/520. Previously, when you made a shallow clone of Magit, then byte-compiling Magit would produce a warning `Cannot determine Magit's version`. This caused problems for users of the package manager `straight.el`, for example (see the linked issue). With the new code, Magit will not print this warning, and when `magit-version` is invoked, the commit hash will be reported. Behavior is only changed in the case that no tags are available in the repository, and none of the other version checks succeed.


raxod502:
This is a proof of concept. Is it satisfactory to you @tarsius, or would you like for the check to be more granular?

tarsius:
Thanks!

PR #4135, opened 2020-06-01
magit-branch-checkout: Ignore submodule modifications

andschwa:
Submodules can be completely up-to-date, yet still show as modified
because of a dirty worktree. Even if they weren't up-to-date, switching
branches with Git on the command-line ignores submodules, so we should
too. Otherwise this prevents the user from switching branches unless
they completely update and clean every submodule, which is unfriendly.


tarsius:
Thanks!

PR #4140, opened 2020-06-10
Fix edebug spec for ‘magit-insert-section’.

phst:
As described in the docstring, the form for ‘eval’ is ‘(eval FORM)', and FORM
doesn’t need to be a symbol.

Without this, functions like ‘magit-insert-recent-commits' can’t be edebugged
if ‘magit-insert-section' is edebugged.


tarsius:
Thanks!

PR #4147, opened 2020-06-16
magit-diff-wash-diff: add cases for both added and removed in local

kidd:
using [m]erge [p]review, it seems not all possibilities for conflicts are supported. I just added the 2 I found so that they don't break hard, but I don't know exactly how should the 'proper' fix be.

Even the labels like "l.deleted" get easily missaligned. So this might not be the path to fix them, but at least is a quick pin at the relevant code area.

Cheers!


kidd:
thanks!

tarsius:
I had to revert that but will give it another try later.

Issue #4157, opened 2020-06-27
magit-remote-add fails with error if no "origin"


bennorth:
Starting with a freshly-init'd repo, attempting to interactively add a remote with `magit-remote-add` fails with an error just after providing the new remote's name, and before prompting for the new remote's URL.  More generally, a repo without a remote called `origin` causes the same failure from `magit-remote-add`.

Expected behaviour: I am able to enter the information needed to add a remote.

Observed behaviour: Error with the following backtrace:
```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  string-match("\\([^:/]+\\)/[^/]+\\(\\.git\\)?\\'" nil)
  (and (string-match "\\([^:/]+\\)/[^/]+\\(\\.git\\)?\\'" origin) (replace-match remote t t origin 1))
  (magit-read-url "Remote url" (and (string-match "\\([^:/]+\\)/[^/]+\\(\\.git\\)?\\'" origin) (replace-match remote t t origin 1)))
  (list remote (magit-read-url "Remote url" (and (string-match "\\([^:/]+\\)/[^/]+\\(\\.git\\)?\\'" origin) (replace-match remote t t origin 1))) (transient-args (quote magit-remote)))
  (let ((origin (magit-get "remote.origin.url")) (remote (magit-read-string-ns "Remote name"))) (list remote (magit-read-url "Remote url" (and (string-match "\\([^:/]+\\)/[^/]+\\(\\.git\\)?\\'" origin) (replace-match remote t t origin 1))) (transient-args (quote magit-remote))))
  call-interactively(magit-remote-add nil nil)
  command-execute(magit-remote-add)
```

How to reproduce:
- create empty temporary directory
- in that directory, run `git init .` from command line
- change to `magit` source dir, run `make emacs-Q`
- in the Emacs which appears, `find-file` the same temporary directory
- `M-x magit status`
- attempt `magit-remote-add` with `M`, `a`, then type a dummy remote name, say `github`, press return; get above error.

Magit version:`Magit v2.90.1-1018-g33052629, Git 2.20.1, Emacs 26.3, gnu/linux`

I have a small PR which seems to fix this problem for me; will add link in a comment shortly.

bennorth:
PR in #4158.

bennorth:
@tarsius: Thanks for merging the PR; closing.

lljr:
I'm still getting this error in Guix's Magit (v 2.90.1-6.7f486d4)...

[...]

PR #4158, opened 2020-06-27
magit-remote-add: Handle lack of "origin" remote

bennorth:
It is possible for a repo to have no remote called "origin".  In this case, magit-remote-add fails when attempting to construct a suggestion for the to-be-added remote's URL.

Check that we have an "origin" remote URL before using it.


bennorth:
Addresses #4157.

tarsius:
Thanks!

PR #4162, opened 2020-07-01
Rename magit-insert-submodules in doc to magit-insert-modules.

treese:
It appears that when the function `magit-insert-submodules` was renamed to `magit-insert-modules`, the documentation was not updated.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -6613,7 +6613,7 @@ Also see [[man:git-submodule]]
 The command ~magit-list-submodules~ displays a list of the current
 repository's submodules in a separate buffer.  It's also possible to
 display information about submodules directly in the status buffer of
-the super-repository by adding ~magit-insert-submodules~ to the hook
+the super-repository by adding ~magit-insert-modules~ to the hook
 ~magit-status-sections-hook~ as described in [[*Status Module Sections]].
 
 - Command: magit-list-submodules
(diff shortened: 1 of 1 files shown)


Documentation/magit.org
```diff
@@ -6613,7 +6613,7 @@ Also see [[man:git-submodule]]
 The command ~magit-list-submodules~ displays a list of the current
 repository's submodules in a separate buffer.  It's also possible to
 display information about submodules directly in the status buffer of
-the super-repository by adding ~magit-insert-submodules~ to the hook
+the super-repository by adding ~magit-insert-modules~ to the hook
```
kyleam, line 6616: AFAICS `magit-insert-modules-overview` is actually what came of `magit-insert-submodules` (83104db0), but, with the introduction of `magit-insert-modules` (b92454b7, a couple of commits before in the same series), it seems like that'd be the one to recommend here.

So, this looks like a good change.

Documentation/magit.org
```diff
@@ -6637,7 +6637,7 @@ the super-repository by adding ~magit-insert-submodules~ to the hook
   has to return a string to be inserted or nil.  PROPS is an alist
   that supports the keys ~:right-align~ and ~:pad-right~.
 
-- Function: magit-insert-submodules
+- Function: magit-insert-modules
```
kyleam, line 6640: As mentioned above, I think this actually corresponds to `magit-insert-modules-overview`.  But that (and `magit-insert-modules`) are already included under "Status Module Sections", so I'd say this entry should just be deleted.

treese:
Thanks for finishing it up, Kyle. I got hung up trying to understand how to properly amend the pull request, and didn't quite to it. And thanks for all your work on the project!

kyleam:
> I got hung up trying to understand how to properly amend the pull request, and didn't quite to it.

Apologies for not checking if you were interested in following up but stalled.  Please don't let the workflow snag prevent you from making other contributions.  Happy to guide you through any issues there.

Issue #4160, opened 2020-06-30
Stashes are not saved when using envrc


glasserc:
I recently started work on a project and I've noticed that pressing `z z` to save a stash is generating "empty" stashes that nevertheless erase my changes. It seems to be an interaction with [envrc](https://github.com/purcell/envrc) although I haven't tracked it down exactly.

Steps to reproduce:

- Add a `.envrc` file to a project. The file can be empty.
- I also add `.envrc` to `.git/config/exclude` so that it won't be considered part of a stash.
- Turn on `envrc-global-mode` and `direnv allow` the file using `envrc-allow`.
- Change a file. I've been adding `Test` to the top of the project's README.
- `magit-status`
- `z z RET`

Expected behavior:

The created stash should have contents.

Actual behavior:

The stash appears empty. Applying it or popping it does not restore the modifications.

![Screenshot from 2020-06-30 10-23-32](https://user-images.githubusercontent.com/17006/86137920-cbdb0280-babb-11ea-95d1-adba0c3f52a3.png)

Other observations:

- A stash is created.
- Doing `git stash` from the command line works OK.
- In `magit-process` for the repository, I see `  0 git … update-ref --create-reflog -m On\ master\:\ 5dcbb9d8\ Need\ to\ set\ up\ autocommit refs/stash b93d90cbf849168ed30d505b77b8830c4e22ff65 ''`.
- Because no error condition is hit, no backtrace is available.
- `magit-version` says `Magit 20200627.806, Git 2.25.1, Emacs 26.3, gnu/linux`
- `magit-debug-git-executable` looks normal (`git` from `/usr/bin/git`, `2.25.1`).
- Everything works correctly if `envrc-global-mode` is off (either never turned on, or turned on and then turned off).
- Everything works correctly if `envrc-global-mode` is on but there is no `.envrc` file in any parent directory.
- Everything works correctly if `envrc-global-mode` is on but the `.envrc` file has not been allowed, or even if an allowed file is denied using `envrc-deny`.

tarsius:
> It seems to be an interaction with envrc although I haven't tracked it down exactly.

Does envrc do anything related to the location of the git datadir and workdir?

glasserc:
I don't *think* so. To the best of my knowledge, it should only be updating the environment, mainly by copying environment variables in. However, I copied `process-environment` before/after and the only variables `envrc` adds are `DIRENV_DIFF`, `DIRENV_DIR`, and `DIRENV_WATCHES`.

[...]

PR #4169, opened 2020-07-15
magit-git: always use caller's process-environment when invoking git

purcell:
In certain circumstances, the calling buffer's process-environment may
be lost when using `with-temp-buffer`, so we explicitly prevent that
from happening in order to ensure that `git` will be invoked as
expected.

See #4160 for more information.


manuel-uberti:
@tarsius friendly ping. :)

tarsius:
Thanks!

Merged after renaming it as a private macro.

purcell:
> Merged after renaming it as a private macro.

Makes sense.

And yay, thanks!

PR #4173, opened 2020-07-25
Add project.el integration

spwhitton:
Add some integration with project.el.  The `C-x p` map is the main way I navigate within and between projects nowadays, and so this is very nice to have.  It's modelled on the vc-dir integration which comes with project.el.  I also thought about somehow replacing vc-dir for git projects, but I think in the end it is more useful to have a separate key.

### DIFF
--- a/lisp/magit-extras.el
+++ b/lisp/magit-extras.el
@@ -30,6 +30,8 @@
 (require 'magit)
 
 (declare-function dired-read-shell-command "dired-aux" (prompt arg files))
+;; For `magit-project-status'.
+(declare-function project-root "project" (project))
 
 (defvar ido-exit)
 (defvar ido-fallback)
@@ -130,6 +132,20 @@ like pretty much every other keymap:
   (setq ido-fallback 'magit-status)                ; for Emacs >= 26.2
   (with-no-warnings (setq fallback 'magit-status)) ; for Emacs 25
   (exit-minibuffer))
+
+;;;###autoload
+(defun magit-project-status ()
+  "Run `magit-status' in the current project's root."
+  (interactive)
+  (magit-status-setup-buffer (project-root (project-current t))))
+
+(with-eval-after-load 'project
+  ;; Only more recent versions of project.el have `project-prefix-map' and
+  ;; `project-switch-commands', though project.el is available in Emacs 25.
+  (when (boundp 'project-prefix-map)
+    (define-key project-prefix-map "m" #'magit-project-status))
+  (when (boundp 'project-switch-commands)
+    (add-to-list 'project-switch-commands '(?m "Magit" magit-status))))
 
 ;;;###autoload
 (defun magit-dired-jump (&optional other-window)


spwhitton:
Hmm, the CI failed, but I'm not sure how best to respond.  project.el is included in Emacs 27, but in the meantime could the build obtain it from GNU Elpa?  I would hope that this did not have to wait on Emacs 27 being released.  Thanks!

tarsius:
It fails on Emacs 27 with the same reason:

> magit-project.el:31:1:Error: Symbol’s value as variable is void: project-prefix-map

but that probably just means that the build we use is too old. @npostavs could you update please?

tarsius:
IMO it is overkill to add a separate library for this. Instead the command should probably be place next to `ido-enter-magit-status`, which serves a similar purpose.

basil-conto:
> project.el is included in Emacs 27

It's been around since Emacs 25, but is only now seeing more active development and integration.

> but in the meantime could the build obtain it from GNU Elpa?

I'm not familiar with Magit's CI, but in theory, yes.

> I would hope that this did not have to wait on Emacs 27 being released.

> but that probably just means that the build we use is too old

Note that `project-prefix-map` and co. are included with what will be Emacs 28, not 27. If you want those new `project.el` features before then, you have to pull them from GNU ELPA.

lisp/magit-project.el
```diff
@@ -0,0 +1,43 @@
...
+
+;; Integration with project.el commands.
+
+;;; Code:
+
+(require 'magit)
+(require 'project)
+
+(defun magit-project-status ()
+  "Run `magit-status' in the current project's root"
```
basil-conto, line 34: Nit: missing full stop.

lisp/magit-project.el
```diff
@@ -0,0 +1,43 @@
...
+  (interactive)
+  (magit-status (project-root (project-current t))))
+
+(define-key project-prefix-map "m" #'magit-project-status)
+
+(add-to-list 'project-switch-commands '(?m "Magit" magit-project-status))
```
basil-conto, line 40: Doesn't `project-switch-project` already bind `default-directory` appropriately, so that you can directly place `magit-status` on `project-switch-commands`?

spwhitton:
Thank you both for the feedback.  I think there are two options: either I add a dependency on a newer version of project.el from GNU Elpa, or I wrap the code in `(when (fboundp 'project-prefix-map))` and similarly for `project-switch-project-commands`.  Which do you prefer?

lisp/magit-project.el
tarsius: @spwhitton DId you look into this? If this is the case, then (given the other complications) the best approach might be enough to just add this to documentation.

```elisp
(define-key project-prefix-map "m" #'magit-status)
(add-to-list 'project-switch-commands '(?m "Magit" magit-status))
```

Or not, but it seems important to answer this question before we proceed.

spwhitton: @tarsius The project-switch-commands suggestion will work indeed, but not for project-prefix-map -- that still needs a custom function.

spwhitton:
@tarsius just to confirm, do you mean the fboundp wrapper, or the magit-insinuate-project function?  Thanks!

tarsius:
Uhm... "the middle"... aka the boundp wrapper. :grin:

spwhitton:
@tarsius okay, pushed that.

tarsius:
We probably shouldn't actually require `project`. (Or is that always loaded anyway?)

Even in comments, please begin sentences with capital letters and end them with a period.

spwhitton:
It's not always loaded.  So shall I wrap the code in an `(eval-after-load 'project` then?

Sorry about the comments; didn't think of that.

tarsius:
Please also deal with these two warnings:

```
In magit-project-status:
magit-extras.el:137:4: Warning: ‘magit-status’ is for interactive use only;
    use ‘magit-status-setup-buffer’ instead.

In end of data:
magit-extras.el:697:1: Warning: the function ‘project-root’ is not known to be
    defined.
```

Use `declare-function` to deal with the latter. Look at some existing uses of that function in our codebase first.

spwhitton:
Thanks for the feedback.  Think I've solved those warnings.

lisp/magit-extras.el
```diff
@@ -30,6 +30,8 @@
 (require 'magit)
 
 (declare-function dired-read-shell-command "dired-aux" (prompt arg files))
+;; For `magit-project-status'
```
tarsius, line 33: Here too, end with a period.

tarsius:
> Think I've solved those warnings.

Use `make all test` to verify:

```
In magit-project-status:
magit-extras.el:140:4: Warning: ‘magit-status-internal’ is an obsolete
    function (as of Magit 3.0.0); use ‘magit-status-setup-buffer’ instead.
```

lisp/magit-extras.el
spwhitton: Will do so, but I was trying to follow other uses of declare-function in magit where there is no period.

spwhitton:
I was foolishly bytecompiling against older magit, not git master, so I didn't see the magit-status-setup-buffer error.  Sorry about that.  Should be fixed now.

tarsius:
Thanks!

spwhitton:
Thank you for your patience!

-- 
Sean Whitton

basil-conto:
The variable `project-switch-commands` was recently promoted to a user option and its format changed: https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=51698f77dd6356049fcacdb01ebe80cfe4c67272

There's no deprecation warning yet, but there may be once the dust settles.

tarsius:
Thanks for the heads up. I probably won't investigate further until the byte-compiler starts whining.

PR #4180, opened 2020-08-03
add new function: magit-find-file-in-other-worktree.

yangyingchao:
This function can be used to find and open file in other worktree quickly.

### DIFF
--- a/lisp/magit-worktree.el
+++ b/lisp/magit-worktree.el
@@ -179,6 +179,23 @@ If there is only one worktree, then insert nothing."
               (insert ?\n))))
         (insert ?\n)))))
 
+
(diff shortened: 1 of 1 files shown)


tarsius:
Let's not add this to Magit.  I am sure it is useful sometimes, but if we add this, then it would make sense to add many more `*-in-other-worktree` variants of existing commands. We have to draw the line somewhere and I think this command lands on the wrong side.

So you'll have to put this in your init file.  You could also add it to https://github.com/magit/magit/wiki/Tips-and-Tricks if you think others will find this useful.

I recommend you move all the code that reads user input inside the `interactive` form.

yangyingchao:
ok for me.

Issue #4185, opened 2020-08-07
Allow helm users to use marked candidates in completing-read when needed


thierryvolpiatto:
Hi,
I have added a new variable to helm that allows external packages to allow their completing-read taking advantage of marks, for example when using "z k" for stashs, you can mark stashes and delete all marked at once.
If user use something else than helm for its completion it will have no effect and the completing-read will return a single element instead of a list. The only thing missing in magit is that `magit-completing-read` returns an error when result is not a string:

```diff
diff --git a/lisp/magit-utils.el b/lisp/magit-utils.el
index 464dbccb..63088774 100644
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -477,7 +477,7 @@ acts similarly to `completing-read', except for the following:
                           predicate
                           require-match initial-input hist def)))
       (setq this-command command)
-      (if (string= reply "")
+      (if (or (null reply) (equal reply ""))
           (if require-match
               (user-error "Nothing selected")
             nil)
```
Once this fixed, you can bind the new variable `helm-comp-read-use-marked` where needed, I see that in some places you already allow a list because you allow user to mark a region (stashes for example) so there is no much changes to make.

tarsius:
> I have added a new variable to helm that allows external packages to allow their completing-read taking advantage of marks, for example when using "z k" for stashs, you can mark stashes and delete all marked at once.

How does that work in a generic fashion? Or do callers somehow have to opt in?

> If user use something else than helm for its completion it will have no effect and the completing-read will return a single element instead of a list.

What does "it" refer to? Can't be a helm feature because that *obviously* has no effect if helm isn't being used. I think you are referring to the change you are describing next, but the order is confusing, so I am asking just to make sure.

> The only thing missing in magit is that `magit-completing-read` returns an error when result is not a string:

I think we can make that change in Magit but why does that become relevant only now? Do you continue to return the empty string for "no completion" and when do you return nil instead?

thierryvolpiatto:
Jonas Bernoulli <notifications@github.com> writes:

>     I have added a new variable to helm that allows external packages to allow their completing-read taking advantage of marks, for example when using "z k" for stashs,
>     you can mark stashes and delete all marked at once.
>    
> How does that work in a generic fashion? Or do callers somehow have to opt in?

    ;; With helm-mode
    (helm-mode 1)

    ;; Mark a, b, c and press RET
    (let ((helm-comp-read-use-marked t))
[...]

PR #4187, opened 2020-08-10
Allow marking stashes to delete in helm stashes completion (#4185)

thierryvolpiatto:
Fixes issue #4185. I only bound the variable for stashes deletion.
Please review before merging, I don't know how to use `--if-let`, maybe it can be used to bind the variable instead of adding an extra let, dunno.  This is a proof of concept, maybe other people will find it useful and use it elsewhere in Magit.
Thanks.

### DIFF
--- a/lisp/magit-stash.el
+++ b/lisp/magit-stash.el
@@ -224,6 +224,8 @@ and forgo removing the stash."
       (magit-stash-drop stash)
     (magit-run-git "stash" "apply" stash)))
 
+(defvar helm-comp-read-use-marked)
+
 ;;;###autoload
 (defun magit-stash-drop (stash)
   "Remove a stash from the stash list.
--- a/lisp/magit-stash.el
+++ b/lisp/magit-stash.el
@@ -231,7 +231,8 @@ When the region is active offer to drop all contained stashes."
   (interactive
    (list (--if-let (magit-region-values 'stash)
              (magit-confirm 'drop-stashes nil "Drop %i stashes" nil it)
-           (magit-read-stash "Drop stash"))))
+           (let ((helm-comp-read-use-marked t))
+             (magit-read-stash "Drop stash")))))
   (dolist (stash (if (listp stash)
                      (nreverse (prog1 stash (setq stash (car stash))))
                    (list stash)))
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -477,7 +477,7 @@ acts similarly to `completing-read', except for the following:
                           predicate
                           require-match initial-input hist def)))
       (setq this-command command)
-      (if (string= reply "")
+      (if (equal reply "")
           (if require-match
               (user-error "Nothing selected")
             nil)


lisp/magit-stash.el
```diff
@@ -231,7 +231,8 @@ When the region is active offer to drop all contained stashes."
   (interactive
    (list (--if-let (magit-region-values 'stash)
              (magit-confirm 'drop-stashes nil "Drop %i stashes" nil it)
-           (magit-read-stash "Drop stash"))))
+           (let ((helm-comp-read-use-marked t))
```
tarsius, line 234: You'll have to declare that function, so that those who don't use helm don't get a warning about this variable being undefined.

basil-conto: More importantly, so that it's marked as special (dynamically bound).

thierryvolpiatto:

> @tarsius commented on this pull request.
>
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
>
> In lisp/magit-stash.el:
>
>> @@ -231,7 +231,8 @@ When the region is active offer to drop all contained stashes."
>    (interactive
>     (list (--if-let (magit-region-values 'stash)
>               (magit-confirm 'drop-stashes nil "Drop %i stashes" nil it)
> -           (magit-read-stash "Drop stash"))))
> +           (let ((helm-comp-read-use-marked t))
>
> You'll have to declare that function, so that those who don't use helm
> don't get a warning about this variable being undefined.

Of course, done.

-- 
Thierry

thierryvolpiatto:
Basil L. Contovounesios <notifications@github.com> writes:

> @basil-conto commented on this pull request.
>
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
>
> In lisp/magit-stash.el:
>
>> @@ -231,7 +231,8 @@ When the region is active offer to drop all contained stashes."
>    (interactive
>     (list (--if-let (magit-region-values 'stash)
>               (magit-confirm 'drop-stashes nil "Drop %i stashes" nil it)
> -           (magit-read-stash "Drop stash"))))
> +           (let ((helm-comp-read-use-marked t))
>
> More importantly, so that it's marked as special (dynamically bound).

If this is what you mean, you don't need to declare such variable
explicitely as special as long it is declared globally.

-- 
Thierry

basil-conto:
@thierryvolpiatto 

> If this is what you mean, you don't need to declare such variable
explicitely as special as long it is declared globally.

What I meant is that there needs to be a `(defvar helm-comp-read-use-marked)` in one of the enclosing scopes, either within `magit-stash-drop` or at top-level, so that the byte-compiler doesn't think it's a local lexically bound variable. You have now done so at top-level, which is fine.

tarsius:
I am closing this proof-of-concept for now. We probably won't merge this as-is and we can look at it even if it is closed. The issue #4185 is where most of the conversation should take place.

tarsius:
@kyleam merged something similar after looking into it some more: 53f7cfe64f6df9fd67c171b3d2e4cb99d173cb7a.

PR #4191, opened 2020-08-17
Show message after creating patch

conao3:
Previously, users couldn't see the completion of their work
because magit didn't show a message after the patch was created.
This message lets the user know that magit have done their job
correctly.

### DIFF
--- a/lisp/magit-patch.el
+++ b/lisp/magit-patch.el
@@ -127,7 +127,8 @@ which creates patches for all commits that are reachable from
             (or (--some (and (string-match "\\`--output-directory=\\(.+\\)" it)
                              (expand-file-name (match-string 1 it) topdir))
                         args)
-                topdir))))))))
+                topdir))))))
+    (message "Creating patch done.")))
 
 (transient-define-argument magit-format-patch:--in-reply-to ()
   :description "In reply to"


tarsius:
Why's this necessary? Creating a patch should be very quick.

conao3:
Yes, too quick and too quiet.
For other magit commands, I can see that the command has been executed by changing the `magit-status` screen.
For example, if I change the branch, the HEAD display changes, and if I commit, the staged changes disappear and the commit are added to the commit list.
But `magit-patch-create` is so quiet in Emacs screen that I couldn't see what was happening.
I think it would be best to show the message as `Saved {{file-name}}.patch`, but I couldn't figure out how to get the file name of the created patch.

conao3:
Yes, too quick and too quiet.
For other magit commands, I can see that the command has been executed by changing the `magit-status` screen.
For example, if I change the branch, the HEAD display changes, and if I commit, the staged changes disappear and the commit are added to the commit list.
But `magit-patch-create` is so quiet in Emacs screen that I couldn't see what was happening.
I think it would be best to show the message as `Saved {{file-name}}.patch`, but I couldn't figure out how to get the file name of the created patch.

tarsius:
This seems unnecessary to me. Sorry.

conao3:
OK, I understand.

PR #4197, opened 2020-08-25
magit-fetch: Add -u for --unshallow option

aspiers:
Add support for --unshallow.  Set this to level 7 in transient, since
this is a relatively uncommon command, and is the inverse of a shallow
clone which magit already has at level 7.

----

#


tarsius:
Thanks!

PR #4204, opened 2020-09-01
git merge --X <option> -> git merge -X

simonpintarelli:
git merge -X <option> only uses a a single `-` according to https://git-scm.com/docs/git-merge


kyleam:
Thanks for catching that.  (I've tweaked the subject of your commit to follow this project's conventions more closely.)

PR #4207, opened 2020-09-03
Fix autoloading of magit-file-mode-map

peniblec:
Hello!

This is a followup to [the discussion on gitter](https://gitter.im/magit/magit?at=5eab4ca67a24ff01b0f415b0) about letting <kbd>C-x g</kbd> and friends work without calling `(require 'magit)` first (I sifted through the issue tracker as [suggested](https://gitter.im/magit/magit?at=5ee8ec92a85de30394190568), but I could not find a report about this).

While looking at recent commits to the Emacs repository, I noticed that one can just [autoload arbitrary forms](https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=5142149758333cfddc25c8c696e0e6f322e37d62).  I thus tried the following:

1. In `~/.emacs.d/elpa/magit-…/magit-files.el`, I added an `;;;###autoload` on the code that activates `global-magit-file-mode`.

2. I byte-compiled `magit-files.el` (<kbd>B</kbd> on `magit-files.el` in Dired).

3. `M-: (package-generate-autoloads "magit" "~/.emacs.d/elpa/magit-…/")`

4. `emacs -Q`

5. `M-x package-initialize`

6. Open a file inside a Git repository.

7. Lo!  <kbd>C-x g</kbd> et al. work without `requir`ing `'magit` first.

I'm far from certain that this is TRT; I didn't notice any side-effect when testing, but I'm not very familiar with the autoload machinery so I might be missing something.

CCing @npostavs who took part in the discussion back then.

### DIFF
--- a/lisp/magit-files.el
+++ b/lisp/magit-files.el
@@ -347,6 +347,7 @@ Currently this only adds the following key bindings.
 ;; variable but does not cause the mode function to be called, and we
 ;; cannot use `:initialize' to call that explicitly because the option
 ;; is defined before the functions, so we have to do it here.
+;;;###autoload
 (cl-eval-when (load eval)
   (when global-magit-file-mode
     (global-magit-file-mode 1)))


lisp/magit-files.el
```diff
@@ -347,6 +347,7 @@ Currently this only adds the following key bindings.
 ;; variable but does not cause the mode function to be called, and we
 ;; cannot use `:initialize' to call that explicitly because the option
 ;; is defined before the functions, so we have to do it here.
+;;;###autoload
 (cl-eval-when (load eval)
   (when global-magit-file-mode
     (global-magit-file-mode 1)))
```
kyleam, line 353: I don't think this is a good direction to go.  This change is autoloading a form that calls an autoloaded function, so loading the autoloads file loads magit.

<details>
<summary>demo</summary>

```elisp
(add-to-list 'load-path "~/src/emacs/with-editor/")
(add-to-list 'load-path "~/src/emacs/transient/lisp/")
(add-to-list 'load-path "~/src/emacs/dash.el/")
(add-to-list 'load-path "~/src/emacs/magit-devel/lisp/")
(load "magit-autoloads.el")
```

```console
$ # on master (a2599c3b)
$ make magit-autoloads.el   # from magit's lisp/
$ emacs -Q --batch -l /path/to/above --eval "(message \"%s\" (featurep 'magit))"
nil

$ # on this pr (3f7907a6)
$ make magit-autoloads.el   # from magit's lisp/
$ emacs -Q --batch -l /path/to/above --eval "(message \"%s\" (featurep 'magit))"
t
```

</details>

Jonas suggested a way around this in the conversation you linked to:

> We would need a separate file (name suggestions?) that only defined the key-binding-only modes and did not depend on other parts of Magit.

peniblec: Mmm, unless I'm mistaken, `magit` only gets autoloaded if `global-magit-file-mode` is `t`? If I add `(setq global-magit-file-mode nil)` before `(load "magit-autoloads.el")`, then I get `nil` when running your recipe.

I'm not against the separate-file solution; I even put in [a few names](https://gitter.im/magit/magit?at=5eac51467a24ff01b0f6bcca). Just to make sure I understand, the idea would be to move `magit-file-mode` and `magit-blob-mode` (and their keymaps) to a new library, and then… do what? [Jonas said to "autoload-require"](https://gitter.im/magit/magit?at=5eac371a5cd4fe50a3e2efff) it, but I'm not sure how to do that?

Thanks for the review, and sorry for being so slow :/

peniblec:
Yikes, I really am out of my depth with this autoloading stuff. I've just realized that with this PR, `magit-section-visibility-indicator` is set to `("…" . t)` even `(when (window-system))`.

I don't mind working on the "move keybinding-only modes to a dedicated library" approach, but I'm not sure I understand what @tarsius meant when he said to ["autoload-require"](https://gitter.im/magit/magit?at=5eac371a5cd4fe50a3e2efff) this would-be library?

tarsius:
> "autoload-require" 

"Put code in an autoload file, which when evaluated behaves similar to how `require` would."

peniblec:
Thanks. I'll close this PR and open an issue instead, as I suspect I'll have more questions before I have a clear idea of what to do.

PR #4211, opened 2020-09-13
Fix display of magit-blame's show-lines option for Emacs 27.1

kljohann:
In Emacs 27.1, the `:extend t` face attribute needs to be applied in order
to extend it until the edge of the window, see Emacs' NEWS file.

See also 891ebdca where this was done for other faces.

IIUC it's not necessary to hide this behind a version check, since earlier versions of Emacs will simply ignore the unknown face attribute.


tarsius:
There's probably a reason why I added a version check. My guess is that while recent releases silently ignore unknown attributes, older ones instead blow up. So for the sake of getting things done lets use a version check here too.

kljohann:
Sounds reasonable → updated.

tarsius:
Thanks!

PR #4213, opened 2020-09-15
Enabling `magit-file-mode-map` automatically

peniblec:
Hello again!

This is a followup to #4207, where I tried `;;;###autoload`ing forms willy-nilly, and it turned out not to be a great idea.  I have now moved `magit-file-mode` and `magit-blob-mode` from `magit-files.el` into a new file (tentatively called `magit-file-modes.el`), as suggested.  To be more specific, I moved everything under `;;; File Mode`, `;;; Blob Mode` and `;;; File Commands` to this new library.

In addition to actually enabling `magit-file-mode-map` when `global-magit-file-mode` is unset and defaults to `t`, this PR makes startup a bit faster when it *is* customized, IIUC by virtue of not loading `magit-files.el`, which `require`s `'magit`.  FWIW, on my laptop:

branch    | `global-magit-file-mode`¹  | startup time² (s) | keymap enabled?
----------|----------------------------|-------------------|----------------
master    | unset                      | 0.72              | ❌
this PR   | unset                      | 0.92              | ✔
master    | nil                        | 1.49              | ❌
this PR   | nil                        | 0.93              | ❌ 
master    | t                          | 1.48              | ✔
this PR   | t                          | 0.92              | ✔

(NB: the default case is *slower* though, since the new library where the globalized mode is defined is always loaded now.  More on that below)

I'll submit this as a draft because I'm sure there will be things I messed up; to start off, here are some things I'm unsure about:

1. While my final set of `require`s, `defvar`s and `declare-function`s quenches all compiler warnings, I ended up having to `(require 'magit-process)` despite `magit-file-modes.el` not using this library directly, because
    - `magit-file-mode-turn-on` calls `magit-inside-worktree-p` (defined in `magit-git.el`),
    - `magit-inside-worktree-p` eventually calls `magit-process-file` (defined in `magit-process.el`),
    - since `magit-process.el` `require`s `'magit-git`, we can't `(require 'magit-process)` in `magit-git.el`, otherwise we get circular imports;
    - my workaround does not feel very satisfying; should more time be spent untying this knot?

2. Should we actually enable `global-magit-file-mode` by default?  It does slow things down a bit in the default case, IIUC because Emacs loads `magit-autoloads.el` before customizations are applied, so the autoloaded `when` condition is always true, therefore Emacs runs `(global-magit-file-mode 1)` unconditionally (before turning it back off if it's customized to nil), which I assume loads `magit-file-modes.el` and everything it `require`s.  Here are the options I see:

    1. Keep the autoload and stop thinking about this.
    2. Keep the autoload, and try to shave off some milliseconds by `requir`ing fewer things.
    3. Drop the autoload, drop `:init-value t`, amend "Minor Mode for Buffers Visiting Files" in `magit.org`.
    4. Find another way to ensure `magit-file-mode-map` is enabled when `global-magit-file-mode` is enabled by default?

(FWIW I've [discussed the issue with Stefan Monnier on help-gnu-emacs](https://lists.gnu.org/archive/html/help-gnu-emacs/2020-09/msg00130.html); from my understanding, while out-of-core minor modes are not *strictly forbidden* to auto-enable themselves, it does seem to be something of a legally gray area.)

Aside from these points, I hope I didn't make too much of a mess with this PR.  Thank you very much for your patience.

## Footnotes

¹ "unset" = uncustomized; otherwise, set with `custom-set-variables`.

² With `/tmp/.emacs`:

``` elisp
(defun pr/startup ()
  (princ
   (format "startup: %s\n"
           (float-time (time-subtract after-init-time before-init-time)))
   'external-debugging-output))

(add-to-list 'load-path "/path/to/with-editor/")
(add-to-list 'load-path "/path/to/transient/")
(add-to-list 'load-path "/path/to/dash/")
(add-to-list 'load-path "/path/to/magit/lisp/")

(load "magit-autoloads.el")

;; Depending on the setup:
;; (custom-set-variables '(global-magit-file-mode …))
```

Then run:

``` sh
HOME=/tmp emacs --eval '(pr/startup)' --eval '(kill-emacs)'
```

### DIFF
--- a/lisp/Makefile
+++ b/lisp/Makefile
@@ -12,6 +12,7 @@ all: lisp
 git-commit.elc:
 magit-utils.elc:
 magit-section.elc:
+magit-file-modes.elc:
 ifeq "$(BUILD_MAGIT_LIBGIT)" "true"
 magit-libgit.elc:
 magit-git.elc:          magit-utils.elc magit-section.elc magit-libgit.elc
--- a/lisp/magit-file-modes.el
+++ b/lisp/magit-file-modes.el
@@ -1,0 +1,100 @@
+;;; magit-file-modes.el --- modes for files and blobs  -*- lexical-binding: t -*-
+
+;; Copyright (C) 2020  The Magit Project Contributors
+;;
+;; You should have received a copy of the AUTHORS.md file which
+;; lists all contributors.  If not, see http://magit.vc/authors.
+
+;; Author: Jonas Bernoulli <jonas@bernoul.li>
+;; Maintainer: Jonas Bernoulli <jonas@bernoul.li>
+
+;; Magit is free software; you can redistribute it and/or modify it
+;; under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 3, or (at your option)
+;; any later version.
+;;
+;; Magit is distributed in the hope that it will be useful, but WITHOUT
+;; ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+;; or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public
+;; License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with Magit.  If not, see http://www.gnu.org/licenses.
+
+;;; Commentary:
+
+;; This library defines modes useful in buffers visiting files and
+;; blobs.
+
+;;; Code:
+
+(require 'cl-lib)
+
+(declare-function magit-inside-worktree-p "magit-git")
+
+;;; File Mode
+
+(defvar magit-file-mode-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map "\C-xg"    'magit-status)
+    (define-key map "\C-x\M-g" 'magit-dispatch)
+    (define-key map "\C-c\M-g" 'magit-file-dispatch)
+    map)
+  "Keymap for `magit-file-mode'.")
+
+(defvar magit-file-mode-lighter "")
+
+(define-minor-mode magit-file-mode
+  "Enable some Magit features in a file-visiting buffer.
+
+Currently this only adds the following key bindings.
+\n\\{magit-file-mode-map}"
+  :package-version '(magit . "2.2.0")
+  :lighter magit-file-mode-lighter
+  :keymap  magit-file-mode-map)
+
+(defun magit-file-mode-turn-on ()
+  (and buffer-file-name
+       (magit-inside-worktree-p t)
+       (magit-file-mode)))
+
+;;;###autoload
+(define-globalized-minor-mode global-magit-file-mode
+  magit-file-mode magit-file-mode-turn-on
+  :package-version '(magit . "2.13.0")
+  :link '(info-link "(magit)Minor Mode for Buffers Visiting Files")
+  :group 'magit-essentials
+  :group 'magit-modes
+  :init-value t)
+;; Unfortunately `:init-value t' only sets the value of the mode
+;; variable but does not cause the mode function to be called, and we
+;; cannot use `:initialize' to call that explicitly because the option
+;; is defined before the functions, so we have to do it here.
+(cl-eval-when (load eval)
+  (when global-magit-file-mode
+    (global-magit-file-mode 1)))
+
+;;; Blob Mode
+
+(defvar magit-blob-mode-map
+  (let ((map (make-sparse-keymap)))
+    (set-keymap-parent map magit-file-mode-map)
+    (define-key map "p" 'magit-blob-previous)
+    (define-key map "n" 'magit-blob-next)
+    (define-key map "b" 'magit-blame-addition)
+    (define-key map "r" 'magit-blame-removal)
+    (define-key map "f" 'magit-blame-reverse)
+    (define-key map "q" 'magit-kill-this-buffer)
+    map)
+  "Keymap for `magit-blob-mode'.")
+
+(define-minor-mode magit-blob-mode
+  "Enable some Magit features in blob-visiting buffers.
+
+Currently this only adds the following key bindings.
+\n\\{magit-blob-mode-map}"
+  :package-version '(magit . "2.3.0"))
+
+;;; _
+(provide 'magit-file-modes)
+;;; magit-file-modes.el ends here
--- a/lisp/magit-file-modes.el
+++ b/lisp/magit-file-modes.el
@@ -66,13 +66,27 @@ Currently this only adds the following key bindings.
   :group 'magit-essentials
   :group 'magit-modes
   :init-value t)
+
 ;; Unfortunately `:init-value t' only sets the value of the mode
 ;; variable but does not cause the mode function to be called, and we
 ;; cannot use `:initialize' to call that explicitly because the option
 ;; is defined before the functions, so we have to do it here.
+
+;;;###autoload
+(defvar global-magit-file-mode--activated nil)
+
 (cl-eval-when (load eval)
-  (when global-magit-file-mode
+  (when (and global-magit-file-mode
+             (not global-magit-file-mode--activated))
+    (setq global-magit-file-mode--activated t)
     (global-magit-file-mode 1)))
+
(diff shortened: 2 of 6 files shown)


lisp/Makefile
```diff
@@ -10,6 +10,7 @@ all: lisp
 ## Build order #######################################################
 
 git-commit.elc:
+magit-file-modes:       magit-git.el magit-process.el
```
tarsius, line 13: The elk ran off.

tarsius:
It hasn't occurred to me before but we might also want to look into using a custom theme instead of (or "around") minor modes. Something similar was suggested at https://lists.gnu.org/archive/html/emacs-devel/2020-09/msg01289.html in a different context. Or maybe not, its 2 am...

lisp/Makefile
peniblec: 🤦 Let's see if we can lure it back…

peniblec:
> It hasn't occurred to me before but we might also want to look into using a custom theme instead of (or "around") minor modes. 

From what I read in "(emacs)Custom Themes", that does sound like a good fit. IIUC we'd drop `:init-value t`, and leave it to themes to enable `global-magit-file-mode`?

As far as I'm aware "non-face" themes are somewhat underused, so I'm curious to see how this thread pans out. FWIW I do think washing our hands of the autoloading stuff would make the most sense.

(If that's where we want to take things, I could amend this PR to completely get rid of the `(cl-eval-when (load eval) …` form.)

tarsius:
Wouldn't it be enough to put only the keymaps and modes into the new file and leave all the commands where they are now? If so, then that would also turn all the problems with circular dependencies into non-issues.

peniblec:
> Wouldn't it be enough to put only the keymaps and modes into the new file and leave all the commands where they are now?

Sure, that works!  I've updated the PR.
And it does wonders for load times too:

branch | `global-magit-file-mode`¹ | startup time² (s) | keymap enabled?
-------|---------------------------|-------------------|----------------
master | unset                     | 0.71              | ❌
oldpr  | unset                     | 0.93              | ✔
pr     | unset                     | 0.72              | ✔
       |                           |                   |  
master | nil                       | 1.51              | ❌
oldpr  | nil                       | 0.94              | ❌
pr     | nil                       | 0.71              | ❌
       |                           |                   |  
master | t                         | 1.52              | ✔
oldpr  | t                         | 0.94              | ✔
pr     | t                         | 0.71              | ✔

¹ ²: see OP.

> If so, then that would also turn all the problems with circular dependencies into non-issues.

Well, `magit-file-mode-turn-on` needs to call `magit-inside-worktree-p`.  To make it available, we have to either `(require 'magit-git)` (and thus `magit-process`), or slap an autoload on it (and on `magit-process-file` too, since `magit-inside-worktree-p` ends up calling it).

Or leave `magit-file-mode-turn-on` in `magit-files` and autoload it?  But that would mean loading `magit-files` when turning on `global-magit-file-mode`, which would slow startup down.

I've gone with the autoload solution in this revision; does that sound like a reasonable way to proceed?

Thank you for your time.

lisp/magit-file-modes.el
```diff
@@ -70,9 +68,12 @@ Currently this only adds the following key bindings.
...
 ;; is defined before the functions, so we have to do it here.
-(cl-eval-when (load eval)
-  (when global-magit-file-mode
-    (global-magit-file-mode 1)))
+;;;###autoload
+(progn
+  (require 'cl-lib)
+  (cl-eval-when (load eval)
+    (when global-magit-file-mode
+      (global-magit-file-mode 1))))
```
tarsius, line 76: Not sure, but maybe it would make more sense to leave the old code and add:

```elisp
;;;###autoload(when global-magit-file-mode (global-magit-file-mode 1))
```

Do loaddef files ever get compiled?

tarsius:
> I've gone with the autoload solution in this revision; does that sound like a reasonable way to proceed?

It does.

lisp/magit-file-modes.el
peniblec: > Not sure, but maybe it would make more sense to leave the old code and add [a new autoloaded form]

"Leave" as in, keep as-is, or "get rid of"? If "leave", IIUC (and from some quick testing), this means that Emacs will first run the added `(when ...)` form when loading `magit-autoloads.el`, and then the old `(when ...)` form when loading `magit-file-modes.el`, running `(global-magit-file-mode 1)` twice.

> Do loaddef files ever get compiled?

I'll go out on a limb and say "probably not", since AFAICT `autoload.el` [always adds `no-byte-compile: t` to their local variables](https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/emacs-lisp/autoload.el?h=emacs-27.1#n381).

peniblec:
> It does.

OK, then I'll polish up this PR (e.g. commit messages) as soon as I'm sure I understand what we should do with this autoloaded `(when ...)` form (cf. code review).

peniblec:
> I am pretty sure that happens in either case.

Mmm… more or less IIUC¹.

> So maybe:

That seems to work well; shall I change my second commit to use this, maybe…

1. with a comment explaining that we duplicate the form to avoid `requir`ing `cl-lib` in the loaddefs file?
2. factoring the `when` form into a function (e.g. `global-magit-file-mode--activate`)?

Thank you for helping me figure this stuff out!

# Footnotes

¹ Using the `.emacs` in the OP (with `global-magit-file-mode` unset), I get:

- With the current second commit:

   ``` elisp
   ;;;###autoload
   (progn
     (require 'cl-lib)
     (cl-eval-when (load eval)
       (when global-magit-file-mode
         (message "when in doubt")
         (global-magit-file-mode 1))))
   ```

  ⇒

  > when in doubt [2 times]


- With your initial suggestion:

  ``` elisp
  (cl-eval-when (load eval)
    (when global-magit-file-mode
      (message "when in doubt (cl-eval-when)")
      (global-magit-file-mode 1)))
  ;;;###autoload
  (when global-magit-file-mode
    (message "when in doubt (autoload)")
    (global-magit-file-mode 1))
  ```

  ⇒

  > when in doubt (autoload)
  > when in doubt (cl-eval-when)
  > when in doubt (autoload)

- And with your new suggestion (slightly tweaked to avoid an undefined variable error for `global-magit-file-mode--activated`):

  ``` elisp
  ;;;###autoload
  (defvar global-magit-file-mode--activated nil)
  (cl-eval-when (load eval)
    (when (and global-magit-file-mode
               (not global-magit-file-mode--activated))
      (setq global-magit-file-mode--activated t)
      (message "when in doubt (cl-eval-when)")
      (global-magit-file-mode 1)))
  ;;;###autoload(when (and global-magit-file-mode
  ;;;###autoload           (not global-magit-file-mode--activated))
  ;;;###autoload  (setq global-magit-file-mode--activated t)
  ;;;###autoload  (message "when in doubt (autoload)")
  ;;;###autoload  (global-magit-file-mode 1))
  ```

  ⇒

  > when in doubt (autoload)

tarsius:
> That seems to work well; shall I change my second commit to use this, maybe…

Yes.

> 1. with a comment explaining that we duplicate the form to avoid `requir`ing `cl-lib` in the loaddefs file?

Yes, but keep it short.

> 2. factoring the `when` form into a function (e.g. `global-magit-file-mode--activate`)?

Nah, I think that's not necessary and might complicate things.

peniblec:
Let me know if something needs tweaking, be it comments, commit messages, or anything.

Again, thanks for your patience!

tarsius:
I have just opened #4237, which instead *removes* `{global-,}magit-file-mode`.

Even if we end up going that route, I think this pull-request was still useful. If nothing else it forced us both to think about some details that turned out being kinda interesting. (We might also end up bringing back these modes to provide opt-in features in the future.)

The reason I now want to give up on the file modes is that however hard we try, we end up still having to load parts of Magit (because of `magit-inside-worktree-p`) and that in turn risks triggering loading all kinds of thing (due to user configuration, which we cannot control).

peniblec:
I've had a look at #4237; that looks like a much less hackish solution. I'm all for simplicity, so by all means, don't hesitate to close this PR once you're set on merging #4237.

Thanks for taking the time to step back and ponder some more!

PR #4223, opened 2020-10-04
Bug fix for `magit--tramp-asserts'

phil-s:
Looks like a typo.  I think this fix saves a call to `(magit-git-version)` (and a consequent `git` process) for every single `magit-status` in a remote directory?


phil-s:
n.b. I haven't tested this at all.

tarsius:
Thanks!

PR #4224, opened 2020-10-15
Install magit-git.el when BUILD_MAGIT_LIBGIT is false

yasuhirokimura:
Currently `make install` doesn't install magit-git.el when BUILD_MAGIT_LIBGIT is false. So make it install.


tarsius:
Thanks!

This is not quite correct, the `magit-git` is required even when `magit-libgit` is also used.

PR #4226, opened 2020-10-19
magit-clone-internal: Fix user-error arguments

basil-conto:
### Change Log

* `lisp/magit-clone.el` (`magit-clone-internal`): Fix typo in `user-error` format string and pass it the expected format argument.


tarsius:
Thanks!

PR #4230, opened 2020-10-24
magit-file-rename: Allow destination directory as second argument

### LINKED ISSUE #4122: RFE: magit-file-rename should allow directory destination
Today I discovered `magit-file-rename`, which is helpful indeed.

But usually I find myself _moving_ files around not actually renaming them.
So I would like suggest adding an additional command `magit-file-move`
which would take a directory as the second argument.

Or maybe better also accept a directory for the second argument of `magit-file-rename`?
This would parallel the behavior of `git mv`.

ian-mi:
Gives magit-file-rename similar behavior to rename-file when newname is a directory name by moving the source file into the destination directory.

Closes #4122

### DIFF
--- a/lisp/magit-files.el
+++ b/lisp/magit-files.el
@@ -430,20 +430,27 @@ the same location in the respective file in the working tree."
 ;;; File Commands
 
 (defun magit-file-rename (file newname)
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks!

PR #4236, opened 2020-11-02
`magit-read-stash': colorful prompt

nbfalcon:
`magit-read-stash`, used by `magit-stash-pop`, ... listed all stashes without highlighting the stash@... part as the magit stashes section does. This is fixed by my PR.

Currently, there is one inconsistency: the Stashes section does not display a colon after stash@..., while `magit-read-stash` does now. Should I fix that?

Here is a screenshot of my changes in action:

Note how the "Pop stash" prompt now has colorized stash@ parts.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -2276,11 +2276,30 @@ out.  Only existing branches can be selected."
                          require-match nil 'magit-revision-history
                          (magit-tag-at-point)))
 
+(defun magit--propertize-stash (stash)
+  "Propertize STASH, a line as returned by git stash list.
+Add `magit-hash' as the `font-lock-face' to the first occurrence
+of stash@{...}. Additional property transformations may be added
+in the future."
+  (save-match-data
+    (string-match "\\(stash@{[[:digit:]]+}\\)" stash)
+    (when-let ((start (match-beginning 1))
+               (end (match-end 1)))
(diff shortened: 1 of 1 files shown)


nbfalcon:
Also, another thing to consider: should we perhaps `user-error` out in `magit-read-stash` if there are no stashes, instead of displaying an empty prompt?

tarsius:
> `magit-read-stash`, used by `magit-stash-pop`, ... listed all stashes without highlighting the stash@... part as the magit stashes section does. This is fixed by my PR.

IMO there was no bug that needed fixing. There was room for adding some tinsel, which one may consider an improvement or an unnecessary distraction. I've concluded doing this is small improvement but I was on the fence.

> Currently, there is one inconsistency: the Stashes section does not display a colon after stash@..., while `magit-read-stash` does now. Should I fix that?

I agree it would be a good idea that the two places looked the same (iff we highlight the revision in the first place), but I don't think it should be done by changing one place *and then also the other*. Instead we should just not add the colon altogether. (We should remove a space though.)

I have implemented it like this myself. The biggest difference to your implementation however is that I did not add two additional helper functions with documentation and everything.

IMO doing those things while commendable in principal, here actually *decrease* readability in the long run. Just inlining the code in existing `magit-read-stash` means that only people who are  interested in the inner workings of this particular function ever have to deal with the (small) added complexity.

PR #4237, opened 2020-11-02
Add three global key bindings by default

tarsius:
For the longest time we have tried hard to avoid doing this but
it is just not worth the effort and leads to an inferior result.

Now we add three bindings by default.

```elisp
  ;;;###autoload
  (when magit-define-global-key-bindings
    (let ((map (current-global-map)))
      (define-key map (kbd "C-x g")   'magit-status)
      (define-key map (kbd "C-x M-g") 'magit-dispatch)
      (define-key map (kbd "C-c M-g") 'magit-file-dispatch)))
```

As you can see is easy to prevent these bindings from being added
and I hope having to do so won't upset anyone too much.  While we
did not add global bindings we already added these bindings in all
file-visiting buffers and we did not get any complaints about that.

This completely replaces `magit-file-mode` and the global variant,
which we define as an alias of the new variable.

---

This is an alternative to #4213, which goes in the opposite direction.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -8,7 +8,7 @@
 #+TEXINFO_DIR_CATEGORY: Emacs
 #+TEXINFO_DIR_TITLE: Magit: (magit).
 #+TEXINFO_DIR_DESC: Using Git from Emacs with Magit.
-#+SUBTITLE: for version 2.90.1 (v2.90.1-1078-gff8a616cd+1)
+#+SUBTITLE: for version 2.90.1 (v2.90.1-1083-gb18be6c14+1)
 
 #+TEXINFO_DEFFN: t
 #+OPTIONS: H:4 num:3 toc:2
@@ -25,7 +25,7 @@ directly from within Emacs.  While many fine Git clients exist, only
 Magit and Git itself deserve to be called porcelains.
 
 #+TEXINFO: @noindent
-This manual is for Magit version 2.90.1 (v2.90.1-1078-gff8a616cd+1).
+This manual is for Magit version 2.90.1 (v2.90.1-1083-gb18be6c14+1).
 
 #+BEGIN_QUOTE
 Copyright (C) 2015-2020 Jonas Bernoulli <jonas@bernoul.li>
@@ -282,13 +282,8 @@ tutorial.  Alternatively you can use an existing local repository, but
 if you do that, then you should commit all uncommitted changes before
 proceeding.
 
-To display information about the current Git repository, type ~M-x
-magit-status RET~.  You will be using this command a lot, and should
-therefore give it a global key binding.  This is what we recommend:
-
-#+BEGIN_SRC emacs-lisp
-  (global-set-key (kbd "C-x g") 'magit-status)
-#+END_SRC
+Type ~C-x g~ to display information about the current Git repository in
+a dedicated buffer, called the status buffer.
 
 Most Magit commands are commonly invoked from the status buffer.  It
 can be considered the primary interface for interacting with Git using
@@ -351,28 +346,24 @@ local branch onto the remote configured as the push-remote.  (If the
 push-remote is not configured yet, then you would first be prompted
 for the remote to push to.)
 
-So far we have mentioned the commit, push, and log transient prefix
-commands.  These are probably among the transients you will be using
-the most, but many others exist.  To show a transient that lists all
-other transients (as well as the various apply commands and some other
-essential commands), type ~h~.  Try a few.
+So far we have mentioned the commit, push, and log menu commands.
+These are probably among the menus you will be using the most, but
+many others exist.  To show a menu that lists all other menus (as well
+as the various apply commands and some other essential commands), type
+~h~.  Try a few.  (Such menus are also called "transient prefix
+commands" or just "transients".)
 
-The key bindings in that transient correspond to the bindings in Magit
+The key bindings in that menu correspond to the bindings in Magit
 buffers, including but not limited to the status buffer.  So you could
-type ~h d~ to bring up the diff transient, but once you remember that
-"d" stands for "diff", you would usually do so by just typing ~d~.  But
-this "prefix of prefixes" is useful even once you have memorized all
-the bindings, as it can provide easy access to Magit commands from
-non-Magit buffers.  You should create a global key binding for this
-command too:
+type ~h d~ to bring up the diff menu, but once you remember that "d"
+stands for "diff", you would usually do so by just typing ~d~.  But this
+"prefix of prefixes" is useful even once you have memorized all the
+bindings, as it can provide easy access to Magit commands from
+non-Magit buffers.  The global binding is ~C-x M-g~.
 
-#+BEGIN_SRC emacs-lisp
-  (global-set-key (kbd "C-x M-g") 'magit-dispatch)
-#+END_SRC
-
-In the same vein, you might also want to enable ~global-magit-file-mode~
-to get some more Magit key bindings in regular file-visiting buffers
-(see [[*Minor Mode for Buffers Visiting Files]]).
+In file visiting buffers ~C-c M-g~ brings up a similar menu featuring
+commands that act on just the visited file, see [[*Commands for Buffers
+Visiting Files]].
 
 It is not necessary that you do so now, but if you stick with Magit,
 then it is highly recommended that you read the next section too.
@@ -2534,11 +2525,10 @@ The log transient also features several reflog commands.  See [[*Reflog]].
 
   Show log for all references and ~HEAD~.
 
-
-Two additional commands that show the log for the file or blob that
-is being visited in the current buffer exists, see [[*Minor Mode for
-Buffers Visiting Files]].  The command ~magit-cherry~ also shows a log,
-see [[*Cherries]].
+Two additional commands that show the log for the file or blob that is
+being visited in the current buffer exists, see [[*Commands for Buffers
+Visiting Files]].  The command ~magit-cherry~ also shows a log, see
+[[*Cherries]].
 
 *** Refreshing Logs
 
@@ -2950,7 +2940,7 @@ in another buffer:
   Show all diffs of a stash in a buffer.
 
 Two additional commands that show the diff for the file or blob that
-is being visited in the current buffer exists, see [[*Minor Mode for
+is being visited in the current buffer exists, see [[*Commands for
 Buffers Visiting Files]].
 
 *** Refreshing Diffs
@@ -3808,10 +3798,7 @@ argument.
 Also see [[man:git-blame]]
 
 To start blaming invoke the ~magit-file-dispatch~ transient prefix
-command by pressing ~C-c M-g~.  (This is only the default binding and
-the recommended binding is ~C-c g~.  Also neither binding may be
-available if you disabled ~global-magit-file-mode~.  Also see [[*Minor
-Mode for Buffers Visiting Files]].)
+command by pressing ~C-c M-g~.
 
 The blaming suffix commands can be invoked from the dispatch
 transient.  However if you want to set an infix argument, then you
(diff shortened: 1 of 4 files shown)


kyleam (approved):
I agree with the rationale.

> As you can see is easy to prevent these bindings from being added
> and I hope having to do so won't upset anyone too much. While we
> did not add global bindings we already added these bindings in all
> file-visiting buffers and we did not get any complaints about that.

Right, IIUC the argument against this is that just (auto)loading Magit shouldn't register keybindings (or make similar user-visible changes).  (I recall there was a forge discussion about this and know that I'm guilty of registering -annex/-tbdiff/-imerge keys on load :/)

But, as you mention, what this PR does isn't too far from what we already do with an auto-enabled global-magit-file-mode.  I suppose the main difference is that, with this PR, there's no in-session toggle like there is for global-magit-file-mode.  Given the complications revealed by gh-4213, that seems like an okay tradeoff.

TxGVNN:
Hi @tarsius,

Seems this feature makes my configs crash.

My config before this feature, it works well. 
```
(use-package magit
  :ensure t
  :bind
  ("C-x g g" . magit-status)
  ("C-x M-g" . magit-dispatch)
  ("C-c M-g" . magit-file-dispatch)
  (:map magit-file-mode-map ("C-x g") nil))
```

After this feature, I updated as bellow
```
(use-package magit
  :ensure t
  :init (setq magit-define-global-key-bindings nil)
  :bind
  ("C-x g g" . magit-status)
  ("C-x M-g" . magit-dispatch)
  ("C-c M-g" . magit-file-dispatch))
```
Then emacs throw a warning `Error (use-package): magit/:catch: Key sequence C-x g g starts with non-prefix key C-x g`
I think because magit-define-global-key-bindings can be `nil` before magit. Could you give me some suggestions?


P/S: Because I map some functions in git-gutter into `C-x g` prefix, so that why I don't want use `C-x g` is magit-status
```
  ("C-x g p" . git-gutter:previous-hunk)
  ("C-x g n" . git-gutter:next-hunk)
  ("C-x g s" . git-gutter:stage-hunk)
  ("C-x g r" . git-gutter:revert-hunk)
```

Thanks

tarsius:
`magit-define-global-key-bindings`'s doc-string mentions that it has to be set to `nil` not only before `magit` is loaded but even before it is loaded.

> To prevent this, you must set this variable to nil *before*
> `magit` is loaded or autoloaded, afterwards it has no effect.

This is because the code that adds the bindings has to be autoloaded itself. It looks like this:

```elisp
;;;###autoload
(when magit-define-global-key-bindings
  (let ((map (current-global-map)))
    (define-key map (kbd "C-x g")   'magit-status)
    (define-key map (kbd "C-x M-g") 'magit-dispatch)
    (define-key map (kbd "C-c M-g") 'magit-file-dispatch)))
```

But I realize now that this is just asking for trouble (in the form of many bug reports, just like yours). We probably have to perform the same dance as for `magit-auto-revert-mode`. Well not the same, but something like:

```elisp
;;;###autoload
(progn
  (defun magit-maybe-define-global-key-bindings ()
    (when magit-define-global-key-bindings
      (let ((map (current-global-map)))
        (unless (lookup-key map (kbd "C-x g"))
          (define-key map (kbd "C-x g") 'magit-status))
        (unless (lookup-key map (kbd "C-x M-g"))
          (define-key map (kbd "C-x M-g") 'magit-dispatch))
        (unless (lookup-key map (kbd "C-c M-g"))
          (define-key map (kbd "C-c M-g") 'magit-file-dispatch)))))
  (if after-init-time
      (magit-maybe-define-global-key-bindings)
    (add-hook 'after-init-hook 'magit-maybe-define-global-key-bindings t)))
```

That way users can disable the default bindings anywhere in their init file and we can even avoid adding bindings that conflict with the users bindings without the user having to do anything about it.

Before I merge that I would like to hear from @kyleam one more time and @peniblec might want to take a look too (so much for "I'm all for simplicity" :grin:).

And here is a heretic thought: If we take great care like this not to mess with existing bindings does that mean we could bind `C-c g`?

TxGVNN:
> But I realize now that this is just asking for trouble (in the form of many bug reports, just like yours). We probably have to perform the same dance as for magit-auto-revert-mode. Well not the same, but something like:

Thanks, it works well. I think that is a good solution to help user can easily customize

peniblec:
@tarsius
> (so much for "I'm all for simplicity" 😁)

Well, I still prefer this new dance to the one we ended up with #4213; at least here we won't need comments explaining why we duplicate forms!

I just have one question and two idle musings (all of which you should feel free to ignore; this solution seems to do the job just fine, especially since the defcustom will work without users needing to mess with the early-init file):

- The THEN branch of the `if` is only there to ensure bindings are auto-enabled right after installing the package (without restarting), right?  Off the top of my head, I can't think of another situation where `magit.el` would be loaded without Emacs having already run its autoloads.

- The more I think about this issue, the more I find myself leaning toward [rg.el's approach](https://rgel.readthedocs.io/en/2.0.1/configuration.html#function-rg-enable-default-bindings):
    - don't load anything by default;
    - let the user call the setup function in their init file; 
    - offer a defcustom to pick the bindings that the setup function will use.

  The promise of everything Just Working™ out of the box is very alluring, and IMO Magit certainly belongs among the group of packages which should get away with defining global bindings by default (if such a group exists); in the end though I wonder if it's worth the hassle, and if having users opt-in wouldn't be an acceptable resolution…

- One thing `magit-file-mode` had going for it was that a minor mode can be easily toggled; bindings are not impossible to remove, but the process is less user-friendly.

kyleam:
[ sorry, wrote this a bit ago and got distracted before posting; haven't read @peniblec's response yet]

Eh, sorry for not thinking this through last time around.  In my view, what you propose is still simpler than the end result of gh-4213.

(Is it too late to say "here, add these bindings to your config if you want them"? :x)

tarsius:
> The promise of everything Just Working™ out of the box is very alluring, and IMO Magit certainly belongs among the group of packages which should get away with defining global bindings by default (if such a group exists); in the end though I wonder if it's worth the hassle, and if having users opt-in wouldn't be an acceptable resolution…

I have of course asked myself the same question—again and again over the years. Ultimately it boils down to Think Of The Children™.

IMO Magit is one of those packages that (along with many built-in packages) deserve a few global key bindings.

Magit is one of those packages that brings new users to Emacs; some of them because they have heard "it's worth using Emacs just for Magit" often enough to decide to give it a try. Some of those beginners will initially have a low threshold for things not working out of the box and I don't want to (continue to) scare them off by immediately forcing them to learn how to add key bindings and what that even means.

There's a lot of talk about making Emacs friendlier for beginners and this is a small step in that direction.

Many other packages also get global bindings by default; `vc` for example uses `C-x v ...`. This includes packages that a lot of Emacs users don't use and some of those users end up having to remove the respective key bindings and that is okay. Magit differs from those package in that it is not built-in but I don't believe that that means that it undeserving of some global bindings.

minad:
@tarsius Are the global keybindings really necessary? Is it not easier to tell the user to simply define three global keybindings in order to use magit? Now I have to disable the default keybindings and still have to define my own bindings.

tarsius:
I am considering improving this like so:

```elisp
;;;###autoload
(defcustom magit-define-global-key-bindings t
  "Whether to bind some Magit commands in the global keymap.

If this variable is non-nil, then the following bindings may
be added to the global keymap.  The default is t.

key             binding
---             -------
C-x g           magit-status
C-x M-g         magit-dispatch
C-c M-g         magit-file-dispatch

These bindings may be added when `after-init-hook' is called.
Each binding is added if and only if at that time no other key
is bound to the same command and no other command is bound to
the same key.  In other words we try to avoid adding bindings
that are unnecessary, as well as bindings that conflict with
other bindings.

Adding the above bindings is delayed until `after-init-hook'
is called to allow users to set the variable anywhere in their
init file, without having to make sure to do so before `magit'
is loaded or even just autoloaded.

Setting this variable after the hook has already been called
has no effect.

We recommend that you bind \"C-c g\" instead of \"C-c M-g\" to
`magit-file-dispatch'.  The former is a much better binding
but the \"C-c <letter>\" namespace is strictly reserved for
users; preventing Magit from using it by default.

Also see info node `(magit)Commands for Buffers Visiting Files'."
  :package-version '(magit . "3.0.0")
  :group 'magit-essentials
  :type 'boolean)

;;;###autoload
(progn
  (defun magit-maybe-define-global-key-bindings ()
    (when magit-define-global-key-bindings
      (let ((map (current-global-map)))
        (dolist (elt '(("C-x g"   . magit-status)
                       ("C-x M-g" . magit-dispatch)
                       ("C-c M-g" . magit-file-dispatch)))
          (let ((key (kbd (car elt)))
                (def (cdr elt)))
            (unless (or (lookup-key map key)
                        (where-is-internal def (make-sparse-keymap) t))
              (define-key map key def)))))))
  (if after-init-time
      (magit-maybe-define-global-key-bindings)
    (add-hook 'after-init-hook 'magit-maybe-define-global-key-bindings t)))
```

minad:
Yes, but why? This is easy to do in the init.el. I am using something like the following.

~~~ elisp
(use-package magit
  :bind (("C-x g" . magit-status)
         ("C-x M-g" . magit-dispatch)
         ("C-c g" . magit-file-dispatch))
~~~

What do you have in mind as installation method? In my experience, packages which bind keys are much more of a hassle than packages which don't and can be freely configured when installing them. Particularly bad are packages which allow to modify their keymap prefices via variables, which must be defined before loading the package. This doesn't play well together with the use-package style configs and one has to be careful to not load things too early somewhere else. The configuration issue, is somehow the same issue you are working around here with the after-init-hook. It would be great if you reconsider this!

minad:
It seems you are trying to improve the experience when installing via package-install and things should work directly afterwards. I wonder if this is a common use case? Most people seem to manage their configs or use preconfigured distributions.

tarsius:
I have just pushed the above, with slightly improved documentation. Please update magit and revert all the changes you have already made to deal with this. With this additional commit I hope that most existing users won't even notice that magit may now add global bindings itself.

We can of course discuss this further, but please try this first.

minad:
Thank you, I will try! 

Independent of everything works nicely now - there is just one thing I want to question - I think you said that "C-c g" is the preferred keybinding (I am also using that), yet you define "C-c M-g" in order to not violate the keybinding guidelines for packages. This is what confuses me. It would be nice to have some kind of plug-n-play setup, where everything works away, but then the user has to live with the inferior keybinding? In order to improve the situation they have to undo first the defaults and then define their own better bindings? Why even have the defaults then?

minad:
@tarsius Unfortunately it does not seem to work for me, "C-c M-g" is still bound even if I disable the magit default bindings. Probably there is something wrong with my config and I have to debug more carefully where magit is required etc etc. Maybe the hook interacts somehow badly with the lazy loading, I don't know. I am simply unbinding the unwanted keys now. But this somehow reflects my experience with packages which define unwanted keybindings - these produces more work in contrast to packages which don't do that.

But please don't feel bothered by me! After all magit is nice, so I trust your judgement here.

kyleam:
> In order to improve the situation they have to undo first the defaults and then define their own better bindings?

With 1ea8a70f, the `C-c M-g => magit-file-dispatch` binding should not be set if the user has given `magit-file-dispatch` a binding, including the recommended `C-c g` binding, so there should be nothing to undo.  Testing with `(global-set-key (kbd "C-c g") 'magit-file-dispatch)` and an otherwise vanilla config, that "don't add binding command if it is already bound" logic seems to be working as intended on my end.

stsquad:
I have no issue with magit changing how it approaches it's key bindings, however I was using magit-file-mode for my own tweaks:

    (use-package magit-files
      :commands global-magit-file-mode
      :bind (:map magit-file-mode-map
                  ("<f5>" . my-counsel-git-grep)
                  ("C-x v l" . magit-log-buffer-file)
                  ("C-x v d" . magit-diff-buffer-file))
      :defer 2
      :config (global-magit-file-mode))

So allowing my f5 functionality to bypass the normal dance I did to determine if the file was in git (because magit-file-mode implied that) and also re-binding vc-mode into the magit equivalent functions. Is there any avenue in magit for keeping these sort of tweaks or should I just move stuff to the local-mode-map via an after-file-visit hook or some such?

madscientist:
@minad The key space C-c N for any character N is reserved by Emacs convention to always be available for users to define their own keybindings, and no package should ever create keybindings in that space.

This is why magit cannot create the keybinding C-c g by default: it intrudes on the user's reserved keybinding space.

tarsius:
@stsquad I was thinking about keeping `{global-,}magit-file-mode`, expecting that it might come in hand after all.

If we bring back these modes then we would not enable them by default anymore, though. But since you now have to add something to the hook or keymap anyway, it seems okay if you have to explicitly enable the mode now.

If we do that then `global-magit-file-mode` could no longer be an alias for `magit-define-global-key-bindings`, which would be problematic because it would force users who opted out of the former to now also opt out of the latter. That might be somewhat upsetting.

I am not sure if we should bring back the mode for your case because IMO it would be better if you just used `magit-file-dispatch`. `magit-{log,diff}-buffer-file` are already available from that transient.

```elisp
(use-package magit
  :defer t
  :bind (("C-c g" . magit-file-dispatch)) ; as good as C-x v and certainly better than C-c M-g
  :config
  (transient-append-suffix 'magit-file-dispatch "e"
    '("<f5>" "grep" my-counsel-git-grep)))
```

andschwa:
Just FYI, I was using `(bind-key "C-c g" #'magit-file-dispatch magit-file-mode-map)` to define `C-c g` (as recommended) and removing `magit-file-mode-map` (or maybe aliasing it?) broke that. It's ok, I've bound it globally now (which actually makes more sense if, say, I wanted to stage a file so I think using `magit-file-mode-map` was an err on my part in the first place). Anyway, happy to see the default bindings be added, and appreciated the open discussion of rationale. It was pretty easy to figure out what happened when I updated Magit.

tarsius:
@bbatsov has written about these bindings [on this blog](https://emacsredux.com/blog/2020/12/10/essential-magit-file-commands). :pray:

joewreschnig:
I'm probably way too late to this (and I totally agree with the default binds), but I only just upgraded various packages after several months, and also got bit by the removal of `magit-file-mode-map`.

Using the transient might be "nicer" than `C-x v l`, but this is muscle memory that I've had "forever", and I still work with SVN repos regularly, so the following has been in my magit `use-package` since I started using it:
```
  :bind (:map magit-file-mode-map
         ("<remap> <vc-diff>" . magit-diff-buffer-file)
         ("<remap> <vc-print-log>" . magit-log-buffer-file)
         ("<remap> <vc-print-root-log>" . magit-log-all)
         ("<remap> <vc-annotate>" . magit-blame-addition))
```

For now I will fix it with my own global mode checking `magit-file-tracked-p` (even a bit more accurate for this than `magit-inside-worktree-p`) - but this is well beyond the Lisp skills of several other Emacs users I know who borrowed / extended that part of my config.

Issue #2985, opened 2017-02-03
Rethink parsing, caching, and refreshing


tarsius:
For a variety of reasons, not least in order to improve performance, it is necessary to rethink how Magit parses and caches complex `git` output, and how sections that are created from such output are being updated and cached.

Here I will discuss this with a focus on `git diff` output.  Other "complex `git` output" includes that of `git log` and `git blame`.  I am excluding `git log` from the discussion here because I am undecided on whether I want to continue to parse the output of that sub-command or switch to using `libgit2` for logs.

"Non-complex `git` output" is output that can easily be turned into a list of items or a single value, i.e. lists and values that we in the future want to get by calling a `libgit2` function.  Some of the things said below also apply to such "non-complex" output, for example because the output is simple, it might never-the-less be expensive to calculate it, so doing it asynchronously would be desirable here also.

---

These are some things I want to change:

1. All parsing of `git` output should happen in a separate buffer, not in the buffer into which the parsed output will ultimately be inserted (a process currently known as "washing").

  This has a few benefits:

   - We know whether there actually is anything to be inserted before we start doing so.  Currently it can happen that a section has to be "canceled", after its heading was already inserted.

   - Once we also start updating sections asynchronously, not doing the parsing in a separate buffer would mean that multiple processes may be inserting into the same buffer at the same time.

   - Once we start using Emacs' upcoming thread support, we can only take full advantage of that if the various threads operate in separate buffers.

   - The output of multiple `git` invocations, as well as other information, can more easily be combined into one amalgamation.  Doing it in separate buffers will also make it easier to delay the "refinement" steps.

     For example we might want to combine the output of `git diff` and `git diff --word-diff` (see {TODO separate issue}), and also the syntax-highlighting extracted from the corresponding file-/blob-visiting buffers (see #2942).

   - The parsing buffers can be used as a cache.

   - Using a separate buffer that the user never sees means that it can contain partially parsed output, which doesn't even have to be hidden.

2. All parsing should happen asynchronously.

  On the command line `git diff` gives immediate results, even for huge changes.  That's because only enough differences have to be formatted to fill the screen.  The rest is calculated asynchronously and by the time the user starts scrolling down, it is ready.

[...]

PR #4244, opened 2020-11-11
Defer Diff Calculation Until Section Expansion

ian-mi:
Adds an alternative diff function, magit-insert-name-status-diff, which uses git diff --name-status output to generate the appropriate file sections. These sections are initially left empty and populated with a separate git diff call when the section is first shown. This change greatly improves the performance of magit-status when dealing with large changes such as a complicated merge or large changes to vendored or generated code so long as most sections are left unexpanded.

In order to accomplish this I've added a new magit-section method, magit-section-insert-body, which is responsible for computing the section's body on initial expansion. After doing so I realized that this is mostly duplicating the functionality of washer although I ended up needing to perform additional steps such as fiddling with the parent's end markers and fixing the body's section and keymap text properties. If desired I can add these additional steps around the section washer and refactor the unexpanded file sections to use washer instead.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -820,8 +820,23 @@ and `:slant'."
 ;;; Section Classes
 
 (defclass magit-file-section (magit-section)
-  ((source   :initform nil)
-   (header   :initform nil)))
+  ((file      :initform nil)
+   (source    :initform nil)
+   (status    :initform nil)
+   (diff-args :initform nil)
+   (header    :initform nil)))
+
+(defmethod magit-section-insert-body ((section magit-file-section))
+  (with-slots (file source diff-args) section
+    (apply 'magit-git-wash
+           (lambda (args)
+             (when (cl-member-if (lambda (arg) (string-prefix-p "--color-moved" arg)) args)
+               (require 'ansi-color)
+               (ansi-color-apply-on-region (point-min) (point-max)))
+             (magit-wash-sequence
+              (apply-partially 'magit-diff-wash-file-diff section))
+             (delete-region (point) (point-max)))
+        "diff" `(,@diff-args "--" ,source ,file))))
 
 (defclass magit-module-section (magit-file-section)
   ())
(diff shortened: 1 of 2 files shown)


tarsius:
This is one of the things I was considering doing to improve performance related to diffs. It's not the one that goes the farthest though and I am not sure doing this now is the best thing forward. It might be better to start from a simpler implementation when finally undertaking the task of writing the ultimate version. On the other hand I don't really plan to work on that any time soon.

So maybe this is simple and effective enough to merge it now. Before I look at this though I have to ask you whether you are willing to assign your copyright to the fsf and if google is going to be cool with that.

basil-conto:
> if google is going to be cool with that.

I've heard that Google have an umbrella CA with the FSF for all their employees, but I could have heard wrong.

ian-mi:
Yes, I am happy to assign the copyright of any code I contribute to the FSF and Google already has a CLA on file.

I'm hoping that this PR will help to unblock more of the improvements listed in #2985 such as parsing diff output asynchronously and in separate buffers. If the additional information provided by git diff --raw would be useful, I think it would be relatively straightforward to modify magit-insert-name-status-diff to parse git diff --raw output instead of git diff --name-status output. Another possible improvement would be to directly parse git status --porcelain output to create the list of staged and unmerged files as currently the status of every unmerged file listed by git diff must be checked with a separate call to git status.

Please let me know if there are changes I could make that would make this PR better align with your vision of how this should all eventually look.

tarsius:
I am afraid I am going to have to pass on your offer to implement this.

> Please let me know if there are changes I could make that would make
> this PR better align with your vision of how this should all
> eventually look.

I did not attempt to implement this myself yet because I expect it to
be difficult and time consuming, and because I expect that once I am
done I would decide to not merge it after all anyway (and go straight
for the complete rewrite).

What you have implemented here is to far what would be needed and I
expect it would take me more time to guide you in implementing this
than to do it itself and I have already decided not to do the latter.

Sorry!

PR #4247, opened 2020-11-12
magit-tag: Add --message argument

olav35:
This adds the `--message` argument to `magit-tag`, so that it is possible to specify tag messages in the mini buffer. The reason I made this change, is that I need a way to make empty annotated tag messages and (ma)git does not permit making them interactively, that is with `t t`.

Ok?

### DIFF
--- a/lisp/magit-tag.el
+++ b/lisp/magit-tag.el
@@ -40,7 +40,8 @@
    ("-f" "Force"    ("-f" "--force"))
    ("-a" "Annotate" ("-a" "--annotate"))
    ("-s" "Sign"     ("-s" "--sign"))
-   (magit-tag:--local-user)]
+   (magit-tag:--local-user)
+   (magit-tag:--message)]
   [["Create"
     ("t"  "tag"     magit-tag-create)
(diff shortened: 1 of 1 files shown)


kyleam:
Thanks for the PR.  In my view this use case probably isn't common enough to warrant adding `--message` to the magit-tag transient.  Users that want `--message` can add it easily with `transient-append-suffix`.

olav35:
Fair enough. Personally I use this all the time, but if I understand if you think it is a too uncommon use case. I will look into adding it to my personal configuration with `transient-append-suffix`.

olav35:
For the record, I added the following code to my configuration to add a `-m` argument as described above.
```lisp
(transient-define-argument magit-tag:--message ()
  :description "Use message"
  :class 'transient-option
  :shortarg "-m"
  :argument "--message="
  ;; Empty (annotated)tag messages must be permitted because it is
  ;; impossible to create them interactively.
  :allow-empty t)
(transient-append-suffix
  'magit-tag
  "-u"
  '(magit-tag:--message)))
```

PR #4249, opened 2020-11-13
Match yes-or-no prompts not ending in ? or :

sideshowcoder:
A magit subprocess can prompt for input by a user with a prompt not matching a
prompt ending in `?` or `:`.

Example arc [phabricator](https://www.phacility.com/phabricator/) provided via [magit-arcanist](https://github.com/sideshowcoder/magit-arcanist):
```
 Amend HEAD with lint patches? [Y/n]
```
This patch makes the ending in `?` or `:` optional.


tarsius:
Thanks!

PR #4253, opened 2020-11-22
manual: Fix typo

kraai:


kyleam:
Thanks.  (I've pushed an amended commit with the regenerated texi file).

Issue #4172, opened 2020-07-24
Support regions in git-rebase-kill-line


Stebalien:
To reproduce, open the interactive rebase buffer, highlight a region of commits to remove, then run `git-rebase-kill-line`. Currently, this will kill the commit on the next line, not the commits in the selected region. I'd expect it to kill all selected commits.

aesyondu:
Same, I'm doing something like this:

<img width="117" alt="image" src="https://user-images.githubusercontent.com/57334943/96336703-3b4d2000-10b4-11eb-8553-abc135d51f05.png">

GIVEN BRANCHES master AND feature

WHEN CHECKOUT master (lb + mov to master + bb)
WHEN REBASE -INTERACTIVELY ELSEWHERE (mov to feature + r-ie)
WHEN SELECT ALL LINES EXCEPT LAST LINE (C-SPACE nnnnnnn)
WHEN DROP (k)

THEN all selected lines should be dropped.

----

Or maybe I'm doing it the wrong way and there is a much better way? I basically want to move the commit that the branch points to to a separate branch (picking), while hard resetting the branch there at the same time.

```
From this:
A - B - C (master)
  \
    - D (feature-other) - E (feature)

To this:
A - B - C (master) - E (feature)
  \
    - D (feature-other)
```

kyleam:
> Or maybe I'm doing it the wrong way and there is a much better way?

I think `magit-rebase-subset` (<kbd>rs</kbd>) would work nicely for your situation.

---

Teaching `git-rebase-kill-line` (and other related commands) to act on the region makes sense to me, though it's also not an itch I've ever had (despite having a pretty rebase-heavy workflow).  A few weeks ago a started putting something together for this.  I _think_ it was mostly working but had some rough edges.  Anyway, I'll try to revisit it soon-ish.

aesyondu:
Works nicely, thanks for the tip!

PR #4258, opened 2020-11-29
git-rebase: Region support for -kill-line and commit action commands

kyleam:
This series updates `git-rebase-kill-line` to support killing lines selected with the region (as proposed in gh-4172).  It also adds region support to commands that operate on "commit" lines (`git-rebase-pick`, `git-rebase-reword`, `git-rebase-edit`, `git-rebase-squash`, and `git-rebase-fixup`).

---

@Stebalien @aesyondu Care to give this a try?

I'll plan to merge this in a few days if there are no objections.

### DIFF
--- a/Documentation/RelNotes/3.0.0.org
+++ b/Documentation/RelNotes/3.0.0.org
@@ -141,6 +141,10 @@
 - New command ~git-rebase-break~ inserts a "break" action in the
   rebase to-do sequence (available as of Git v2.20).  #3762
 
+- ~git-rebase-kill-line~ and the commands for changing the action of a
+  commit line (e.g., ~git-rebase-squash~) learned to work on all lines
+  selected by the region.  #4172
+
 - The ~--color-moved~ diff argument is supported now, but isn't
   available from the diff transients by default.  To enable it
   use "C-x l" in those transients.  #3424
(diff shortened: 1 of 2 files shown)


Stebalien:
Works:

* Normal emacs region.
* Normal evil region `v`.

Doesn't work:

* Evil line selection `V`. This drops the commits in the region, but also drops the commit following the region. Does this require a patch to evil-magit as well?
* Evil column selection `C-v`. It doesn't do anything but honestly, I didn't expect it to do anything so that's probably fine.

The fact that line selection doesn't work with evil is a bit annoying given that that's the command I'll usually reach for when working with lines, but otherwise this patch seems like a step in the right direction.

kyleam:
Sorry, I'm not an evil user.  If someone (@justbur?) can point to a simple change that would make this play better with it, that'd be great.  Anything extensive, though, would have to be done by someone else (and would likely fall outside of Magit proper).

tarsius:
Looks good.

> To avoid duplicating the region handling, move the custom kill logic into git-rebase-current-line as a special case.

Ah, the dangers of looking at hunk headings when writing commit messages... :grin:

kyleam:
> Ah, the dangers of looking at hunk headings when writing commit messages...

Doh, indeed.  Thank you for spotting that.

pes2020kk (commented):
#ipad

PR #4265, opened 2020-12-12
Show magit manual for magit-dispatch

Szunti:
Had no manual defined before, so nothing lost, but it is more newbie
friendly to end after repetitive panic "?"-s in the manual.

At least it would have been a tiny bit easier for me.


tarsius:
Thanks!

PR #4266, opened 2020-12-12
Check XDG convention for elpa dirs in default.mk

Szunti:
Since Emacs 27.1 user-emacs-directory can also be in XDG_CONFIG_HOME or (~/.config).

### DIFF
--- a/default.mk
+++ b/default.mk
@@ -137,7 +137,15 @@ endif
 
 ifndef LOAD_PATH
 
-ELPA_DIR ?= $(HOME)/.emacs.d/elpa
+USER_EMACS_DIR = $(HOME)/.emacs.d
+ifeq "$(wildcard $(USER_EMACS_DIR))" ""
+  XDG_CONFIG_DIR = $(or $(XDG_CONFIG_HOME),$(HOME)/.config)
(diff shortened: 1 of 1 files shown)


tarsius:
Instead of `find`, it would IMO be better to use `ifeq ($(wildcard $(XDG_CONFIG_HOME)),) ...`.

Szunti:
Something like this?

tarsius:
Yes, but I think the inner `ifneq` doesn't do much good.

Previously we could end up with `ELPA_DIR` being `/home/me/.emacs.d/elpa` even when that does not exist, and that is okay.

With your change the fall through is `/elpa`. I think e.g. `/home/me/.emacs.d/elpa`. We can get that by simply not checking whether `XDG_CONFIG_DIR` exists.

Szunti:
Sorry, for taking this many attempts for a few lines. This should give the default /home/me/.emacs.d/elpa location if nothing exists.

tarsius:
Np. We both improved upon the other's ideas.

I rebased and dropped the commend before merge.

tarsius:
Thanks!

Issue #4127, opened 2020-05-26
Commit body indicator in the log


jazzzz:
It would be nice to have an indicator for each commit with a body in the message, like in GitHub here: ![image](https://user-images.githubusercontent.com/290626/82930212-4d9fb380-9f85-11ea-888d-fc363242e65b.png)
It would allow to see directly which commits have more than a subject.

tarsius:
Personally I feel like this feature would not be much of an improvement. (Or worse would make things slightly worse by wasting some space.) So that's one reason why that feature doesn't exist.

The much more important reason however is that it would make parsing of the git output more difficult and more expensive. We would have to use `%b` in `log --format` and then we would have to move over all those lines just so we can know whether there are any such lines.

Considering that I think that this feature is not very useful and that people are already complaining about Magit being to slow, I am afraid I have to say no.

jazzzz:
Well, it makes me sad, because I love magit, but it's the only tool that don't let me know if there's a commit body, which is essential for reviewing merge/pull requests.

jazzzz:
Wouldn't be possible to least add an option to allow this?

tarsius:
> which is essential for reviewing merge/pull requests.

That sounds like a huge exaggeration. "It's convenient" sure, I can see that, but "essential"? Besides, when reviewing a pull request you have to review every commit anyway so you might as well review each commits message at the same time as you review its diff.

> Wouldn't be possible to least add an option to allow this?

No it would just be to much work for IMO to little gain.

kyleam:
> but it's the only tool that don't let me know if there's a commit body, which is essential for reviewing merge/pull requests.

I suspect you already know, and it's just not close enough to the indicator that you want, but just in case: Are you aware of the log transient's `++header` flag and the associated `magit-log-revision-headers-format` option?  That'd let you see the full message without needing to move the cursor to each commit you're interested in.

tarsius:
I was going to suggest that too but then I tried it out and it looked awful. Is it a regression that the graph is not drawn next to the commit body?

kyleam:
> I was going to suggest that too but then I tried it out and it looked awful. 

Heh, true, it's not beautiful.

> Is it a regression that the graph is not drawn next to the commit body?

It smells like one, though I'm not positive.  I'm don't know the last time I've used `++header` myself.

kyleam:
> > Is it a regression that the graph is not drawn next to the commit body?
> 
> It smells like one, though I'm not positive.

[...]

PR #4269, opened 2020-12-18
magit-log: add marker " .." for non-empty bodies

qwerbzuio:
Show " .." at the end of subject if a text body is available. The term
"%% b" will insert the (truncated) body with a space in front, which
effectively expands to " ..". If no body is present, the space will also
be omitted.

----

# 

I think this is a useful addition as one can immediately see which commits have
more explanation available.

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -435,7 +435,8 @@ the upstream isn't ahead of the current branch) show."
    ("=S" "Show signatures"     "--show-signature") ;1
    ("-h" "Show header"         "++header")         ;4
    ("-p" "Show diffs"          ("-p" "--patch"))   ;2
-   ("-s" "Show diffstats"      "--stat")]          ;2
+   ("-s" "Show diffstats"      "--stat")           ;2
+   ("-b" "Show body indicator" "++body")]
   [["Log"
     ("l" "current"        magit-log-current)
     ("o" "other"          magit-log-other)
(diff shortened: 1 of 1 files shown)


kyleam:
Thanks for the patch.

This topic came up in gh-4127.  There were two main objections there: 1) the subjective question of whether such a marker is an improvement/desirable and 2) the added complication in parsing.

As you'd expect, the second is the bigger sticking point, and I think what you suggest is a clever solution that sidesteps the problem (and it looks like the needed placeholders are supported in our minimal git version).  Based on quickly testing, it seems to mostly work, but it doesn't properly handle `++header`.

My views align with @tarsius's on the first.  I personally wouldn't want to see these in the log buffer; it would be a waste of characters given how I work.  So, if we do decide to support this, I'd vote for having it as optional behavior.

qwerbzuio:
Let me know if it is ok to put the option into the "Formatting" section of the magit-log transient. Any alternatives?

tarsius:
> Let me know if it is ok to put the option into the "Formatting" section of the magit-log transient. Any alternatives?

I guess this would make sense, but I wouldn't want to confuse users with this option, so it should be hidden by putting it on a level that is hidden by default: `(7 "-b" "Show body indicator"     "++body")`. You would then have to reveal it like so:

tarsius:

tarsius:
But I think we should not add every conceivable feature and I just don't think this very useful. Obviously you disagree, but for every feature there is someone who wants it and adding everything is too expensive.

I guess now we have a reasonable implementation ready in case this somehow becomes a popular feature request, but until that happens, I don't think we should add this.

PR #4270, opened 2020-12-18
copy-section-value: copy added/removed lines

qwerbzuio:
Make it possible to copy only added/removed lines via a
positive/negative prefix arg.

----

### DIFF
--- a/lisp/magit-extras.el
+++ b/lisp/magit-extras.el
@@ -589,27 +589,20 @@ When ARG is a list, strip the outer diff marker column from
 INPUT. With a positive/negative numeric ARG, strip the outer diff
 marker column and copy only added/removed lines, as well as
 unchanged lines. When ARG is nil, return INPUT unchanged."
-  (if arg
-      (let* ((magit-diff-sign-regexp
-              (cond
-               ((listp arg) "-\\|\\+")
-               ((or (eq '- arg) (< (prefix-numeric-value arg) 0)) "-")
-               (t "\\+")))
-             (magit-diff-filter-func
-              (lambda (str)
-                (if (listp arg) t
-                  (if (string-match-p
-                       (format "^\\(%s\\| \\)" magit-diff-sign-regexp) str)
-                      t
-                    nil)))))
-        (mapconcat
-         (lambda (str)
-           (replace-regexp-in-string
-            (format "^\\(%s\\| \\)" magit-diff-sign-regexp) "" str))
-         (-filter (apply-partially magit-diff-filter-func)
-                  (split-string input "\n"))
-         "\n"))
-    input))
+  (if (null arg)
+      input
+    (mapconcat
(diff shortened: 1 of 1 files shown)


lisp/magit-extras.el
```diff
@@ -582,6 +582,35 @@ the minibuffer too."
...
 
+(defun magit-filter-hunk-substring (input arg)
+  "Maybe strip some diff markers and lines from INPUT, depending on ARG.
+
+When ARG is a list, strip the outer diff marker column from
+INPUT. With a positive/negative numeric ARG, strip the outer diff
+marker column and copy only added/removed lines, as well as
+unchanged lines. When ARG is nil, return INPUT unchanged."
+  (if arg
+      (let* ((magit-diff-sign-regexp
```
tarsius, line 593: No need to namespace this. Since this can only be accessed within this function something like `re` is good enough.

lisp/magit-extras.el
```diff
@@ -582,6 +582,35 @@ the minibuffer too."
...
+              (cond
+               ((listp arg) "-\\|\\+")
+               ((< (prefix-numeric-value arg) 0) "-")
+               (t "\\+")))
+             (magit-diff-filter-func
+              (lambda (str)
```
tarsius, line 599: Just define this function where it is used below instead of binding some variable to it.

lisp/magit-extras.el
```diff
@@ -582,6 +582,35 @@ the minibuffer too."
...
+                (if (listp arg) t
+                  (if (string-match-p
+                       (format "^\\(%s\\| \\)" magit-diff-sign-regexp) str)
+                      t
+                    nil)))))
+        (mapconcat
+         (lambda (str)
+           (replace-regexp-in-string
+            (format "^\\(%s\\| \\)" magit-diff-sign-regexp) "" str))
+         (-filter (apply-partially magit-diff-filter-func)
```
tarsius, line 609: Use `seq-filter` instead.

It makes no sense to use `apply-partially` and then not provide any arguments.

tarsius (changes requested):

tarsius:
IMO this simpler version would be better: https://github.com/magit/magit/compare/keep-region-side.

kyleam:
@qwerbzuio Nice idea.  Seems like an obvious improvement, but it didn't occur to me when writing the initial "strip markers" variant or afterwards when using the functionality fairly regularly.

> IMO this simpler version would be better: https://github.com/magit/magit/compare/keep-region-side.

Looks good, aside from using cl-minusp instead `(< x 0)` :)

Perhaps already planned with the final patch, but just in case: the manual needs an update as well.

tarsius:
> Looks good, aside from using cl-minusp instead `(< x 0)` :)

Okay, I fixed that. :P

PR #4277, opened 2020-12-24
magit-copy-section-value: Fix copying of last line

qwerbzuio:
+/- are not properly stripped from the last line in the active region due to no
trailing newline being in that region.

### DIFF
--- a/lisp/magit-extras.el
+++ b/lisp/magit-extras.el
@@ -654,7 +654,7 @@ argument."
                    (region-beginning)
                    (region-end))
        (replace-regexp-in-string
-        (format "^\\%c.*\n" (if (< (prefix-numeric-value arg) 0) ?+ ?-))
+        (format "^\\%c.*\n?" (if (< (prefix-numeric-value arg) 0) ?+ ?-))
         "")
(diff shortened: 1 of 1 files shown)


kyleam (approved):
Thanks, makes sense.  I think it'd be good for your PR's description to be in the commit message, so I'll push an amended commit.

---

Here's the example that I verified is fixed with your patch:

If I have

```
modified   foo
@@ -1,2 +1,3 @@
 one
 two
+three
\ No newline at end of file
```

and I call `magit-copy-section-value` with a negative prefix argument and the region spanning the beginning of "two" and end of "three", it incorrectly copies

```
two
three
```

With this patch, it correctly copies

```
two
```

qwerbzuio:
Ok, thanks for the amend. Note that the `\ No newline at end of file` shouldn't even be necessary to reproduce the issue. Even if it exists, the newline is in any way not part of the transient region.

kyleam:
> Note that the `\ No newline at end of file` shouldn't even be necessary to reproduce the issue

Right, thanks for noting that.  That was a leftover from an initial misstep when I wasn't quite sure how to trigger it.

Issue #4279, opened 2020-12-27
git-commit-mode highlights pseudo headers incorrectly


alphapapa:
* Magit 20201225.4, Git 2.7.4, Emacs 26.3, gnu/linux
* Tested in a clean Emacs configuration (i.e. `emacs-sandbox.sh -i magit`)

I noticed that `git-commit-mode`'s highlighting of pseudo headers doesn't quite work correctly.  Here's a screenshot:

![python-4](https://user-images.githubusercontent.com/601365/103180320-58dbfa00-485a-11eb-9dcd-c24f9a34babb.png)

In case it helps, I'm attaching the text of the screenshot: [text.txt](https://github.com/magit/magit/files/5745315/text.txt)

As always, thanks for your work on Magit.

tarsius:
I don't understand. That is `git log` output--what is it doing in the buffer that is being used to write the commit message and why do you expect `git-commit.el` to know how to prettify log output?

alphapapa:
That's not a `magit-log` buffer, that's a `COMMIT_EDITMSG` buffer.  I was writing the commit message for a squash commit, and the `COMMIT_EDITMSG` buffer looked like that, with some parts highlighted using the `git-commit-pseudo-header` face, and others not, without any apparent pattern to where the boundaries are.  I guess the font-lock keywords need adjustment, but I'm not sure.

tarsius:
> That's not a `magit-log` buffer, that's a `COMMIT_EDITMSG` buffer. 

I figured so much when I mentioned "the buffer that is being used to write the commit message".

> I was writing the commit message for a squash commit, and the `COMMIT_EDITMSG` buffer looked like that, 

I don't think I have ever seen a squash command do that, that's why I asked how that text ended up in that buffer. Did a magit command do that or did you squash on the command line? How?

Anyways...

For me the font-lock sometimes goes berserk even when that buffer contains more conventional text. I don't know why and also why no one has reported that until now. I haven't looked into it. I guess now I finally have to do it.

That being said, if this is a different issue that only kicks in when the buffer contains this (until now at least to me) unexpected text, then I am probably fine with that. We cannot properly propertize everything that might somehow end up in that buffer.

I would appreciate help with this. I have ignored it for so long, I will probably continue to do unless you or someone else helps me figure out what's going on. (I am gonna be busy for a while.)

alphapapa:
[...]

PR #4281, opened 2020-12-28
Add/Fix: (git-commit-mode) Highlighting of log headers

alphapapa:
See #4279.

The result now looks like this: 
fixed


tarsius:
Thanks a lot! That looks much better.

You didn't follow the commit message conventions that we use but that turned out to be a good thing, because when I went to change them I noticed that it triggered the issue I was talking about *and* that after evaluating your changes that issue was gone. :grinning:

PR #4284, opened 2021-01-01
bisect: Support specifying alternate terms

kyleam:
I find alternate terms for bisecting very useful, and it's a common reason I resort to the command line.  I've tried to force the "good" and "bad" terms onto bisects a good number of times, and in most cases my brain isn't capable of keeping things straight.

This series adds support for parsing the status buffer when bisecting with alternate terms (second commit) and for running a bisect with alternate terms (third commit).

While I think more people than just myself will find this useful, I doubt most users will, given that I don't recall support for alternate terms being requested.  For this reason, I've put this functionality at a hidden level (6).

(The texi file hasn't been regenerated yet.)

### DIFF
--- a/Documentation/RelNotes/3.0.0.org
+++ b/Documentation/RelNotes/3.0.0.org
@@ -186,6 +186,10 @@
 
 - ~git bisect~ is now run asynchronously.  #3802
 
+- ~magit-bisect~ now supports specifying alternate terms.  The new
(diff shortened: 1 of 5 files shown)


tarsius (approved):
Makes sense and looks good.

Happy new year!

kyleam:
Thanks for the review.

> Happy new year!

Same to you!

PR #4287, opened 2021-01-11
Don't use a completion function with selectrum-mode

rakyi:
We shouldn't override selectrum sorting either.

### DIFF
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -519,7 +519,8 @@ acts similarly to `completing-read', except for the following:
   (prompt choices &optional predicate require-match initial-input hist def)
   "Magit wrapper for standard `completing-read' function."
   (unless (or (bound-and-true-p helm-mode)
-              (bound-and-true-p ivy-mode))
+              (bound-and-true-p ivy-mode)
+              (bound-and-true-p selectrum-mode))
     (setq prompt (magit-prompt-with-default prompt def))
     (setq choices (magit--completion-table choices)))
   (cl-letf (((symbol-function 'completion-pcm--all-completions)


kyleam (changes requested):
Thanks.  The change itself is probably fine, but the rationale is confusing.

> We shouldn't override selectrum sorting either.

The code that this is touching isn't about not overriding sorting of completion frameworks.  We _do_ want completion to respect the order we choose for the collection.  As explained in 60d381c6 (Don't use a completion function with {ivy,helm}-mode, 2018-06-08), not using a completion function here was necessary to work around a helm issue and was "why do it if you don't need it" for ivy.  selectrum probably falls into one of those two categories, but it would be good for the commit message to say which.

More importantly in terms of user experience, it'd be good to know if selectrum preserves the order or if special handling is needed there.

rakyi:
I might have misunderstood what is happening there. My problem is that I'm using selectrum with prescient.el for sorting and filtering the candidates. Without these changes the list of candidates won't be sorted by prescient, which I expect to happen in every completion.

That means that for example when `magit-read-repository` calls `magit-completing-read` with list of repos, they are sorted in some static order instead of by recency and frequency etc.

I can explicitly set `magit-completing-read-function` instead of course, I just thought that this way we'd respect user's preferences better, because I was surprised that magit completion is working differently than the rest.

kyleam:
I'd recommend you configure things how you want on your end because at this point [*] the alternative would be interfering with the spots where magit is using meaningfully sorted candidates (e.g. see `magit-list-refs-sortby`).

[*] We could audit all the completion call sites and decide for each one if we consider the candidates to have a meaningful order.  For those that don't, we could, say, let-bind an internal variable that indicates that `magit-builtin-completing-read` and `magit-completing-read-multiple` shouldn't do the things they currently do to preserve the order.  That'd be added complexity and a good amount of effort, but perhaps it'd be worth doing.  Not sure.

tarsius:
I recently used selectrum for a few days and liked it but eventually went back to ivy with the intention of waiting out the phase where all the accidentally broken functionality gets reported and fixed. We should also wait before jumping in adjusting magit, things (hopefully) are still in flux.

Issue #4285, opened 2021-01-06
Suggestion: Replace magit-maybe-save-repository-buffers etc. in magit-status with "Unsaved buffers" section


CyberShadow:
Hi @tarsius, and thank you for all your work on magit. I would like to make a suggestion which I hope is not too egregious.

Hopefully we can be agree that inserting modal prompts which appear only *sometimes* in the middle of a process which the user would otherwise do in seconds is not good UX. The "Save file" prompts are currently my #​1 annoyance with magit, and the most frequent cause for getting into a confused state where I have to hammer G-g and C-] to get out of. Though `magit-save-repository-buffers` exists, the options offered are to never save (which is desired rarely), always save (which is desired often but not always, and is not undoable), and the prompts.

I would like to suggest the following changes, which would be in my opinion an improvement:

- Only show the save prompts before operations where saving modified buffers would affect the outcome, such as staging files, pushing a stash, or diffing against the work tree in a context other than `magit-status`.
- In `magit-status`, add a section before "Unstaged changes" which enumerates modified buffers for files belonging to the current repository.
  - `k` (`magit-delete-thing`) would revert the buffer.
  - `C-x C-s` (`write-buffer`) with point on a file's entry would save it.
  - For bonus coolness, show the diff between the file on disk and the buffer contents, by piping the buffer contents into `git diff`'s standard input.

The new section should still clearly inform the user that they have unsaved buffers without interrupting them. Being able to see a diff between the buffer and the file could be a useful feature by itself.

I'm happy to attempt an implementation :)

tarsius:
> always save (which is desired often but not always, and is not undoable)

Never-the-less that is what I use and recommend to others. Actually I don't remember the last time I felt that the automatic saving was a problem.

> Only show the save prompts before operations where saving modified buffers would affect the outcome, such as [...]

It would also mean that we should have to go through all operations we offer and determine whether it is one of those, and whenever we added a new operation we would have to think about again. And if we stop doing always, then we would probably end up having to do it by explicitly calling the "maybe save" function in a hundred or more places.

Since I think "everyone" would be best adviced to just go with the "always save" setting, I don't think this effort is justified.

> I'm happy to attempt an implementation :)

You could do so by advising `magit-maybe-save-repository-buffers` to only do anything if `this-command` is not one of a certain set of commands. Coming up with that set would constitute most of the work.

> add a section [...] which enumerates modified buffers for files belonging to the current repository. 

[...]

PR #4290, opened 2021-01-12
Add "Unsaved changes" section inserter, and accompanying functionality

CyberShadow:
#4285

Implements:

- [x] Section construction
  - [X] Enumerate modified buffers belonging to the repository
    - [x] Also add entries for buffers visiting files that don't yet exist - show them as "new file"
  - [X] Show diff between buffer and file on disk
- [x] Section operations
  - [x] <kbd>RET</kbd> (`magit-visit-thing`)
    - [ ] ~~on section~~
    - [X] on file (go to buffer)
    - [X] on hunk (go to line in buffer)
      - [X] on deleted line in hunk (go to line in original file on disk)
  - [x] <kbd>k</kbd> (`magit-delete-thing` / `magit-discard`)
    - [x] on section (revert/kill all)
    - [x] on file
      - [x] existing files (revert)
      - [x] new files (kill buffer)
    - [x] on hunk (using a temporary file)
  - [x] <kbd>a</kbd> (`magit-cherry-apply` / `magit-apply`)
    - [x] on section (save all)
    - [x] on file (save file)
    - [x] on hunk

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -2758,6 +2758,12 @@ It the SECTION has a different type, then do nothing."
 
 (add-hook 'magit-section-goto-successor-hook #'magit-hunk-goto-successor)
 
+(defvar magit-unsaved-section-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map [remap magit-delete-thing] 'magit-discard)
+    map)
+  "Keymap for the `unsaved' section.")
+
 (magit-define-section-jumper magit-jump-to-unsaved "Unsaved changes" unsaved)
 
 (defun magit--insert-unsaved-diff (buffer)
@@ -2850,12 +2856,12 @@ It the SECTION has a different type, then do nothing."
   "Return the diff type of SECTION.
 
 The returned type is one of the symbols `staged', `unstaged',
-`committed', or `undefined'.  This type serves a similar purpose
-as the general type common to all sections (which is stored in
-the `type' slot of the corresponding `magit-section' struct) but
-takes additional information into account.  When the SECTION
-isn't related to diffs and the buffer containing it also isn't
-a diff-only buffer, then return nil.
+`unsaved', `committed', or `undefined'.  This type serves a
+similar purpose as the general type common to all sections (which
+is stored in the `type' slot of the corresponding `magit-section'
+struct) but takes additional information into account.  When the
+SECTION isn't related to diffs and the buffer containing it also
+isn't a diff-only buffer, then return nil.
 
 Currently the type can also be one of `tracked' and `untracked'
 but these values are not handled explicitly everywhere they
@@ -2887,7 +2893,7 @@ Do not confuse this with `magit-diff-scope' (which see)."
                    (t 'committed))))
           ((derived-mode-p 'magit-status-mode)
            (let ((stype (oref it type)))
-             (if (memq stype '(staged unstaged tracked untracked))
+             (if (memq stype '(staged unstaged unsaved tracked untracked))
                  stype
                (pcase stype
                  ((or `file `module)
@@ -2949,7 +2955,7 @@ actually a `diff' but a `diffstat' section."
         (`(file  ,_  ,_  ,_) 'file)
         (`(module   t   t nil) 'files)
         (`(module  ,_  ,_  ,_) 'file)
-        (`(,(or `staged `unstaged `untracked)
+        (`(,(or `staged `unstaged `unsaved `untracked)
            nil ,_ ,_) 'list)))))
 
 (defun magit-diff-use-hunk-region-p ()
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -2761,6 +2761,7 @@ It the SECTION has a different type, then do nothing."
 (defvar magit-unsaved-section-map
   (let ((map (make-sparse-keymap)))
     (define-key map [remap magit-delete-thing] 'magit-discard)
+    (define-key map [remap magit-cherry-apply] 'magit-apply)
     map)
   "Keymap for the `unsaved' section.")
 
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -2750,30 +2750,45 @@ It the SECTION has a different type, then do nothing."
 
 (magit-define-section-jumper magit-jump-to-unsaved "Unsaved changes" unsaved)
 
+(defun magit--insert-unsaved-diff (buffer)
+  "Insert diff for unsaved changes in BUFFER."
+  (let* ((file-name     (buffer-file-name buffer))
+         (relative-name (magit-file-relative-name file-name))
+         (args          (-flatten (list "diff" magit-buffer-diff-args
+                                        "--no-index" "--no-prefix" "--"
+                                        relative-name "-")))
+         (beg           (point)))
+    (apply #'magit--git-insert-with-input buffer args)
+    (when (< beg (point))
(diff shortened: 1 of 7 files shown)


lisp/magit-git.el
```diff
@@ -440,15 +440,19 @@ signal `magit-invalid-git-boolean'."
   "Execute Git with ARGS, inserting its output at point.
 If Git exits with a non-zero exit status, then show a message and
 add a section in the respective process buffer."
+  (apply #'magit--git-insert-with-input nil args))
+
+(defun magit--git-insert-with-input (input &rest args)
```
CyberShadow, line 445: Here and elsewhere, the `&rest args` have posed an obstacle to extending the process functions.

Some alternative approaches:

- Pass the input via a global :nauseated_face: 
- If the first item in `args` is a buffer or `nil`, assume it's the optional `input` argument
- Introduce `fun-1` variants which don't use `&rest`, and thus are extensible
- Just change the function signature? It would be a breaking change but I don't know what impact that would have in practice.

CyberShadow:
I believe this is ready for initial review.

Outstanding points:

- The `&rest args` issue above.
- I made all new function declarations private (`magit--`). Perhaps some of these could be public and documented in the manual.
- One detail in the logic for finding buffers to diff depends on the resolution of #4292, which lies in code with a lot of history that I don't have the experienced to tackle.
- I'm not sure if `magit--process-input` needs to do the same `encode-coding-region` call as `magit-run-git-with-input`. For one, we are now piping in data to diff, as opposed to diffs to apply, but of course that's not reflected in the process creation API, so maybe that call needs to be higher up? For two, `magit-run-git-with-input` assumed that its input buffer was temporary and thus was safe to modify, but that doesn't apply to `magit--process-input`.

lisp/magit-git.el
CyberShadow: > Here and elsewhere

Oops, actually, just here (if we don't count `magit--git-wash-fun`). There were more, but I refactored them away.

tarsius:
I have applied the first tree commits after regenerating `magit.texi`, so you'll have to rebase.

CyberShadow:
Thanks, done!

tarsius:
I was not expecting something like this and while I appreciate the careful work that went into this I am undecided whether to merge this and what parts I would want to be changed. (But there's hope.)

For now I think the first commit should implement a much simpler version of `magit-insert-unsaved-changes` that just lists the unsaved buffers (using the file-name but without a diff). `RET` should visit the buffer. But that's about it. I would then cherry-pick that. I don't have the time to have a discussion about the rest right now.

CyberShadow:
> For now I think the first commit should implement a much simpler version of `magit-insert-unsaved-changes` that just lists the unsaved buffers (using the file-name but without a diff). `RET` should visit the buffer. But that's about it.

Okay, done :)

lisp/magit-diff.el
```diff
@@ -2720,6 +2720,33 @@ It the SECTION has a different type, then do nothing."
 
 (add-hook 'magit-section-goto-successor-hook #'magit-hunk-goto-successor)
 
+(magit-define-section-jumper magit-jump-to-unsaved "Unsaved changes" unsaved)
+
+(defun magit-insert-unsaved-changes ()
+  "Insert section showing unsaved changes in buffers."
```
tarsius, line 2763: Not yet. Please adjust.

lisp/magit-diff.el
```diff
@@ -2720,6 +2720,33 @@ It the SECTION has a different type, then do nothing."
...
+  (when-let ((topdir (magit-toplevel)))
```
tarsius: Why `when-let`? Once we get here it is well established that we are inside a git repository.

tarsius (changes requested):

lisp/magit-diff.el
CyberShadow: Alrighty, done (I used "a list of unsaved buffers")

lisp/magit-diff.el
CyberShadow: You're right, done

CyberShadow:
> I don't have the time to have a discussion about the rest right now.

No problem. Thank you.

> magit-files: Implement {orig-worktree} pseudo-revision

I think I would like to revisit this before this is merged, as this concerns slightly more than the idea of unsaved buffers. Thinking about this in terms of user story, I see two distinct intentions: 1) go to a buffer (at this point in the diff) which contains (exactly) this side of the diff that the user can view (and copy code out of), and 2) go as closely as possible to this point in the diff in a buffer that the user can *edit*. So, <kbd>RET</kbd> in a diff would do the former, but <kbd>C-RET</kbd> would always do the latter.

My initial thought was that creating an additional buffer should probably be avoided if possible, as it's another buffer that the user has to manage (or at least kill), so I was thinking that `{worktree}` should just do what `{orig-worktree}` does when a buffer exists visiting that file and it has modified. However, that would mess with muscle memory, and the user would now need to stop and think before killing the buffer if that's a buffer they want to keep or kill. (Even if the buffer is unmodified, killing it would lose the undo history, for example.) In any case, instead of `{orig-worktree}`, there should perhaps be `{worktree-or-buffer}` OSLT (used by <kbd>C-RET</kbd>), which always tries to create or find a suitable buffer and point within it, even if it has been modified.

tarsius:
IMO this implements some sort of secondary index. It is like Git's index but obviously much less mature. I think this is a mistake as it adds a lot of complexity and yet to be discovered bugs/missing features that I would have to fix/implement over time.

I am sorry but I have to reject this.

CyberShadow:
> IMO this implements some sort of secondary index. It is like Git's index but obviously much less mature.

I think this is not an accurate description of the functionality added in this pull request. It does represent the state in buffers as an additional layer above the working tree; if anything, the working tree is now more like an index than the buffers. The index is two layers down from Emacs buffers.

> I think this is a mistake as it adds a lot of complexity and yet to be discovered bugs/missing features that I would have to fix/implement over time.

I would have been happy to address those.

> I am sorry but I have to reject this.

Okay.

However, I am personally very happy with the functionality added here, and so would like to continue using Magit with this functionality, which means I will be maintaining my own fork of Magit.

I would like to make this fork of Magit public. Do you have any advice on the most proper way to do so? Or, do you know of an existing fork of Magit that I can contribute this improvement to instead?

I may be interested in helping maintain other improvements to Magit that you have rejected in the past. Any suggestions?

tarsius:
> I would like to make this fork of Magit public. Do you have any advice on the most proper way to do so? Or, do you know of an existing fork of Magit that I can contribute this improvement to instead?

Well it would be best if it were an *extension* instead of a *fork*, but that probably won't work here. In some cases you might be able to add a *new* command and add it in the appropriate new section specific keymap, but I doubt that would work for all or at least nearly all of what you are adding here. So that's probably not an option.

So a fork it is. Please clearly document the differences and tell users to ask you for help first. Other than that I cannot think of any advice or requests right now.

> I may be interested in helping maintain other improvements to Magit that you have rejected in the past. Any suggestions?

I have tried using the `wontadd` label when I reject a contribution or suggestion.

tarsius:
Don't try to add it to Melpa under a different name. That would only be accepted if you changed all symbol names and that would make it extremely frustrating to compare your fork with the canonical version.

If you want to make it installable using `package.el`, then you should provide your own elpa archive, which users can add in a way that shadows Melpa.

CyberShadow:
Great, thank you!

Issue #4295, opened 2021-01-14
File names are too short in diff buffer


LeandroCardoso:
The output of _diff --stat_ in the begging of diff buffer are truncated to 78 characters making the visualization of long file names and/or deep directory structure difficult.
It would be better if the output of _diff --stat_ was set to _(window-width)_. _--stat-width_ can be used to this purpose.
I tried to setup this myself using the _magit-diff-default-arguments_, but it doesn't work for me. Anyway, the long lines output should be the default option IMHO.

LeandroCardoso:
I sent the patch #4297 to address this issue.

PR #4297, opened 2021-01-17
Set the diffstat width to window's width with new customization

LeandroCardoso:
Before this patch the filename + graph lines of diffstat are limited to 78 characters by
git (according with the git documentation this is 80 characters, but the limit is actually 78 in my
tests). This causes part of the directory structure and/or filename to be abbreviated to 78
characters even if the buffer has enough space to display the full directory structure and filename.

This patch set the default width of diffstat to the same as the window's width and add 3 new custom
variables to customize this behavior in the same way as available when using git in a terminal.

This patch affects the magit-diff, magit-revision and the magit-log buffers.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -2024,14 +2024,16 @@ Staging and applying changes is documented in info node
   ;; Don't add any of width arguments unless --stat is present - adding
   ;; any of these arguments implies a --stat.
   (if (member "--stat" args)
-      (let ((diffstat-width (if (functionp magit-diffstat-width)
-                                (funcall magit-diffstat-width)
-                              magit-diffstat-width)))
-        (when-let ((graph-width (nth 2 diffstat-width)))
-          (push (format "--stat-graph-width=%d" graph-width) args))
-        (when-let ((name-width (nth 1 diffstat-width)))
-          (push (format "--stat-name-width=%d" name-width) args))
-        (push (format "--stat-width=%d" (car diffstat-width)) args))
+      (if-let ((window (get-buffer-window (current-buffer) 'visible)))
+          (with-selected-window window
+            (let ((diffstat-width (if (functionp magit-diffstat-width)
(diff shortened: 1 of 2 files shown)


tarsius:
See #4295 for discussion.

tarsius:
I don't think we should have only one option. Allowing an arbitrary functions seems like a good idea, it should return a *list* of one or more of these arguments. We should offer an option that uses `window-width` out of the box.

LeandroCardoso:
I agree with your idea of only one option, so I replaced the three options by one that accepts a list or a function that returns a list with all the three options.

tarsius:
The outer function should be named `magit-diffstat-maybe-add-width-arguments`.

And I think the inner function should be expected to return a list of one or more arguments (strings) instead of integers/nil. The allowed arguments should be mentioned in the option's doc-string. IMO that simplifies things.

LeandroCardoso:
All the suggestions were implemented.

We now have a more flexible approach and just realized that --compact-summary can also be added to the custom variable. I added the --compact-summary to the doc-string, but I didn't do any other change related to it.

Do you think it is a good idea to improve the support to the --compact-summary argument? We can add some font-lock-face to it in the same way magit does in the graph part - e.g. making the "(new)" green and the "(gone)" red. We can also provide a second function as an option to the magit-diffstat-additional-arguments with the compact-summary. Like:

(defun magit-diffstat-use-window-width-and-compact-summary ()
  "Use window width as the diffstat width."
  (list (format "--stat-width=%d" (window-width)) "--compact-summary"))

tarsius:
I have pushed a somewhat simpler implementation.

> Do you think it is a good idea to improve the support to the --compact-summary argument?

No. In a stand-alone diff it is kinda redundant because if one collapses the diff that follows already presents this information in a readable fashion. And logs are already slow enough. IMO allowing the use of `--compact-summary` is good enough.

PR #4298, opened 2021-01-18
magit-rebase-remove-commit: override core.commentChar

nacl:
When "magit-rebase-remove-commit" (r k) is called, it replaces the line with
"noop" followed by a commented-out version of the original "pick" command on a
separate line. However, the comment character is hardcoded to "#", which does
not work as expected when the "core.commentChar" option is set to something
other than "#" in .gitconfig files.

This change overrides the value of "core.commentChar" when
"magit-rebase-interactive-1" is called within "magit-rebase-remove-commit", so
that the rebase operation can properly delete the commit, and further ensures
that it set when "magit-log-select" called in "magit-rebase-interactive-1".

(Formerly):

# ~magit-rebase--perl-editor: add awareness of core.commentChar~

~When "magit-rebase-remove-commit" (r k) is called, it replaces the line with~
~"noop" followed by a commented-out version of the original "pick" command on a~
~separate line.  However, the comment character is hardcoded to "#", which does~
~not work as expected when the "core.commentChar" option is set to something~
~other than "#".~

~This change replaces the hardcoded comment character in~
~"magit-rebase--perl-editor" with a format string aware of "core.commentChar",~
~using functionality similar to that found in other parts of magit.~

### DIFF
--- a/lisp/magit-sequence.el
+++ b/lisp/magit-sequence.el
@@ -669,8 +669,13 @@ START has to be selected from a list of recent commits."
                                  (unless (member "--root" args) commit)))
     (magit-log-select
       `(lambda (commit)
-         (magit-rebase-interactive-1 commit (list ,@args)
-           ,message ,editor ,delay-edit-confirm ,noassert))
+         ;; In some cases (currently just magit-rebase-remove-commit), "-c
+         ;; commentChar=#" is added to the global arguments for git.  Ensure
+         ;; that the same happens when we chose the commit via
+         ;; magit-log-select, below.
+         (let ((magit-git-global-arguments (list ,@magit-git-global-arguments)))
+           (magit-rebase-interactive-1 commit (list ,@args)
+             ,message ,editor ,delay-edit-confirm ,noassert)))
       message)))
 
 (defvar magit--rebase-published-symbol nil)
@@ -755,10 +760,14 @@ START has to be selected from a list of recent commits."
   "Remove a single older commit using rebase."
   (interactive (list (magit-commit-at-point)
                      (magit-rebase-arguments)))
-  (magit-rebase-interactive-1 commit args
-    "Type %p on a commit to remove it,"
-    (apply-partially #'magit-rebase--perl-editor 'remove)
-    nil nil t))
+  ;; magit-rebase--perl-editor assumes that the comment character is "#".
+  (let ((magit-git-global-arguments
+         (nconc (list "-c" "core.commentChar=#")
+                magit-git-global-arguments)))
+    (magit-rebase-interactive-1 commit args
+      "Type %p on a commit to remove it,"
+      (apply-partially #'magit-rebase--perl-editor 'remove)
+      nil nil t)))
 
 (defun magit-rebase--perl-editor (action since)
   (let ((commit (magit-rev-abbrev (magit-rebase--target-commit since))))


lisp/magit-sequence.el
```diff
@@ -767,7 +767,7 @@ START has to be selected from a list of recent commits."
             commit
             (cl-case action
               (edit   "edit")
-              (remove "noop\n# pick")
+              (remove (format "noop\n%s pick" (or (magit-get "core.commentChar") "#")))
```
kyleam, line 770: Thanks for spotting this issue and for the PR.  There are some cases that this fix can't handle: anything makes the perl snippet constructed by `magit-rebase--perl-editor` invalid (e.g., "$") and the special case of "auto".  For the latter case, I think it's likely that spots in magit already don't play well with "auto", but it'd still be good to not add another.

All `magit-rebase-interactive-1` call sites in the code base only use a non-interactive command for `editor`, so I wonder if instead we should just override `core.commentchar` in these cases.  What about something like this?

```diff
diff --git a/lisp/magit-sequence.el b/lisp/magit-sequence.el
index 6ca0e9de..d4339222 100644
--- a/lisp/magit-sequence.el
+++ b/lisp/magit-sequence.el
@@ -658,7 +658,11 @@ (defun magit-rebase-interactive-1
                   commit delay-edit-confirm
                   (--some (string-prefix-p "--rebase-merges" it) args))))
   (if (and commit (not confirm))
-      (let ((process-environment process-environment))
+      (let ((magit-git-global-arguments
+             (nconc (and editor
+                         (list "-c" "core.commentchar=#"))
+                    magit-git-global-arguments))
+            (process-environment process-environment))
         (when editor
           (push (concat "GIT_SEQUENCE_EDITOR="
                         (if (functionp editor)
```

nacl: Good points.  Thanks for taking a look!

I like the idea, however when I tried `magit-rebase-reword-commit` (`r w`), the syntax highlighting for comments in the resulting commit message buffer were no longer font-lock'd as comments (`comment-start` is unconditionally overridden in `git-commit-setup-font-lock` in `git-commit.el`).  At that point, it's not clear what are comments, and users could easily inadvertently lose data in their commit messages when they finish rebasing.

Some options that come to mind off the top of my head:

1) Make my original regexp formatting more aware of Perl syntax (breaks `auto`).
2) Wrap the call to `magit-rebase-interactive-1` in `magit-rebase-remove-commit` in a `let` expression that adds `-c core.commentChar=#"` to `magit-git-global-arguments` .

WDYT? I like 2 the best out of these.

kyleam: > I like the idea, however when I tried `magit-rebase-reword-commit` (`r w`) [...]

Bah, true, I overlooked that command when thinking about it.

> WDYT? I like 2 the best out of these.

Yeah, I like the idea of let-binding in `magit-rebase-remove-commit` better than trying to ensure we produce valid perl.  And it's conceptually cleaner and more targeted than my misguided proposal of doing it in `magit-rebase-interactive-1` based on `editor`.  The complication with it is that the binding won't make it through when `magit-log-select` calls `magit-rebase-interactive-1`.  One way to deal with that would be to modify `magit-log-select` lambda to splice the `magit-git-global-arguments` value into another let-binding within the lambda.

nacl: Cool.  I did this in 2dd9220, and it seems that the entire `magit-rebase-remove-commit` workflow works properly.  This is much cleaner too.  Thanks for the suggestions!

I didn't need to use more than one `let`-binding in the `magit-log-select` `lambda` in `magit-rebase-interactive-1`, though.  Please take a look.

kyleam: > I didn't need to use more than one `let`-binding in the `magit-log-select` `lambda` in `magit-rebase-interactive-1`, though

Yeah, by "another let-binding within the lambda" I just meant one in addition to the let-binding in `magit-rebase-remove-commit`, so you landed on what I had in mind.

Please squash the fixup commit into the initial one.  Aside from that, this looks good to me, and works on all of the cases I've tested.  I'll wait another day or two for others to give feedback before merging.

Thanks again.

lisp/magit-sequence.el
```diff
@@ -667,11 +667,16 @@ START has to be selected from a list of recent commits."
                 process-environment))
         (magit-run-git-sequencer "rebase" "-i" args
                                  (unless (member "--root" args) commit)))
-    (magit-log-select
-      `(lambda (commit)
-         (magit-rebase-interactive-1 commit (list ,@args)
-           ,message ,editor ,delay-edit-confirm ,noassert))
-      message)))
+      (magit-log-select
```
tarsius, line 670: The old indentation was correct.

tarsius (changes requested):

lisp/magit-sequence.el
nacl: Fixed; thanks for pointing this out.

PR #4299, opened 2021-01-18
magit-process: Respect process decoding user-setting

qwerbzuio:
Otherwise, on Windows an issue shows up with processes that output
CRLF-endings (e.g. python): the process-output is deleted due to the
subsequent code (which tries to improve output of counters).

---

On windows I encounter an issue that I don't see any output from processes (namely python), when executed via magit's "Run shell command". Python is supposed to output `\r\n` as line ending on windows (see e.g. https://bugs.python.org/issue13119).

I wrote a patch that fixes the issue, while still properly showing counting processes (which only output `\r` without `\n`, in order to erase the current line).

However, I'm not sure if this is the right place to fix the issue, or if it should be rather fixed outside magit (I have set `default-process-coding-system` to `(utf-8-dos . utf-8-unix)` if that matters).

Could you please have a look and tell me if this is the right way to go?

I'm sorry if this is somehow a mix between a merge request and an issue description; as I already had a possible fix available, I wrote it up like this. If requested I can of course transfer it to an issue.

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -577,7 +577,7 @@ Magit status buffer."
     (when (eq system-type 'windows-nt)
       ;; On w32, git expects UTF-8 encoded input, ignore any user
       ;; configuration telling us otherwise.
-      (set-process-coding-system process 'utf-8-unix))
+      (set-process-coding-system process nil 'utf-8-unix))
     (process-put process 'section section)
     (process-put process 'command-buf (current-buffer))
     (process-put process 'default-dir default-directory)


tarsius:
@npostavs please have a look.

npostavs:
> However, I'm not sure if this is the right place to fix the issue, or if it should be rather fixed outside magit (I have set `default-process-coding-system` to `(utf-8-dos . utf-8-unix)` if that matters).
> 
> Could you please have a look and tell me if this is the right way to go?

I think the correct fix would be to narrow the scope of our `magit--process-coding-system` choice, so that it only affects `git` processes.

It should work to add a `process-name` argument to `magit--process-coding-system` and have it just return `default-process-coding-system` if the argument doesn't match `"git"`. Alternatively, let-bind `process-coding-system-alist` with a `("git" .  ,(magit--process-coding-system))` entry rather than binding `default-process-coding-system`, but let-binding an extra alist entry is a bit awkward compared to just binding a single value.

----

This bit in `magit-start-process` should be narrowed as well probably (though it practically doesn't matter at the moment since only calls using git would come with input):

```elisp
    (when (eq system-type 'windows-nt)
      ;; On w32, git expects UTF-8 encoded input, ignore any user
      ;; configuration telling us otherwise.
      (set-process-coding-system process 'utf-8-unix))
```

qwerbzuio (commented):
Hi @npostavs, thanks for your feedback.

> This bit in `magit-start-process` should be narrowed as well probably (though it practically doesn't matter at the moment since only calls using git would come with input):
> 
> ```emacs-lisp
>     (when (eq system-type 'windows-nt)
>       ;; On w32, git expects UTF-8 encoded input, ignore any user
>       ;; configuration telling us otherwise.
>       (set-process-coding-system process 'utf-8-unix))
> ```

I wonder if there's a mistake in this code; it actually hardcodes the decoding-system for all processes to `utf-8-unix`, while it should probably set the encoding-system instead (at least that's what I understand from looking at 2bb9f281fcf899b6a1eb51cc8945dabd41c44edf). 

Shouldn't it be like this?:

```emacs-lisp
    (when (eq system-type 'windows-nt)
      ;; On w32, git expects UTF-8 encoded input, ignore any user
      ;; configuration telling us otherwise.
      (set-process-coding-system process nil 'utf-8-unix))
```

With this change, my issue actually vanishes. I'll push the according changes to my branch.

qwerbzuio:
Note: to reproduce the issue on Windows, you can e.g. `magit-run` the command `python -c "print('foo\nbar\n')"`; output will be empty.

qwerbzuio:
Hi @kyleam I cleaned up the pull request and reduced it to what I think should be the fix for the issue. I'd be glad to get your feedback on the change once you find the time. Thanks!

kyleam:
I think I may have been pinged by accident, but fwiw...

> I wonder if there's a mistake in this code; it actually hardcodes the decoding-system for all processes to `utf-8-unix`, while it should probably set the encoding-system instead (at least that's what I understand from looking at [2bb9f28](https://github.com/magit/magit/commit/2bb9f281fcf899b6a1eb51cc8945dabd41c44edf)).

... your assessment looks correct to me.  

It'd be good to amend d3ade15c to include an explanation in the commit message.  Perhaps something like

```
magit-start-process: Fix setting of input encoding on Windows

As described in gh-3250, Windows git requires UTF-8 encoded input.
However, magit-start-process mistakenly calls
set-process-coding-system with utf-8-unix as the second argument,
specifying the output decoding.  Instead pass utf-8-unix as the third
argument to specify the input encoding.
```

qwerbzuio:
> I think I may have been pinged by accident, but fwiw...

You're right. Sorry for dragging you into this - not sure what went on in my mind. Thanks for having a look anyway.

> > I wonder if there's a mistake in this code; it actually hardcodes the decoding-system for all processes to `utf-8-unix`, while it should probably set the encoding-system instead (at least that's what I understand from looking at [2bb9f28](https://github.com/magit/magit/commit/2bb9f281fcf899b6a1eb51cc8945dabd41c44edf)).
> 
> ... your assessment looks correct to me.
> 
> It'd be good to amend [d3ade15](https://github.com/magit/magit/commit/d3ade15c724aee510fa06a82a5be75a3c4db59c3) to include an explanation in the commit message. Perhaps something like
> [...]

Thanks for your suggestion. I amended the commit.

PR #4302, opened 2021-01-20
Fixing GPG key suggestions

samwhitlock:
Prior: would suggest the first key for each keyring

Current: only suggests any subkey that is capable of signing

-----

A common setup for GPG is to have the primary (root) key only have certification capabilities.
The primary key is then used to certify subkeys which typically do the actual functions.
This is common because it allows the primary key to not be in the gpg keychain; it is left stored away in safe-keeping to expire (renew) keys, create new keys, etc.

The previous behavior would suggest the key ID of the primary key, _even if it didn't have signing capabilities_.

This new behavior will suggest all subkey combinations of each keychain that contain a sign-capable private key, including

* a user having more than one signing key per keychain (unwise, but possible)
* a user having more than one key in their keychain (possible e.g. to sign for work vs. personal commits)

Please do suggest `cl-loop` or formatting fixes. I tried my best to format it nicely and use what I could figure out about the `cl-loop` construct, but if there is a better way of writing this I'm happy to give it another go.

### DIFF
--- a/lisp/magit-commit.el
+++ b/lisp/magit-commit.el
@@ -161,7 +161,7 @@ Also see `git-commit-post-finish-hook'."
                                                  (epg-decode-dn id-obj))))
                               append (cl-loop for subkey in (epg-key-sub-key-list key)
                                               if (and (memq 'sign (epg-sub-key-capability subkey))
-                                                      (time-less-p (float-time nil) (epg-sub-key-expiration-time subkey)))
+                                                      (time-less-p (float-time) (epg-sub-key-expiration-time subkey)))
                                               collect (let ((subkey-id (epg-sub-key-id subkey)))
                                                         (propertize
                                                          subkey-id 'display (string-join (list subkey-id author) " "))))))
--- a/lisp/magit-commit.el
+++ b/lisp/magit-commit.el
@@ -160,7 +160,8 @@ Also see `git-commit-post-finish-hook'."
                                                    id-str
                                                  (epg-decode-dn id-obj))))
                               append (cl-loop for subkey in (epg-key-sub-key-list key)
-                                              if (memq 'sign (epg-sub-key-capability subkey))
+                                              if (and (memq 'sign (epg-sub-key-capability subkey))
+                                                      (> (epg-sub-key-expiration-time subkey) (time-convert nil 'integer)))
                                               collect (let ((subkey-id (epg-sub-key-id subkey)))
                                                         (propertize
                                                          subkey-id 'display (string-join (list subkey-id author) " "))))))
--- a/lisp/magit-commit.el
+++ b/lisp/magit-commit.el
@@ -161,7 +161,7 @@ Also see `git-commit-post-finish-hook'."
                                                  (epg-decode-dn id-obj))))
                               append (cl-loop for subkey in (epg-key-sub-key-list key)
                                               if (and (memq 'sign (epg-sub-key-capability subkey))
-                                                      (> (epg-sub-key-expiration-time subkey) (time-convert nil 'integer)))
+                                                      (time-less-p (float-time) (epg-sub-key-expiration-time subkey)))
                                               collect (let ((subkey-id (epg-sub-key-id subkey)))
                                                         (propertize
                                                          subkey-id 'display (string-join (list subkey-id author) " "))))))
--- a/lisp/magit-commit.el
+++ b/lisp/magit-commit.el
@@ -151,19 +151,20 @@ Also see `git-commit-post-finish-hook'."
 
 (defun magit-read-gpg-secret-key (prompt &optional initial-input history)
   (require 'epa)
-  (let* ((keys (mapcar
-                (lambda (obj)
-                  (let ((key (epg-sub-key-id (car (epg-key-sub-key-list obj))))
-                        (author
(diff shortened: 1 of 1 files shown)


lisp/magit-commit.el
```diff
@@ -151,18 +151,21 @@ Also see `git-commit-post-finish-hook'."
...
+  (let* ((all-keys (epg-list-keys (epg-make-context epa-protocol) nil t))
+         (by-subkeys (cl-loop for key in all-keys
+                              for author = (when-let ((id-obj (car (epg-key-user-id-list key))))
+                                             (let ((id-str (epg-user-id-string id-obj)))
+                                               (if (stringp id-str)
+                                                   id-str
+                                                 (epg-decode-dn id-obj))))
+                              append (cl-loop for subkey in (epg-key-sub-key-list key)
+                                              if (and (memq 'sign (epg-sub-key-capability subkey))
+                                                      (> (epg-sub-key-expiration-time subkey) (time-convert nil 'integer)))
```
basil-conto, line 164: Note that `time-convert` was added in Emacs 27.

samwhitlock: Ah I didn't realize that!

What is the canonical way to get UNIX time (in seconds) using `(current-time)`?

If there is no easy way to get this, I can just remove this check. It's not really necessary as the signing will fail with an expired key, but it helps filter out potentially expired keys.

basil-conto: > What is the canonical way to get UNIX time (in seconds) using `(current-time)`?

You mean [`float-time`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Time-of-Day.html#index-float_002dtime)?

I'm not familiar with `epg`, but most of the time you should leave all time comparisons up to [`time-less-p`, `time-subtract`, etc.](https://www.gnu.org/software/emacs/manual/html_node/elisp/Time-Calculations.html) rather than worrying about specific time representations in Emacs, since there are many of them.

samwhitlock: I just pushed another commit to reflect these suggestions, which uses `(float-time)` and `(time-less-p)`. Thank you for the advice!

lisp/magit-commit.el
```diff
@@ -151,18 +151,21 @@ Also see `git-commit-post-finish-hook'."
...
+  (let* ((all-keys (epg-list-keys (epg-make-context epa-protocol) nil t))
+         (by-subkeys (cl-loop for key in all-keys
+                              for author = (when-let ((id-obj (car (epg-key-user-id-list key))))
+                                             (let ((id-str (epg-user-id-string id-obj)))
+                                               (if (stringp id-str)
+                                                   id-str
+                                                 (epg-decode-dn id-obj))))
+                              append (cl-loop for subkey in (epg-key-sub-key-list key)
+                                              if (and (memq 'sign (epg-sub-key-capability subkey))
+                                                      (time-less-p (float-time) (epg-sub-key-expiration-time subkey)))
```
basil-conto, line 164: Since Emacs 25 (which is the minimum required version for Magit), passing `nil` stands for the "current time".

samwhitlock: Pushed another fix. Thanks so much for this help! I only have my local Emacs 27 to test, so I am not so aware about these backwards compat issues :(

basil-conto: > Pushed another fix. Thanks so much for this help! I only have my local Emacs 27 to test, so I am not so aware about these backwards compat issues

Using `nil` isn't needed for backward compatibility; it was just a tiny simplification suggestion.  The only reason I mentioned Emacs 25 is because Emacs' time API was less consistent in these things before then.

basil-conto: What I meant is writing `(time-less-p nil ...)` instead of `(time-less-p (float-time) ...)`.

samwhitlock: Ah yes I see now. I pushed another commit to change that comparison to your suggestion.

tarsius:
Sorry for not replying earlier.

> The previous behavior would suggest the key ID of the primary key, even if it didn't have signing capabilities.
> This new behavior will suggest all subkey combinations of each keychain that contain a sign-capable private key

The primary key is usually used to certify that it and the subkeys belong to a certain person. That doesn't change just because the primary key doesn't have the relevant capability, here signing.  Sure the signing subkey will be used to do the actual signing, but it is still better and customary to specify the primary key, not the particular subkey directly.  The proposed patch breaks that.

> * a user having more than one signing key per keychain (unwise, but possible)

Gpg uses the newest subkey with signing capability.  I don't think we should explicitly support using another subkey.  If the user really has a reason to have multiple signing subkeys in a single certificate, then they can already pick the desired subkey, just without the subkey fingerprints being offered as completion candidates.

Filtering certificates that don't have the signing capability (neither on the primary nor any of the subkeys) does make sense.  I'll do that.

Not sure yet about doing the same for expired keys.

tarsius:
See https://github.com/magit/magit/compare/read-signing?expand=1.

samwhitlock:
This is an excellent fix! I just tried it out and it works great.

One consideration: including the full fingerprint (as opposed to just the id) can eat up a lot of screen space on smaller laptops.
However, I do like the id highlighting in the fingerprint.

samwhitlock:
Superceded by @tarsius `read-signing` branch

PR #4303, opened 2021-01-22
Replace Travis CI with GitHub Actions

### LINKED ISSUE #4300: Migrate from Travis CI to GitHub Actions?
As the future of Travis CI [seems to be, at best, uncertain](https://news.ycombinator.com/item?id=25338983), it may make sense to migrate to another CI service. As magit is hosted on GitHub, GitHub Actions seem to be a good fit.

npostavs/emacs-travis can be replaced with [purcell/setup-emacs](https://github.com/purcell/setup-emacs).

A related task would be to also enable code coverage reports.

Today I migrated [Emacs-D-Mode](https://github.com/Emacs-D-Mode-Maintainers/Emacs-D-Mode) to GitHub Actions, and added support for GitHub Actions to [undercover.el](https://github.com/undercover-el/undercover.el) (which we use to [upload coverage reports to Coveralls](https://coveralls.io/github/Emacs-D-Mode-Maintainers/Emacs-D-Mode?branch=master)). So, I would be happy to submit a pull request implementing a similar setup for the magit repository.

CyberShadow:
Closes #4300.

Preview: https://github.com/CyberShadow/magit/actions

README.md badge preview: test (replaces Build Status)

GitHub Actions .yml files don't have an equivalent for Travis CI's `notifications` section. Instead, notification settings are configured per-user in their GitHub settings:
- https://docs.github.com/en/github/managing-subscriptions-and-notifications-on-github/configuring-notifications#github-actions-notification-options
- https://github.com/settings/notifications#notifications-github-actions-settings

The GitHub Actions equivalent of `allow_failures` is to configure branches targeted by PRs as [protected](https://docs.github.com/en/github/administering-a-repository/about-protected-branches), and then, mark all matrix combinations except those in `allow_failures` as "required". Because each matrix combination shows up as an individual entry, this would show the failure on the PR page, but won't block merging the PR.


tarsius:
So this always fails on ubuntu-16.04, right? Then I would just squash the update commit into the migration commit. It's not like there was a reason to stick to xenial anyway.

.github/workflows/test.yml
```diff
@@ -0,0 +1,39 @@
...
+        $CURL -O ${GHRAW}/magit/transient/master/lisp/transient.el
+        $CURL -O ${GHRAW}/magit/with-editor/master/with-editor.el
+        emacs -Q --batch -L . -f batch-byte-compile dash.el transient.el with-editor.el
+        emacs --version
+    - name: Test
+      run: |
+        git config --global user.name "A U Thor"
+        git config --global user.email a.u.thor@example.com
+        git tag 0
+        make lisp DASH_DIR=$PWD
```
npostavs, line 38: Perhaps it would make sense to put this in a separate step called `Compile`? (the `git config` stuff could go in `Install`)

npostavs:
> It's not like there was a reason to stick to xenial anyway.

Yeah, that was there just to make sure we used the same machine type for building Emacs as running the magit tests (i.e., a quirk of the Travis setup).

.github/workflows/test.yml
CyberShadow: Sure. Done.

CyberShadow:
> So this always fails on ubuntu-16.04, right? Then I would just squash the update commit into the migration commit. It's not like there was a reason to stick to xenial anyway.

Yes, just wanted to make that possibly-important change explicit. Done.

CyberShadow:
Added test coverage collection as a second commit.

Preview: https://coveralls.io/builds/36543993

This one requires set-up. After merging, a repository owner will need to go to https://coveralls.io/repos/new and flick the switch for the magit/magit repository.

tarsius:
> After merging, a repository owner will need to [...] flick the switch for the magit/magit repository.

Done.

Looks good to me. Is there anything else that needs to be done or discussed before I can merge?

CyberShadow:
> Is there anything else that needs to be done or discussed before I can merge?

Hmm, here are a few things that might be worth mentioning.

- Another service similar to Coveralls is CodeCov. I think it's more popular, though I haven't used it personally, so I can't compare. Coveralls seemed adequate so far. Undercover does support both.

- Undercover relies on the edebug specifications for macros to be correct to work. When the specifications are wrong, instrumented code crashes at runtime. Because it's common to test several previous Emacs versions, Undercover includes [its own small set of up-to-date edebug specifications](https://github.com/undercover-el/undercover.el/blob/0bc3583065e49647db47d8a595fec13cb517d12f/undercover.el#L115). I had to add one to get magit to run, and it's not impossible that more may need to be added in the future.

- Undercover [seems to have some issues instrumenting magit on Emacs 25](https://github.com/CyberShadow/magit/runs/1753610808?check_suite_focus=true#step:7:12). The only effect is reduced observed test coverage for that Emacs version, so I'm not sure it's worth the time to investigate it.

- Makefile changes OK?

tarsius:
>     * Another service similar to Coveralls is CodeCov. I think it's more popular, though I haven't used it personally, so I can't compare. Coveralls seemed adequate so far. Undercover does support both.
> 
>     * Undercover relies on the edebug specifications for macros to be correct to work. When the specifications are wrong, instrumented code crashes at runtime. Because it's common to test several previous Emacs versions, Undercover includes [its own small set of up-to-date edebug specifications](https://github.com/undercover-el/undercover.el/blob/0bc3583065e49647db47d8a595fec13cb517d12f/undercover.el#L115). I had to add one to get magit to run, and it's not impossible that more may need to be added in the future.
> 
>     * Undercover [seems to have some issues instrumenting magit on Emacs 25](https://github.com/CyberShadow/magit/runs/1753610808?check_suite_focus=true#step:7:12). The only effect is reduced observed test coverage for that Emacs version, so I'm not sure it's worth the time to investigate it.

I'm okay with all of the above.

>     * Coveralls will post commit statuses to this repository and possibly complain about coverage decreases. If the plan is to have it in fully stealth mode for now that may be undesirable.

Yeah, I don't want that for now. Please split the change in two local-only vs. overall. The latter we could still apply at some later time.

(Personally I always found it quite annoying when I contributed some minor fix (including typo fixes) to some project and that is marked has having failed testing because some algorithm being a bit random.)

>     * Makefile changes OK?

Hmmm. That's not gonna work locally:

```
	(load-file \"shut-up.el\")\
	(load-file \"undercover.el\")\
```

I'm gonna apply just the first commit for now.

tarsius:
Tests passed! :partying_face: 
Thanks!

I have removed the second commit from your branch, so that I could merge this pr. You'll have to create a new branch for that. Or the two commits to replace it, if you go down that road.

CyberShadow:
> Yeah, I don't want that for now.

Well, it's easily disabled on Coveralls' side. (See screenshot below.)

> Personally I always found it quite annoying when I contributed some minor fix (including typo fixes) to some project and that is marked has having failed testing because some algorithm being a bit random.

This particular nuisance is solved easily enough by configuring a threshold here:

image

> That's not gonna work locally:

What do you think about using Cask for managing dependencies locally and on CI?

tarsius:
> Well, it's easily disabled on Coveralls' side.
> This particular nuisance is solved easily enough by configuring a threshold here:

Okay then I am open to the idea again. Please open another pull-request.

> What do you think about using Cask for managing dependencies locally and on CI?

I don't like it. Yes, what Magit is doing is ugly but it works. Maybe one day I will sit down and think about a better alternative, but Cask ain't it.

Using Cask on CI may be acceptable, but doing locally I consider it a mistake. Compiling a package against one version of its dependency and then running it against another is fundamentally flawed and using Cask ensures that that frequently happens.

Since the current hack works works both locally and on CI I see no reason to add a second implementation. That would lead to two implementations having to be maintained.

CyberShadow:
> Compiling a package against one version of its dependency and then running it against another

I was not aware of this problem. In what circumstances does it manifest?

Is it something that perhaps could be fixed in Cask?

One thing that stood out to me about the current process is that by default it seems to assume that magit was installed from MELPA with package.el, whereas lately alternative package managers such as straight.el appear to have been gaining popularity.

> That's not gonna work locally:

Okay. Considering that the current process doesn't seem to have a precedent for development-only dependencies, how would you like to approach this? I can see a few options:

- Do the same thing as with dash / `DASH_DIR`
- Move the curl invocations from `test.yml` into the makefile (and add to `.gitignore`)
- Leave the code as-is and document that users must place (copy/symlink) `undercover.el` and `shut-up.el` into the current directory (and add to `.gitignore`)

tarsius:
> > Compiling a package against one version of its dependency and then running it against another
> 
> I was not aware of this problem. In what circumstances does it manifest?

Always.
When you compile a project with Cask, then it loads any dependencies from `<project>/.cask`.
When the user manually starts Emacs, then that loads dependencies from the `load-path`, which does not include `<project/.cask`.

tarsius:
> One thing that stood out to me about the current process is that by default it seems to assume that magit was installed from MELPA with package.el, whereas lately alternative package managers such as straight.el appear to have been gaining popularity.

Funny that you mention that. I am the author of one of these package managers, `borg`, which served as a major inspiration for `straight`.

The way Magit references its dependencies was defined to work well with such package managers! Dependencies are expected to be located at `../<dependency>[/<subdir]`, relative to where Magit is checked out / installed. Later I extended it so that it could also work with packages installed by `package.el`.

CyberShadow:
> Always.

I see what you mean.

Hmm, I don't think that is a solvable problem. As far as I can see, the only way to remove that redundancy / discrepancy would be to somehow use the user's package management setup but not their other customizations. So, it would require that:

1. All package managers support a mode in which they only load dependencies, but do nothing else
2. All Emacs users (at least interested in contributing to Magit) structure their Emacs init to support such a mode

So, it doesn't seem to me that this redundancy is avoidable. (But probably you have thought about this for much longer than I have.)

I also see a few points in its favor:

- Development-only dependencies. For example, there isn't really any reason for undercover.el to be loaded in an interactive graphical Emacs session. Undercover is in MELPA probably mainly because Cask and similar projects use MELPA as a package source.

- If I have a customized (or intentionally older) version of some package set up with my package manager of choice, I think I would prefer that `make test` in Magit's directory uses a private, clean copy instead of whatever garbage I have checked out. That would give me more peace of mind that I don't accidentally submit code which "works on my machine" but fails for everyone else.

So, I would say that I personally agree with Cask's philosophy.

> Funny that you mention that. I am the author of one of these package managers, `borg`, which served as a major inspiration for `straight`.

That is indeed very interesting :)

tarsius:
https://github.com/CyberShadow/magit/commit/efc74b924e1ae4179555304263e4672f6d4b516e is the current draft, right?

CyberShadow:
Yes. I'm not sure how to address the Makefile issue, so I was hoping you would provide some direction here:

> Considering that the current process doesn't seem to have a precedent for development-only dependencies, how would you like to approach this?

tarsius:
Please create a (wip) pull-request for that (after rebasing). That would make it easier for me to reason about it.

PR #4305, opened 2021-01-29
magit-update-index: make it work over TRAMP

### LINKED ISSUE #4286: Wrong options passed to `git update-index --cacheinfo`
1. ##### What behavior I expected
   After a `magit-ediff-dwim` session, I am asked to confirm whether the index should be updated according to my changes to the center ediff buffer (`<my-file>.~{index}~`) and whether to save the worktree file. I confirm both queries with `y` and expect the index as well as my worktree to be updated accordingly.
2. ##### What behavior I observed
   The index update does not occur because of an error:
```
129 git … update-index --cacheinfo 100644 <my-file>
error: option 'cacheinfo' expects <mode>,<sha1>,<path>
```
3. ##### How you can reproduce the issue.
   I did the whole process over TRAMP, so maybe that case messed up the options passed to `git update-index`.

    M-x magit-version RET
`Magit 20210105.1030, Git 2.18.4, Emacs 27.1.50, gnu/linux`

pritambaral:
Fix #4286 

1. `magit-update-index` calls `make-temp-file` to create a temporary file to store the edited buffer for the index.
2. [`make-temp-file`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Unique-File-Names.html#index-make_002dtemp_002dfile) create a file relative to the special variable [`temporary-file-directory`](index-temporary_002dfile_002ddirectory-1)
3. The special variable `temporary-file-directory` is not automatically set by Emacs/TRAMP, so it points to the local machine's TMPDIR/TMP/TEMP directory. Thus, `make-temp-file` creates a file on the local machine.
4. `magit-update-index` thus writes the new index contents to a local file.
5. `magit-update-index` calls `git hash-object ... -- /tmp/indexXXX` on the remote machine, but of course the remote machine does not have the index file `/tmp/indexXXX`.

Fortunately, TRAMP already has functions for transparently using remote & local files:
`(make-temp-file ...)` => `(make-nearby-temp-file ...)`
`("git" "hash-object" ... index)` => `("git" "hash-object" ... (file-local-name index))`

### DIFF
--- a/lisp/magit-files.el
+++ b/lisp/magit-files.el
@@ -203,21 +203,25 @@ is done using `magit-find-index-noselect'."
     (unless (equal magit-buffer-refname "{index}")
       (user-error "%s isn't visiting the index" file))
     (if (y-or-n-p (format "Update index with contents of %s" (buffer-name)))
-        (let ((index (make-temp-file "index"))
+        (let ((index (make-temp-name (magit-git-dir "magit-update-index-")))
               (buffer (current-buffer)))
           (when magit-wip-before-change-mode
             (magit-wip-commit-before-change (list file) " before un-/stage"))
-          (let ((coding-system-for-write buffer-file-coding-system))
-            (with-temp-file index
-              (insert-buffer-substring buffer)))
-          (magit-with-toplevel
-            (magit-call-git "update-index" "--cacheinfo"
-                            (substring (magit-git-string "ls-files" "-s" file)
-                                       0 6)
-                            (magit-git-string "hash-object" "-t" "blob" "-w"
-                                              (concat "--path=" file)
-                                              "--" index)
-                            file))
+          (unwind-protect
+              (progn
+                (let ((coding-system-for-write buffer-file-coding-system))
+                  (with-temp-file index
+                    (insert-buffer-substring buffer)))
+                (magit-with-toplevel
+                  (magit-call-git
+                   "update-index" "--cacheinfo"
+                   (substring (magit-git-string "ls-files" "-s" file)
+                              0 6)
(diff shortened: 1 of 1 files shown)


lisp/magit-files.el
```diff
@@ -216,7 +216,7 @@ is done using `magit-find-index-noselect'."
                                        0 6)
                             (magit-git-string "hash-object" "-t" "blob" "-w"
                                               (concat "--path=" file)
-                                              "--" index)
+                                              "--" (file-local-name index))
```
kyleam, line 219: Thanks for the analysis and patch.  `make-nearby-temp-file` and `file-local-name` weren't added until 26.1, but Magit supports 25.1.  Instead of `file-local-name`, `magit-convert-filename-for-git` would be appropriate.  Perhaps instead of `make-temp-file`, we can place the file under `.git/`, taking care to clean it up.  `magit-with-temp-index` provides an example of this.  (In the case of `magit-with-temp-index`, it is actually writing an index file, and the file given to `read-tree --index-output=` should be on the same filesystem.)

Also, as noted in the pull request template, please avoid creating the commit through GitHub.

kyleam (changes requested):

kyleam:
Thanks for the update.  Looks good to me and worked well when I tested `magit-ediff-show-stage` over TRAMP.

I pushed a commit on top with some minor tweaks.  The main change is to avoid using "index..." as the temp file name to make it clearer to readers that this isn't a git index file (especially since we're now working in .git/).

I'll plan to squash that commit in and merge tomorrow unless objections come in.

kyleam (approved):

kyleam:
@pritambaral Thanks again!

PR #4314, opened 2021-02-07
Collect test coverage using Undercover and Coveralls

CyberShadow:
Continuation of #4303.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,6 +28,8 @@ jobs:
         $CURL -O ${GHRAW}/magnars/dash.el/master/dash.el
         $CURL -O ${GHRAW}/magit/transient/master/lisp/transient.el
         $CURL -O ${GHRAW}/magit/with-editor/master/with-editor.el
+        $CURL -O ${GHRAW}/cask/shut-up/master/shut-up.el
+        $CURL -O ${GHRAW}/undercover-el/undercover.el/master/undercover.el
         emacs -Q --batch -L . -f batch-byte-compile dash.el transient.el with-editor.el
         emacs --version
         git config --global user.name "A U Thor"
@@ -39,3 +41,19 @@ jobs:
     - name: Test
       run: |
         make test DASH_DIR=$PWD
+    - name: Coverage
+      env:
+        COVERALLS_FLAG_NAME: Emacs ${{ matrix.emacs_version }}
+        COVERALLS_PARALLEL: 1
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        make coverage DASH_DIR=$PWD
+  finalize:
+    runs-on: ubuntu-20.04
+    if: always()
+    needs: test
+    steps:
+    - run: >
+        url=$(printf -- "https://coveralls.io/webhook?repo_name=%s&repo_token=%s" \
+              "$GITHUB_REPOSITORY" "${{ secrets.GITHUB_TOKEN }}")
+        curl "$url" -d "payload[build_num]=$GITHUB_RUN_NUMBER&payload[status]=done"
--- a/Makefile
+++ b/Makefile
@@ -122,6 +122,16 @@ emacs-Q: clean-lisp
 	(require 'magit)\
 	(global-set-key \"\\C-xg\" 'magit-status))"
 
+coverage:
+	@find -name '*.elc' -delete
+	@$(BATCH) --eval "(progn\
+	(load-file \"shut-up.el\")\
+	(load-file \"undercover.el\")\
+	(undercover \"lisp/*.el\")\
+        $$suppress_warnings\
+	(load-file \"t/magit-tests.el\")\
+	(ert-run-tests-batch-and-exit))"
+
 ## Clean #############################################################
 
 clean: clean-lisp clean-docs clean-archives


.github/workflows/test.yml
```diff
@@ -39,3 +41,16 @@ jobs:
...
+        COVERALLS_PARALLEL: 1
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        make coverage DASH_DIR=$PWD
+  finalize:
+    runs-on: ubuntu-20.04
+    if: always()
+    needs: test
+    steps:
+    - run: curl "https://coveralls.io/webhook?repo_name=$GITHUB_REPOSITORY&repo_token=${{ secrets.GITHUB_TOKEN }}" -d "payload[build_num]=$GITHUB_RUN_NUMBER&payload[status]=done"
```
tarsius, line 56: You don't use the GITHUB_TOKEN from above. Good, that seems unnecessary anyway.

Can this be spread across multiple lines so that each one isn't longer than 79 characters, or at least no longer than ~95.

tarsius: Also, what does this do?

CyberShadow: > You don't use the GITHUB_TOKEN from above. Good, that seems unnecessary anyway.

If you mean the environment variable, it is not visible in this scope here. Above, it is necessary to allow Undercover to submit the results for that job (Emacs version) to Coveralls.

> Can this be spread across multiple lines so that each one isn't longer than 79 characters, or at least no longer than ~95?

I'll look into it.

> Also, what does this do?

This sends a final ping to Coveralls that tells it that all coverage reports from all parallel jobs in this build have been submitted, and no more will be submitted, so it can go ahead and finalize and generate the report.

CyberShadow: > Can this be spread across multiple lines so that each one isn't longer than 79 characters, or at least no longer than ~95?

Seems to be working.

tarsius:
I am closing this because (a) I don't see much of a benefit (b) it would be quite a bit of work for me to add it.

For example, it looks like I would have to update the make target myself. That would mean installing `undercover` and its dependency `shut-up` locally. Compiling the latter results in warnings and I don't allow that in packages that I keep installed, so I would have to fix that...

Alternatively I could help you implement everything but I feel that might be more work than just doing it myself.

Elsewhere you said:

> Well, it's difficult to improve what can't be measured.

I disagree. One can add a test and then know that there now are more tests than before without measuring it. I do not buy into the quantitative research fetish.

CyberShadow:
> Compiling the latter results in warnings and I don't allow that in packages that I keep installed, so I would have to fix that...

I don't understand why you chose to say that in the same message as closing the pull request. I would have been happy to work on that if you had merely pointed out the issues that need solving, though I am now less inclined to do so.

> Elsewhere you said:
> 
> > Well, it's difficult to improve what can't be measured.
> 
> I disagree. One can add a test and then know that there now are more tests than before without measuring it.

Well, this is just not true for very obvious reasons.

- The test may be completely redundant, testing code that is already tested elsewhere.
- The test may not actually be executing the code that was intended to be tested. In my experience, this happens more often than one may think.
- Changes in code may make some tests ineffective (and pass vacuously). Tracking regressions in test coverage is thus helpful.

More importantly in this particular case, coverage is extremely helpful in writing a test suite for a project which lacks one, as it is the easiest way to find code that is not tested.

> I do not buy into the quantitative research fetish.

My experience (with test coverage, not "quantitative research") shows otherwise. A sufficiently high test coverage allows the project to be almost defined by tests, which allows major refactoring or overhauls to be done quickly and frequently, without having to expend great care to avoid breaking this or that edge case. This is especially true for weakly typed languages like Emacs Lisp. (But, as the project's creator and maintainer, I respect that you of course have the ultimate say, especially on such managerial topics.)

tarsius:
I apologize for just shutting this down.

There certainly is value in this but currently this looks like this is going to add more work than it is going to save. Of course in the long run it may safe a lot of work, due to fewer regressions. Or not, if this doesn't actually result in people suddenly starting to contribute tests.

I have so much to do and I will never be done and that stresses me out. Sometimes I deal with that by delaying a decision or then by rushing a decision without thinking everything through again.

I am not against adding test coverage per se (though obviously I am not exactly excited about it), but I doubt it will save me much time in the short- to midterm, if any, so I would rather work on things that do. If you improve this to the point where I can run `make coverage` and it just works with my setup, then that would be great.

And you seem to imply that you would have been willing to do that, and now that I have shut it down you are less willing. That's understandable. On the other hand just getting comfortable saying "no" is important for my well being, so maybe I should try not feel bad about it. Maintaining Magit and a few dozen other packages is just too much work to always give every helpful contribution the attention and fair chance it deserves. Sometimes I have to go purely with my instincts or even just momentary unwillingness to put in the work.

tarsius:
Actually I would like test coverage, but I really really don't want to have anything to do with the implementation. Not even with the "configure coverall so that it doesn't block merges" part. I mean I would be the one who would have to click on things on their webpage, but I would like to do that from a screenshot that would show me exactly what I had to do without having to read any documentation whatsoever.

That's a lot to ask. That's why I haven't done so.

But that were basically the choices I saw. Reject this altogether or make the rather unreasonable request to not have to think about it at all. It just didn't seem realistic to essentially expect you to read my mind, so I went with the "easy" option.

CyberShadow:
Understood.

> If you improve this to the point where I can run make coverage and it just works with my setup, then that would be great.

I would like to say that I may do so in the future, and submit it along with an initial batch of tests which raise the coverage (as I now understand to be an important part of this), but unfortunately at the moment I cannot commit to such a promise. Perhaps if/when something requires me to spend time on Magit / Emacs Lisp again.

Thank you very much on all your work on magit, it is very highly appreciated.

PR #4330, opened 2021-02-22
magit-branch/magit-checkout: add recurse-submodules option

insi-eb:
# Change description and rationale

When using submodules, having to switch the submodules manually after each (branch and) checkout operation is quite cumbersome, even with the nice "o u" keyboard shortcuts.

This PR adds the git option "--recurse-submodules" to the magit-branch (sub-) commands that checkout branches, visibility controlled by transient level.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -8,7 +8,7 @@
 #+TEXINFO_DIR_CATEGORY: Emacs
 #+TEXINFO_DIR_TITLE: Magit: (magit).
 #+TEXINFO_DIR_DESC: Using Git from Emacs with Magit.
-#+SUBTITLE: for version 2.90.1 (v2.90.1-1171-gb4cf9fa9c+1)
+#+SUBTITLE: for version 2.90.1 (v2.90.1-1193-g49fbb7c2d+1)
 
 #+TEXINFO_DEFFN: t
 #+OPTIONS: H:4 num:3 toc:2
@@ -25,7 +25,7 @@ directly from within Emacs.  While many fine Git clients exist, only
 Magit and Git itself deserve to be called porcelains.
 
 #+TEXINFO: @noindent
-This manual is for Magit version 2.90.1 (v2.90.1-1171-gb4cf9fa9c+1).
+This manual is for Magit version 2.90.1 (v2.90.1-1193-g49fbb7c2d+1).
 
 #+BEGIN_QUOTE
 Copyright (C) 2015-2021 Jonas Bernoulli <jonas@bernoul.li>
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -4749,8 +4749,8 @@ command.
 - User Option: magit-branch-direct-configure
 
   This option controls whether the transient command ~magit-branch~ can
-  be used directly change the values Git variables.  This defaults to
-  ~t~ (to avoid changing key bindings).  When set to ~nil~, then no
+  be used to directly change the values of Git variables.  This defaults
+  to ~t~ (to avoid changing key bindings).  When set to ~nil~, then no
   variables are displayed by that transient command, and its suffix
   command ~magit-branch-configure~ has to be used instead to view and
   change branch related variables.
--- a/lisp/magit-branch.el
+++ b/lisp/magit-branch.el
@@ -201,6 +201,10 @@ has to be used to view and change branch related variables."
 (transient-define-prefix magit-branch (branch)
   "Add, configure or remove a branch."
   :man-page "git-branch"
+  ["Arguments"
+   (7 "-r" "Recurse submodules when checking out an existing branch"
+      "--recurse-submodules"
+      :if (lambda () (version<= "2.13" (magit-git-version))))]
   ["Variables"
    :if (lambda ()
          (and magit-branch-direct-configure
@@ -232,18 +236,22 @@ has to be used to view and change branch related variables."
   (interactive (list (magit-get-current-branch)))
   (transient-setup 'magit-branch nil nil :scope branch))
 
+(defun magit-branch-arguments ()
+  (transient-args 'magit-branch))
+
 ;;;###autoload
-(defun magit-checkout (revision)
+(defun magit-checkout (revision &optional args)
   "Checkout REVISION, updating the index and the working tree.
 If REVISION is a local branch, then that becomes the current
 branch.  If it is something else, then `HEAD' becomes detached.
 Checkout fails if the working tree or the staging area contain
 changes.
 \n(git checkout REVISION)."
-  (interactive (list (magit-read-other-branch-or-commit "Checkout")))
+  (interactive (list (magit-read-other-branch-or-commit "Checkout")
+                     (magit-branch-arguments)))
   (when (string-match "\\`heads/\\(.+\\)" revision)
     (setq revision (match-string 1 revision)))
-  (magit-run-git "checkout" revision))
+  (magit-run-git "checkout" args revision))
 
 ;;;###autoload
 (defun magit-branch-create (branch start-point)
@@ -254,12 +262,13 @@ changes.
   (magit-refresh))
 
 ;;;###autoload
-(defun magit-branch-and-checkout (branch start-point)
+(defun magit-branch-and-checkout (branch start-point &optional args)
   "Create and checkout BRANCH at branch or revision START-POINT."
-  (interactive (magit-branch-read-args "Create and checkout branch"))
+  (interactive (append (magit-branch-read-args "Create and checkout branch")
+                       (list (magit-branch-arguments))))
   (if (string-match-p "^stash@{[0-9]+}$" start-point)
       (magit-run-git "stash" "branch" branch start-point)
-    (magit-call-git "checkout" "-b" branch start-point)
+    (magit-call-git "checkout" args "-b" branch start-point)
     (magit-branch-maybe-adjust-upstream branch start-point)
     (magit-refresh)))
 
@@ -336,7 +345,7 @@ when using `magit-branch-and-checkout'."
            (t
             (list choice (magit-read-starting-point "Create" choice))))))
   (if (not start-point)
-      (magit-checkout branch)
+      (magit-checkout branch (magit-branch-arguments))
     (when (magit-anything-modified-p t)
       (user-error "Cannot checkout when there are uncommitted changes"))
     (magit-branch-and-checkout branch start-point)
--- a/Documentation/magit.texi
+++ b/Documentation/magit.texi
@@ -31,7 +31,7 @@ General Public License for more details.
 @finalout
 @titlepage
 @title Magit User Manual
-@subtitle for version 2.90.1 (v2.90.1-1171-gb4cf9fa9c+1)
+@subtitle for version 2.90.1 (v2.90.1-1193-g49fbb7c2d+1)
 @author Jonas Bernoulli
 @page
 @vskip 0pt plus 1filll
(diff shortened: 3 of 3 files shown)


lisp/magit-branch.el
```diff
@@ -232,18 +235,21 @@ has to be used to view and change branch related variables."
...
-(defun magit-checkout (revision)
+(transient-define-suffix magit-checkout (revision args)
   "Checkout REVISION, updating the index and the working tree.
 If REVISION is a local branch, then that becomes the current
 branch.  If it is something else, then `HEAD' becomes detached.
 Checkout fails if the working tree or the staging area contain
 changes.
 \n(git checkout REVISION)."
-  (interactive (list (magit-read-other-branch-or-commit "Checkout")))
+  (interactive (list (magit-read-other-branch-or-commit "Checkout") (magit-branch-arguments)))
```
tarsius, line 249: Add this on a new line.

lisp/magit-branch.el
```diff
@@ -254,12 +260,12 @@ changes.
   (magit-refresh))
 
 ;;;###autoload
-(defun magit-branch-and-checkout (branch start-point)
+(defun magit-branch-and-checkout (branch start-point args)
   "Create and checkout BRANCH at branch or revision START-POINT."
-  (interactive (magit-branch-read-args "Create and checkout branch"))
+  (interactive (list (magit-branch-read-args "Create and checkout branch") (magit-branch-arguments)))
```
tarsius, line 265: This gives `((a b) c)` when we want `(a b c)`.

lisp/magit-branch.el
```diff
@@ -336,7 +342,7 @@ when using `magit-branch-and-checkout'."
            (t
             (list choice (magit-read-starting-point "Create" choice))))))
   (if (not start-point)
-      (magit-checkout branch)
+      (magit-checkout branch (transient-args 'magit-branch))
```
tarsius, line 345: Since you added `magit-branch-arguments`, also use it here.

lisp/magit-branch.el
```diff
@@ -232,18 +235,21 @@ has to be used to view and change branch related variables."
   (interactive (list (magit-get-current-branch)))
   (transient-setup 'magit-branch nil nil :scope branch))
 
+(defun magit-branch-arguments ()
+  (transient-args 'magit-branch))
+
 ;;;###autoload
-(defun magit-checkout (revision)
+(transient-define-suffix magit-checkout (revision args)
```
tarsius, line 242: Let's make that optional to avoid having to adjust callers.

lisp/magit-branch.el
```diff
@@ -201,6 +201,9 @@ has to be used to view and change branch related variables."
 (transient-define-prefix magit-branch (branch)
   "Add, configure or remove a branch."
   :man-page "git-branch"
+  :value '("--recurse-submodules")
+  ["Arguments"
+   ("-r" "Recurse submodules when checking out an existing branch" "--recurse-submodules")]
```
tarsius, line 206: This shouldn't be enabled by default because it detaches HEAD and I think it should even be hidden by default.

tarsius (changes requested):

lisp/magit-branch.el
insi-eb: ok, I can see that detaching HEAD might break other people's workflows. I'll check how I can make default-enabling and default-showing this configurable via the magit-group options, as I'd very much like to have that for my workflow.

btw, what is the rationale for default-hiding? just uncluttering the interface for a probably rare use-case or something eles in addition to that?
It would probably also make sense to list the option to default-show this option underneath the magit-submodule documentation in addition to the magit-branch documentation, wouldn't it? at least that is where I looked first for something like this.

kyleam:
> If there are uncommitted changes on a submodule, the update will fail, so it should be safe to default-enable this as far as I can see.

Unfortunately there's still an outstanding bug in git where `git checkout --recurse-submodules ...` will discard modifications in submodules.

https://lore.kernel.org/git/20200525094019.22padbzuk7ukr5uv@overdrive.tratt.net/

> I could not find any other way to enable this option with the existing magit configuration possibilities.

Depending on the user's preferences/workflow, one option is to set `submodule.recurse` to true.  That of course isn't desirable for those that don't always want to recurse with all the commands that that option affects.

lisp/magit-branch.el
kyleam: Additional things to note when considering whether to show this by default:

  * the outstanding bug that I linked to in my other reply

  * this isn't available until 2.13 (which is more recent than magit's current minimum git version)

    Perhaps it's worth guarding this option in the same way the bisect transient does:
    `:if (lambda () (version<= "2.13" (magit-git-version)))`

insi-eb: thanks for the feedback, I'll try to guard the option and not enable/show this by default as soon as I've fixed some of the other findings here

lisp/magit-branch.el
insi-eb: done

lisp/magit-branch.el
insi-eb: done

lisp/magit-branch.el
insi-eb: it looks like I need to dig through the elisp manual (and perhaps my good old copy of Abelson and Sussmann), as my first attempts here failed ... I will fix this and (re-)learn something

insi-eb: fixed

lisp/magit-branch.el
insi-eb: fixed

lisp/magit-branch.el
insi-eb: I've added user options to show and default-enable this, in addition the git-version is checked as proposed. If that is ok, I'd take a shot at updating the user documentation so that people are aware of these options.

lisp/magit-branch.el
```diff
@@ -232,18 +260,22 @@ has to be used to view and change branch related variables."
   (interactive (list (magit-get-current-branch)))
   (transient-setup 'magit-branch nil nil :scope branch))
 
-;;;###autoload
-(defun magit-checkout (revision)
+(defun magit-branch-arguments ()
+  (transient-args 'magit-branch))
+
+;;;###autoload (autoload 'magit-checkout "magit" nil nil t)
+(transient-define-suffix magit-checkout (revision &optional args)
```
insi-eb, line 267: changing this from defun to transient-define-suffix adds warnings to the build
`Warning: ‘magit-checkout’ is for interactive use only`

is there any way to keep the defun and create a transient-suffix just from the existing function? I've skimmed the transient documentation for something like this, but nothing really popped out. If there is a specific section in that documentation I should read in detail for this, could someone point me to it?

tarsius:
These two options are not necessary. This essentially accomplishes the same:

```diff
-   ("-r" "Recurse submodules when checking out an existing branch" "--recurse-submodules")]
+  (7 "-r" "Recurse submodules when checking out an existing branch" "--recurse-submodules")]
```

tarsius:

tarsius:
The infix arguments used by a certain transient popups can be [set/saved](https://magit.vc/manual/transient/Saving-Values.html) for future invocations.  TL;DR Enter the transient as usual, enable and/or disable arguments, and finally type <kbd>C-x s</kbd> or <kbd>C-x C-s</kbd>.

insi-eb:
> These two options are not necessary. This essentially accomplishes the same:
> 
> ```diff
> -   ("-r" "Recurse submodules when checking out an existing branch" "--recurse-submodules")]
> +  (7 "-r" "Recurse submodules when checking out an existing branch" "--recurse-submodules")]
> ```

Thanks for the hint.

I actually did look at the levels feature of transient some weeks before I started adding the functionality, but promptly forgot about it afterwards. I will update both the code and documentation.

lisp/magit-branch.el
insi-eb: I believe this is solved now. I will resolve the GitHub conversation if no one complains soon

lisp/magit-branch.el
```diff
@@ -254,12 +263,12 @@ changes.
   (magit-refresh))
 
 ;;;###autoload
-(defun magit-branch-and-checkout (branch start-point)
+(defun magit-branch-and-checkout (branch start-point &optional args)
   "Create and checkout BRANCH at branch or revision START-POINT."
-  (interactive (magit-branch-read-args "Create and checkout branch"))
+  (interactive (append (magit-branch-read-args "Create and checkout branch") (magit-branch-arguments)))
```
tarsius, line 268: And now it is `(a b c d)` when we want `(a b (c d))`.

Documentation/magit.org
```diff
@@ -5162,6 +5162,17 @@ default.
...
+- Key: - r
+
+  This argument enables the corresponding git option. The option is only
+  available since git 2.13 and somewhat dangerous as it
+  a) detaches HEAD of the submodule and
+  b) as of git 2.30 still has a bug when switching to a branch that
+  does not have the submodule registered, where it will silently
+  discard any changes made.
+
+
```
tarsius, line 5175: We don't document individual arguments in the manual.

lisp/magit-branch.el
```diff
@@ -201,6 +201,11 @@ has to be used to view and change branch related variables."
 (transient-define-prefix magit-branch (branch)
   "Add, configure or remove a branch."
   :man-page "git-branch"
+  :value '("--recurse-submodules")
```
tarsius, line 204: You are still enabling this by default. Please don't.

lisp/magit-branch.el
```diff
@@ -201,6 +201,11 @@ has to be used to view and change branch related variables."
...
+  ["Arguments"
+   (7 "-r" "Recurse submodules when checking out an existing branch" "--recurse-submodules"
+    :if (lambda ()
```
tarsius, line 207: Merge the next line into this one and indent correctly.

lisp/magit-branch.el
tarsius: (And this should be two lines, especially once you have added the `(list ...)`.

lisp/magit-branch.el
tarsius: Why do you want to use `transient-define-suffix`? I don't see a need for that here. You don't need to do this to use `transient-args`.

Documentation/magit.org
insi-eb: to be honest, I would find that sometimes helpful. But I can remove it again, of course.

lisp/magit-branch.el
insi-eb: right, will remove

lisp/magit-branch.el
insi-eb: I'm not sure I understand what's wrong here now, sorry for possibly being dense.

The previous finding was 
> This gives ((a b) c) when we want (a b c).
and the code really was not working right.

now magit-branch-read-args produces (a b), magit-branch-arguments produces (c), append should produce (a b c), and that is as it should be, shouldn't it, as there are 3 arguments to the function (one being optional)?

tarsius: In the future there might be two arguments. Then we would get `(a b "--foo" "--bar")`, which does not match the function signature. We need:

```lisp
(append (magit-branch-read-args "Create and checkout branch")
        (list (magit-branch-arguments)))
```

or

```lisp
`(,@(magit-branch-read-args "Create and checkout branch")
  ,(magit-branch-arguments))
```

insi-eb: ok, understood, thanks!

just out of curiosity, adding `&rest` to the signature would also make this as future-proof, wouldn't it?

tarsius: Yes, but that's not how we do it for the "arguments for git" `ARGS` argument.

insi-eb: hopefully fixed now

PR #4332, opened 2021-02-26
Only use local paths in magit-gitignore-in-subdir

snogge:
The magit-gitignore-in-subdir command fails when DIRECTORY is a remote dir. 
Use `file-local-name` to resolve it to a local name before giving it to git add.

### DIFF
--- a/lisp/magit-gitignore.el
+++ b/lisp/magit-gitignore.el
@@ -68,8 +68,8 @@ repository.  Also stage the file."
 
 ;;;###autoload
 (defun magit-gitignore-in-subdir (rule directory)
-  "Add the Git ignore RULE to a \".gitignore\" file.
-Prompted the user for a directory and add the rule to the
+  "Add the Git ignore RULE to a \".gitignore\" file in DIRECTORY.
+Prompt the user for a directory and add the rule to the
 \".gitignore\" file in that directory.  Since such files are
 tracked, they are shared with other clones of the repository.
 Also stage the file."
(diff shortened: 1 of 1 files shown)


lisp/magit-gitignore.el
```diff
@@ -68,8 +68,8 @@ repository.  Also stage the file."
 
 ;;;###autoload
 (defun magit-gitignore-in-subdir (rule directory)
-  "Add the Git ignore RULE to a \".gitignore\" file.
-Prompted the user for a directory and add the rule to the
+  "Add the Git ignore RULE to a \".gitignore\" file in DIRECTORY.
+Prompt the user for a directory and add the rule to the
```
tarsius, line 72: To keep checkdoc happy, each argument should appear in the doc-string. So use `Prompt the user for a DIRECTORY and ...`.

lisp/magit-gitignore.el
```diff
@@ -78,7 +78,7 @@ Also stage the file."
   (magit-with-toplevel
     (let ((file (expand-file-name ".gitignore" directory)))
       (magit--gitignore rule file)
-      (magit-run-git "add" file))))
+      (magit-run-git "add" (file-local-name file)))))
```
tarsius, line 81: Use `magit-convert-filename-for-git` instead.

lisp/magit-gitignore.el
snogge: Checkdoc is happy here on Emacs 28 when I added the 'in DIRECTORY on the first line.  But I'm happy to upcase all instances of DIRECTORY if that's how you want it.

lisp/magit-gitignore.el
snogge: Done

lisp/magit-gitignore.el
tarsius: Errr I misread the diff. I somehow thought you removed "... in DIRECTORY".

tarsius:
Thanks!

PR #4333, opened 2021-02-26
Improve suggestion for magit after-save hook

ilyagr:
This changes the suggestion to set the hook only after magit is loaded.

Currently, copy-pasting from the manual leads to messages like 
`Symbol’s function definition is void: magit-after-save-refresh-status` when trying to
exit emacs and, even worse, strange debugger errors after files are saved.
I wasn't enough of an Emacs expert to know how hooks work, so this was unexpected
and annoying to me.

This commit replaces the suggested way to implement the hook with the answer from
https://emacs.stackexchange.com/questions/56309/auto-refresh-magit-status-only-when-magit-is-running.

==========

After making this, I saw that this doesn't quite fit the contribution guidelines in that I edited the file in Github and didn't add a feature branch. I hope that is OK for such a small change, but feel free to dispose of my edit and treat this as a bug report instead.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -645,7 +645,8 @@ whenever a buffer is saved to a file inside the respective repository
 by adding a hook, like so:
 
 #+BEGIN_SRC emacs-lisp
-  (add-hook 'after-save-hook 'magit-after-save-refresh-status t)
+  (with-eval-after-load 'magit-mode
+    (add-hook 'after-save-hook 'magit-after-save-refresh-status t))
 #+END_SRC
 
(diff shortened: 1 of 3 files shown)


ilyagr:
It turns out that the "weird debugger errors" I mentioned are unrelated (https://emacs.stackexchange.com/questions/39042/exiting-insert-mode-in-evil-mode-and-saving-immediately-ctrl-w-causes-er).  Sorry I blamed Magit for it. :)

tarsius:
Thanks!

I trimmed down the commit message.

ilyagr:
Thank you!

PR #4342, opened 2021-03-10
log: search messages case insensitive

danielfleischer:
I added an option to the log screen `-i` to be able to grep commit messages in a case insensitive way. The keyboard shortcut is `-i` (it was available). There's no documentation to update.

I tried combining the `-F` key to be both a transient-option and a transient-switch, but I think it's simpler as a toggle. 

Thank!

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -407,7 +407,8 @@ the upstream isn't ahead of the current branch) show."
    (7 "=s" "Limit to commits since" "--since=" transient-read-date)
    (7 "=u" "Limit to commits until" "--until=" transient-read-date)
    (magit-log:--grep)
-   (7 "-I" "Invert search pattern"  "--invert-grep")
+   (7 "-i" "Search case-insensitive" ("-i" "--regexp-ignore-case"))
+   (7 "-I" "Invert search pattern"   "--invert-grep")
    (magit-log:-G)     ;2
    (magit-log:-S)     ;2
    (magit-log:-L)     ;2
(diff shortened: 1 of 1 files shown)


lisp/magit-log.el
```diff
@@ -407,6 +407,7 @@ the upstream isn't ahead of the current branch) show."
    (7 "=s" "Limit to commits since" "--since=" transient-read-date)
    (7 "=u" "Limit to commits until" "--until=" transient-read-date)
    (magit-log:--grep)
+   (7 "-i" "Search case-insensitive"           "-i")
```
kyleam, line 410: For the second "-i": I think it's better to use the long argument (`--regexp-ignore-case`).

(If you look around the code base, you also see `(<short arg>, <long arg>)` used ([relevant manual section](https://magit.vc/manual/transient/Suffix-Specifications.html#Suffix-Specifications)), though the result is the same in terms of what's displayed.)

kyleam (changes requested):
Thanks for the PR.  In my view this is a good addition and using `-i` is fine.

Some meta-comments:

  * Please squash fixup commits down into the initial commit.

  * As mentioned in the pull request template, we prefer that you don't commit via GitHub.  It messes with the trailing new line, among other things.

  * Take a look at some of the commits that touch the magit-log transient to get a sense for this project's conventions.  You'll notice that most of the commit subjects have a `<scope>:` prefix to make it more obvious what area is affected (e.g. `log: Add remaining history simplification arguments`, `magit-log{,-refresh}: Add --invert-grep`).

lisp/magit-log.el
tarsius: The (s l) form should be used.

PR #4344, opened 2021-03-11
Expand symlinks in top-level test directory

cedarbaum:
This change ensures symlinks in the base test directory for toplevel tests are expanded within the `magit-with-test-directory` macro. Doing this ensures that the fully expanded paths returned from `magit-toplevel` will match expanded relative paths (e.g., `(expand-file-name "repo/")`).

This issue was found on macOS (11.2.2) where temp files/directories are created under `/var/...` which is symlinked to `/private/var/...`.

### DIFF
--- a/t/magit-tests.el
+++ b/t/magit-tests.el
@@ -29,7 +29,7 @@
        (push "GIT_AUTHOR_EMAIL=a.u.thor@example.com" process-environment)
        (condition-case err
            (cl-letf (((symbol-function #'message) (lambda (&rest _))))
-             (let ((default-directory ,dir))
+             (let ((default-directory (file-truename ,dir)))
                ,@body))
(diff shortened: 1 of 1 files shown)


tarsius:
What do you mean by *toplevel* tests? Can we just leave out that word without loosing any important information?

Speaking of information, please briefly describe the issue in the commit message.

cedarbaum:
> What do you mean by toplevel tests? Can we just leave out that word without loosing any important information?

I meant this problem only effects the tests `magit-toplevel:basic`, `magit-toplevel:tramp`, and `magit-toplevel:submodule`. I can try to rephrase to be clearer.

> Speaking of information, please briefly describe the issue in the commit message.

I've reworded the commit to include more details. Please let me know if it is still unclear.

PR #4347, opened 2021-03-14
libgit: add additional test infrastructure

cedarbaum:
This change is meant to make testing usages of libgit functions within magit easier and work within the CI environment.

- Add macros for defining magit tests that run with the Git CLI,
libgit, or both implementations.

- Add simple tests for `magit-bare-repo-p` and run them using both the
Git CLI and libgit.

- Add and use new `test-in-ci` MakeFile target that installs and
sources the libgit2 Emacs package before running tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,7 @@ help:
 	$(info )
 	$(info make test             - run tests)
 	$(info make test-interactive - run tests interactively)
+	$(info make test-in-ci       - run tests in CI environment)
 	$(info make emacs-Q          - run emacs -Q plus Magit)
 	$(info )
 	$(info Release Management)
@@ -108,6 +109,13 @@ install-info: info
 test:
 	@$(BATCH) --eval "(progn\
         $$suppress_warnings\
+	(load-file \"t/magit-tests.el\")\
+	(ert-run-tests-batch-and-exit))"
+
+test-in-ci:
+	@$(BATCH) --eval "(progn\
+		$$suppress_warnings\
+	(setq magit--ensure-libgit-tests-run t)\
 	(load-file \"t/magit-tests.el\")\
 	(ert-run-tests-batch-and-exit))"
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,9 +1,5 @@
 name: test
 on: [ push, pull_request ]
-env:
-  CURL: "curl -fsSkL --retry 9 --retry-delay 9"
-  GHRAW: "https://raw.githubusercontent.com"
-  BUILD_MAGIT_LIBGIT: "false"
 jobs:
   test:
     runs-on: ubuntu-20.04
@@ -19,17 +15,21 @@ jobs:
         - 27.1
         - snapshot
     steps:
-    - uses: purcell/setup-emacs@master
+    - uses: cachix/install-nix-action@v12
       with:
-        version: ${{ matrix.emacs_version }}
+        nix_path: nixpkgs=channel:nixos-unstable
+    - uses: cachix/cachix-action@v8
+      with:
+        name: emacs-ci
     - uses: actions/checkout@v2
     - name: Install
       run: |
-        $CURL -O ${GHRAW}/magnars/dash.el/master/dash.el
-        $CURL -O ${GHRAW}/magit/transient/master/lisp/transient.el
-        $CURL -O ${GHRAW}/magit/with-editor/master/with-editor.el
-        emacs -Q --batch -L . -f batch-byte-compile dash.el transient.el with-editor.el
+        # Build and install Emacs (+ magit dependencies) using Nix
+        emacs_ci_version=$(echo "emacs-${{ matrix.emacs_version }}" | sed -e "s/\./-/g")
+        nix-env -f ./t/default.nix -iA $emacs_ci_version
         emacs --version
+
+        # Configure Git
         git config --global user.name "A U Thor"
         git config --global user.email a.u.thor@example.com
         git tag 0
@@ -38,4 +38,4 @@ jobs:
         make lisp DASH_DIR=$PWD
     - name: Test
       run: |
-        make test DASH_DIR=$PWD
+        make test-in-ci DASH_DIR=$PWD
--- a/t/default.nix
+++ b/t/default.nix
@@ -1,0 +1,18 @@
+let
+  emacs-overlay = import (builtins.fetchTarball { url = https://github.com/nix-community/emacs-overlay/archive/master.tar.gz; });
+  emacs-ci = import (builtins.fetchTarball { url = https://github.com/purcell/nix-emacs-ci/archive/master.tar.gz; });
+
+  pkgs = import <nixpkgs> { overlays = [ emacs-overlay ]; };
+in
+builtins.mapAttrs
+  (version: emacs:
+    (pkgs.emacsPackagesGen emacs).emacsWithPackages
+      (emacsPackages: [
+        emacsPackages.dash
+        emacsPackages.transient
+      ] ++ (with emacsPackages.melpaPackages; [
+        libgit
+        with-editor
+      ])
+      ))
+  emacs-ci
--- a/t/magit-tests.el
+++ b/t/magit-tests.el
@@ -39,9 +39,56 @@
   (declare (indent 0) (debug t))
   `(magit-with-test-directory (magit-test-init-repo ".") ,@body))
 
+(cl-defmacro magit-deftest-with-git (name () &body docstring-keys-and-body)
+  "Defines a magit test where all Git commands are run using the Git CLI."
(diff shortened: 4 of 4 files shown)


cedarbaum:
There is still an issue in the CI environment with libssl not being found when building libgit. I am looking into this, but in the meantime would appreciate any feedback regarding the general approach (i.e. is what this PR adds useful for future development work on libgit integrations).

Makefile
```diff
@@ -116,6 +118,23 @@ test-interactive:
 	(load-file \"t/magit-tests.el\")\
 	(ert t))"
 
+--ensure-libgit:
```
cedarbaum, line 121: This should probably be moved to test.yml.

npostavs:
>     * Add macros for defining magit tests that run with the Git CLI,
>       libgit, or both implementations.

I seem to recall that replacing `ert-deftest` with a differently named macro makes trouble when trying to jump to test definitions. I think it's usually better to make a macro to generate the body rather replacing the whole `ert-deftest`.

cedarbaum:
> > ```
> > * Add macros for defining magit tests that run with the Git CLI,
> >   libgit, or both implementations.
> > ```
> 
> I seem to recall that replacing `ert-deftest` with a differently named macro makes trouble when trying to jump to test definitions. I think it's usually better to make a macro to generate the body rather replacing the whole `ert-deftest`.

Good point - it does seem that jumping to the test from the ERT results buffer just goes to the top of the test file instead of the definition. As an aside, I was thinking about how to more generically enable parameterized testing with ERT and things like this are important to consider.

I still think there are a couple strong benefits to the current approach:

1. Less boilerplate macros and indentation levels in the test body, especially if every test in the future needs to run with both Git and libgit.

2. A clearer view from ERT's runner as to which Git implementation is problematic, since each has its own test definition. E.g.,

```
   FAILED   5/22  magit-in-bare-repo-with-git (0.182633 sec)
   passed   6/22  magit-in-bare-repo-with-libgit (0.194090 sec)
```

However, I am happy to defer to maintainers here as to which approach makes more sense for this project.

cedarbaum:
A couple updates:

1.) @tarsius - I've split this into 3 commits, so it should hopefully be easier to review now.

2.) The issue building `libegit` in CI turned out to be more complicated than I initially thought. Basically, it seems that since the `setup-emacs` GitHub action uses Emacs builds from Nix, Emacs can have trouble interoperating with dynamic modules built outside of Nix (see similar issue here: https://github.com/purcell/setup-emacs/issues/6). For `libegit`, this manifests as not being able to find `libssl.so.1.1`, even when it is available on the system.

Trying to point Emacs to `libssl.so.1.1` via `LD_LIBRARY_PATH` and the `-L` command line argument both fail, with the latter being a well-known issue: https://github.com/NixOS/nix/issues/902.

My current solution for this is to fork the GitHub action and build Emacs alongside `libegit` in Nix and use these builds instead. Another solution would be to pull in the Nix build of `libegit` at runtime on the build machines, but I am not sure how to do this. Alternatively, having a way to bundle arbitrary [M]ELPA packages with the `setup-emacs` action (i.e. in the YAML config) would also work, but this might be complex functionality to add.

I must also admit that I am far from a Nix expert, so I might be missing something obvious here. @purcell, if you have any insights or suggestions regarding above, it would be greatly appreciated!

purcell:
> Alternatively, having a way to bundle arbitrary [M]ELPA packages with the `setup-emacs` action (i.e. in the YAML config) would also work, but this might be complex functionality to add.

With Nix you can pretty easily create a custom Emacs bundled with arbitrary packages using the [emacs overlay](https://github.com/nix-community/emacs-overlay). I worry it could be a debugging and support nightmare if I tried to expose that functionality via the `setup-emacs` action's configuration though: the bundling functionality might well break messily with older Emacsen. But it might be something I could look at.

Forking the action seems like a massive amount of overhead and ongoing maintenance, and I wouldn't recommend it.

cedarbaum:
> With Nix you can pretty easily create a custom Emacs bundled with arbitrary packages using the [emacs overlay](https://github.com/nix-community/emacs-overlay). 

Overlays seem like an interesting solution, though I am not sure what the right way to implement it is. For some context, my current solution runs the below Nix code to augment the Emacs built from `setup-emacs`:

```nix
{ emacsAttr
}:
let
  pkgs = import (import ./nix/sources.nix).nixpkgs {};
  emacs = (import ./emacs-versions.nix).${emacsAttr};
in
(pkgs.emacsPackagesFor emacs).emacsWithPackages (
  epkgs: [
    epkgs.seq
    epkgs.dash
    epkgs.transient
  ] ++ (with epkgs.melpaPackages; [
    libgit
    with-editor
  ])
)
```
So with an overlay, I would have to somehow compose this with the Nix code being run by the `setup-emacs` action. This seems possible by dropping an `overlays.nix` file in `~/.config/nixpkgs/` before the install action is run, but I am currently struggling a bit with the syntax for this (assuming it is even possible 😄 ).

Also, I am not sure how emacs-overlay fits into this. Would this help with augmenting builds from the `setup-emacs` action or was this meant to be a completely separate build process?

Do you have any example code or other pointers for achieving this build composition in Nix?

> Forking the action seems like a massive amount of overhead and ongoing maintenance, and I wouldn't recommend it.

Completely agree, definitely looking for a better solution.

purcell:
I think the solution here (absent me providing magical extra functionality in setup-emacs, which I'm not _completely_ ruling out yet) is to work directly with Nix in your CI. There's no way to "get inside" the nix expressions bundled into setup-emacs in the way you're thinking.

So if you have a default.nix, you can make versions of each of the upstream CI Emacsen that have packages from the emacs overlay added to them:

```nix
let
  emacs-overlay = import (builtins.fetchTarball { url = https://github.com/nix-community/emacs-overlay/archive/master.tar.gz; });
  emacs-ci = import (builtins.fetchTarball { url = https://github.com/purcell/nix-emacs-ci/archive/master.tar.gz; });

  pkgs = import <nixpkgs> { overlays = [ emacs-overlay ]; };
in
builtins.mapAttrs
  (version: emacs:
    (pkgs.emacsPackagesGen emacs).emacsWithPackages
      (epkgs: [
        epkgs.seq
        epkgs.dash
        epkgs.transient
      ] ++ (with epkgs.melpaPackages; [
        libgit
        with-editor
      ])
      ))
  emacs-ci
```
and then you can either use `nix-env` to "install" augmented Emacs into your local environment, or run commands inside a nix-shell, e.g. `nix-shell -A emacs-26-1 --run "my command here"` in the directory with the default.nix.

I tinkered a little to come up with the above and it doesn't quite work, since the packages don't show up as installed, but I've got a cold and my brain is foggy, so I need to stop now -- hopefully it is a useful clue as to the approach.

P.S. to get caching of pre-built emacsen you'd use the `use-cachix` action with the `emacs-ci` cache name. Any compiled modules would probably still get compiled afresh each CI run, but you could yourself use the free cachix service to avoid this if you get everything working.

cedarbaum:
Thanks so much @purcell - this approach looks to be working in CI! Hope you feel better soon!

A few notes on the latest commit:

- The `setup-emacs` action is completely removed in favor of the custom Nix install defined above.
- Nix is now installed using the [install-nix-action](https://github.com/cachix/install-nix-action).
- `cachix` is installed & configured in the "Install" step. **TODO:** use the `cachix/cachix-action` GitHub action instead.
- All `magit` dependencies (`transient`, `dash`) are now installed using Nix. Previous `curl` commands have been removed. **TODO:** removed unused `CURL` & `GHRAW` ENV variables.
- Add shell script to convert `matrix.emacs_version` to the full attribute name expected by `nix-emacs-ci` (e.g., 26.1 => emacs-26-1). This was previously done in the `setup-emacs` action.

purcell:
>     * All `magit` dependencies (`transient`, `dash`) are now installed using Nix.

Note that the emacs overlay has a thing called "emacsWithPackagesFromPackageRequires" (which I contributed), so you could conceivably use that to parse magit.el and add those dependencies automatically. The one flaw with that suggestion is that magit's dependencies are listed in magit-pkg.el, which isn't supported. But maybe support for that upstream in the overlay would be good...

tarsius:
This seems to complicate things. Can't we just add a new dimension to the test matrix?

1. dimension: emacs versions
2. dimension: git implementations

cedarbaum:
> This seems to complicate things. Can't we just add a new dimension to the test matrix?

I think this would work well. Do you think it would be useful to differentiate tests that need to be run with both implementations (i.e. label such tests or put them in a separate file) or should everything run twice?

tarsius:
Just run everything twice. Less work for us and it might actually reveal issues in unexpected places.

cedarbaum:
Makes sense. Also, since this PR has diverged quite a bit from its original design, I opened a new one (#4361) and will close this.

PR #4349, opened 2021-03-25
magit-extras: Do not modify customized project-switch-commands

mpolden:
Problem: Magit messes with my `project-switch-commands`.

The variable `project-switch-commands` is customizable, and I configure it to contain the commands I want. Since I use `use-package` to load `magit`, `project-switch-commands` may be set before either `magit` or `magit-extras` is loaded. 

However, once `magit-extras` is loaded it unconditionally modifies `project-switch-commands`. This is not ideal.

This PR only adds the "Magit" command if the variable hasn't been customized. If also ensures that the "Magit" command is added _after_ the default ones, which is a better default in my opinion.

### DIFF
--- a/lisp/magit-extras.el
+++ b/lisp/magit-extras.el
@@ -149,7 +149,7 @@ like pretty much every other keymap:
              (equal project-switch-commands
                     (custom--standard-value 'project-switch-commands)))
     (define-key project-prefix-map "m" #'magit-project-status)
-    (add-to-list 'project-switch-commands '(magit-project-status "Magit"))))
+    (add-to-list 'project-switch-commands '(magit-project-status "Magit") t)))
 
 ;;;###autoload
 (defun magit-dired-jump (&optional other-window)
--- a/lisp/magit-extras.el
+++ b/lisp/magit-extras.el
@@ -144,7 +144,10 @@ like pretty much every other keymap:
 (with-eval-after-load 'project
   ;; Only more recent versions of project.el have `project-prefix-map' and
   ;; `project-switch-commands', though project.el is available in Emacs 25.
-  (when (boundp 'project-prefix-map)
+  (when (and (boundp 'project-prefix-map)
+             ;; Only modify if it hasn't already been modified.
+             (equal project-switch-commands
+                    (custom--standard-value 'project-switch-commands)))
     (define-key project-prefix-map "m" #'magit-project-status)
     (add-to-list 'project-switch-commands '(magit-project-status "Magit"))))
 


lisp/magit-extras.el
```diff
@@ -144,9 +144,11 @@ like pretty much every other keymap:
 (with-eval-after-load 'project
   ;; Only more recent versions of project.el have `project-prefix-map' and
   ;; `project-switch-commands', though project.el is available in Emacs 25.
-  (when (boundp 'project-prefix-map)
+  (when (and (boundp 'project-prefix-map)
+             ;; Only modify project-switch-commands if it hasn't already been customized
+             (equal project-switch-commands (eval (car (get 'project-switch-commands 'standard-value)))))
```
basil-conto, line 149: `eval` should probably be called with a non-`nil` second argument.

basil-conto (commented):
Would it be desirable to control this with a user option, as with `magit-define-global-key-bindings`?

mpolden:
You mean a variable for controlling the project.el integration?

basil-conto:
> You mean a variable for controlling the project.el integration?

Yes, in place of the `standard-value` check.

I'm just thinking out loud and curious about the pros/cons though, and have no strong opinion either way.

mpolden:
Yes, I thought about doing that, but there were no other such variables in `magit-extras`. I'm also not sure whether Magit has any guidelines for what should be customizable and not.

No strong preference on my part either way.

tarsius:
Let's go with this approach and add the option once/if someone actually wants to use it.

lisp/magit-extras.el
```diff
@@ -144,9 +144,12 @@ like pretty much every other keymap:
 (with-eval-after-load 'project
   ;; Only more recent versions of project.el have `project-prefix-map' and
   ;; `project-switch-commands', though project.el is available in Emacs 25.
-  (when (boundp 'project-prefix-map)
+  (when (and (boundp 'project-prefix-map)
+             ;; Only modify if it hasn't already been modified.
+             (equal project-switch-commands
+                    (custom--standard-value 'project-switch-commands)))
```
galeo, line 150: It seems that `custom--standard-value` is a new function introduced in Emacs master branch.

kyleam: Thanks for noticing.  I've inlined its definition in 527842a3.

tarsius: Which, by the way, is what @mpolden did before I messed it up. I though I did check that it was present in 26.1 before using it but somehow I knew I would mess it up, which is why I added myself as a co-author. :wink:

Issue #4293, opened 2021-01-14
Issue defering git-commit


Ergus:
Hi: 

Recently I have seen an increase in my loading time when opening files in git repositories (like: emacs myfile). The increase seems to be related with the loading of many packages/dependencies I cannot defer properly because they are loaded when calling `git-commit-setup-check-buffer`. So different questions come to me now:

1) `global-git-commit-mode` is `t` by default... this makes it very difficult to disable it without autoloading the package first (dependencies like log-edit are very "heavy"). The code goes in the quickstart but also I am not sure if it is right to enable by default global external packages modes that are not in vanilla... is this intended?? So only installing the package will affect the user.

2) I also see that the function `git-commit-setup-check-buffer` is somehow a manual implementation for what `auto-mode-alist` does and needs to be added to `find-file-hook`... which means that it always autoloads the git-commit package. Is this intended?? Isn't it just the same and simpler to add the git-commit-setup and the related hooks to a minor mode and set that to auto-mode-alist?? That may take care of the autoloading properly deffering that and specially does not affect the loading time just because the user installed (probably indirectly) git-log as a magit dependency.

3) As a consequence of 1 `package-quickstart` has this code:
```
(defcustom global-git-commit-mode t 
  :set #'custom-set-minor-mode 
  :initialize 
    (lambda (symbol exp) 
      (custom-initialize-default symbol exp) 
      (when global-git-commit-mode (add-hook 'find-file-hook 'git-commit-setup-check-buffer))) 
  :group 'git-commit 
  :type 'boolean)
```
Which means that `global-git-commit-mode` needs to be set to nil in the early-init to avoid the unconditional autoloading. Otherwise `global-git-commit-mode` will always be true. This problem is somehow recent so it seems it is related with some recent change in git-log or magit... is it??

This also breaks my `use-package` config that used to be very simple and efficient.

Thanks in advance,
Ergus

tarsius:
> Which means that `global-git-commit-mode` needs to be set to nil in the early-init to avoid the unconditional autoloading. Otherwise `global-git-commit-mode` will always be true.

IMO that's acceptable.

> This problem is somehow recent so it seems it is related with some recent change in git-log or magit... is it??

I don't think it is.

basil-conto:
> Which means that `global-git-commit-mode` needs to be set to nil in the early-init to avoid the unconditional autoloading.

(Or package activation deferred until `user-init-file` via `package-enable-at-startup`.)

Ergus:
[...]

PR #4352, opened 2021-03-27
git-commit: (Re)enable autoloading without all dependencies

wentasah:
As a response to #4293, commit 13f20763 ("No longer autoload
global-git-commit-mode", 2021-02-06) disabled autoloading of
git-commit.el to avoid loading its many dependencies at Emacs
initialization time. This made it harder to use git-commit.el as
$EDITOR invoked by `git commit`.

This commit re-enables the autoloads again, but also splits
git-commit.el to two files. For initialization, only a small file
without any dependencies is needed. The rest of git-commit.el and all
dependencies are loaded later, when needed.

Note that this change would require updating MELPA (and maybe other) receipes to include the new file. Alternatively, the same result could be achieved by "copying" the needed code to `git-commit-autoloads.el` by wrapping the needed functions in `progn` and prefixing them with autoload cookies. Let me know what is preferable.

### DIFF
--- a/lisp/git-commit-global.el
+++ b/lisp/git-commit-global.el
@@ -1,0 +1,78 @@
+;;; git-commit-global.el --- Support for autoloading git-commit.el -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2010-2021  The Magit Project Contributors
+;;
+;; You should have received a copy of the AUTHORS.md file which
+;; lists all contributors.  If not, see http://magit.vc/authors.
+
+;; Authors: Jonas Bernoulli <jonas@bernoul.li>
+;;	Sebastian Wiesner <lunaryorn@gmail.com>
+;;	Florian Ragwitz <rafl@debian.org>
+;;	Marius Vollmer <marius.vollmer@gmail.com>
+;; Maintainer: Jonas Bernoulli <jonas@bernoul.li>
+
+;; This file is not part of GNU Emacs.
+
+;; This file is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 3, or (at your option)
+;; any later version.
+
+;; This file is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


tarsius:
> This made it harder to use git-commit.el as $EDITOR invoked by `git commit`.

But not much. As I said at https://github.com/magit/magit/issues/4293#issuecomment-777486312:

> > The autoloads triggered full loading as soon as the user created
> > any buffer.  The autoloads had to do that to be of any use, which
> > means that they had nearly the same effect as loading the library
> > explicitly, which unfortunately is very slow.
> 
> > Now that the autoloads are gone, users can actually choose whether
> > they want to pay that price.  Most users likely won't notice that
> > anything has changed as this only matter when one sometimes commits
> > from the command-line without loading `magit' first.
> 
> In other words just `(require 'git-commit)`, which isn't any worse than the old autoloads.

So yes, the `(require 'git-commit)` is needed now, but as I said I think that is okay. And loading the dependencies takes its time (like in the past).

In a sense what you are proposing is the best of the two approaches, I can see that, but I am still skeptical. This is the sort of arms race that I have seen in this code base before, when we try to accommodate all sides. If this works as advertised, then it would be fine, but we cannot be certain of that. Maybe in a few months we discover that in a few edge-cases the autoloads trigger the full library and its dependencies to be loaded anyway. We could just revert then, but some users would already have removed the require form. So we are committed now and have to find another approach, which could get messy. Or some user (or I) decides that it is ugly that we had to split up the library for this and opens a new pull-request that I have to review, which e.g. does the same by adding some fake major-mode to auto-mode-alist, which then somehow unloads itself and causes the mode that is configured by git-commit-major-mode to be loaded instead. And that too works well for a while until we discover that actually no, it doesn't.

I think the right time to stop on this patch down the rabbit hole is now, before merging this pull-request.

wentasah:
I understand that making things working reliably for everybody can be difficult so I'm fine with not merging this. I think I'll keep using this in my `init.el`. If needed I'll be able to provide more feedback, should you change your mind later.

PR #4356, opened 2021-04-11
bookmark: mending incompatibilities with bookmark+ package

danielfleischer:
Currently there is an issue in both **creating** and **visiting** bookmarks for Magit buffers while using the Bookmark+ package. Upon investigation, a quick fix can be made that's compatible with both vanilla bookmarks and Bookmark+.

- Manually creating alist instead of using the (bookmark+ based)`bookmark-prop-set` function, which expects
    the bookmark to have the structure `(bookmark-name . alist)`, which is not ready yet at the time of that call.

- Explicitly calling bookmark+ jumping function. In bookmark+, custom handlers are expected to handle
   the buffer jumping.

After this PR was created, a patch was made to the bookmark+ package that solves the first issue. The original PR was composed of 2 fixes to the `magit-bookmark` file. I then forced pushed in order to have only the 2nd fix for the visiting issue.

### DIFF
--- a/lisp/magit-bookmark.el
+++ b/lisp/magit-bookmark.el
@@ -84,6 +84,9 @@ with the variables' values as arguments, which were recorded by
                       hidden)
               (magit-section-hide child)
             (magit-section-show child)))))
+    ;; Compatibility with `bookmark+' package.  See #4356.
+    (when (bound-and-true-p bmkp-jump-display-function)
+      (funcall bmkp-jump-display-function (current-buffer)))
     nil))
 
 (cl-defgeneric magit-bookmark-name ()
--- a/lisp/magit-bookmark.el
+++ b/lisp/magit-bookmark.el
@@ -40,7 +40,11 @@ Input values are the major-mode's `magit-bookmark-name' method,
 and the buffer-local values of the variables referenced in its
 `magit-bookmark-variables' property."
   (if (plist-member (symbol-plist major-mode) 'magit-bookmark-variables)
-      (let ((bookmark (bookmark-make-record-default 'no-file)))
+      ;; `bookmark-make-record-default's return value does not match
+      ;; (NAME . ALIST), even though it is used as the default value
+      ;; of `bookmark-make-record-function', which states that such
+      ;; functions must do that.  See #4356.
+      (let ((bookmark (cons nil (bookmark-make-record-default 'no-file))))
         (bookmark-prop-set bookmark 'handler  'magit--handle-bookmark)
         (bookmark-prop-set bookmark 'mode     major-mode)
         (bookmark-prop-set bookmark 'filename (magit-toplevel))


tarsius:
I am very confused. There was this other issue, which I believe to have replied to as shown below, but when search terms that would take me to that reply, then I get no useful search results. Is this in fact that "other" issue and I just forgot to submit reply?

----

Previously discussed at https://www.reddit.com/r/emacs/comments/mhuex8/setting_a_bookmark_in_magit/.

Wouldn't this get the job done too?:

```diff
diff --git a/lisp/magit-bookmark.el b/lisp/magit-bookmark.el
index 91380a7e2..ba726456f 100644
--- a/lisp/magit-bookmark.el
+++ b/lisp/magit-bookmark.el
@@ -40,7 +40,8 @@ (defun magit--make-bookmark ()
 and the buffer-local values of the variables referenced in its
 `magit-bookmark-variables' property."
   (if (plist-member (symbol-plist major-mode) 'magit-bookmark-variables)
-      (let ((bookmark (bookmark-make-record-default 'no-file)))
+      ;; `bookmark+'s redefined `bookmark-prop-set' expects this nil.
+      (let ((bookmark (cons nil (bookmark-make-record-default 'no-file))))
         (bookmark-prop-set bookmark 'handler  'magit--handle-bookmark)
         (bookmark-prop-set bookmark 'mode     major-mode)
         (bookmark-prop-set bookmark 'filename (magit-toplevel))
```

danielfleischer:
Hi Jonas, thank you for connecting the dots with the reddit discussion and offering a fix. 

I've been in contact with the bookmark+ author for the last couple of days, thinking how to solve it. Today he pushed a new version specifically to fix the first issue (creating a bookmark), without needing my previous, "manual" cons-building. It had to do with some differences in `bookmark-prop-set` between vanilla and bookmark+. 

That's why I changed the issue to draft and force pushed only the second patch (visiting bookmark) utilizing explicit jumping if using bookmark+. Forgot to change the description though. 

I think it's ready for review. If you have any question feel free to ask.

tarsius:
* The first incompatibility was due to some ambiguity in the original `bookmark.el`: `bookmark-make-record-function`'s doc-string says that the function must return `(NAME . ALIST)`, but the default value for that variable `bookmark-make-record-default` does not do that *and* there is code that deals with the NAME missing. Yet I was going to fix that on my end.

* The other case seems more complicated and I don't really feel like investing the time needed to understand it. It seems to me that the problem can't be specific to `magit--handle-bookmark`. If this function has to be modified like this for the benefit of `bookmark+` that this would be necessary for any such third-party function.

  But if Drew says that this cannot be handled on `bookmark+`'s side and this is the appropriate way to handle it, then I am willing to merge this without really understanding it. Please add a one-line comment mentioning `bookmark+` and #4356.

tarsius:
`bound-and-true-p` would be more elegant I think.

danielfleischer:
Hi, you got it right in both cases. In the second case, the issue is due to bookmark+ design; buffer jumping is implemented only in a default bookmark handler letting others develop new custom handlers that generalize the idea of a bookmark even having a buffer to jump to. One solution we tried was calling the default handler right after magit custom handler but it wasn't appropriate because magit's buffers are non-file buffers which raises more issues. Anyway, that's the solution offered eventually. I added the comment, PR mention and better predicate. Thanks!

tarsius:
Thanks!

PR #4357, opened 2021-04-11
magit-extras: do not load magit/project integration by default

sirikid:
I propose to make the default integration setup an autoloaded function and let people who need it enable it manually:

```elisp
(eval-after-load 'project #'magit-project-maybe-setup-defaults)
```

or

```elisp
(with-eval-after-load 'project
  (magit-project-maybe-setup-defaults))
```

### DIFF
--- a/lisp/magit-extras.el
+++ b/lisp/magit-extras.el
@@ -141,16 +141,15 @@ like pretty much every other keymap:
   (interactive)
   (magit-status-setup-buffer (project-root (project-current t))))
 
-(with-eval-after-load 'project
+;;;###autoload
+(defun magit-project-maybe-setup-defaults ()
+  "Setup the default magit and project integration."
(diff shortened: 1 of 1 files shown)


sirikid:
This topic was already discussed in #4349

tarsius:
What I had in mind is an opt-out. I've just added one.

sirikid:
> What I had in mind is an opt-out. I've just added one.

Why not an opt-in though?

tarsius:
If the feature was not enabled by default then many users might never know it exists. I see the inconvenience of having to disable an undesired feature as the lesser evil.

sirikid:
Having this feature enabled by default is really bad Custom users. Project-Switch-Command will inevitably be broken again and there is no way to fix it without restart as far as I know.

PR #4361, opened 2021-04-15
Run tests using both Git and libgit implementations in CI

cedarbaum:
- Add simple tests for `magit-bare-repo-p`, which is implemented with both Git and libgit.
- Install Emacs with libegit in CI via a custom Nix install.
- Run tests using both Git and libgit functions in CI by adding a `git_impl` dimension to the build matrix.

### DIFF
--- a/t/magit-tests.el
+++ b/t/magit-tests.el
@@ -14,11 +14,11 @@
 
 (require 'magit)
 
-(defun magit-test-init-repo (dir)
+(defun magit-test-init-repo (dir &rest args)
   (let ((magit-git-global-arguments
          (nconc (list "-c" "init.defaultBranch=master")
                 magit-git-global-arguments)))
-    (magit-git "init" dir)))
+    (magit-git "init" args dir)))
 
 (defmacro magit-with-test-directory (&rest body)
   (declare (indent 0) (debug t))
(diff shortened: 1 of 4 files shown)


t/magit-tests.el
```diff
@@ -20,6 +20,15 @@
...
     (magit-git "init" dir)))
 
+(defun magit-test-init-repo (dir &rest args)
+  (let ((magit-git-global-arguments
+         (nconc (list "-c" "init.defaultBranch=master")
+                magit-git-global-arguments))
+        (args (if args
+                  (append (cons "init" args) (list dir))
+                (list "init" dir))))
+    (apply 'magit-git args)))
```
tarsius, line 30: You are redefining this function that was defined just above instead of modifying that definition. Also the dance concerning `args` is not necessary because Magit's low-level functions for running `git` all ultimately flatten the arguments and prune nil elements. I'll push a fixed version shortly.

t/magit-tests.el
```diff
@@ -35,6 +44,10 @@
                 (signal (car err) (cdr err))))
        (delete-directory ,dir t))))
 
+(defmacro magit-with-test-repository-custom-init (init-args &rest body)
```
tarsius, line 47: I'm gonna replace this with a new `magit-with-bare-test-repository` function.

tarsius:
Thanks! (And sorry it took me so long to review and merge.)

PR #4374, opened 2021-05-03
Respect vc-tor when invoicing a subprocess

phikal:
The user option `vc-tor` was [added in Emacs 27.1](http://git.savannah.gnu.org/cgit/emacs.git/commit/?id=626e0a6aad6b0f3c7348f37c9b2c9854b06b449b) that routes all VC commands through torsocks.

This patch makes magit check if the option exists and is enabled, in which case it replicates VC's behaviour.

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -401,6 +401,9 @@ Process output goes into a new section in the buffer returned by
 Identical to `process-file' but temporarily enable Cygwin's
 \"noglob\" option during the call and ensure unix eol
 conversion."
+  (when (bound-and-true-p vc-tor)
+    (push process args)
+    (setq process "torsocks"))
   (when magit-process-extreme-logging
     (let ((inhibit-message t))
       (message "$ %s" (magit-process--format-arguments process args))))


tarsius:
It might be better to instead have our own option. What to you think @kyleam?

tarsius:
And maybe we should add `magit-run-git-async-with-network` and `magit-run-git-async-with-network-and-editor` to limit this to `git` invocations that actually benefit from that. Of course we should come up with better names. If we do this, then the `(run-hooks 'magit-credential-hook)` should be moved into these new functions.

phikal:
Having a separate option would certainly work, but I think that respecting `vc-tor` is still nice, as it is already meant to apply to every VC system.

kyleam:
@tarsius:
> It might be better to instead have our own option.

@phikal:
> Having a separate option would certainly work, but I think that respecting `vc-tor` is still nice, as it is already meant to apply to every VC system.

I think a separate option would be a good idea.  It seems more discoverable, particularly for those that don't use vc.  It'd also avoid opening the door to requests that we respect various vc configuration knobs.  We could set the initial value based on `vc-tor` so that vc users don't need to configure both.

@tarsius:
> And maybe we should add `magit-run-git-async-with-network` and `magit-run-git-async-with-network-and-editor` to limit this to `git` invocations that actually make network connections.

Something along those lines was my initial thought too.  My main concern with that is the risk that we miss a network operation.  Someone that wants to set this option is probably willing to trade unnecessary costs for a bit more confidence that all operations go through `torsocks`.

Fwiw here are some timings for `git status` and `git log`:

<details>
<summary>timings</summary>


```sh
repo="${1?pass repository to clone}"

cd "$(mktemp -d "${TMPDIR:-/tmp}"/git-torsocks-XXXXXXX)" || exit 1

git version
torsocks --version

git clone --quiet "$repo" ./

cat >bf <<'EOF'
-q torsocks git status
-q git status
EOF
sleep 1
git status >/dev/null 2>&1
multitime -n 20 -b bf

cat >bf <<'EOF'
-q torsocks git log -n100
-q git log -n100
EOF
sleep 1
git log -n100  >/dev/null 2>&1
multitime -n 20 -b bf
```

```
git version 2.31.1.659.g12c5fe8677
Torsocks 2.3.0
===> multitime results
1: -q torsocks git status
            Mean        Std.Dev.    Min         Median      Max
real        0.010       0.004       0.004       0.012       0.016
user        0.005       0.003       0.001       0.005       0.012
sys         0.005       0.003       0.000       0.005       0.012

2: -q git status
            Mean        Std.Dev.    Min         Median      Max
real        0.004       0.002       0.002       0.003       0.007
user        0.002       0.002       0.000       0.002       0.007
sys         0.002       0.003       0.000       0.000       0.007
===> multitime results
1: -q torsocks git log -n100
            Mean        Std.Dev.    Min         Median      Max
real        0.011       0.004       0.006       0.012       0.023
user        0.006       0.004       0.001       0.005       0.019
sys         0.005       0.003       0.000       0.005       0.010

2: -q git log -n100
            Mean        Std.Dev.    Min         Median      Max
real        0.007       0.003       0.004       0.005       0.017
user        0.003       0.003       0.000       0.004       0.008
sys         0.003       0.003       0.000       0.003       0.012
```

</details>

tarsius:
Agree with everything @kyleam said, but:

> We could set the initial value based on `vc-tor` so that vc users don't need to configure both.

Can we sure that is always defined when magit is loaded? Otherwise a user might inspect the default value, which this time around is `t` and then they do something else and the "default" becomes `nil`.

Also our option should have a better name, maybe `magit-use-torsocks`.

phikal:
No, it is not bound by default, but it loaded whenever any vc command is invoked.

tarsius:
Okay then, lets not derive the value from `vc-tor`. (Also note that I reworded my previous post, just in case.)

phikal:
Nevertheless, using `(bound-and-true-p vc-tor)` or something like that as the default value still seems like a good approach, as the default value of `vc-tor` is nil, and the only interesting one is if it _is_ bound, in which case it is most likely `t`.

tarsius:
Right.

:P

tarsius:
After thinking about it some more I am not comfortable adding such an option at all.

`magit-process-file` does not cover all of Magit's calls to `git` or some other executable. I wouldn't want to complicate every place where we start a subprocess and that's what we would have to be on the safe side. And even if I didn't mind that, then we would still risk forgetting about it and putting people at risk at a later time.

IMO if you must use tor, then you should not rely on "every" application explicitly handing that and should instead use a system wide mechanism

kyleam:
> IMO if you must use tor, then you should not rely on "every" application explicitly handing that and should instead use a system wide mechanism

True, doing it at this level doesn't really seem like a viable approach.  

Fwiw on emacs-devel there is a thread with some discussion of making (nearly all) subprocesses go through tor:
https://lists.gnu.org/archive/html/emacs-devel/2020-11/msg00679.html (mid:E1ketIz-0008Np-C3@fencepost.gnu.org)

phikal:
Then I'll close this pull request.

PR #4378, opened 2021-05-10
Remove dash

sirikid:
I removed dash, mostly replaced it by seq and/or pcase, cl-lib in some cases.

Why? I dislike dash a lot and I want to remove it from my Emacs completely someday.


tarsius:
> Why? I dislike dash a lot and I want to remove it from my Emacs completely someday.

I feel somewhat similarly about `s` and `f`, and I have opened pull-requests to remove these dependencies from packages I use.  These packages do provide some useful functionality but in many cases it appeared that authors used them merely because of the hype.  Many of the functions from these package are basically aliases for the respective build-in functions, in which case I see no reason to use them.  Some functions provide additional functionality and if a package uses more than a X such functions (depending on size of project), then I feel the use of `s` and/or `f` are justified.

I feel more positively about `dash`, but understand where you are coming from. Also, the authors of `dash` has done a lot of good for emacs-lisp, not least by encouraging/forcing changes/additions to Emacs itself.

Initially I liked the anaphoric functions, now not so much anymore.  Also seehttps://emacsair.me/2021/01/28/llama-0.1/. 

> I removed dash, mostly replaced it by seq and/or pcase, cl-lib in some cases.

I have stopped using `dash` in many of my own smaller packages to reduce dependencies and because for the most part switching to built-in forms, *nowadays* comes with no or little loss of expressiveness.

Even in the case of Magit I have started to slowly drift away from `dash`, but I am only replacing `dash` functions/macros with built-in alternatives on a case by case basis.  For example I have already replaced `--when-let` with `when-let`.

I have not replaced, for example, `-flatten` with `flatten-tree` yet because the latter was not added until Emacs 27.1 and Magit still supports everything up from Emacs 25.1.  If that was the only function that had no replacement that we could use, then we could just write `magit--flatten`, but there is probably more.

----

So in summary I am slowly migrating away from `dash`, a process which will likely take a few more years.  In fact it may never come to an end because I don't actually dislike `dash`, I just think that if possible, then it is preferable to use built-in functions/macros.

----

A tip for your endeavors removing `dash` from other packages: Don't split commits on file boundaries, that's ... not very helpful.  Instead split by `dash` forms.  Begin with trivial things such as `-when-let` => `when-let` (of course the same commit can also deal with `-if-let`).  And end with more complex changes where the removal of `dash` actually complicates matters.  By doing it like that you could possibly demonstrate how little gets complicated by the removal of `dash`.  Or if that isn't the case, then the maintainer could at least accept the trivial changes while rejecting the complicated ones.

sirikid:
Thank you, I'm glad you understand me.

> So in summary I am slowly migrating away from dash, a process which will likely take a few more years. In fact it may never come to an end because I don't actually dislike dash, I just think that if possible, then it is preferable to use built-in functions/macros.

Ok, I will resubmit parts of this patch mixed with actual improvements :^)

> A tip for your endeavors removing dash from other packages: Don't split commits on file boundaries, that's ... not very helpful. Instead split by dash forms. Begin with trivial things such as -when-let => when-let (of course the same commit can also deal with -if-let). And end with more complex changes where the removal of dash actually complicates matters. By doing it like that you could possibly demonstrate how little gets complicated by the removal of dash. Or if that isn't the case, then the maintainer could at least accept the trivial changes while rejecting the complicated ones.

I've tried this approach once and failed miserably. But perhaps this was due to the complexity of the library, and not to the approach itself.

tarsius:
> Ok, I will resubmit parts of this patch mixed with actual improvements

Wait. Better not. I would rather do that myself. But you could help by suggesting a list of functions that could be replaced a long with usage counts, and research since when the replacements are available.

sirikid:
> > Ok, I will resubmit parts of this patch mixed with actual improvements
> 
> Wait. Better not. I would rather do that myself. But you could help by suggesting a list of functions that could be replaced a long with usage counts, and research since when the replacements are available.

Why not? I mean real improvements like abstracting the repeating pattern into a utility function.

tarsius:
Yeah I am not sure what you mean by "actual improvements". :smiley: 

Maybe start with just a few improvements as examples.

tarsius:
That sounds useful but again, please begin with a smaller pull-request.

minad:
@tarsius Is this more realistic now given that Magit uses Compat? You mentioned `flatten-tree` above. I understand that this won't improve anything for the user, it is purely a maintenance issue.

tarsius:
Yes, "this" is more realistic now. Since you have taken over as `compat`'s maintainer, it has more steam again and it has become feasible to quickly make additions to it, should gaps be discovered. Thanks! :grinning: 

Removing the dependency on `dash` still isn't a priority, and doing so will be a long, relaxed process. A benefit of replacing uses of `dash` with builtin functions/macros, is that it gives me an opportunity to look at code I haven't looked at in a long time. I can use that to make further improvements.

At times I find that kind of work quite enjoyable and it is a good thing I have a nice reserve of that waiting for when I feel like spending some time on it. This kind of work can be an unsatisfying chore, when it is something that is forced as soon as it is possible, or if it comes in the form of huge pull-requests that come at a time when there is more important things to do. But when it is done when feeling like doing it, then it can be rewarding and pleasurable, and who knows, I might even spot a few bugs in the process; when it is forced, then it is more likely that I add bugs instead.

Currently my focus is elsewhere. I am mostly concerned with "chore prevention". I am trying to get all my projects into a state where it isn't *necessary* to do anything. That will make it possible to more quickly react to new requirements (such as bugs and feature requests), to take the time to deal with chores of low importance (when feeling like it), and ultimately to work on new exiting things or take existing things to the next level.

Or to put it in other words, I am trying hard to focus my time on things that I either want to get over with or that I expect to have a positive effect in the mid- and longterm (primarily on my motivation, but ultimately that will also increase my productivity). "Using more built-in functionality" is something that I do want to do, and that will be enjoyable to work on for the most part, but I don't think doing it rather sooner than later has the cumulative positive effect that focusing on other work first would have. So, I want to keep this as some simple work to do at my leisure, instead of forcing it now and thus turning it into a chore.

(The next task I will likely do, that was enabled by `compat` additions, is to cherry-pick https://github.com/magit/magit/commit/af3e748c6973d458b744d2efafaa858808d377fd.)

minad:
I fully agree. There is no hurry but such low priority trivial work can be quite satisfying - like some work meditation ;)

sirikid:
> I fully agree. There is no hurry but such low priority trivial work can be quite satisfying - like some work meditation ;)

Especially when you hate dash and have OCR

minad:
Well then you should probably train to to control your OCD :-P

I think there are technical arguments to be made that third-party "standard libraries" like dash are not the best idea. Bit the situation also improved considerably, it may have been a good idea back then. Nowadays for the whole ecosystem it is better to follow an ideomatic style. This will make it overall easier to read code and to contribute, in particular if we talk about major packages like Magit.

PR #4384, opened 2021-05-14
magit-read-file-trace: By default, suggest to trace the evolution of …

helixhorned:
…the current line

----

Rationale: assumption that this is the most frequently use case for `magit-log`/`-L`. Otherwise, it is necessary to glance at the mode line to find out the current line and enter it manually, which is error prone (for me), especially with source files with line counts in the thousands.

So, in the sequence of keystrokes to trace the evolution of the current line,

1. `magit-log` (can be bound to a keystroke)
2. `-L`
3. (accept the default -- current file)
4. (find out the current line and enter it)
5. `l`

this turns point 4 into just pressing Enter.

On a general note: I haven't found anything in the documentation concerning how to make sequences of input events like these into single key chords. (If it's a short Lisp function, that would be fine too.) My assumption is just that you'd want in general to have frequently used features bound to shorter key sequences.

On a positive note: `magit-log`/`-L` is really, really useful to me. Thanks for all the effort!

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -579,11 +579,12 @@ the upstream isn't ahead of the current branch) show."
   :reader 'magit-read-file-trace)
 
 (defun magit-read-file-trace (&rest _ignored)
-  (let ((file  (magit-read-file-from-rev "HEAD" "File"))
-        (trace (magit-read-string
-                "Trace" nil nil
-                ;; By default, suggest to trace the evolution of the current line.
-                (concat (number-to-string (1+ (current-line))) ",+1"))))
+  (let* ((file  (magit-read-file-from-rev "HEAD" "File"))
+         (trace (magit-read-string
+                 "Trace" nil nil
+                 (if (string= file (magit-current-file))
+                     ;; Suggest to trace the evolution of the current line.
+                     (format "%i,+1" (line-number-at-pos))))))
     (concat trace ":" file)))
 
 ;;;; Setup Commands
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -580,7 +580,10 @@ the upstream isn't ahead of the current branch) show."
 
 (defun magit-read-file-trace (&rest _ignored)
   (let ((file  (magit-read-file-from-rev "HEAD" "File"))
-        (trace (magit-read-string "Trace")))
+        (trace (magit-read-string
+                "Trace" nil nil
+                ;; By default, suggest to trace the evolution of the current line.
+                (concat (number-to-string (1+ (current-line))) ",+1"))))
     (concat trace ":" file)))
 
 ;;;; Setup Commands


tarsius:
Have you tried `C-c M-g t` to show the trace for the current definition. It's for the current definition, not line, but isn't that better?

There's no `current-line` function and by using `format` we can make it smaller: `(format "%i,+1" (1+ (line-number-at-pos)))`.

It only makes sense to use the location in the current file, if the user selected the current file at the preceding prompt, so we would have to detect if that is the case.

helixhorned:
> Have you tried `C-c M-g t` to show the trace for the current definition.

Unfortunately, when attempting this for an out-of-line definition of a C++ class member function, I'm getting:

```
user-error: No function at point found
```

`narrow-to-defun` works fine (I'm assuming they're related; haven't looked into Magit code).

> It's for the current definition, not line, but isn't that better?

That depends on the kind of code one is looking at, I would say. It might work if functions are short. In any case, the evolution of  one line is really what I'm querying for most of the time, as I'm interested in just one very specific change.

> There's no `current-line` function and by using `format` we can make it smaller: (...)

Sure, looks great! `current-line` points me to `/usr/share/emacs/26.3/lisp/array.el.gz`, but I guess that's an Emacs version thing then. I'm all for portability, of course.

> It only makes sense to use the location in the current file, if the user selected the current file at the preceding prompt, so we would have to detect if that is the case.

Ah, true. I'll have a look at that.

tarsius:
Try `C-c M-g l` after marking the region. I think that does exactly what you want.

helixhorned:
> Try `C-c M-g l` after marking the region. I think that does exactly what you want.

As of `magit-20210512.1949` from MELPA, as far as I can see, this is a shortcut for `magit-log`. It might be nice if having an active selection automatically activated `-L`, but it doesn't for me. Thanks for the hint though -- I'm learning new things about Magit as I go; I'm certainly not a power user.

helixhorned:
@tarsius Addressed your comments and manually tested the new code. Note: `1+` is now not needed -- the function you suggested already gives us a one-based line number.

kyleam:
> as far as I can see, this is a shortcut for `magit-log`

The keybinding that @tarsius mentioned (`C-c M-g l`) maps to `magit-log-buffer-file` by default.

> It might be nice if having an active selection automatically activated `-L`, but it doesn't for me

`magit-log-buffer-file` constructs the `-L` value based on the region bounds:

https://github.com/magit/magit/blob/fcd50dd8ae7cb33332eb7c6d90932be41b869cae/lisp/magit-log.el#L709-L714

helixhorned:
> > as far as I can see, this is a shortcut for `magit-log`
> 
> The keybinding that @tarsius mentioned (`C-c M-g l`) maps to `magit-log-buffer-file` by default.
> 
> > It might be nice if having an active selection automatically activated `-L`, but it doesn't for me
> 
> `magit-log-buffer-file` constructs the `-L` value based on the region bounds:
> 
> https://github.com/magit/magit/blob/fcd50dd8ae7cb33332eb7c6d90932be41b869cae/lisp/magit-log.el#L709-L714

True, can confirm now! Earlier, I had erroneously bound my own convenience key chord to `magit-dispatch` instead of `magit-file-dispatch`.

I guess that makes this PR redundant for the purpose I had intended for it, since the existing key sequence seems shorter [*] and on top more general.

[*] (when `magit-file-dispatch` is invoked by a key chord which is faster to type than `C-c M-g`)

Still, I think the suggested default for `magit-log`/`-L` is nice to have as a hint. Naturally, I'll leave the ultimate decision to you.

tarsius:
I'll leave it as is.

PR #4388, opened 2021-05-21
Consolidate git/libgit CI steps

lambdadog:
Merges the "Test with libgit" and "Test with git" CI steps into a singular "Test" step that uses string substitution to select the `make` target.

There should be no functional change here other than removing one disabled step for every test in the Actions interface.


tarsius:
Thanks!

PR #4395, opened 2021-05-30
Add indirect buffer feature to magit-ediff

geza-herman:
If the same buffer is open in ediff and normally, the normal version also gets ediff's highlights.  With `magit-ediff-use-indirect-buffers` get to `t`, ediff will use an indirect buffer, so the normal buffer won't get ediff's highlights.

Note I know that this PR is incomplete currently. If you're open to this idea, I can finish it:
 - add indirect support to other ediff functions (not just compare)
 - describe the new variable properly.

### DIFF
--- a/lisp/magit-ediff.el
+++ b/lisp/magit-ediff.el
@@ -101,6 +101,17 @@ tree at the time of stashing."
   :group 'magit-ediff
   :type 'boolean)
 
+(defcustom magit-ediff-use-indirect-buffers nil
+  "Whether use indirect buffers when magit executes an ediff session.
+
+If non-nil, magit will create an indirect buffer for each input buffer
+to ediff.  This can be useful if an input buffer is also visible in a
+non-ediff window, because the highlights done by ediff won't affect
+non-ediff windows."
+  :package-version '(magit . "3.0.0")
+  :group 'magit-ediff
+  :type 'boolean)
+
 ;;; Commands
 
 (defvar magit-ediff-previous-winconf nil)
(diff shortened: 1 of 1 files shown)


tarsius:
> If you're open to this idea

:+1:

geza-herman:
I've implemented something, hopefully it is more or less correct. I'm not sure about the solution of `magit-ediff-stage`, this function does some workaround for read-only-ness, I'm not sure that it still works correctly with my version.

tarsius:
> I've implemented something, hopefully it is more or less correct.

I've also implemented something, but have not tested it once, so it is probably broken. Anyways, you can see it in the `ediff` branch. I am mostly posting this now, so that I did not not reply for three weeks. :roll_eyes:

> I'm not sure about the solution of `magit-ediff-stage`

I've just not added the feature to that and `magit-ediff-resolve`. I have some vague feeling that one would not want it in those special cases. But I am not sure, I don't actually use Ediff.

Which reminds me, could you please explain *why* we want indirect buffers in the first place? It seems that was clear to me three weeks ago, but now I need a refresher.

Anyway, a problem with the existing code is all the repeated boilerplate, which is due to limitations of Ediff's API. I am usually not against a bit of repetition but here it already was bad and with the addition of the indirect buffer stuff it is getting out of hand. So my version tries to reduce the boilerplate first. Using a macro, which I would normally also avoid, but you know... the api.

geza-herman:
> I've also implemented something, but have not tested it once, so it is probably broken. Anyways, you can see it in the `ediff` branch.

I had a cursory look on it, it seems to work.

> I've just not added the feature to that and `magit-ediff-resolve`. I have some vague feeling that one would not want it in those special cases. But I am not sure, I don't actually use Ediff.

Yep, that's a solution. But a user could expect that this new option works everywhere.

 
> Which reminds me, could you please explain _why_ we want indirect buffers in the first place? It seems that was clear to me three weeks ago, but now I need a refresher.

Well, I didn't give the reason, so let me explain where this can be useful (at least, for me). Suppose that you want to apply a diff to your worktree (this diff is a result of diffing the worktree with some git rev), and you want to make several commits out of the diff. So, you apply some part of the diff, maybe do some little additional modifications, and commit. Then, refresh the diff, and repeat the process. Until the diff becomes empty. Also, you may want to check the worktree before commiting that it is in OK condition (for example, it compiles and runs OK).

I use this process, for example, when I want to split a commit into separate commits: in one tab, I have the diff (created by some magit diffing command), and I execute `ediff`. In another tab, I compile the code. If the code fails to compile, the failing source code is loaded. So I may end up having the same file in the first tab (ediffing the file) and in the second tab (fixing the compilation error). I like to use tabs, because if I didn't (and just start compilation from `ediff`), then the compilation process destroys the window config that `ediff` setup. Sure, this could be fixed somehow. But I think that tabs make this process more clean (when fixing compilation errors, I don't want to see the diff).

This way of working is also helpful when resolving complicated conflicts: in one tab I see the `ediff` of the conflict, and in another tab I can compile the current resolved version and fix compilation errors.

The problem that this PR tries to solve is that without indirect buffers, I see `ediff`'s highlights in the second tab. With this PR, there are no `ediff` highlights in the second tab. It's just a cosmetic issue, of course.

> Anyway, a problem with the existing code is all the repeated boilerplate, which is due to limitations of Ediff's API. I am usually not against a bit of repetition but here it already was bad and with the addition of the indirect buffer stuff it is getting out of hand. So my version tries to reduce the boilerplate first. Using a macro, which I would normally also avoid, but you know... the api.

Yep, I saw the repeated code. I just thought that it already had some repetition, so it's not much worse if I added a little bit more :)
Anyways, feel free to close this PR, I suppose that your `ediff` branch is a better way to solve this problem.

tarsius:
I've merged my implementation after working some more on it.

The only command that doesn't support using indirect buffers is `magit-ediff-resolve`. It doesn't have to do so because it uses `smerge-ediff`, which itself creates new buffers.

tarsius:
@geza-herman please test this a bit and report back if you find any issues. Thanks!

geza-herman:
Sure, thank you for implementing this, I'll start using it right away!

geza-herman:
I found some issues: https://github.com/magit/magit/issues/4440

(I would have fixed these myself, but unfortunately I'm not skilled enough in lisp, the new code is beyond my reach, I need to learn how macros work)

PR #4396, opened 2021-05-31
Fix git credential manager core by adding a newline to the chosen option

### LINKED ISSUE #4393: git-credential-manager-core helper doesn't ask for access token
I added the `magit-process-git-credential-manager-core` to the `magit-process-prompt-functions` hook. And then `push`ed inside `magit-status`, I was asked to select an authentication method, with prompt `option`. I entered `2` for Personal access token:

```
Select an authentication method for 'https://github.com/':
  1. Web browser (default)
  2. Personal access token
option (enter for default): 2
```
I expected `magit` to ask for the `token`:

```
Enter GitHub personal access token for 'https://github.com/'...
Token:
```

But after I entered `2` as the option, nothing happened. No errors, no operations, no messages or output.

For reproducing, setup the `git-credential-manager-core`, then push inside a project.

Magit version:
Magit 20210525.844, Git 2.31.1, Emacs 27.2, gnu/linux

Unrelated to this issue, `magit-emacs-Q-command` lacked the `-L magit-section` package.

CIAvash:
fixes #4393

- Add a newline to the option user entered, so that it gets submitted to git prompt.
- Insert the option in the process buffer, so user sees what option was chosen. Also newline puts the next git message on a newline.

I tried using `magit-process-kill-on-abort`, so the git process gets killed when user cancels with `C-g`. But was unsuccessful. Canceling any of `Option` or `Token` prompts, do not kill the git process.

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -846,9 +846,10 @@ To use this function add it to the appropriate hook
   (and (string-match "^option (enter for default): $" string)
        (progn
          (magit-process-buffer)
-         (process-send-string
-          process
-          (format "%c" (read-char-choice "Option: " '(?\r ?\j ?1 ?2)))))))
+         (let ((option (format "%c\n"
+                               (read-char-choice "Option: " '(?\r ?\j ?1 ?2)))))
+           (insert-before-markers-and-inherit option)
+           (process-send-string process option)))))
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks!

PR #4399, opened 2021-06-06
Snappy, speedy magit-list-repositories

Boruch-Baum:
The first time I evaluated function `magit-list-repositories' it took
so long (> 20 seconds) that I thought it had hung. This commit
addresses that and adds functionalities to the feature.

The code defines a user-configurable set of presentation styles and
defaults to one that has a very short response-time and that gives the
basic information to let the user decide which repository to jump to /
get the status of (the only major-mode-specific keybinding of the
result buffer).

A keybinding 'n' is configured to new function
`magit-list-repositories-next' which cycles through presentation
styles. The default configuration defines styles with 'quick'
response-times first. This new function can be used as a replacement
for `magit-list-repositories'.

A "working..." message is added to let the user know this operation
can take a while.

The default sort column for the result buffer was changed from
being hard-coded to being a defcustom. This was the only change
made to function `magit-repolist-mode'. FWIW, my opinion is that
the default should be "Name", not "Path": 1) Paths can be long, and their
most important (right-most) part is most likely to be truncated by the
buffer window width; 2) Duplicate names become easier to spot.

A new function `magit-repolist-column-flags' offers a column to
present all the flags of variable
`magit-repolist-column-flag-alist' instead of only the first, as
offered by the current function `magit-repolist-column-flag'.

A feature was added to the present style specification to allow the
column width field to be an expression that evaluates to a numeric
column width.

A follow-up commit could supplement function
`magit-repolist-column-flags' to include columns for stashes (heading
'z'), unpulled (heading '↓'), and unpushed (heading '↑'). For each
these, the actual value would be either: blank, for none; an integer
1-9; or '+' for greater than 9. I like this presentation
because it's horizontally compact.

### DIFF
--- a/lisp/magit-repos.el
+++ b/lisp/magit-repos.el
@@ -153,7 +153,7 @@ Each element of the list should be a CAR value of variable
                                    (memq v (mapcar 'car magit-repolist-styles)))))
   :group 'magit-repolist)
 
-(defvar-local magit-repolist-current-style (car magit-repolist-desired-styles))
+(defvar-local magit-repolist-current-style nil)
 
 (defvar magit-repolist-status-flag-alist
   '(("N" . magit-untracked-files)
--- a/lisp/magit-repos.el
+++ b/lisp/magit-repos.el
@@ -118,6 +118,47 @@ as the value of `magit-repolist-column-flag'."
   :type '(alist :key-type (function :tag "Predicate Function")
                 :value-type (string :tag "Flag")))
 
+;; (defcustom magit-repolist-sort-column "Path"
+(defcustom magit-repolist-sort-column "Name"
+ "Default sort column for function `magit-list-repositories'.
+  This string should be the CAR of an entry in variable
+  `magit-repolist-columns'."
+  :type 'string
+  :group 'magit-repolist)
+
+(defcustom magit-repolist-styles
+  '((simple .
+      (("Name" 25 magit-repolist-column-ident nil)
+       ("Path" 99 magit-repolist-column-path nil)))
+    (versioned .
+      (("Name" 25 magit-repolist-column-ident nil)
+       ("Version" 25 magit-repolist-column-version nil)))
+    (status .
+      (("Unpushed" 8 magit-repolist-column-unpushed-to-upstream
+        ((:right-align t)
+         (:help-echo "Local changes not in upstream")))
+       ("Unpulled" 8 magit-repolist-column-unpulled-from-upstream
+        ((:right-align t)
+         (:help-echo "Upstream changes not in branch")))
+       ("Flags"    (length magit-repolist-column-flag-alist) magit-repolist-column-flags nil)
+       ("Name"     25  magit-repolist-column-ident nil)
+       ("b"         2  magit-repolist-column-branches ((:right-align t)))
+       ("branch"   15  magit-repolist-column-branch   nil)
+       ("stashes"   2  magit-repolist-column-stashes  ((:right-align t))))))
+  "List of display styles for function `magit-list-repositories'.
+Each entry should be a CONS whose car is a symbol identifying the
+style, and whose CDR is of a form suitable for variable
+`magit-repolist-columns'."
+  :type 'list
+  :group 'magit-repolist)
+
+(defcustom magit-repolist-current-style 0
+  "Index into variable `magit-repolist-styles'.
+It will be the initial style presented when evaluating function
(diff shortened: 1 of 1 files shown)


tarsius:
* I think `magit-repolist-column-flags` should be added in a separate commit, before all the rest. (And use `pcase-let` like in `magit-repolist-column-flag`. The line-break after `mapconcat` seems unnecessary.)
* `magit-repolist-sort-column` should probably also be added in a separate commit.

Boruch-Baum:
On 2021-06-09 13:32, Jonas Bernoulli wrote:
> * I think magit-repolist-column-flags should be added in a separate
>   commit, before all the rest. (And use pcase-let like in
>   magit-repolist-column-flag. The line-break after mapconcat seems
>   unnecessary.)

> * magit-repolist-sort-column should probably also be added in a
>   separate commit.

OK, then. I just finished re-doing those as separate commits and pull
requests and also a third for the "working.." message.

You didn't give me feedback on the main operation for all this effort,
which was to offer a speedy and snappy default listing, with option to
subsequently display data that can have a long wait time (in my case ~20
seconds, for only about a dozen repositories, none particularly large,
all on the same local hdd).

--
hkp://keys.gnupg.net
CA45 09B5 5351 7C11 A9D1  7286 0036 9E45 1595 8BC0

tarsius:
> You didn't give me feedback on the main operation for all this effort,

Yes, sorry. I'll try to get to it soon.

tarsius:
I am attempting to review again. As you can see I have pushed a few times. To make it easier for both of us to see what has happened and who did it, I reverted these changes. Then I rebased on `master` and pushed that, except for that everything is how you submitted it.

I appreciate your effort to make things easier to consume for me, but I am afraid the result was the opposite. There is a middle ground between "several somewhat related changes in one commit" and "one pull-request per topic" and it is "on COMMIT per change/topic".

tarsius:
Supporting different styles seems a worthwhile goal.

I am less convinced of the cycle idea.

The refs buffer used cycling to allow switching between informative and fast. The result was really awful; to go to "fast" one had to first cycle through various increasingly fast variants. I.e. in order to get to "fast" one had to endure several instances of "slow", which defeated the purpose entirely.

Magit's blaming support comes with a cycling command, which should be inspected for inspiration.

A different approach could be to refactor existing code so that users could variants of `magit-list-repository`, whose body would essentially consist of `(magit--list-repository MY-COLUMNS)`.

Even if we provide some variants out of the box (though I am not convinced we should) we might want to do so using a dispatcher instead.

In short, I am very undecided about pretty much all of this, and am not motivated to investigate and decide now. I will try to pick out some bits that don't require major design decisions, like the new flag inserter variants.

tarsius:
b19263006a1964fe8f9ae85d43429aa05023206d

Boruch-Baum:
On 2021-07-21 08:58, Jonas Bernoulli wrote:
>    Closed [1]#4399 via [2]0ea7868.

Thanks.

So if now the feature *is* snappy and speedy, my closing suggestion is
to promote its use front-and-center in the documentation to prominence
on par with the other entry points (ie. magit-status, magit-dispatch,
magit-file-dispatch).

When I saw that this command existed, I envisioned it as a nice
alternate entry point to magit. All the other entry points presume
something about the current buffer, so one needs to first navigate
there. OTOH, this entry point gives a nice global overview that allows a
user to quickly swoop in to any configured repo from anywhere in emacs
without any of that. For me, that's a much more valuable entry point
than the other three, and I expect many other users will agree.

These past few weeks I've been distracted with "life-stuff" and haven't
been following the play-by-play for this PR and the related PR's. That's
why you haven't been hearing from me, and why I haven't been
collaborating with you in real-time as you made your commits
play-by-play. I hope to review your finished product Sunday but even
that I can't commit.



--
hkp://keys.gnupg.net
CA45 09B5 5351 7C11 A9D1  7286 0036 9E45 1595 8BC0

tarsius:
Currently you would still have to add something like this to your init file:

```elisp
(defun magit-list-repositories-simple ()
  (interactive)
  (magit-repolist-setup
   '(("Name" 25 magit-repolist-column-ident nil)
     ("Path" 99 magit-repolist-column-path nil))))

(defun magit-list-repositories-versioned ()
  (interactive)
  (magit-repolist-setup
   '(("Name"    25 magit-repolist-column-ident nil)
     ("Version" 25 magit-repolist-column-version nil))))
```

PR #4407, opened 2021-06-10
[crm] Override split-string more safely

jakanakaevangeli:
(To be applied to the crm branch.)

In `magit-completing-read-multiple*`, this avoids changing the return value of
`split-string`. With this, selectrum's crm function works correctly instead of
signaling an error.

(The first commit doesn't change behaviour, only formatting, to make the second
commit easier.)


tarsius:
Thanks!

I have applied this to the `crm` branch to avoid having to keep rebasing two branches until I finally get around to finishing this up.

PR #4409, opened 2021-06-14
magit-repolist-mode: configurable sort-column

Boruch-Baum:
+ The default sort column for the result buffer was changed from being
  hard-coded to being a defcustom.

+ The default sor column was changed from "Path" to "Name"

  1) Paths can be long, and their most important (right-most) part is
     most likely to be truncated by the buffer window width;

  2) Duplicate names become easier to spot.

+ This is the first of several expected PR's to redo #4399 which

### DIFF
--- a/lisp/magit-repos.el
+++ b/lisp/magit-repos.el
@@ -118,6 +118,13 @@ as the value of `magit-repolist-column-flag'."
   :type '(alist :key-type (function :tag "Predicate Function")
                 :value-type (string :tag "Flag")))
 
+(defcustom magit-repolist-sort-column "Name"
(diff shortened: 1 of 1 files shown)


tarsius:
I cannot decide on this until after I have decided on whether I want something like `magit-repolist-styles`. If I decide in favor of the latter, then the sort column should be configurable per style, i.e. as part of  `magit-repolist-styles`, not using a separate option.

tarsius:
Actually, let's merge it anyway. It's not a big deal if we have to remove it later.

tarsius:
I've replaced the option, see the above commit.

PR #4410, opened 2021-06-14
Add message to indicate working in progress

Boruch-Baum:
+ This is the second of several expected PR's to redo #4399 which

+ The first time I evaluated function `magit-list-repositories' it took
  so long (> 20 seconds) that I thought it had hung. This commit
  is one step in addressing that in just simply letting the user know
  the operation may take a while.

  + A follow-up could be to set-up a timer to add a period every
    two-seconds until the operation is complete. I didn't include that
    in this commit because in the final product of #4399 the initial
    run of the command is so much speedier, but it could be done
    anyway.

  + A follow-up could be to figure out why the operation is so slow to
    begin with, but that wasn't the direction of #4399.

### DIFF
--- a/lisp/magit-repos.el
+++ b/lisp/magit-repos.el
@@ -129,6 +129,7 @@ repositories are displayed."
   (interactive)
   (if magit-repository-directories
       (with-current-buffer (get-buffer-create "*Magit Repositories*")
+        (message "magit-list-repositories: working...")
         (magit-repolist-mode)
         (magit-repolist-refresh)
         (tabulated-list-print)


tarsius:
I've used a more conventional message, including the final "...done" message. And I have done the same for `magit-list-submodules`.

Boruch-Baum:
On 2021-06-14 05:50, Jonas Bernoulli wrote:
>    I've used a more conventional message, including the final "...done"
>    message. And I have done the same for magit-list-submodules.

Great. Whatever works.

--
hkp://keys.gnupg.net
CA45 09B5 5351 7C11 A9D1  7286 0036 9E45 1595 8BC0

PR #4411, opened 2021-06-14
New options for columns

Boruch-Baum:
+ This is the third of several expected PR's to redo #4399 which

+ A new function `magit-repolist-column-flags' offers a column to
  present all the flags of variable `magit-repolist-column-flag-alist'
  instead of only the first, as offered by the current function
  `magit-repolist-column-flag'.

+ A new function `magit-repolist-status-flags' offers a column to
  present a more complrehensive collection of flags than does function
  `magit-repolist-column-flags', to include columns for stashes (heading
  'z'), unpulled (heading '↓'), and unpushed (heading '↑'). For each
  these, the actual value is either: blank, for none; an integer 1-9; or
  '+' for greater than 9.

  + The motivation for this presentation option is that it's
    horizontally compact, comprehensive, and hopefully with the blanks
    instead of zeros won't be often terribly crowded to read.

+ This commit DOES NOT actually force these options onto the user.

  + That is expected to be another redo of aforesaid PR #4399
    which included this and more, bundled into a single neat
    complete package with a few more features yet (see there).

### DIFF
--- a/lisp/magit-repos.el
+++ b/lisp/magit-repos.el
@@ -37,6 +37,17 @@
 
 (defvar x-stretch-cursor)
 
+(defvar magit-repolist-status-flag-alist
+  '(("N" . magit-untracked-files)
+    ("U" . magit-unstaged-files)
+    ("S" . magit-staged-files)
+    (" " . (lambda (x) " "))
+    ("↓" . magit-repolist-column-unpulled-from-upstream)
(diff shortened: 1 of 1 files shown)


tarsius:
I've added your `magit-repolist-column-flag`.

Instead of `magit-repolist-status-flags` I took a more flexible approach, see 8005b8b2df89e51a721d887418fb116ac89ac831 (and the two commits leading up to that).

Issue #4394, opened 2021-05-28
Expose git diff's -R (swap two inputs) functionality


geza-herman:
(This is a very minor feature request.)

Magit already has a similar functionality to `git diff -R`, flip revisions. But this only works when comparing two revisions. If I'd like to compare the worktree to a revision, this option is not available.

I'd like to use `-R` in the following scenario: suppose that I'd like to apply hunks from another commit to the worktree. I do it by using `magit-diff-working-tree`. But in the diff, I need to use `reverse` instead of `apply` to copy a hunk to the worktree, it needs backward thinking. It would be good if I could switch the roles of the worktree and the commit. Because in this case I'm interested in "what the other revision contains that my worktree doesn't", instead of "what modifications my worktree has compared to some other revision".

PR #4413, opened 2021-06-15
reverse diff (against working tree)

dickmao:
git diff -R


tarsius:
Also see #4394.

tarsius:
I've added the argument but did it a bit differently and hide it by default.

tarsius:

Issue #3940, opened 2019-08-09
allow jumping to fork-point in log buffer


edgimar:
It would be convenient in the log buffer to have a magit command which performs a `git merge-base --fork-point <rev>` on the currently selected commit, and then jumps to that commit in the log buffer.  This is convenient for determining the branch that another was forked from, knowing how long ago the fork occurred, and knowing what the initial state of the branch was.

edgimar:
One other idea might be to have this be a set of commands for navigating in the buffer along a particular branch.  So currently up/down just go to the newer/older commits in the graph log view, but it would also be nice to have a version of 'up' and 'down' which stay on the current branch, and a version of 'Ctrl-up' and 'Ctrl-down' that stay on the current branch.  The 'Ctrl-down' variant would move to the fork-point, and the 'Ctrl-up' variant would move to HEAD of the currently selected branch.

tarsius:
The arguments are `git merge-base --fork-point <ref> [<commit>]`. We would use the commit at point as `<commit>`. How is Magit to know what should be used for `<ref>`?

edgimar:
There will be one or more <ref> associated with <commit>, so if there is one, just choose it, and if more than one, allow the user to select which one.

tarsius:
Maybe `magit-log-move-to-parent` is all you need?

edgimar:
No, `magit-log-move-to-parent` isn't the same -- what I'm proposing walks back over many commits until it finds where the chosen branch was forked off of another branch, but `magit-log-move-to-parent` doesn't seem to do this.

edgimar:
I realize that what I'm looking for isn't exactly the same as what `merge base --fork-point` performs.  Really what I'm hoping for is something that for a current/selected branch looks at all of the most recent common-ancestors with all other branches, and chooses the most recent of them all.

tarsius:
Something like this?

```emacs-lisp
(defun magit-log-previous-crotch ()
  (interactive)
  (forward-line -1)
  (if (re-search-backward "|/" nil t)
      (forward-line)
    (user-error "No more crotches")))

(defun magit-log-next-crotch ()
  (interactive)
  (if (re-search-forward "|/" nil t)
      (forward-line)
    (user-error "No more crotches")))
```

edgimar:
How would that work if you are viewing a log with multiple branches?  The goal would be to stay on the current branch, even when there are parallel branches that have more recent fork-points with other parallel branches.

tarsius:
[...]

Issue #3853, opened 2019-04-30
Jump to a branch in log buffer


yurikhan:
I’m in a log buffer (`l a` with `--graph --color --decorate -n"4096"`, no `--simplify-by-decoration`), pruning old local branches. In another window, I have a list of local branches (`y`).

In the list of branches, I see a branch name. In the log buffer, I want to jump to that branch, so that I see it in the context of the commit graph — whether it has been merged into the development branch, whether it has been rebased; basically, whether I can safely delete it.

Unsatisfying workarounds:

* I can start an isearch and type the branch name, but that can give lots of false positives if the branch name is a word commonly used in commit message summaries.

* I can start an isearch and type in a part of a commit message summary, but that is a lot of typing and inconvenient. It can also give false positives if the branch has been rebased a few times.

The obvious user interface would be a key binding in `magit-log-mode`, first prompting for any commitish (a prefix of a commit hash, a branch name, a tag name, a remote ref name, the symbolic reference `HEAD`, or any history expression; with completion for any references) and moving point to that commit’s line and decoration if applicable. I’d maybe even bind it to something starting with `M-g`.

Bonus points for a key binding in any magit buffer mode where a commitish can be displayed, to grab the commitish thing at point and jump to it in the appropriate log buffer.

tarsius:
> whether it has been merged into the development branch, whether it has been rebased; basically, whether I can safely delete it.

These questions can be answered by using the features available in the refs buffer. Some of those features are disabled by default, so read the documentation: https://magit.vc/manual/magit/References-Buffer.html.

As for the "jump to this ref in a log" command that you are proposing, that sounds useful and simple enough, but doing it properly would likely be a lot of work. I don't intend to do that anytime soon, so here is a minimal viable implementation instead:

```emacs-lisp
(defun magit-log-jump-to-branch (rev)
  (interactive (list (magit-read-branch-or-commit "In log, jump to")))
  (with-current-buffer
      (cond ((derived-mode-p 'magit-log-mode)
             (current-buffer))
            ((when-let ((buf (magit-get-mode-buffer 'magit-log-mode)))
               (pop-to-buffer-same-window buf)))
            (t
             (apply #'magit-log-all-branches (magit-log-arguments))))
    (magit-log-goto-commit-section (magit-rev-abbrev branch))))
```

tarsius:
I've added this command as `magit-log-move-to-revision`, see #4418.

PR #4418, opened 2021-06-21
magit-log-jump: new function: jump to branch/tag/commit in log buffer

danielfleischer:
- Function to **jump to different location in a log buffer** based on branches, tags (with completion)
   and commits (without completion).

- If the item selected (based on `magit-read-other-branch-or-commit`) does not appear in the log
  buffer, do nothing, notifying the user. Perhaps other function can be used to limit the selections
  only to content in the current log buffer.

- Shortcut was chosen as `j` because in other buffers it's also associated with jumping.

- Documentation updated.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -2616,6 +2616,11 @@ the status buffer.
   first parent, but a numeric prefix can be used to specify another
   parent.
 
+- Key: j, magit-log-jump
+
+  Cursor jump to a given branch, tag or commit. If item does not
+  appear in the current log buffer, do nothing. 
+
 - Key: SPC, magit-diff-show-or-scroll-up
 
   Update the commit or diff buffer for the thing at point.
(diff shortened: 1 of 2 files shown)


tarsius:
I'm not sure I want to add that.

Here's another implementation: https://github.com/magit/magit/issues/3853.

Also somewhat related: https://github.com/magit/magit/issues/3940.

danielfleischer:
The other implementation is similar. I reckon the feature is useful as a way to orient yourself inside large graphs. It's just a jumping functionality, with a possible improvement to the UX. 

Do you not like the keybinding or whether it has a universal appeal or missing some edge cases in the given code?

Either way, you decide, so feel free to close the issue and anyone can add the code to their personal config.

tarsius:
I've merge my version. After adding the "no such ref" warning from yours. Cheers!

PR #4420, opened 2021-06-23
Update our completing-read-multiple replacement/wrapper

tarsius:


lisp/magit-utils.el
```diff
@@ -584,37 +592,52 @@ into a list."
...
-          (split-string input crm-separator t)))
-    (remove-hook 'choose-completion-string-functions
-                 'crm--choose-completion-string)))
+TABLE and take an additional argument NO-SPLIT, which causes
+the user input to be returned as a single unmodified string.
+Also work around various misfeatures of various third-party
+completion frameworks."
+  (cl-letf*
+      (;; To implement NO-SPLIT we have to manipulate the respective
+       ;; `split-string' invocation.  We cannot simply advice it to
```
kyleam, line 605: s/advice/advise/

lisp/magit-utils.el
```diff
@@ -584,37 +592,52 @@ into a list."
 
 (defun magit-completing-read-multiple*
     (prompt table &optional predicate require-match initial-input
-            hist def inherit-input-method)
+            hist def inherit-input-method
+            no-split)
```
kyleam, line 596: Perhaps not worth worrying about, but to future-proof this against conflicts if an argument is every added in the upstream signature, I wonder if we should instead let-bind a defvar to signal `no-split`.

lisp/magit-utils.el
```diff
@@ -584,37 +592,52 @@ into a list."
...
+       ;; Prevent `IVY' from messing up our existing order. c7af78726
+       (ivy-sort-matches-functions-alist nil)
+       ;; Prevent `HELM' from messing up our existing order.  6fcf994bd
+       (helm-completion-in-region-default-sort-fn nil)
+       ;; Prevent `HELM' from automatically appending the separator,
+       ;; which is counterproductive when NO-SPLIT is non-nil and/or
+       ;; when reading commit ranges. 798aff564
+       (helm-crm-default-separator
+        (if no-split nil helm-crm-default-separator))
+       ;; And now, the moment we have all been waiting for...
```
kyleam, line 636: :)

kyleam (commented):
I haven't tested this out yet (I should be able to get to that tomorrow night), but the direction looks good to me.  Thanks for revisiting this.

> We cannot keep using the former name, because of the differences in the function signatures.

"cannot" is a bit strong :), but, yeah, moving to a compatible signature sounds good to me.  Of the reasons in the commit message, I find the "don't bypass advice" one the most compelling.

> Eventually we will inverse of direction of the deprecation.

I don't understand what this is referring to.  `magit-completing-read-multiple` is marked as obsolete and will be removed in a later release.  What's being inverted?

> Because this commit shuffles things around quite a bit, making history inspection harder, I have added a code comment for each issue that our wrapper addresses.

Nice, thanks.

tarsius:
>> Eventually we will inverse of direction of the deprecation.

> I don't understand what this is referring to. 

Does this make it clearer?:

Now the time has come to deprecate `magit-completing-read-multiple`
in favor of `magit-completing-read-multiple*`.  We should stop using
the former name for now, because of the differences in the function
signatures.  Eventually we would like to use the former name again,
but if we abruptly changing the signature of an existing function,
that would risk breaking third-party packages.  The use of an
intermediate function name minimizes the risk to third-parties that
ignore warnings for several releases.

lisp/magit-utils.el
tarsius: Yeah, I was thinking about that too but concluded that it is worth the risk. I am betting on "if they add a new argument, then they do that by first creating a new function, which uses keyword arguments instead of a long list of positional arguments". I think I saw some discussion that was going in that direction but might completely misremember.

kyleam:
> Does this make it clearer? [...]

Yes, thanks.

---

I've tested crm sorting (mostly via `magit-log-other`) as well as non-crm sorting (mostly via `magit-checkout`) with the following combinations and didn't notice any issues: `emacs-{25,27} x {built-in,helm-mode,ivy-mod} x magit-list-refs-sortby={nil,"-creatordate"}`

kyleam (approved):

tarsius:
And I've tested with `vertico-mode` and `selectrum-mode`.

PR #4423, opened 2021-06-28
Fix magit-log-select-quit: actually call quit func

tecosaur:
Unless I'm missing something, this is a rather trivial fix.

Since `magit-log-select-quit-function` is a buffer-local variable, killing the buffer then calling the function means that nothing is ever called. The quit function should be temporarily bound so it isn't lost.


tarsius:
Thanks!

PR #4429, opened 2021-07-06
magit-log: add option to show notes

FirstLoveLife:
magit-log cannot show notes in now, add an option to support it.


tarsius:
IMO It's better to just add `%+N` to the default value of `magit-log-revision-headers-format`.

PR #4441, opened 2021-07-13
magit-blame: Add --first-parent

To1ne:
Passing the --first-parent option to git-blame(1) will make git only
follow the first parent commit upon seeing a merge commit. This option
can be used to determine when a line was introduced to a particular
integration branch, rather than when it was introduced to the history
overall.

### DIFF
--- a/lisp/magit-blame.el
+++ b/lisp/magit-blame.el
@@ -895,6 +895,7 @@ instead of the hash, like `kill-ring-save' would."
   ["Arguments"
    ("-w" "Ignore whitespace" "-w")
    ("-r" "Do not treat root commits as boundaries" "--root")
+   ("-P" "Follow only first parent" "--first-parent")
    (magit-blame:-M)
    (magit-blame:-C)]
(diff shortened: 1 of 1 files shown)


To1ne:
`git blame --first-parent` was introduced in https://github.com/git/git/commit/95a4fb0eac

tarsius:
We should probably bind `-P` because `-p` is `--porcelain`. Other arguments are `-C` and `-M`, so `-P` fits nicely. Also I would drop the "upon seeing merge commit", that seems unnecessary and makes that line quite long.

To1ne:
> We should probably bind -P because -p is --porcelain. Other arguments are -C and -M, so -P fits nicely.

Done.

> Also I would drop the "upon seeing merge commit", that seems unnecessary and makes that line quite long.

Sure. Modified.

@tarsius Please have another look.

strickinato:
Thank you for doing this!!

Just came here to make an issue about it - and sure enough - Boom!!

PR #4445, opened 2021-07-16
magit-extras: Check that `project-switch-commands' is defined

leungbk:
This works around old versions of `project.el`.

### DIFF
--- a/lisp/magit-extras.el
+++ b/lisp/magit-extras.el
@@ -157,6 +157,7 @@ to nil before loading Magit to prevent \"m\" from being bound.")
   ;; `project-switch-commands', though project.el is available in Emacs 25.
   (when (and magit-bind-magit-project-status
              (boundp 'project-prefix-map)
+             (boundp 'project-switch-commands)
              ;; Only modify if it hasn't already been modified.
              (equal project-switch-commands
                     (eval (car (get 'project-switch-commands 'standard-value))


kyleam:
Thanks for the PR.  That seems safe enough and was checked before fb25d4e9 (magit-extras: Update project.el integration, 2021-03-11, gh-4343).

But I'm confused on how this error can be triggered.  It looks like `project-prefix-map` was added in Emacs's 2f231fcfb7 (Add global bindings for project commands, 2020-06-18), which would have been part of the `project` release following 0.4.0.  And the `project-switch-commands` rename happened in 70824683fd (Rename 'project-switch-menu' to 'project-switch-commands', 2020-05-26), which would have been part of the `project` release following 0.3.0.  So, as far as I can tell, `project-prefix-map` was introduced _after_ `project-switch-commands`.  Why isn't the existing `(boundp 'project-prefix-map)` a sufficient guard against the issue you're running into?

@basil-conto, any thoughts (on this change in general or my confusion above)?

---

[ The test failures are due to an unrelated regression from d477163d (2021-07-12) ]

basil-conto:
> But I'm confused on how this error can be triggered.

Which error, though?  @leungbk, could you please describe in more detail what problem you encountered, in which version, etc.?

> So, as far as I can tell, `project-prefix-map` was introduced *after* `project-switch-commands`.

Yes, `project-prefix-map` was introduced in the following commit:

> Add global bindings for project commands
> 2f231fcfb7 2020-06-18 02:05:31 +0300
> https://git.sv.gnu.org/cgit/emacs.git/commit/?id=2f231fcfb763626b8a4ede7da0f80da14a122ca9

At which point `project-switch-commands` already existed, so checking that it's `boundp` is a harmless no-op.

So without more information, I can only speculate that @leungbk is either facing some installation snafu, perhaps by having multiple versions of `project` lying around, or is trying to get Magit's integration working with an older version of `project`.  I don't see how the current patch would address either of those, though.

Issue #4433, opened 2021-07-07
OSX Mojave - x24 speed increase


gerrywastaken:
Reading [FAQ -> Issues & Errors -> Magit is slow -> Performance](https://magit.vc/manual/magit/Performance.html#Performance) gave me no hints that I should set the absolute path to git.

```
|0.173s|git diff-files -z --name-only --ignore-submodules --
|0.009s|/usr/local/bin/git log --format=%h%x0c%D%x0c%x0c%aN%x0c%at%x0c%s --decorate=full -n256 --use-mailmap --no-prefix ..@{upstream} --
|0.170s|git rev-parse --short HEAD~
|0.175s|git rev-parse --short HEAD
|0.009s|/usr/local/bin/git log --format=%h%x0c%D%x0c%x0c%aN%x0c%at%x0c%s --decorate=full -n256 --use-mailmap --no-prefix @{upstream}.. --
|0.006s|/usr/local/bin/git rev-parse @{upstream}
|0.007s|/usr/local/bin/git merge-base --is-ancestor HEAD origin/main
|0.171s|git rev-parse --verify refs/stash
|0.007s|/usr/local/bin/git diff --ita-visible-in-index --cached --no-ext-diff --no-prefix --
|0.171s|git rev-parse --is-bare-repository
|0.008s|/usr/local/bin/git diff --ita-visible-in-index --no-ext-diff --no-prefix --
|0.185s|git status -z --porcelain --
|0.169s|git rev-parse --git-dir
|0.169s|git describe --contains HEAD
|0.173s|git describe --long --tags
|0.168s|git rev-parse --verify origin/main
|0.168s|git show --no-patch --format=%s origin/main^{commit} --
|0.167s|git rev-parse --verify --abbrev-ref main@{upstream}
|0.168s|git show --no-patch --format=%h %s HEAD^{commit} --
|0.165s|git symbolic-ref --short HEAD
|0.170s|git config --list -z
|0.167s|git rev-parse --show-toplevel
|0.173s|git rev-parse --verify HEAD
|0.016s|/usr/local/bin/git update-index --refresh
---
24 calls in 2.963s
```

Many hours later I discovered [this](https://github.com/magit/magit/pull/4069/files#diff-b1ef1c7610b3f95772eda8d124a9f4e887f9ef4edd272f1a84118e0eeb919de7R7630-R7635)

```
|0.007s|/usr/local/bin/git diff-files -z --name-only --ignore-submodules --
[...]

PR #4447, opened 2021-07-19
Use separate option for remote git executable

tarsius:
...and absolute path on macOS.


tarsius:
Closes #4433.

kyleam (commented):
This direction sounds good to me.  

In addition to my inline comments: I wonder if the variable->function conversion was missed in `magit-patch-id`.

PR #4457, opened 2021-07-28
magit-repolist-sort-reverse: New option

fredericgiquel:
This PR is a complement of the commit 7f666751c0226f37a65f3e9d7dba4c02aa3c252f to be able to reverse the sort order.

For exemple : to have repositories with the most local changes unpushed to upstream at the top of the list, you need to set `magit-repolist-sort-column` to "B>U" and reverse the result. But there is no option to do the reverse.

The implementation is very simple: it just change the cdr of the `tabulated-list-sort-key` variable. According to the documentation of this variable, that's the way to reverse the sort order.

### DIFF
--- a/lisp/magit-repos.el
+++ b/lisp/magit-repos.el
@@ -131,6 +131,12 @@ This has to be the key of an entry in `magit-repolist-columns'."
   :group 'magit-repolist
   :type 'string)
 
+(defcustom magit-repolist-sort-reverse nil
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks.

I changed the doc-string a bit.

tarsius:
Actually I have discarded this again and am currently looking at a solution that does not need an additional option.

tarsius:
Done.

PR #4464, opened 2021-08-05
Ensure that whitespaces are not decorated

delehef:
Hi,

This is a tiny PR, which only ensures that in `magit-status`, separating spaces are not part of font-locked texts. Thus, themes altering the background color does not make the formatting spill outside of the formatted text _per se_.

Example here (https://imgur.com/a/zoB9Z2i) with doom-flatwhite – the space between `sample.git` and `remote [...]` shouldn't be font-locked in red background.

### DIFF
--- a/lisp/magit-status.el
+++ b/lisp/magit-status.el
@@ -612,12 +612,13 @@ arguments are for internal use only."
               ((magit--valid-upstream-p remote merge)
                (if (equal remote ".")
                    (concat
-                    (propertize merge 'font-lock-face 'magit-branch-local)
-                    (propertize " does not exist"
+                    (propertize merge 'font-lock-face 'magit-branch-local) " "
+                    (propertize "does not exist"
                                 'font-lock-face 'font-lock-warning-face))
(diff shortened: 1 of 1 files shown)


lisp/magit-status.el
```diff
@@ -668,10 +672,11 @@ arguments are for internal use only."
                    (tag count face)
                    (insert (concat (propertize tag 'font-lock-face 'magit-tag)
                                    (and (> count 0)
-                                        (format " (%s)"
-                                                (propertize
-                                                 (format "%s" count)
-                                                 'font-lock-face face)))))))
+                                        (concat " "
+                                                (format "(%s)"
```
tarsius, line 676: I don't think this makes any difference.

delehef: You're right, I reversed this chunk.

tarsius:
Similar to how you saw a need to make the output prettier, I had to do the same for the code. :grinning:

delehef:
Thanks for merging!

PR #4495, opened 2021-09-06
magit-version: more flexible for magit.el not exists (magit.el.gz and magit.elc exists)

sunlin7:
Empty result from function `magit-version` when magit is distributed with `*.elc` and `*.el.gz` without `*.el`.

This PR try to make `magit-version` work with this situation.


tarsius:
Thanks!

Issue #4458, opened 2021-07-29
allow to use smerge-keep-lower/upper on Magit Status buffer


FirstLoveLife:
It will report `Point not in conflict region` now.

markgdawson:
I've added a PR that does adds some rudimentary support for this. No doubt it could be refined. In particular the way the key bindings are done could be improved. But it's usable and hopefully can be a good starting point.

https://github.com/magit/magit/pull/4502

tarsius:
I wanted this for a long time but never got around to implementing it. Recently it has come up a few times again and I bumped it up on my todo list. After seeing your melpa pr yesterday, I decided to do it now.

Your code looks sane in principal, but many details needed work, so I decided to start from scratch. You can see my version in the `smerge` branch. Like your version it does provide 1-to-1 one mappings to smerge commands. Additionally `k` on an unmerged hunk now behaves similarly to `k` on a unmerged file.

I am undecided whether to keep the 1-to-1 mappings and in any case I have to write documentation before I can merge.

markgdawson:
Very nice to see this @tarsius. And very instructive to see how you've done it.

Just one quick comment. To my mind it would be more intuitive/consistent if I hit `k` whilst point is inside a conflict that I want to get rid of that conflict (and keep the other). Admittedly, this would mean ignoring the base in a diff3 scenario. But maybe that's the less common use case anyway?

tarsius:
> To my mind it would be more intuitive/consistent if I hit `k` whilst point is inside a conflict that I want to get rid of that conflict (and keep the other). Admittedly, this would mean ignoring the base in a diff3 scenario. But maybe that's the less common use case anyway?

True, it would be nice if this were more consistent, but I think we only get to choose what to be consistent with at the cost of what else. `k` on a file already lets the user choose a side and that already existed for a long time. In the pick on of n; n>2, we have no choice and IMO its also important that is consistent with the n=2 case.

PR #4502, opened 2021-09-13
WIP: Add magit-smerge.el to support magit-smerge from the diff buffer

markgdawson:
Adds support for calling some `smerge-mode` commands from a hunk.


tarsius:
See #4458.

PR #4544, opened 2021-11-16
git-rebase.el: Add new face to customize look of rebase actions

bastianbeischer:
- Add new face to customize look of "pick/reword/..." actions
- Fix group of git-rebase-comment-heading face

### DIFF
--- a/Documentation/RelNotes/3.4.0.org
+++ b/Documentation/RelNotes/3.4.0.org
@@ -10,6 +10,9 @@
 - Refreshing a repository or submodule list buffer now restores the
   previous position.
 
+- Added new face ~git-rebase-action~ to allow customization of the face
+  used for the action words in git-rebase-todo files.
+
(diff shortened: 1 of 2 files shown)


kyleam (changes requested):
Thanks, changes look fine to me.  

Could you please move the `:group` correction for `git-rebase-comment-heading` into its own commit?

tarsius:
The name of the face should not end with `-face' and the definition should not be placed in between two other faces which are clearly related to each other but not to the new face.

tarsius:
> Could you please move the `:group` correction for `git-rebase-comment-heading` into its own commit?

I have just created that commit yourself. So you should rebase before you address the other feedback, @bastianbeischer.

bastianbeischer:
Thanks for your comments! I updated the branch accordingly.

tarsius:
Thanks!

PR #4551, opened 2021-12-05
manual : Fix Evil URL

fredericgiquel:
Evil repository is no longer available on Bitbucket.


kyleam:
Thanks.  

I've regenerated the associated texi file and amended your commit.  I've also dropped the space between "manual" and ":" in the commit message subject to more closely match this project's conventions.

PR #4557, opened 2021-12-14
fix filename converter for tramp in Windows

ShuguangSun:
For tramp in Windows, expanding the file name first before to get the local
name. Otherwise "c:" will be added to the filename.

For example, "/plink:user@host:~/R/" should be converted as
"/home/R/" and not "c:/home/R/"

Signed-off-by: Shuguang Sun <shuguang79@qq.com>

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1028,9 +1028,9 @@ Sorted from longest to shortest CYGWIN name."
                 (cl-rassoc filename magit-cygwin-mount-points
                            :test (lambda (f win) (string-prefix-p win f))))
           (concat cyg (substring filename (length win)))
-        (expand-file-name
-         (or (file-remote-p filename 'localname)
-             filename)))
+        (let ((expanded (expand-file-name filename)))
+          (or (file-remote-p expanded 'localname)
+              expanded)))
     filename))
 
 (defun magit-decode-git-path (path)


tarsius:
@npostavs sorry to bother you but I think we'll need your help here.

lisp/magit-git.el
```diff
@@ -1028,9 +1028,8 @@ Sorted from longest to shortest CYGWIN name."
                 (cl-rassoc filename magit-cygwin-mount-points
                            :test (lambda (f win) (string-prefix-p win f))))
           (concat cyg (substring filename (length win)))
-        (expand-file-name
-         (or (file-remote-p filename 'localname)
-             filename)))
+        (or (file-remote-p (expand-file-name filename) 'localname)
+            (expand-file-name filename)))
```
npostavs, line 1032: Probably makes sense to let-bind the result of `(expand-file-name filename)` to avoid evaluating it twice in case of local names.

npostavs (commented):
I guess the example result should be "/home/user/R/", but yes, the change seems to be doing the right thing.

(with the minor caveat, that whether `expand-file-name` on a remote filename does the right thing may depend on Emacs version. I vaguely recall some changes around this on recentish versions)

lisp/magit-git.el
tarsius: I've done that.

tarsius:
Thanks!

PR #4561, opened 2021-12-15
Add libgit to the list of required libraries.

umanwizard:

Github is an insufficient interface for editing files and creating commits.  Please don't do it when contributing to Magit.

What you should write here

I have found that the emacs libgit package is necessary for building, so add it to the instructions.

How to update the manual

If you make changes to the manual, then edit only "magit.org".  Do not manually edit "magit.texi".  The latter has to be generated from the former.  If you want to do that yourself, then follow the instructions at [2].  Otherwise a maintainer will do it and amend that to your commit.

### DIFF
--- a/Documentation/magit.org
+++ b/Documentation/magit.org
@@ -167,7 +167,7 @@ Now see [[*Post-Installation Tasks]].
 
 ** Installing from the Git Repository
 
-Magit depends on the ~dash~, ~transient~ and ~with-editor~ libraries
+Magit depends on the ~dash~, ~transient~, ~with-editor~ and ~libgit~ libraries
 which are available from Melpa and Melpa-Stable.  Install them using
 ~M-x package-install RET <package> RET~.  Of course you may also install
 them manually from their repository.


Documentation/magit.org
```diff
@@ -167,7 +167,7 @@ Now see [[*Post-Installation Tasks]].
 
 ** Installing from the Git Repository
 
-Magit depends on the ~dash~, ~transient~ and ~with-editor~ libraries
+Magit depends on the ~dash~, ~transient~, ~with-editor~ and ~libgit~ libraries
```
kyleam, line 170: Thanks, it does seem like `libgit` should be mentioned here given cloners will need to do *something* about it before they can build successfully.

Two things:

  * `libgit` isn't required if `BUILD_MAGIT_LIBGIT=false` (i.e. the "something" above can be "set `BUILD_MAGIT_LIBGIT=false`" rather than "make `libgit` available").  That seems worth mentioning.
  * commit subjects in this project tend to start with "$scope: ..." to signal where the change touches.  If you look through the history touching magit.org, you'll see a common pattern for this file is "manual: ...".

kyleam (changes requested):

tarsius:
I don't think I intended `BUILD_MAGIT_LIBGIT=true` to be the default and will change that in a bit.

PR #4562, opened 2021-12-15
Support pull --ff-only

umanwizard:
Support equivalent functionality to `git pull --ff-only` .


tarsius:
Thanks!

I put it before `--rebase` because I feel that makes slightly more sense. "Only do it if it is safe" vs. "if it isn't safe, then fix it first" :grin: .

I probably would have added this myself had I not a long time stopped using `git pull` altogether. (Nothing wrong with fetching, then looking at what changed and then making a decision, though I suppose `git pull --ff-only` is similar.)

PR #4568, opened 2021-12-28
magit--shell-command: Use with-connection-local-variables macro

ndreys:
Use with-connection-local-variables macro to make sure that
`shell-file-name` and `shell-command-switch` have values appropriate
for the machine we are trying to run the command on.

Consider the scenario where we are using Emacs + Magit running on a
Windows machine while manipulating a remote repo on a Linux machine
with Tramp + plink. Without `with-connection-local-variables` we'll
have `shell-file-name` pointing to `cmdproxy.exe` which, for obvious
reasons, won't work on the remote Linux box.

### DIFF
--- a/lisp/magit-bisect.el
+++ b/lisp/magit-bisect.el
@@ -204,7 +204,9 @@ bisect run'."
        (magit-process-git-arguments
         (list "bisect" "start" bad good args)))
       (magit-refresh)))
-  (magit-git-bisect "run" (list shell-file-name shell-command-switch cmdline)))
+  (magit--with-connection-local-variables
+   (magit-git-bisect "run" (list shell-file-name
+                                 shell-command-switch cmdline))))
 
 (defun magit-git-bisect (subcommand &optional args no-assert)
   (unless (or no-assert (magit-bisect-in-progress-p))
--- a/lisp/magit-utils.el
+++ b/lisp/magit-utils.el
@@ -1198,6 +1198,31 @@ See <https://github.com/raxod502/straight.el/issues/520>."
         (setq filename target))))
   (file-chase-links filename))
 
+;;; Kludges for older Emacs versions
+
+(if (fboundp 'with-connection-local-variables)
+    (defalias 'magit--with-connection-local-variables
+      'with-connection-local-variables)
+  (defmacro magit--with-connection-local-variables (&rest body)
+    "Abridged `with-connection-local-variables' for pre Emacs 27 compatibility.
+Bind shell file name and switch for remote execution.
+`with-connection-local-variables' isn't available until Emacs 27.
+This kludge provides the minimal functionality required by
+Magit."
+    `(if (file-remote-p default-directory)
+         (let* ((vec (tramp-dissect-file-name default-directory))
+                (shell-file-name (tramp-get-method-parameter
+                                  vec
+                                  'tramp-remote-shell))
+                (shell-command-switch (mapconcat
+                                       #'identity
+                                       (tramp-get-method-parameter
+                                        vec
+                                        'tramp-remote-shell-args)
+                                       " ")))
+           (progn ,@body))
+       ,@body)))
+
 ;;; Miscellaneous
 
 (defun magit-message (format-string &rest args)
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1959,12 +1959,13 @@ Return a list of two integers: (A>B B>A)."
     (cdr (split-string it))))
 
 (defun magit-patch-id (rev)
-  (magit--with-temp-process-buffer
-    (magit-process-file
-     shell-file-name nil '(t nil) nil shell-command-switch
-     (let ((exec (shell-quote-argument (magit-git-executable))))
-       (format "%s diff-tree -u %s | %s patch-id" exec rev exec)))
-    (car (split-string (buffer-string)))))
+  (magit--with-connection-local-variables
+   (magit--with-temp-process-buffer
+     (magit-process-file
+      shell-file-name nil '(t nil) nil shell-command-switch
+      (let ((exec (shell-quote-argument (magit-git-executable))))
+        (format "%s diff-tree -u %s | %s patch-id" exec rev exec)))
+     (car (split-string (buffer-string))))))
 
 (defun magit-rev-format (format &optional rev args)
   (let ((str (magit-git-string "show" "--no-patch"
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -447,8 +447,9 @@ is run in the top-level directory of the current working tree."
   (let ((default-directory (or directory default-directory))
         (process-environment process-environment))
     (push "GIT_PAGER=cat" process-environment)
-    (magit-start-process shell-file-name nil
-                         shell-command-switch command))
+    (magit--with-connection-local-variables
+     (magit-start-process shell-file-name nil
+                          shell-command-switch command)))
   (magit-process-buffer))
 
 (defun magit-read-shell-command (&optional toplevel initial-input)


ndreys:
AFAICT, there a couple of other users of `shell-file-name` that _might_ benefit from a similar fix, but I am not quite sure and it wasn't obvious to me how to invoke those codepath from Magit. Happy to try something out on the aforementioned Win + Linux setup if need be.

kyleam:
Thanks, the change itself looks correct.  However, `with-connection-local-variables` is only available since Emacs 27.1, and `magit--minimal-emacs` is currently 25.1.  There would need to be some sort of compatibility kludge (such as a wrapper on Magit's side that copies `with-connection-local-variables`, though I haven't checked to see if everything in the definition of `w-c-l-v` is available in Emacs 25.1).  Or, since it looks like `w-c-l-v` was a rename and rework of another macro, perhaps that could be used if available instead of copying the definition.

> there a couple of other users of `shell-file-name` that _might_ benefit from a similar fix, but I am not quite sure and it wasn't obvious to me how to invoke those codepath from Magit.

```
$ git grep -n shell-file-name | nl
     1	lisp/magit-bisect.el:207:  (magit-git-bisect "run" (list shell-file-name shell-command-switch cmdline)))
     2	lisp/magit-git.el:313:         ;; The user may have customized `shell-file-name' to
     3	lisp/magit-git.el:318:         (shell-file-name (if (and (eq system-type 'windows-nt)
     4	lisp/magit-git.el:324:                            shell-file-name)))
     5	lisp/magit-git.el:1964:     shell-file-name nil '(t nil) nil shell-command-switch
     6	lisp/magit-process.el:683:   ((and args (equal program shell-file-name))
     7	lisp/magit.el:450:    (magit-start-process shell-file-name nil
     8	test/magit-tests.el:83:                      shell-file-name)))
```

 * 7: that's the one in this PR

 * 8: shouldn't need a change

 * 2, 3, 4: These are within `magit-with-editor`.  I'm guessing that's operating on the local side regardless of whether the repo is on the remote side, so I think that could be okay as is, but I'm not sure.

 * 6: This is in `magit-process--format-arguments`.  The fix probably would be relevant, but I think it's purely cosmetic and could be left as is.

 * 1, 5: I think these should probably get the fix.

   You could trigger the first (in `magit-bisect-run`) with `B s` in a remote repo.

   The second (`magit-patch-id`) could be triggered by doing an interactive rebase in a remote repo.

ndreys:
@kyleam, thanks for review! Added the kludge and tested it on Emacs 25.1. Hopefully this limited version of it is enough for this to be merged. I also added a fix for #1 from your list since I was able to repro the problem. However, I don't see `magit-patch-id` invoked when I do an interactive rebase on a remote machine. I didn't to anything special, just started a rebase and then marked on of the commits to be "reworded". I think I need a more specific scenario to land on that codepath. Any ideas?

Moreso, I _think_ `magit-patch-id` is broken in a couple of other ways:

```
(defun magit-patch-id (rev)
  (magit--with-temp-process-buffer
    (magit-process-file
     shell-file-name nil '(t nil) nil shell-command-switch
     (let ((exec (shell-quote-argument (magit-git-executable))))
       (format "%s diff-tree -u %s | %s patch-id" exec rev exec)))
    (car (split-string (buffer-string)))))
```

that `shell-quote-argument` won't be doing correct quoting, so, I think we might need to use `tramp-shell-quote-argument` for remote repos. But more importantly, I am not sure how that piping operation `|` is going to work on Windows at all. AFAIU `cmdproxy.exe` won't emulate that, so maybe that should be rewritten to be handled via ephemeral alias like this one https://github.com/magit/magit/blob/master/lisp/magit-git.el#L148 ?

lisp/magit-utils.el
```diff
@@ -1223,6 +1223,27 @@ Like `message', except that `message-log-max' is bound to nil."
          (goto-char (or ,pos 1))
          ,@body))))
 
+(unless (fboundp 'with-connection-local-variables)
+  (defmacro with-connection-local-variables (&rest body)
```
kyleam, line 1227: The approach you chose isn't too uncommon, but I'd vote to avoid defining a macro with the same name.  Instead, define something with a `magit-` prefix (e.g., `magit--with-connection-local-variables`) that makes it clearer where it's coming from.  When `with-connection-local-variables` is available, you can just link it up with `defalias`.  So, `(if bound alias <custom definition>)`.

This leads to a little more churn once `magit--minimal-emacs` is at least 27 and `magit--with-connection-local-variables` is dropped for a plain `with-connection-local-variables`, but I think it's worth it.

lisp/magit-utils.el
```diff
@@ -1223,6 +1223,27 @@ Like `message', except that `message-log-max' is bound to nil."
          (goto-char (or ,pos 1))
          ,@body))))
 
+(unless (fboundp 'with-connection-local-variables)
```
kyleam, line 1226: Rather than putting this code in the "Miscellaneous" section of magit-utils, perhaps move it alongside the other "Kludges ..." sections with a new section called "Kludges for older Emacs version" or something like that.

lisp/magit-utils.el
```diff
@@ -1223,6 +1223,27 @@ Like `message', except that `message-log-max' is bound to nil."
...
+  (defmacro with-connection-local-variables (&rest body)
+    "A poor implementation of `w-c-l-v' for pre Emacs 27.1 compatibility
+
+This is a kludge implementing bare minimum of `w-c-l-v' used by
+Magit for backwards compatibility with Emacs 25 and 26. Only
+`shell-file-name' and `shell-command-switch' are modified"
```
kyleam, line 1232: Minor docstring comments:

 * I'd avoid spelling out Emacs 25 in the docstring because it will be stale if `magit--minimal-emacs` gets bumped to 26.1.  I think something along the lines of the "pre Emacs 27.1 compatibility" you say before that is sufficient.

 * Please follow the Emacs convention of adding a period after the first line of the docstring and using two spaces between sentences.  (See `(info "(elisp)Documentation Tips")`.)

 * The last sentence is also missing a period.

 * It's probably better to expand `w-c-l-v` for readers to be able to easily follow it (e.g., with `describe-function`).  That would probably make your first line a bit long, so it might require rewording.

Considering those points, maybe something like

```
Bind shell file name and switch for remote execution.
`with-connection-local-variables' isn't available until Emacs 27.
This kludge provides the minimal functionality required by
Magit.
```

lisp/magit-bisect.el
```diff
@@ -204,7 +204,8 @@ bisect run'."
        (magit-process-git-arguments
         (list "bisect" "start" bad good args)))
       (magit-refresh)))
-  (magit-git-bisect "run" (list shell-file-name shell-command-switch cmdline)))
+  (with-connection-local-variables
+   (magit-git-bisect "run" (list shell-file-name shell-command-switch cmdline))))
```
kyleam, line 208: Nit: Please split this across multiple lines to be under 80 characters.

kyleam (changes requested):
I don't have access to a Windows machine, but I tried out `magit-shell-command-topdir` and `magit-bisect-run` over Tramp with a local `shell-file-name` that wasn't available on the remote.  On master, those commands failed due to the shell not being on the remote.  With this PR, they passed as expected (tested with Emacs 25 and 27).

This series looks in pretty good shape to me.  I've left some minor comments.

---

> Subject: [PATCH 3/3] magit-bisect-run: Use with-connection-local-variables
>  macro
> 
> Use with-connection-local-variables macro in order to select correct
> shell to run a bisec script if operating against a remote repo.

s/bisec/bisect/

kyleam:
> However, I don't see `magit-patch-id` invoked when I do an interactive rebase on a remote machine. I didn't to anything special, just started a rebase and then marked on of the commits to be "reworded". I think I need a more specific scenario to land on that codepath. Any ideas?

Hmm, yeah, a reword that doesn't lead to the rebase sequence being shown in the status buffer wouldn't trigger it.  It needs to be something that leads to `magit-sequence-insert-sequence` being called.  Just changing that to "edit" should do it.  You could also call `magit-rebase-edit-commit` and select the first commit, or even just evaluate `(magit-patch-id "HEAD")` in a remote repo.

> Moreso, I _think_ `magit-patch-id` is broken in a couple of other ways:
> [...]
> that `shell-quote-argument` won't be doing correct quoting, so, I think we might need to use `tramp-shell-quote-argument` for remote repos.

I may be missing something, but reading `tramp-shell-quote-argument`, I'm not sure it would actually help.  It seems to just be about newlines and limited to "Bourne-like shells".

> But more importantly, I am not sure how that piping operation `|` is going to work on Windows at all. [...]

Your guess is much better than mine here.  I know very little about Windows.

Anyway, thank you for taking a look.  Let's leave that be, then, until someone has the motivation to give a go at a complete fix.

lisp/magit-bisect.el
ndreys: Done

lisp/magit-utils.el
ndreys: Done

lisp/magit-utils.el
ndreys: Done

ndreys:
> Hmm, yeah, a reword that doesn't lead to the rebase sequence being shown in the status buffer wouldn't trigger it. It needs to be something that leads to magit-sequence-insert-sequence being called. Just changing that to "edit" should do it. You could also call magit-rebase-edit-commit and select the first commit, or even just evaluate (magit-patch-id "HEAD") in a remote repo.

Thanks! Used `M-:` + `(magit-patch-id "HEAD")` as a way to test the fix.

>> Moreso, I think magit-patch-id is broken in a couple of other ways:
[...]
that shell-quote-argument won't be doing correct quoting, so, I think we might need to use tramp-shell-quote-argument for remote repos.

> I may be missing something, but reading tramp-shell-quote-argument, I'm not sure it would actually help. It seems to just be about newlines and limited to "Bourne-like shells".

The one nice thing it'd do is set `system-type` to `not-windows`, which should prevent weird escaping, but this might be moot since `(magit-git-executable)` will, most likely, expand to just `"git"`, which doesn't need any escaping to do. I think that escaping code is not correct in theory, but I can't come up with a non made-up use-case to break it, so I'll leave it be.

>> But more importantly, I am not sure how that piping operation | is going to work on Windows at all. [...]

> Your guess is much better than mine here. I know very little about Windows.

I stand corrected. Apparently `cmd.exe` supports `|` for piping 🤯 , so that part of the code is correct.

lisp/magit-utils.el
ndreys: Done

kyleam (approved):
> Used `M-:` + `(magit-patch-id "HEAD")` as a way to test the fix.

Great, thanks looking into that.

---

I've rebased the branch to the latest master, amending the first commit with cosmetic whitespace tweaks to the docstring of `magit--with-connection-local-variables`.

Thanks for the nicely done series.

PR #4569, opened 2021-12-30
Add NonGNU ELPA badge to README.md

skangas:
This adds a NonGNU ELPA badge to README.md.

Unfortunately, there was recently an issue with generating these badges (now resolved), so the badge will just be blank until the next release of Magit. However, the next time Magit is released, this will just work automatically. Please see an example of what it will look like here: http://elpa.nongnu.org/nongnu/mpv.svg

Thanks!


tarsius:
Thanks! I'll merge before the release.

PR #4574, opened 2022-01-05
Fix #4573 when "make" ignores "docs" directory

### LINKED ISSUE #4573: Command "make" ignores "docs" directory
Hi! If I run `make` from the root of Magit then I get the following message:
```
'docs' is up to date.
```
So, `make` does not visit subdirectory `docs` and does not build documentation. The reason is very simple: the name of target `docs` is the same as the name of subdirectory. A fix should be easy: just add `docs` into `.PHONY` targets.

vonavi:


vonavi:
Fix magit/magit#4573

kyleam (approved):
Thanks for the patch.  This was surfaced by the recent rename of Documentation/ to docs/.

The fix looks obviously correct.  (Perhaps Jonas will prefer `docs` placed somewhere else in the `.PHONY` list, but I'm sure he won't mind getting the fix in now and adjusting that later.)

I've rebased the commit on top of the latest master and amended the commit message with a few tweaks, including aligning it more closely with the conventions used in this repo.

PR #4576, opened 2022-01-07
magit-list-refs: exclude case-sensitive HEAD from branch list

### LINKED ISSUE #4575: magit-list-refs should be case-sensitive to support `head` branch
I work in a repo with main branch named `head` (yes, lowercase, no, I didn't choose it). This makes it impossible to set as an upstream in magit. E.g. `b`-`C`-`RET`-`u` refuses to accept `origin/head` despite branch existence. 

I tracked this behavior down to this term 
```
(--remove (string-match-p "\\(\\`\\|/\\)HEAD\\'" it) refs)
```
[here](https://github.com/magit/magit/blob/7d0d7068deff003a02f49fa430153e4611c80175/lisp/magit-git.el#L1694) in `magit-list-refs`. Sadly the behavior of `string-match-p` is dependent on a global and I didn't figure out how to make this case-sensitive. I confirmed I was right by mangling the search string and `head` is happily accepted then.

Could somebody more fluent in elisp fix this?

blackgnezdo:
This allows people who named their branch `head` to still see them.
Fix #4575

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1691,7 +1691,9 @@ SORTBY is a key or list of keys to pass to the `--sort' flag of
                                         ((and val (pred listp)) val)))
                                (or namespaces magit-list-refs-namespaces))))
     (if (member format '("%(refname)" "%(refname:short)"))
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks!

PR #4577, opened 2022-01-10
Replace preserve-merges by rebase-merges

chep:
This replaces --preserve-merges by --rebase-merges in rebase popup.

--preserve-merges is obsolete and new git version gives the error "fatal: --preserve-merges has been replaced by --rebase-merges"

git version 2.34.1


tarsius:
The new argument was already available since c8bb181079a66d5bb39068cffaf8b444b61c5fa4.

But...

tarsius:

tarsius:
I have now updated `magit-rebase` to offer one or both of these arguments based on the Git version in use.

PR #4578, opened 2022-01-10
Resist the Jedi

tarsius:
@kyleam what do you think?

Also, how many translations did I mess up?

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -567,7 +567,8 @@ and Emacs to it."
         (when print-dest
           (princ (format "Magit %s%s, Git %s, Emacs %s, %s"
                          (or magit-version "(unknown)")
-                         (or (and (ignore-errors (version< "2008" magit-version))
+                         (or (and (ignore-errors
+                                    (magit--version>= magit-version "2008"))
                                   (ignore-errors
                                     (require 'lisp-mnt)
                                     (and (fboundp 'lm-header)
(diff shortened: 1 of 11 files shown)


lisp/magit.el
```diff
@@ -567,7 +567,8 @@ and Emacs to it."
         (when print-dest
           (princ (format "Magit %s%s, Git %s, Emacs %s, %s"
                          (or magit-version "(unknown)")
-                         (or (and (ignore-errors (version< "2008" magit-version))
+                         (or (and (ignore-errors ; Is it a Melpa snapshot?
+                                    (magit--version> magit-version "20080000"))
```
kyleam, line 571: I'm confused by the `s/20080000/2008/`.  Intentional?

kyleam (commented):
:upside_down_face: 

> @kyleam what do you think?

Like you, I find the end result easier to parse.  I'm in favor of adding these custom variants.

> Also, how many translations did I mess up?

Zero that I caught :)

Left one question about a change in the Melpa snapshot value that seemed unrelated.

lisp/magit.el
tarsius: It felt weird going from < to >= since that would imply some precision that just isn't there, but not doing that would look like a bug, so made an additional change and added a comment to hint that this is just a heuristic. But it didn't work so I am dropping it.

PR #4584, opened 2022-01-18
Sparse checkout and partial clone support

### LINKED ISSUE #4102: Partial Clone and Sparse Checkout
I've just taken a look at options for handling large files with git/magit. Of course, git-annex (with magit-annex) or git-lfs (with magit-lfs) come to mind. Then I read that the main git project introduced Partial Clone ([1]) at some point, which e.g. Gitlab recommends now for large files ([2],[3]). 

It seems for magit to be able to support these workflows, some keybindings and menu options would have to be added, e.g. for cloning the option to support `clone --filter` or the command `git sparse-checkout`.

Is there any plan to support this?

[1]: https://git-scm.com/docs/partial-clone
[2]: https://about.gitlab.com/blog/2020/03/13/partial-clone-for-massive-repositories/
[3]: https://docs.gitlab.com/ee/topics/git/partial_clone.html

kyleam:
Commits 1-5 add support for `git sparse-checkout`.  The second commit is the main addition, and its commit message has some background on sparse checkouts and explains which aspects the new magit-sparse-checkout.el library targets.

Commit 6 adds partial clone support by adding `--filter` to `magit-clone`.  Although they can work together, partial clones aren't inherently coupled to sparse checkouts, but I decided to include the last commit given that both features share the same issue (gh-4102).

Closes #4102.

cc @aseltmann

---

I'm marking this as a draft, mostly because I've run out of time to read this over tonight (apologies for bad typos or half-thoughts that may have slipped through).  I think it's likely that I won't get back to this until next weekend, but I'd of course be happy for any feedback in the meantime.

- [x] regenerate manual

### DIFF
--- a/lisp/magit-clone.el
+++ b/lisp/magit-clone.el
@@ -129,6 +129,9 @@ the name of the owner.  Also see `magit-clone-name-alist'."
    ("s" "shallow"            magit-clone-shallow)
    ("d" "shallow since date" magit-clone-shallow-since :level 7)
    ("e" "shallow excluding"  magit-clone-shallow-exclude :level 7)
+   (">" "sparse checkout"    magit-clone-sparse
+    :if (lambda () (magit-git-version>= "2.25.0"))
+    :level 6)
    ("b" "bare"               magit-clone-bare)
    ("m" "mirror"             magit-clone-mirror)]
   (interactive (list (or magit-clone-always-transient current-prefix-arg)))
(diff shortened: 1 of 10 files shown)


tarsius:
Sounds good to me.  I haven't actually tested anything but I have read the changes and descriptions and I liked what I saw.  Feel free to merge.

Since the man-page WARNS that things are likely to change, we might want to do the same.  IMO it would be okay if something changes upstream in a backward incompatible way, that we then require users to update to the Git version with the breaking change.  Of course if is trivial to continue to support older versions, then we should do that.  But given the fat upstream warning, we should probably have one too.

> There are other examples of --x/--no-x options along with an associated config knob (e.g., --recurse-submodules in magit-branch) that we don't bother supporting (in terms of having both options available in the transient or considering the configured value).

Yeah, I should really tackle that soon. ;)

tarsius (changes requested):
that makes improves -> that improves

lisp/magit-clone.el
```diff
@@ -117,6 +117,9 @@ the name of the owner.  Also see `magit-clone-name-alist'."
   ["Setup arguments"
    ("-o" "Set name of remote"     ("-o" "--origin="))
    ("-b" "Set HEAD branch"        ("-b" "--branch="))
+   (magit-clone:--filter
+    :if (lambda () (magit-git-version>= "2.17.0"))
+    :level 6)
```
tarsius, line 122: To me that sounds like a "level 7" argument.

kyleam: Okay, will update.

kyleam:
> Since the man-page WARNS that things are likely to change, we might want to do the same. [...]

Right, good idea.  Will do.

> that makes improves -> that improves

Oops, thanks.

PR #4585, opened 2022-01-29
Updates for compatibility with SHA-256 repositories

### LINKED ISSUE #4516: update for extensions.objectFormat=256
We need to eventually audit for sha1 hardcoding.  A quick grep for oid length indicates that things might not be so bad:

```
$ git grep '\b40\b'
lisp/magit-bisect.el:           (done-re "^\\([a-z0-9]\\{40\\}\\) is the first bad commit$")
lisp/magit-bisect.el:           "# first bad commit: \\[\\([a-z0-9]\\{40\\}\\)\\] [^\n]+\n" nil t)
lisp/magit-blame.el:    (unless (looking-at "^\\(.\\{40\\}\\) \\([0-9]+\\) \\([0-9]+\\) \\([0-9]+\\)")
lisp/magit-blame.el:                ((looking-at "^previous \\(.\\{40\\}\\) \\(.+\\)")
lisp/magit-diff.el:              "^  \\([^ ]+\\) +[0-9]\\{6\\} \\([a-z0-9]\\{40\\}\\) \\(.+\\)$")
lisp/magit-git.el:  (--mapcat (and (string-match "^160000 [0-9a-z]\\{40\\} 0\t\\(.+\\)$" it)
```

I've yet to do any testing under a repo created with

```
git init --object-format=sha256
```

kyleam:
This series fixes spots in the code base that are incompatible with Git's recent and experimental SHA-256 support (basically anywhere that hard codes the expectaion of 40-character object names).

I found these spots with two checks:

```
$ git grep '\b40\b'
$ git grep -E '[a-f0-9]{40}'
```

I expected that the empty tree hash might pop up somewhere in there, but `magit-headish` already handles this in a nicely compatible way.

Fixes #4516.

### DIFF
--- a/lisp/magit-bisect.el
+++ b/lisp/magit-bisect.el
@@ -250,7 +250,7 @@ bisect run'."
                       "It appears you have invoked `git bisect' from a shell."
                       "There is nothing wrong with that, we just cannot display"
                       "anything useful here.  Consult the shell output instead.")))
-           (done-re "^\\([a-z0-9]\\{40\\}\\) is the first bad commit$")
+           (done-re "^\\([a-z0-9]\\{40,\\}\\) is the first bad commit$")
            (bad-line (or (and (string-match done-re (car lines))
                               (pop lines))
                          (--first (string-match done-re it) lines))))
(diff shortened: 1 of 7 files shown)


tarsius (approved):
I somehow feared the status quo would be worse. Great to see it's actually quite harmless. Thanks!

Issue #4572, opened 2022-01-03
Add hint for `K` "Untrack file" command


cstby:
Magit supports untracking files, but there is no hint in the "popup of popups" for this command, which can be called with `K`.

In situations where a user has updated their .gitignore file and needs to untrack the files from their repo, they are likely to look at the "popup of popups" for this option. Is the omission of this command intentional?

tarsius:
The "popup of popups" doesn't list all top-level bindings.

It lists the popups (and some other important commands) in "Transient and dwim commands", some essential commands in "Essential commands" and apply commands in "Applying changes".

One could argue that `magit-file-untrack` fits in the last category, but I it seems to me to make sense to not overload that group.  If this was the only command command with a top-level binding that isn't listed in the dispatcher, then I would probably just add it anyway.

Look at the definition of `magit-dispatch`, it contains comments mentioning the commands that are not shown.  Fitting all those (in the existing groups) would be harder.

This does not include commands that have top-level bindings where that binding is not a *alphabetic* binding, e.g. `C-c C-w` for `magit-browse-thing`.  If we include *all* the alphabetic bindings shouldn't also include all the non-alphabetic bindings too; after all they are *less* discoverable.

So the dispatcher provides only a curated subset of the available commands and to get a list of all the available bindings one can use the command that can be used for that purpose in any buffer: `describe-bindings` (`C-h b`).

@kyleam, do you agree with that approach in general? Are there any commands you would add to the dispatcher?

Oh! We could make use of transient levels to give users a choice.

cstby:
Thanks for referencing `magit-dispatch`. I see that not including all commands was deliberate, and I understand your reasoning.

> If we include all the alphabetic bindings shouldn't also include all the non-alphabetic bindings too; after all they are less discoverable.

This is a very good point. One consideration is that the popup currently _appears_ to be inclusive of all non-navigational alphabetic bindings. Making every command discoverable may not be feasible, but you also want to ensure that you're setting user expectations appropriately.

As a user, I wanted to untrack a file, but I didn't know whether magit could do that or which magit command to use. I worked through this situation by googling "magit untrack file", which led me stack exchange where I found that my answer. It didn't take long. I was mildly surprised that the popup didn't include a hint for `K`.  If you consider that flow acceptable, then there's nothing else to do here.

kyleam:
@tarsius:
> @kyleam, do you agree with that approach in general?

[...]

PR #4587, opened 2022-01-29
Make magit-dispatch slightly more discoverable

tarsius:
In response to #4572.

@kyleam what do you think?

Additionally (but optionally) adding all alphabetic bindings would
look like this:

``` diff
index d83ccbb90..19cbd3d30 100644
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -373,7 +373,15 @@ (transient-define-prefix magit-dispatch ()
     ("<tab>" "   toggle section at point"  magit-section-toggle)
     ("<return>" "visit thing at point"     magit-visit-thing)]
    [("C-x m"    "show all key bindings"    describe-mode)
-    ("C-x i"    "show Info manual"         magit-info)]])
+    ("C-x i"    "show Info manual"         magit-info)]]
+  ["Other commands"
+   :level 7
+   ("G" "Refresh repository buffers" magit-refresh-all)
+   ("K" "Untrack file"               magit-file-untrack)
+   ("n" "Goto next section"          magit-section-forward)
+   ("p" "Goto previous section"      magit-section-backward)
+   ("R" "Rename file"                magit-file-rename)
+   ("x" "Reset quickly"              magit-reset-quickly)])
 
 ;;;###autoload
 (defun magit-info ()
```

The problem with that is that those who could benefit the most don't get to see it because it is opt-in, while some of those that likely need it the least get to see it because they have set `transient-default-level` to 7.

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -372,7 +372,14 @@ Also see info node `(magit)Commands for Buffers Visiting Files'."
     ("q" "       bury current buffer"      magit-mode-bury-buffer)
     ("<tab>" "   toggle section at point"  magit-section-toggle)
     ("<return>" "visit thing at point"     magit-visit-thing)]
-   [("C-x m"    "show all key bindings"    describe-mode)]])
+   [("C-x m"    "show all key bindings"    describe-mode)
+    ("C-x i"    "show Info manual"         magit-info)]])
+
+;;;###autoload
+(defun magit-info ()
+  "Show Magit's Info manual."
+  (interactive)
+  (info "magit"))
 
 ;;; Git Popup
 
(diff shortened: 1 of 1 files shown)


kyleam:
> @kyleam what do you think?

All three commits seem like a good idea to me.  Obviously the second commit very much depends on the perception of others, but at least for me I agree "more noticeable" is accurate (then again my mind was already primed by reading the commit subject :>) 

> Additionally [...]  The problem with that is that those who could benefit the most don't get to see it because it is opt-in, while some of those that likely need it the least get to see it because they have set `transient-default-level` to 7.

Right.  I'm not convinced that would help anyone.

tarsius:
> Obviously the second commit very much depends on the perception of others, but at least for me I agree "more noticeable" is accurate (then again my mind was already primed by reading the commit subject :>)

We will likely never know since only those for whom it doesn't work will tell us about it. But it's worth a try. :)

Another benefit of adding the `C-x i` binding is that users are more likely to press `C-x` and learn about the additional bindings that are available after doing that.

PR #4588, opened 2022-01-31
Fix hash/commit summary order in push branch header

geza-herman:
I use `magit-status-show-hashes-in-headers`, and noticed that the order of hash and commit summary is not consistent in the header, the push header has the wrong order. This PR fixes this.


tarsius:
Thanks!

PR #4591, opened 2022-02-03
magit-turn-on-auto-revert-mode-if-desired: do cheaper checks first

spwhitton:
If we are not going to do anything because auto-revert-mode or global-auto-revert-mode are on, don't call executable-find, file-readable-p or magit-toplevel.  This is particular important for remote files, for which calling those functions is expensive.


kyleam (approved):
Thanks, looks good to me.

PR #4598, opened 2022-02-11
Use eq instead of = in magit-log-insert-child-count

cvchaparro:
A call to magit-log-get-commit-limit can return nil, so make sure we don't try
to compare nil to a number. I ran into this when running `magit-status` in a repository because of a call to the `magit-log-insert-child-count` function where we compare the result of calling `magit-log-get-commit-limit` with `=`. It seems to be the case that if `magit-buffer-log-args` is `nil`, then the result of the whole call to `magit-log-get-commit-limit` will also be `nil` and so a call of the form `(= <num> nil)` will cause a failure.


tarsius:
Thanks!

Issue #4065, opened 2020-03-10
Cache and report on git version incompatibilities


phil-s:
Commit f6307724251b1632efb7607d187dd8ccc602ae55 added a repository-local cache of the availability of one specific feature, based on the version of git being used.

It's long been the case that trying to use Magit with an unsupported older version of git causes problems which can be difficult for users to interpret.  Most recently this cropped up in https://github.com/magit/magit/issues/3714#issuecomment-596500488 and so I'm assuming that this kind of problem is still a going concern.

I suggest adapting that repository-local caching approach so that any and all version-related issues can be captured (as a one-time action), and reported on.

If the cache doesn't exist yet, we would check the git version, then cache store all related information.  To start with this would be (a) the git version; (b) whether that version is supported by Magit; and (c) the diff-ita-kludge-p value from the aforementioned commit.

If the version is unsupported then Magit can make a loud complaint to the user pointing out the incompatibility in an unambiguous fashion (and also pointing out the 'zap caches' command, in case the user wishes to forcibly re-check, perhaps after upgrading git).

tarsius:
Sounds like a good idea. Could you implement that please?

kyleam:
> It's long been the case that trying to use Magit with an unsupported older version of git causes problems which can be difficult for users to interpret. Most recently this cropped up in [#3714 (comment)](https://github.com/magit/magit/issues/3714#issuecomment-596500488) and so I'm assuming that this kind of problem is still a going concern.

Hmm, `magit-startup-asserts` should display an informative message when the git version on the local machine isn't new enough.  (I quickly tested by tweaking `magit--minimal-git` to the currently unborn 2.27.0, and I saw the warning.)  At first I thought that perhaps a warning wasn't occurring because the user was operating over Tramp, but then looking closer I think `magit--tramp-asserts` should catch an unsupported git in that case.

> To start with this would be [...]

There are a few other spots that could benefit from this.  There's a cached `update-index-has-ignore-sw-p` as of b857fc1b.  And looking through `magit-git-version` calls, other candidates for cached git version conditions are `magit-fetch-modules`, `magit-rev-name` `magit-update-ref`, `magit--tramp-asserts`, `magit-submodule-add-1`, and `magit-submodule-remove`.

phil-s:
[...]

PR #4603, opened 2022-02-19
Refactor Git version asserts

tarsius:
This is response to #4065 but doesn't go as far as the changes @phil-s linked to in that issue.

I have come to the conclusion that it is not necessary to try to catch the very first call to `git` that a user invokes on a new host using Magit (even if `magit-status` wasn't responsible for that), because I cannot remember the last time that would have been useful and because I did not like the introduction of `magit-process-file-git`.  However if we later decide otherwise, we could simply add that function and add a call to `magit-git-version-assert` there.

The central idea is that callers `magit-git-version<` and `magit-git-version>` do not have to worry about caching and error reporting because these functions use `magit-git-version`, which takes care of all that.  They obviously should still select another code-path or report an error based on the result of the comparison.  `magit-git-version-assert`, which checks for the global minimum, also uses that function.

Unlike @phil-s' implementation, this does not add a central place that records all the functionality that need a more recent Git version than the global default.  I feel that it is more useful to see the needed version along-side the affected code and if one needs a list of such functionality, then one can just grep for `magit-git-version[<>]`.

Still needs some testing but I plan to merge in a few days once I have done that. Please let me know if you see any issues with this approach.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -2053,15 +2053,9 @@ Staging and applying changes is documented in info node
                (remove "--literal-pathspecs" magit-git-global-arguments)))
     ;; As of Git 2.19.0, we need to generate diffs with
     ;; --ita-visible-in-index so that `magit-stage' can work with
-    ;; intent-to-add files (see #4026).  Cache the result for each
-    ;; repo to avoid a `git version' call for every diff insertion.
+    ;; intent-to-add files (see #4026).
     (when (and (not (equal cmd "merge-tree"))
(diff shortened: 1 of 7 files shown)


kyleam (commented):
Sounds good to me.

> Subject: [PATCH 6/8] magit--safe-git-version: New function
>
> Instead of using `magit-git-version' and hacking around that wanting
> to report verbosely on any issues, add a new function for when we want
> to be more quite about it.

typo: quiet

PR #4605, opened 2022-02-21
Add custom date format variable for magit-margin

Takuro-Ito:
# Background
Although you can customize date format with current version of magit like this,
```lisp
(setq magit-status-margin '(t "%Y-%m-%d %H:%M:%S " magit-log-margin-width t 18))
```
~this format will be lost if you cycle styles by pressing `L-l` (`magit-cycle-margin-style`) and also~ you can't set initial style as `age-abbreviated` or `age` along with customized date format.

# Summary
This PR adds ability to set custom date format for `magit-log-margin`.

Let users allow to define custom date format from `customize-group` or like the following.
```lisp
(setq magit-log-margin-date-time-format "%Y-%m-%d %H:%M:%S ") ;; Show seconds
(setq magit-log-margin '(t age-abbreviated magit-log-margin-width t 18)) ;; Initial style for magit-log buffer (`l-l`)
(setq magit-status-margin '(t age-abbreviated magit-log-margin-width t 18)) ;; Initial style for Recent commits
````

### DIFF
--- a/lisp/magit-margin.el
+++ b/lisp/magit-margin.el
@@ -88,7 +88,7 @@ does not carry to other options."
           (`age 'age-abbreviated)
           (`age-abbreviated
            (let ((default (cadr (symbol-value (magit-margin-option)))))
-             (if (stringp default) default "%Y-%m-%d %H:%M ")))
+             (if (stringp default) default magit-log-margin-date-time-format)))
           (_ 'age)))
   (magit-set-buffer-margin nil t))
 
@@ -174,6 +174,12 @@ does not carry to other options."
 
 ;;; Custom Support
 
+(defcustom magit-log-margin-date-time-format "%Y-%m-%d %H:%M "
+  "Format of the author or committer date in `magit-log-mode' buffers"
+  :group 'magit-log
+  :group 'magit-margin
+  :type 'string)
+
 (defun magit-margin-set-variable (mode symbol value)
   (set-default symbol value)
   (message "Updating margins in %s buffers..." mode)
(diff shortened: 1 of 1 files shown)


tarsius:
> this format will be lost if you cycle styles by pressing `L-l`
> (`magit-cycle-margin-style`)

That doesn't seem to be the case.

> and also you can't set initial style as `age-abbreviated` or `age` along with
> customized date format.

But that is.

The proposed patch has issues.  Most importantly the docstring does not succeed in explaining why the option is needed.  Some details are also wrong, such as where the option is placed and its type.

Takuro-Ito:
>> this format will be lost if you cycle styles by pressing L-l
(magit-cycle-margin-style)

>  That doesn't seem to be the case.

I confirmed that it did not that case because of my misconfiguration and I crossed out that sentence in original PR description.

Is there another way to configure initial style as `age-abbreviated` along with custom date time format such as `"%Y-%m-%d %H:%M:%S "` with current version of magit?
If not, I'm happy to fix some issues in my PR!

Takuro-Ito:
> Git "recently" gained support for relative times in logs and Magit's margin code should be rewritten to use that. That is a larger undertaking so I am delaying it indefinitely

That is totally understandable.


> So I have added a new variable magit-margin-default-time-format that you can use as you imagined but don't define it as an option to avoid advertising it.

and 

> magit-margin-default-time-format: New un-advertised variable https://github.com/magit/magit/commit/d5ed87b1382522c3e875fc7ae40e19135e3cc0aa

Thank you for adding the option! I appreciate it!

Issue #4602, opened 2022-02-16
Occasional indefinite hangs when viewing a commit through magit with which-function-mode active


billsacks:
First, thank you very much for all of the work you do to develop and maintain magit! I have been enjoying it immensely, and I really appreciate all of the time and energy that you have put in to creating such an amazing git interface.

I experienced this bug using the latest Emacs 28 pretest build from https://emacsformacosx.com/builds. This is using a clean startup (via moving aside my .emacs.d for this test). For now, I am working around this issue by giving an explicit list of modes in `which-func-modes`, rather than using the default value of `t` for that variable. So this is not critical to my personal use and enjoyment of emacs and magit.

(1) Outside of emacs, clone the git repository where I have noticed this issue:
```
mkdir ~/temporary
git clone https://github.com/ESCOMP/CTSM.git ~/temporary/ctsm_temp
```

(2) Launch emacs

(3) Install magit:
M-x list-packages
Then install magit; I installed the latest current release, 3.3.0. (I have also seen this with the latest development magit version installed through melpa.)

(4) Turn on which-function-mode:
M-x which-function-mode

(5) Open a dired buffer for the CTSM git repository from step (1):
C-x C-f ~/temporary/ctsm_temp

(6) Open magit status buffer:
C-x g

(7) Try to view a particular commit:
M-x magit-show-commit [RET] 90741806a

Emacs now hangs indefinitely with CPU usage near 100%. C-g doesn't help.

Sending emacs a SIGUSR2 generates the following backtrace:

```
  magit-section-goto(#<magit-section nil [commit-message commitbuf] 302-[312-]356>)
  magit-section-backward()
  magit-section--backward-protected()
  magit-section--backward-find(#f(compiled-function () #<bytecode 0x11a398a59e8531fa>))
  magit-imenu--diff-prev-index-position-function()
  imenu-default-create-index-function()
  imenu--make-index-alist(t)
  which-function()
  which-func-update-1(#<window 3 on magit-revision: ctsm_temp>)
  which-func-update()
  apply(which-func-update nil)
  timer-event-handler([t 0 0 500000 t which-func-update nil idle 0 nil])
```

This problem doesn't show up for all commits, but I have seen it for multiple commits. I can't tell what is different about these problematic commits, but at least some of them (including the one given above) are older commits that were originally made in svn and then migrated to git.

[...]

PR #4606, opened 2022-02-21
Refactor Imenu support

tarsius:
I actually noticed that this could be greatly simplified a long time ago but back then I decided it was not worth the effort, but now I have done it because of #4602.

> Should be fixed by e47973968.

That fixed the immediate problem and I am leaving it in place for now because `magit-section-backward` would otherwise also get stuck at the same position. That command does not enter an infinite loop and it is easy to recover by using `previous-line` or something like that, but we should find a better solution, probably by teaching the section movement commands about such parent and child sections that begin at the same location.

The status buffer actually begins with such a combo but here it does not better because there's nothing we could move to anyway and `magit-section--backward-find` also wasn't affected.

`magit-section--backward-find` and `magit-section--backward-protected` were a mistake though, and I am removing them in 63d4dfb5127812f41b2ad47fafd87a0771b0e160.

As it goes, while refactoring, I ran into some other things that needed an overhaul.


kyleam:
Didn't test anything on my end, but looks very nice based on quick read of the series.

---

> Subject: [PATCH 05/10] magit-status-mode: Add more to Imenu index
>
> Explicitly exclude the headers, including everything else.  Previously
> we explicitly we tried to explicitly include everything else, but that

`s/we explicitly we tried to explicitly/we tried to explicitly/`?

PR #4620, opened 2022-03-04
apply blame highlight-overlay to just one line

### LINKED ISSUE #4619: Bug with magit-blame highlight overlay causes incorrect combining
I think I came across a bug that has been quietly present for a long time but can have nasty consequences since 44c58868c997f0b86d66faeed2f0b29064f8d0b9. I only caught it because I recently found a typo in some collaboratively written code, used `magit` to blame it, and then literally (politely) "blamed" the apparent author via email to request a fix, but they didn't remember writing it. Later, a different collaborator took the credit and fixed it. Blaming the file with `git blame` from the command line appropriately attributed the line in question to the correct author, but the blame overlay in `magit` told a different story.

In brief, I think the bug is in [the second line](https://github.com/magit/magit/blob/613682a70631178caee15d1b5a16fae90da4af7b/lisp/magit-blame.el#L597) of `magit-blame--make-highlight-overlay`. I think the intent is to add a "highlight overlay" to the first link of the current chunk (i.e., from `beg` to `eol`, but `(line-end-position)` returns the position of `eol` of wherever `point` currently happens to be rather than the end-of-line corresponding to `beg`. This results in highlight overlays sometimes stretching over vast distances (e.g., `beg` is at top of the file and `point` is at the end). This can then confuse the new logic introduced in 44c58868c997f0b86d66faeed2f0b29064f8d0b9 that was intended to merge neighboring chunks associated with the same revision.

The consequences can be seen by making the following simple history in a fresh git repo. Let file `test` have two revisions. At the root commit it looks like
```
this file
has some lines
this is the last line
```

and then a second commit where the second line is edited so that it reads
```
this file
has three lines
this is the last line
```

If I put point anywhere on the second line and then blame using magit, I get this incorrect result:

image

whereas if I put point on the first or third line and blame, then I get this correct result
image

From carefully stepping through, I think the issue is that when point is on the second line, the bug discussed above results in the highlight overlay (associated with the first line) stretching down to the end of the second line, and then when it's working on the 3rd line the highlight on the 2nd line confuses [`magit-blame-chunk-at`](https://github.com/magit/magit/blob/613682a70631178caee15d1b5a16fae90da4af7b/lisp/magit-blame.el#L550) into thinking that the 2nd line is associated with the root commit (which is also the case for the 3rd commit) and then the chunks get (erroneously) combined under the new logic introduced in 44c58868c997f0b86d66faeed2f0b29064f8d0b9.

This bug can also make highlights display analogously weird (and they probably have done so for a while), but it's not quite as dramatic as when headings are used.

I have a straightforward fix for this that I'll submit as a PR shortly. It also seems to remedy the behavior of highlight display and moreover seems to anecdotally improve performance (perhaps because making all of those giant overlays is expensive).

dankessler:
fix #4619, see there for more detailed discussion

In brief, `magit-blame--make-highlight-overlay` should add an overlay from `beg`
to end-of-line that contains `beg`. Previously, however, it was from `beg` to
end-of-line that contained point, but this should fix that.

Wrapping things in `save-excursion` and `save-restriction` is probably overkill
since `magit-blame--make-highlight-overlay` is currently only called from
`magit-blame--make-overlays` where it is already warpped in `save-excursion` and
`save-restriction`, but it's included to be extra cautious.

### DIFF
--- a/lisp/magit-blame.el
+++ b/lisp/magit-blame.el
@@ -594,7 +594,9 @@ modes is toggled, then this mode also gets toggled automatically.
     (magit-blame--update-heading-overlay ov)))
 
 (defun magit-blame--make-highlight-overlay (chunk beg)
-  (let ((ov (make-overlay beg (1+ (line-end-position)))))
+  (let ((ov (make-overlay beg (save-excursion
+                                (goto-char beg)
+                                (1+ (line-end-position))))))
     (overlay-put ov 'magit-blame-chunk chunk)
     (overlay-put ov 'magit-blame-highlight t)
     (magit-blame--update-highlight-overlay ov)))


tarsius:
Thanks for locating this issue, I think I occasionally saw some odd behavior that was caused by this.

I changed the patch a bit, dropping the save-excursion and moving the rest of the additions.

PR #4627, opened 2022-03-11
Minimal change correcting description to remote.NAME.pushurl variable

angel-devicente:
Minimal change in documentation to change the description of the remote.NAME.pushurl variable


kyleam:
Thanks for spotting that.  

I regenerated the corresponding texi file and amended your commit.  (I also tweaked the commit message to more closely follow this project's conventions.)

PR #4641, opened 2022-03-18
Don’t try to evaluate startup assertions when byte-compiling.

### LINKED ISSUE #4640: Building still fails if Git isn't installed/found
This is about building Magit from a fresh checkout in a CI environment where Git isn't available. To reproduce on GNU/Linux, you can e.g. make Git temporarily unavailable by bind-mounting it to a non-executable file:

```bash
make clean
bwrap --dev-bind / / --ro-bind /etc/hostname /usr/bin/git make lisp LOAD_PATH="-L . -L ../dash/ -L ../with-editor/lisp/"
```

Compiling the first few files will work, but then compiling magit-status will fail:

```none
Compiling magit-status.el

In toplevel form:
magit-status.el:32:1: Error: Searching for program: Permission denied, git
make[1]: *** [Makefile:77: magit-status.elc] Error 1
```

I think the logic in https://github.com/magit/magit/blob/9fa3e4995ef160faf9e3b8587a958e94067b41f2/lisp/magit.el#L618-L620 doesn't catch this case. `(cl-eval-when (load eval) ...)` will work as intended when magit.el is byte-compiled, but not when it's loaded as a dependency of some other file to be byte-compiled. This should probably be changed to something that tests whether byte-compilation is active at all, e.g. whether `byte-compile-current-file` is non-nil.

phst:
Fixes #4640

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -616,7 +616,7 @@ and Emacs to it."
      ;; See comment above.
      "https://github.com/magit/magit/wiki/Don't-set-$GIT_DIR-and-alike"))
   ;; Git isn't required while building Magit.
-  (cl-eval-when (load eval)
+  (unless byte-compile-current-file
     (magit-git-version-assert))
   (when (version< emacs-version magit--minimal-emacs)
     (display-warning 'magit (format "\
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks!

PR #4648, opened 2022-03-31
Fix doc typo

blaenk:
Hope little edits like these aren't considered a nuisance, if so feel free to close, just noticed it while I was reading and figured I'd contribute back this tiny fix.

Thank you for working on Magit!


tarsius:
Thanks!

PR #4658, opened 2022-04-12
Fix imenu/which-func-mode in magit-submodule-list buffers

kljohann:
`tabulated-list-get-entry` returns a vector.  This led to a `(wrong-type-argument listp ["path/to/submodule" ...])` error in `which-func-ff-hook` when using Emacs 28.1.

### DIFF
--- a/lisp/magit-repos.el
+++ b/lisp/magit-repos.el
@@ -32,6 +32,7 @@
 ;;; Code:
 
 (require 'magit-core)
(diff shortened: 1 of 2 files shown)


tarsius:
That's not the only thing that was broken. Luckily simply deleting the broken code does the trick.

kljohann:
I can confirm that this works. Thanks for the quick fix!

PR #4662, opened 2022-04-18
Support revisions that match commit message

telotortium:
Git supports revisions that match a commit message:

       :/<text>, e.g. :/fix nasty bug
           A colon, followed by a slash, followed by a text, names a commit whose commit message matches the specified regular expression. This name returns the
           youngest matching commit which is reachable from any ref, including HEAD. The regular expression can match any part of the commit message. To match
           messages starting with a string, one can use e.g.  :/^foo. The special sequence :/! is reserved for modifiers to what is matched.  :/!-foo performs a
           negative match, while :/!!foo matches a literal ! character, followed by foo. Any other sequence beginning with :/! is reserved for now. Depending on
           the given text, the shell’s word splitting rules might require additional quoting.

However, Magit would not support these, because it appends `^{commit}`
in several places to revisions. To fix this, recognize this special
revision syntax and handle it specially.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -2517,7 +2517,7 @@ Staging and applying changes is documented in info node
     "show" "-p" "--cc" "--format=" "--no-prefix"
     (and (member "--stat" magit-buffer-diff-args) "--numstat")
     magit-buffer-diff-args
-    (concat magit-buffer-revision "^{commit}")
+    (magit--rev-dereference magit-buffer-revision)
     "--" magit-buffer-diff-files))
 
 (defun magit-insert-revision-tag ()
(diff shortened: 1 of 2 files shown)


tarsius:
I have made some cosmetic changes. I am unsure about the name of the new function and also of those of some related existing functions. The time might have come to rethink those (which might take a while).

telotortium:
Sure, take as much time as you need.

kyleam:
> I am unsure about the name of the new function [...]

Going off of how Git refers to the operation, perhaps something with "peel" or "dereference"? `magit-rev-peel-to-commit`?

---

I was curious if the similar `<rev>^{/<text>}` form also fails when `^{commit}` is appended, but that works fine.

tarsius:
I went with `magit--rev-dereference'. Also see the commit message.

PR #4664, opened 2022-04-21
magit-stash: Disable external diff tools

bcc32:
External diff tools may cause git to produce errors like "unrecognized
input" when the output is fed back into `git apply`.  Pass
`--no-ext-diff` to `git diff` where necessary in magit-stash.el.

### DIFF
--- a/lisp/magit-stash.el
+++ b/lisp/magit-stash.el
@@ -317,7 +317,7 @@ current branch or `HEAD' as the start-point."
                            (magit-stash-create message index worktree untracked))
         (if (eq keep 'worktree)
             (with-temp-buffer
-              (magit-git-insert "diff" "--cached")
+              (magit-git-insert "diff" "--cached" "--no-ext-diff")
               (magit-run-git-with-input
                "apply" "--reverse" "--cached" "--ignore-space-change" "-")
               (magit-run-git-with-input


tarsius:
Hard-coding this in `magit-stash-save` seems appropriate.

Doing the same in `magit-stash-insert-section` doesn't. This affects every diff not just stash diffs, so if we were to hardcode it, then `magit--insert-diff` would be a more appropriate location.

However at times one may wish to view the output of the external diff tool even if one cannot apply it (after all that's why one setup an external diff tool in the first place). For that reason Magit already supports adding the `--no-ext-diff` argument when that is desired. Press `D` to show and change the diff arguments used in the current buffer.

bcc32:
> Hard-coding this in `magit-stash-save` seems appropriate.
> 
> Doing the same in `magit-stash-insert-section` doesn't. This affects every diff not just stash diffs, so if we were to hardcode it, then `magit--insert-diff` would be a more appropriate location.

Ah, on second reading I think this just affects displayed stash diffs.  But in any case, it's for display only so it should allow external diff tools.  Thanks.

> However at times one may wish to view the output of the external diff tool even if one cannot apply it (after all that's why one setup an external diff tool in the first place). For that reason Magit already supports adding the `--no-ext-diff` argument when that is desired. Press `D` to show and change the diff arguments used in the current buffer.

tarsius:
Thanks!

PR #4666, opened 2022-04-26
Fix behaviour of magit-visit-ref outside of magit menus

Jell:
I noticed that `magit-visit-ref-behavior` was no longer respected in the Magit Refs buffer, and I narrowed down the issue to this part of the code (`magit-visit-ref`) not working according to its intended behavior (in fact, it was behaving the opposite of the documentation afaiu):

``` 
When invoked from a menu this command always behaves like `magit-show-commit'.
``` 

This solves the issue I had with `magit-visit-ref-behavior` not being respected and should actually implement the documented / intended behavior of the function.


tarsius:
Upsi...

Thanks!

PR #4683, opened 2022-05-08
Resolve magit-find-file calling hooks

elken:
Calling `magit-find-file` in a rust lsp environment causes `#'lsp` to be
called again with the contents of the older file instead.

This commit seems to resolve it, but I'm not sure if I also have to wrap
the let-bind to prevent buffer hooks from being called.

See linked issue[1] for more information and a lengthier discussion

[1] https://github.com/doomemacs/doomemacs/pull/6309


tarsius:
Thanks!

PR #4688, opened 2022-05-11
magit-revert-rev-file-buffer: Use set-auto-mode

elken:
For that newly-opened buffer to have syntax coloring this function
briefly sets buffer-file-name and performs (normal-mode t).  This,
in turn, triggers related major-mode hooks, which at least in the
case of lsp causes issues.

An attempt was made to resolve this in f331092df but it didn't seem
to resolve the issue.

Because of how the buffers are named, set-auto-mode couldn't
work out which mode to use; so this has been swapped around.

This PR is slightly more disruptive than the previous change, so I accept that I may have to make other changes :) 

Closes https://github.com/doomemacs/doomemacs/pull/6309

### DIFF
--- a/lisp/magit-files.el
+++ b/lisp/magit-files.el
@@ -152,7 +152,7 @@ then only after asking.  A non-nil value for REVERT is ignored if REV is
 
 (defun magit-get-revision-buffer (rev file &optional create)
   (funcall (if create #'get-buffer-create #'get-buffer)
-           (format "%s.~%s~" file (subst-char-in-string ?/ ?_ rev))))
+           (format "%s:%s" (subst-char-in-string ?/ ?_ rev) file)))
 
 (defun magit-revert-rev-file-buffer (_ignore-auto noconfirm)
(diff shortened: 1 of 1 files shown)


elken:
> fyi, applying this patch makes syntax coloring not work (e.g. rustic-mode is not starting) 😢

Are you sure? It was working fine for me and you said it was working before :p

Patryk27:
> Are you sure? 

Pretty sure, I've just tried it a few times :x

> you said it was working before :p

Which I've then corrected by saying `Ah wait, sorry - I think I've forgotten to undo my change before applying yours` 😅

elken:
Okay, I'll keep this open but for now it may not work

tarsius:
I'm closing this for now. Please ping me when you are ready.

elken:
@tarsius This has been confirmed to be resolved either by doom or magit; so all's well in the end.

Apologies this was left hanging for so long

PR #4701, opened 2022-06-03
Small doc updates

bennorth:
Two minor doc updates:
* Typo in docstring for `magit-diff-buffer-file`.
* Update for the keybindings used to cycle through previous arguments in a transient (`transient-history-prev` and `transient-history-next`).


tarsius:
Thanks!

PR #4710, opened 2022-06-18
Support bookmark-handler-type for Magit bookmarks

skangas:
Hi Jonas,

```
From etc/NEWS in emacs.git:
  *** 'list-bookmarks' now includes a type column.
  Types are registered via a 'bookmark-handler-type' symbol property on
  the jumping function.
```
This adds support for it to Magit.


tarsius:
Thanks!

PR #4715, opened 2022-06-25
magit-push: Fix magit-push-implicitly--desc

darkfeline:
The previous implementation was basically wrong because it didn't
handle push.default or resolve the remotes and refspecs separately
like git push actually does.

An easy example is if you have a branch `fork` tracking `fork/master`, the previous implementation said it would push to `fork/master` even though the default `push.default` value `simple` means it would push to `fork/fork`.

### DIFF
--- a/lisp/magit-push.el
+++ b/lisp/magit-push.el
@@ -300,15 +300,17 @@ what this command will do.  To add it use something like:
   ;; so it doesn't make sense to talk about "pushing to upstream".
   ;; Depending on the options, you could end up pushing to the
   ;; "upstream" remote but not the "upstream" branch, and vice versa.
-  (let ((branch (magit-get-current-branch))
-        (remote (or (magit-get-push-remote)
-                    (magit-get-remote)
-                    (let ((remotes (magit-list-remotes)))
-                      (cond
-                       ((and (magit-git-version>= "2.27")
-                             (= (length remotes) 1))
-                        (car remotes))
-                       ((member "origin" remotes) "origin"))))))
+  (let* ((branch (magit-get-current-branch))
+         (remote (or (magit-get-push-remote branch)
+                     ;; Note: Avoid `magit-get-remote' because it
+                     ;; filters out the local repo case (".").
+                     (magit-get "branch" branch "remote")
+                     (let ((remotes (magit-list-remotes)))
+                       (cond
+                        ((and (magit-git-version>= "2.27")
+                              (= (length remotes) 1))
+                         (car remotes))
+                        ((member "origin" remotes) "origin"))))))
     (if (null remote)
         "nothing (no remote)"
       (let ((refspec (magit-get "remote" remote "push")))
--- a/lisp/magit-push.el
+++ b/lisp/magit-push.el
@@ -291,34 +291,58 @@ what this command will do.  To add it use something like:
   (magit-run-git-async "push" "-v" args))
 
 (defun magit-push-implicitly--desc ()
-  (let ((default (magit-get "push.default")))
-    (or (and (equal default "nothing")
-             (format "nothing (%s is %s)"
-                     (magit--propertize-face "push.default" 'bold)
-                     (magit--propertize-face default        'bold)))
-        (and-let* ((remote (or (magit-get-remote)
-                               (magit-primary-remote)))
-                   (refspec (magit-get "remote" remote "push")))
-          (format "%s using %s"
-                  (magit--propertize-face remote 'magit-branch-remote)
-                  (magit--propertize-face refspec 'bold)))
-        (and-let* ((upstream (and (not (magit-get-push-branch))
-                                  (magit-get-upstream-branch))))
-          (format "%s aka %s"
-                  (magit-branch-set-face upstream)
-                  (magit--propertize-face "@{upstream}" 'bold)))
-        (and-let* ((push-branch (magit-get-push-branch)))
-          (format "%s aka %s"
-                  (magit-branch-set-face push-branch)
-                  (magit--propertize-face "pushRemote" 'bold)))
-        (and-let* ((push-branch (magit-get-@{push}-branch)))
-          (format "%s aka %s"
-                  (magit-branch-set-face push-branch)
-                  (magit--propertize-face "@{push}" 'bold)))
-        (format "using %s (%s is %s)"
-                (magit--propertize-face "git push"     'bold)
-                (magit--propertize-face "push.default" 'bold)
-                (magit--propertize-face default        'bold)))))
+  ;; This implements the logic for git push as documented.
+  ;; First, we resolve a remote to use based on various remote and
+  ;; pushRemote options.
+  ;; Then, we resolve the refspec to use for the remote based on push
+  ;; and pushDefault options.
+  ;; Note that the remote and refspec to push are handled separately,
+  ;; so it doesn't make sense to talk about "pushing to upstream".
+  ;; Depending on the options, you could end up pushing to the
+  ;; "upstream" remote but not the "upstream" branch, and vice versa.
+  (let ((branch (magit-get-current-branch))
+        (remote (or (magit-get-push-remote)
+                    (magit-get-remote)
+                    (let ((remotes (magit-list-remotes)))
+                      (cond
+                       ((and (magit-git-version>= "2.27")
+                             (= (length remotes) 1))
+                        (car remotes))
+                       ((member "origin" remotes) "origin"))))))
+    (if (null remote)
+        "nothing (no remote)"
+      (let ((refspec (magit-get "remote" remote "push")))
+        (if refspec
+            (format "to %s with refspecs %s"
+                    (magit--propertize-face remote 'bold)
+                    (magit--propertize-face refspec 'bold))
+          (pcase (or (magit-get "push.default") "simple")
+            ("nothing" "nothing (due to push.default)")
+            ((or "current" "simple")
+             (format "%s to %s"
+                     (magit--propertize-face branch 'magit-branch-current)
+                     (magit--propertize-face (format "%s/%s" remote branch)
+                                             'magit-branch-remote)))
+            ((or "upstream" "tracking")
+             (let ((ref (magit-get "branch" branch "merge")))
+               (if ref
+                   (format "%s to %s"
+                           (magit--propertize-face branch 'magit-branch-current)
+                           (cond
+                            ((string-prefix-p "refs/heads/" ref)
+                             (magit--propertize-face
+                              (format "%s/%s" remote
+                                      (substring ref (length "refs/heads/")))
+                              'magit-branch-remote))
+                            ((not (string-match "/" ref))
+                             (magit--propertize-face (format "%s/%s" remote ref)
+                                                     'magit-branch-remote))
+                            (t (format "%s as %s"
+                                       (magit--propertize-face remote 'bold)
+                                       (magit--propertize-face ref 'bold)))))
+                 "nothing (no upstream)")))
+            ("matching" (format "all matching to %s"
+                                (magit--propertize-face remote 'bold)))))))))
 
 ;;;###autoload (autoload 'magit-push-to-remote "magit-push" nil t)
 (transient-define-suffix magit-push-to-remote (remote args)


tarsius:
The part in parenthesis from git-config(1) seems incorrect or at least misleading:

    push.default
           Defines the action git push should take if no refspec is given
           (whether from the command-line, config, or elsewhere).

`push.default` trumps *configured* refspecs:

```
$ git config push.default nothing
$ git config branch.draft.merge refs/heads/draft
$ git config branch.draft.remote origin
$ git push -n
fatal: You didn't specify any refspecs to push, and push.default is "nothing".
```

Your implementation would claim that `magit-push-implicitly` would do something, which is not correct.

(I have just updated my implementation to handle the "nothing" case correctly. Please rebase your implementation on top of that.)

darkfeline:
https://github.com/git/git/blob/39c15e485575089eb77c769f6da02f98a55905e0/builtin/push.c#L242

lisp/magit-push.el
```diff
@@ -291,34 +291,44 @@ what this command will do.  To add it use something like:
...
+  ;; pushRemote options.
+  ;; Then, we resolve the refspec to use for the remote based on push
+  ;; and pushDefault options.
+  ;; Note that the remote and refspec to push are handled separately,
+  ;; so it doesn't make sense to talk about "pushing to upstream".
+  ;; Depending on the options, you could end up pushing to the
+  ;; "upstream" remote but not the "upstream" branch, and vice versa.
+  (let ((remote (magit-get-push-remote)))
+    (unless remote
+      (setq remote (magit-get-remote)))
```
kyleam, line 305: `magit-get-push-remote` and `magit-get-remote` are used, so this won't return a remote when on a detached head, leading to "nothing (no remote)".  That's okay in most cases, because a bare `git push` would fail anyway.  But `git push` works fine with a detached when `push.default=matching`.

Also, a style nit:

```elisp
(let ((remote (or (magit-get-push-remote)
                  (magit-get-remote))))
  ...)
```

lisp/magit-push.el
```diff
@@ -291,34 +291,44 @@ what this command will do.  To add it use something like:
...
+                    (magit--propertize-face remote 'bold)
+                    (magit--propertize-face refspec 'bold))
+          (pcase (or (magit-get "push.default") "simple")
+            ("nothing" "nothing (due to push.default)")
+            ((or "current" "simple")
+             (format "%s\n"
+                     (magit-branch-set-face
+                      (format "%s/%s" remote (magit-get-current-branch)))))
+            ((or "upstream" "tracking")
+             (let ((refspec (magit-get "remote" remote "merge")))
```
kyleam, line 320: `remote.$remote.merge`?  This should look at `branch.$branch.merge`, no?

kyleam (changes requested):
I'll need to take a closer look and will have other comments, but a few notes from an initial look.

lisp/magit-push.el
darkfeline: fixed

lisp/magit-push.el
darkfeline: > `magit-get-push-remote` and `magit-get-remote` are used, so this won't return a remote when on a detached head, leading to "nothing (no remote)". That's okay in most cases, because a bare `git push` would fail anyway. But `git push` works fine with a detached when `push.default=matching`.

That's odd, how is it deciding what remote to use?  Normally it uses the push or default/"merge" remote of the current branch.  Is it defaulting to `origin` or if there's only one remote?

> 
> Also, a style nit:
> 
> ```emacs-lisp
> (let ((remote (or (magit-get-push-remote)
>                   (magit-get-remote))))
>   ...)
> ```

I'd consider that a legitimate fix rather than a style nit, done thanks

lisp/magit-git.el
```diff
@@ -1705,10 +1705,13 @@ according to the branch type."
...
-    (let ((remote (magit-get "branch" branch "remote")))
-      (unless (equal remote ".")
-        remote))))
+  (if (or branch (setq branch (magit-get-current-branch)))
+      (let ((remote (magit-get "branch" branch "remote")))
+        (unless (equal remote ".")
+          remote))
+    (let ((remotes (magit-list-remotes)))
+      (if (= (length remotes) 1) (car remotes)
+        "origin"))))
```
kyleam, line 1714: Let's not modify the behavior of `magit-get-remote`.  Matching Git's behavior isn't necessarily the desirable behavior on Magit's end (and this change, at the very least, breaks `magit-get-some-remote`).

Please instead inline this logic in `magit-push-implicitly--desc`.  However, when there are no remotes, calling `git push` is of no use, so I think it'd clearer to just your original "nothing (no remote)" value in that case.

Two additional notes on the behavior:

 * this doesn't quite match Git's behavior because it will return nil when there is a current branch but `branch.$branch.remote` isn't configured.  (That, in turn, will lead to a type error in `magit-push-implicitly--desc`.)

 * there is version-specific behavior here to consider.  Before 8a649be7e8 (push: default to single remote even when not named origin, 2022-04-29), part of the v2.27 release, Git defaults to "origin" even if there is a single remote with another name.

lisp/magit-push.el
```diff
@@ -291,34 +291,42 @@ what this command will do.  To add it use something like:
...
-          (format "%s aka %s"
-                  (magit-branch-set-face push-branch)
-                  (magit--propertize-face "@{push}" 'bold)))
-        (format "using %s (%s is %s)"
-                (magit--propertize-face "git push"     'bold)
-                (magit--propertize-face "push.default" 'bold)
-                (magit--propertize-face default        'bold)))))
+                  (magit--propertize-face remote 'bold)
+                  (magit--propertize-face refspec 'bold))
+        (pcase (or (magit-get "push.default") "simple")
```
kyleam, line 311: If Git adds another value and a user chooses it, this `pcase` body will return nil, resulting in "(BUG: no description)" as the suffix description.  That seems fine and will lead to a bug report telling us that we need to update the value.  Another option would be to use `pcase-exhaustive`, which would fail a bit more loudly but have the downside of making the push transient non-functional.  So, a plain `pcase` seems like the way to go.  Okay.

lisp/magit-push.el
```diff
@@ -291,34 +291,42 @@ what this command will do.  To add it use something like:
...
+          ("nothing" "nothing (due to push.default)")
+          ((or "current" "simple")
+           (format "%s\n"
```
kyleam, line 314: Hmm, 35da95e9 (magit-push-*--desc: Remove newline at end of description, 2022-06-25) dropped the new lines from the descriptions.  Is adding them back in this PR intentional?

lisp/magit-push.el
```diff
@@ -291,34 +291,42 @@ what this command will do.  To add it use something like:
...
+                   (magit-branch-set-face
+                    (format "%s/%s" remote branch))))
```
kyleam, line 316: This (and the if-arm of the "upstream" case below) would show something like

```
Push
i origin/master
```

in the transient.  That seems confusing to me, because I read "origin/master" as *what* is being pushed (which of course makes no sense) rather than the target of the push.  In contrast, other parts of the transient that explicitly say what is being pushed:

```
Push master to
p ...
u origin/master
e elsewhere

....

Push
o another branch
r explicit refspecs
......
```

So I'd prefer to still say what the current branch is:

```
Push
i master to origin/master
```

lisp/magit-push.el
```diff
@@ -291,34 +291,42 @@ what this command will do.  To add it use something like:
...
-                (magit--propertize-face "git push"     'bold)
-                (magit--propertize-face "push.default" 'bold)
-                (magit--propertize-face default        'bold)))))
+                  (magit--propertize-face remote 'bold)
+                  (magit--propertize-face refspec 'bold))
+        (pcase (or (magit-get "push.default") "simple")
+          ("nothing" "nothing (due to push.default)")
+          ((or "current" "simple")
+           (format "%s\n"
+                   (magit-branch-set-face
```
kyleam, line 315: We already know this is a remote branch, so I'd avoid `magit-branch-set-face`, which does a "remote or local?" check, and instead directly call `(magit--propertize-face ... 'magit-branch-remote)`.

kyleam (changes requested):
A comment on top of my inline comments: the approach used in this project is to reroll the series rather than extend it with fixup commits that resolve issues from previous commits in the series or more generally overwrite things that were just introduced by the series.   Please rebase your branch to avoid having such band-aid commits on top.  How you split things is up to you, but I'd guess the final series would be one to three commits.

Also, as a general remark, I'll just throw out there that I'm not a fan of `magit-push-implicitly--desc`s attempt to say what Git will do, and I think it'd be better to just report the value of `push.default` rather trying capture Git's logic.  I realize that that ship has probably sailed, as `magit-push-implicitly--desc` has been around a while, and I'm probably alone in this opinion.  Regardless, I'm all for this series correcting `magit-push-implicitly--desc`s logic.

lisp/magit-push.el
darkfeline: fixed

darkfeline:
> A comment on top of my inline comments: the approach used in this project is to reroll the series rather than extend it with fixup commits that resolve issues from previous commits in the series or more generally overwrite things that were just introduced by the series. Please rebase your branch to avoid having such band-aid commits on top. How you split things is up to you, but I'd guess the final series would be one to three commits.

I was going to squash when it's ready.  It's easier to see WIP history/diff like this.

> Also, as a general remark, I'll just throw out there that I'm not a fan of magit-push-implicitly--descs attempt to say what Git will do, and I think it'd be better to just report the value of push.default rather trying capture Git's logic. I realize that that ship has probably sailed, as magit-push-implicitly--desc has been around a while, and I'm probably alone in this opinion. Regardless, I'm all for this series correcting magit-push-implicitly--descs logic.

ack

lisp/magit-push.el
darkfeline: done

lisp/magit-push.el
darkfeline: ack

lisp/magit-push.el
darkfeline: done

lisp/magit-git.el
darkfeline: don't understand your note 1

note 2 done

kyleam: > don't understand your note 1

\[ The latest version you pushed doesn't have the problem, but in case you'd still like to understand what I was referring to... \]

The version of the code that I was commenting on looked like this:

``` elisp
(if (or branch (setq branch (magit-get-current-branch)))
      (let ((remote (magit-get "branch" branch "remote")))
        (unless (equal remote ".")
          remote))
    (let ((remotes (magit-list-remotes)))
      (if (= (length remotes) 1) (car remotes)
        "origin")))
```

When there is a current branch, this goes down the `if`-branch. And then when `branch.$branch.remote` is not configured (or if it points to a local upstream), nil is returned.

That differs from Git's logic:

``` c
if (branch && branch->remote_name) {
    if (explicit)
            *explicit = 1;
    return branch->remote_name;
}
if (explicit)
    *explicit = 0;
if (remote_state->remotes_nr == 1)
    return remote_state->remotes[0]->name;
return "origin";
```

### example

Command line output for `git push` with `push.default=current` is

    To /tmp/magit-4715-LcefK96/a
       829f509..abcd18b  other -> other

and for `push.default=matching`

    To /tmp/magit-4715-LcefK96/a
       eab166b..3f2a31d  master -> master
       829f509..abcd18b  other -> other

There is not a remote configured for `other`:

    $ git config branch.other.remote
    $ echo $?
    1

<details>
<summary>setup script</summary>

``` sh
#!/bin/sh

set -eu

tdir=$(mktemp -d "${TMPDIR:-/tmp}"/magit-4715-XXXXXXX)

cd "$tdir"

git init --bare a
git clone a b
(
    cd b
    git commit --allow-empty -mc0
    git push origin HEAD
    git commit --allow-empty -mc2

    git checkout --no-track -b other HEAD
    git commit --allow-empty -mc3
    git push origin HEAD
    git commit --allow-empty -mc4

    git -c push.default=current push -n
    git config push.default matching
    git push -n
)
```

</details>

If I try that with the version of your PR that I was commenting on (992a20f6), `magit-push-implicitly--desc` signals a type error when `push.default` is `matching` and displays an incorrect description ("i nil/other") when `push.default` is `current` or `simple`.

kyleam (commented):
@darkfeline Thanks for the update. At a glance, the code change looks good to me. Please reroll the series how you'd like it to land so that I review the commit messages and structure of the final series. I'll also do another quick round of testing at that point.

lisp/magit-push.el
```diff
@@ -291,34 +291,53 @@ what this command will do.  To add it use something like:
...
+          (pcase (or (magit-get "push.default") "simple")
+            ("nothing" "nothing (due to push.default)")
+            ((or "current" "simple")
+             (format "%s to %s"
+                     (magit--propertize-face branch 'magit-branch-current)
+                     (magit--propertize-face (format "%s/%s" remote branch)
+                                             'magit-branch-remote)))
+            ((or "upstream" "tracking")
+             (let ((refspec (magit-get "branch" branch "merge")))
+               (when refspec
```
kyleam, line 328: When `branch.$branch.merge` isn't set for a branch, this will show "i (BUG: no description)".  That's a valid state, even if `git push` can't do something useful in this case, so it shouldn't be presented as a bug.  So, I think this should be turned into an `if`, with the `else` arm returning a string like "nothing (no upstream)".

lisp/magit-push.el
```diff
@@ -291,34 +291,53 @@ what this command will do.  To add it use something like:
...
+                    (magit--propertize-face refspec 'bold))
+          (pcase (or (magit-get "push.default") "simple")
+            ("nothing" "nothing (due to push.default)")
+            ((or "current" "simple")
+             (format "%s to %s"
+                     (magit--propertize-face branch 'magit-branch-current)
+                     (magit--propertize-face (format "%s/%s" remote branch)
+                                             'magit-branch-remote)))
+            ((or "upstream" "tracking")
+             (let ((refspec (magit-get "branch" branch "merge")))
```
kyleam, line 327: `ref` seems like a better name here than `refspec` here because, unlike `remote.*.{fetch,push}`, `branch.*.merge` doesn't support a full refspec/mapping.

lisp/magit-push.el
```diff
@@ -291,34 +291,53 @@ what this command will do.  To add it use something like:
...
+                 (format "%s to %s"
+                         (magit--propertize-face branch 'magit-branch-current)
+                         (if (string-prefix-p "refs/heads/" refspec)
+                             (magit--propertize-face
+                              (format "%s/%s" remote
+                                      (substring refspec (length "refs/heads/")))
+                              'magit-branch-remote)
+                           (format "%s as %s"
+                                   (magit--propertize-face remote 'bold)
+                                   (magit--propertize-face refspec 'bold)))))))
```
kyleam, line 338: I'm not understanding what this case is handling, I think.  While `branch.*merge` is typically set to a full ref, it can be set to a short refname (though `magit-get-upstream-branch` doesn't support it).  But that's essentially the same case as the "refs/heads/" one above (just with no need to strip out "refs/heads/").

What am I missing?

kyleam (changes requested):
Was hoping to merge this in tonight, but with another read and some testing, have a few more comments/questions.

lisp/magit-push.el
darkfeline: I think `branch.*merge` can be set to a custom/non-head full ref, e.g., I think Gerrit does this for submitting CLs.

https://gerrit-review.googlesource.com/Documentation/user-upload.html#push_create

In any case, this is a fallback that most users won't see.

I've also added support for short refnames.

kyleam (approved):
During my testing this round, noticed an issue for the local upstream case.  I've added a commit on top to deal with that.

(ci failure is recent and unrelated.)

PR #4717, opened 2022-06-30
Default names to non-short format

johanvts:
The `refname:short` format is slow. On my repository (with 1000+ branches) it takes about 3.5seconds to show branches with `short` vs. about 0.2 seconds for the default format. Maybe this should be configurable somehow or depend on number of brances. Having `short` makes magit painful to use on repos with many branches. Tested on multiple versions of git and under windows and linux.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1903,7 +1903,7 @@ SORTBY is a key or list of keys to pass to the `--sort' flag of
             (magit-git-lines "show-ref"))))
 
 (defun magit-list-refnames (&optional namespaces include-special)
-  (nconc (magit-list-refs namespaces "%(refname:short)")
+  (nconc (magit-list-refs namespaces "%(refname)")
          (and include-special
               (magit-list-special-refnames))))
 
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1903,7 +1903,7 @@ SORTBY is a key or list of keys to pass to the `--sort' flag of
             (magit-git-lines "show-ref"))))
 
 (defun magit-list-refnames (&optional namespaces include-special)
-  (nconc (magit-list-refs namespaces "%(refname)")
+  (nconc (magit-list-refs namespaces "%(refname:lstrip=2)")
          (and include-special
               (magit-list-special-refnames))))
 


kyleam (commented):
Thanks for the PR, but I don't this is the right direction to go.  This reason `magit-list-refnames` exists is to return the short name.  `magit-list-refs` already returns `%(refname)` by default.

And any blanket replacement of `magit-list-refnames` with the output of `magit-list-refs` is going to run into problems (beyond just frustrating many users with full refs in all of their prompts) and would need to analyze the downstream code to assess the consequences.  For example, your change here means that `magit-checkout` offers the full refs, causing `git checkout` to check out a detached HEAD rather than a branch.

```
$ git checkout refs/heads/master
HEAD is now at 35da95e9 magit-push-*--desc: Remove newline at end of description
$ git status
HEAD detached at refs/heads/master
nothing to commit, working tree clean
```

---

It's worth noting that, at least in some cases, 1000 refs isn't a point at which things become painfully slow.  In git.git (which currently has 1385 refs, if you clone the maintainer's repo with contributor branches), `magit-list-refnames` takes around 50 ms on my end.  There are a lot of factors to consider, but my point is just that even when you get up into 1000 refs range things aren't necessarily painful, so any steps to improve cases where it is painful needs to take care to not degrade the experience for users that are not hitting those cases.

Also, to be clear, I'm not saying `:short` doesn't add time.  It does, and a large part of that must come from checking for ambiguous refs.  If ambiguous refs aren't a concern, things can be improved a bit by setting `core.warnAmbiguousRefs=false`, but that still doesn't match what you would see from a plain `strip` call.  (The large performance hit over `strip` when `core.warnAmbiguousRefs=false` smells like something that could potentially be improved in git.git.)

```
# $ git for-each-ref | wc -l
# 1385
Benchmark #1: git for-each-ref --format="%(refname:short)"
  Time (mean ± σ):      35.8 ms ±   1.3 ms    [User: 18.9 ms, System: 15.6 ms]
  Range (min … max):    34.1 ms …  37.8 ms    79 runs
 
Benchmark #2: git -c core.warnAmbiguousRefs=false for-each-ref --format="%(refname:short)"
  Time (mean ± σ):      28.7 ms ±   0.5 ms    [User: 15.2 ms, System: 13.1 ms]
  Range (min … max):    27.2 ms …  30.1 ms    101 runs
 
Benchmark #3: git for-each-ref --format="%(refname)"
  Time (mean ± σ):       8.6 ms ±   0.3 ms    [User: 4.3 ms, System: 4.2 ms]
  Range (min … max):     8.2 ms …   9.9 ms    301 runs
 
Benchmark #4: git for-each-ref --format="%(refname:strip=2)"
  Time (mean ± σ):       9.0 ms ±   0.3 ms    [User: 4.6 ms, System: 4.2 ms]
  Range (min … max):     8.5 ms …  10.4 ms    298 runs
 
Summary
  'git for-each-ref --format="%(refname)"' ran
    1.04 ± 0.05 times faster than 'git for-each-ref --format="%(refname:strip=2)"'
    3.33 ± 0.12 times faster than 'git -c core.warnAmbiguousRefs=false for-each-ref --format="%(refname:short)"'
    4.15 ± 0.20 times faster than 'git for-each-ref --format="%(refname:short)"'
```

johanvts:
Thanks for the really nice review. I understand this fix is way too heavy handed, I also had to switch to "%(refname:lstrip=2)" to get correct branch names. I tried "git config core.warnAmbiguousRefs false" but it only lowers the time from ~3.5 seconds to ~3.1 second, so it seems most of the slowness is just gits string formatting. I checked the actual count and there are almost 6.000 branches in my repo, but this should not be so uncommon that it is acceptable that magit crawls. I understand that "%(refname:lstrip=2)" would probably also not be generally acceptable since it doesn't resolve ambiguities but I'm not sure what the right solution is.

tarsius:
It has been reported on the orange site that this was closed without justification. That's not what I usually do but as demonstrated here, it can happen. I'll reopen so we can have another look some other time, but I won't do so now, because just today I started with a "break" (as in giving up to deal with issues as they come in).

I don't see a clear path forward here, but I have just skimmed what has been said so far. I tend to disagree that keeping 6000 branches around is a reasonable thing to do. As a work around you can change the fetch rules to get only the branches *you* are interested in. Maybe there is a reason to preserve all those branches, but any one developer probably doesn't need all of them. There is an opportunity cost in investing time to deal with edge-cases such as this that I expect a relatively small number of users. In such cases it can be reasonable to expect the affected users to use a workaround, at least for a while.

tarsius:
Ah, I cannot actually reopen because the source branch has been deleted. I've added a bookmark.

johanvts:
Thanks @tarsius, just wanted to say that I don't think closing this was unjust after @kyleam 's through response. I just mentioned it on the orange side as it did wonders for me and other users where complaining about potentially similar issues. I have bee using this workaround without issue since July 2022.

tarsius:
If you don't hear back from me in about two months, then feel free to ping me. Since it works for you, others might be willing to make the same tradeoff, so adding an option might be reasonable. But again, I'll have to look at it in depth before I can really have an opinion on that.

kyleam:
@johanvts (and @tarsius) sorry for not following up to last comment in July.

---

@johanvts:
> I also had to switch to "%(refname:lstrip=2)" to get correct branch names.

Right, in contrast to the original commit's switch to a plain `refname`, changing `magit-list-refnames` from `refname:short` to `refname:lstrip=2` preserves the intended behavior of `magit-list-refnames` for the most common namespaces: `refs/heads/foo -> foo`, `refs/remotes/origin/foo -> origin/foo`, and `refs/tags/foo -> foo`.

Some snags/points to consider:

 * The ambiguous ref cases you mention.

 * `:lstrip=2` and `:short` don't behave the same for other namespaces.  For example, it's not uncommon for users to fetch GitHub's `refs/pull/*/head` into a local namespace.  (forge maps these to `refs/pullreqs/*`, and `magit-list-refs-namespaces` includes that by default.)  `:short` would transform `refs/pullreqs/4717` to `pullreqs/4717` while `:lstrip=2` would give `4717`.

   Git commands will accept `pullreqs/4717`, but they won't accept `4717` as a shortened form of `refs/pullreqs/4717`.

   Internal functionality dealing with the `refs/notes/`, `refs/stash/`, `refs/shelved/` namespaces could also be affected, if they use `magit-list-refnames`.  (See below for a `refs/shelved/` example.)

 * Magit's minimum Git version is v2.2.0.  `:strip` isn't available until v2.7.1, and its synonym `:lstrip` was introduced in v2.13.0.  (I wouldn't be against bumping Magit's minimum version, though, given Debian oldoldstable is at v2.11.)

> I understand that "%(refname:lstrip=2)" would probably also not be generally acceptable since it doesn't resolve ambiguities but I'm not sure what the right solution is.

Unlike the original `refname` proposal, I think `refname:strip=2` is worth considering as an opt-in behavior, as @tarsius suggests.  The option would have to make it clear what the potential issues are so that users with performance issues can decide if they want that trade.  That'd of course require a more detailed analysis of downstream code to decide what things we're comfortable just documenting/warning about vs breakage that would need to be addressed in order to add the option.

Here's an example of the latter.  `refname:strip=2` breaks `magit-branch-unshelve`, which assumes it's getting a "shelved/..." value back from `magit-list-refnames`:

https://github.com/magit/magit/blob/28bb411a726f19564dca244b3e9031dffe89e249/lisp/magit-branch.el#L789-L792

The fix in this case could be to use `magit-list-refs` instead.

> I tried "git config core.warnAmbiguousRefs false" but it only lowers the time from ~3.5 seconds to ~3.1 second, so it seems most of the slowness is just gits string formatting.

Thanks for timing/testing on your end.  That looks consistent with my timings.

I don't think it's just string formatting (which `:strip` needs to do too).  I suspect it's mostly due to Git trying to detect name collisions because `core.warnAmbiguousRefs=false` only seems to prevent `for-each-ref` from disambiguating some cases (see demo below).  I'm confused about what the intended behavior is here, but in any case, it's not much help here. 

<details>
<summary>demo</summary>

```
$ git for-each-ref --format='%(refname)'
refs/heads/dup
refs/heads/master
refs/heads/origin/dup
refs/remotes/origin/dup
refs/remotes/origin/master
refs/tags/dup

$ git -c core.warnAmbiguousRefs=true for-each-ref --format='%(refname:short)'
heads/dup
master
heads/origin/dup
remotes/origin/dup
origin/master
tags/dup

$ git -c core.warnAmbiguousRefs=false for-each-ref --format='%(refname:short)'
heads/dup
master
origin/dup
remotes/origin/dup
origin/master
dup

$ git for-each-ref --format='%(refname:lstrip=2)'
dup
master
origin/dup
origin/dup
origin/master
dup
```

</details>

Issue #2145, opened 2015-08-05
Remote urls and spaces


laynor:
Today I changed the remote of a repo using <kbd>M u</kbd>.
By mistake, i inserted a space before the usual git@server:repo url. This resulted the insertion of 

[remote "origin"]
     url = " git@xxxxxxx:yyyy.git"

in .git/config, which, in turn, made git behave "strange", asking the password of the user git (or was it actually ' git'?).

```lisp
(defun magit-remote-set-url (remote url)
  "Change the url of the remote named REMOTE to URL."
  (interactive
   (let  ((remote (magit-read-remote "Set url of remote")))
     (list remote (magit-read-string "Url" (magit-get "remote" remote "url"))))) 
  (magit-set url "remote" remote "url")) ;; trimming the whitespace in `remote'  would be nice!
```

npostavs:
Is an extra space different from any random typo in the URL?

laynor:
well, it's certainly easy to spot as opposed to any random typo. A user id starting with whitespace is certainly fishy.

tarsius:
There are keymaps intended for the minibuffer which make it impossible to insert whitespace. But I don't know if there is a non-hacky way to make `completing-read` and friends use one of those instead of the default.

laynor:
What's so bad about left trimming the remote URLs after the user inserts
them? Am I missing something?

tarsius:
It's no longer possible to type whitespace at these prompts. But it's possible to yank whitespace. If you do that and don't fix it, then you get an error (no automatic trimming).

PR #4723, opened 2022-07-05
magit-remote-add: Allow whitespace in remote URL

undergroundquizscene:
It’s possible to add remote URLs containing whitespace when calling git
directly on the command line (e.g. `git remote set-url something
"host:/path/to/folder name/"`), so I believe magit-remote-add should
also allow whitespace in URLs.

As far as I can see, `'magit-read-url` is only called in
`'magit-remote-add`, so it seems ok to make the modification there.

### DIFF
--- a/lisp/magit-remote.el
+++ b/lisp/magit-remote.el
@@ -90,7 +90,7 @@ has to be used to view and change remote related variables."
   (transient-setup 'magit-remote nil nil :scope remote))
 
 (defun magit-read-url (prompt &optional initial-input)
-  (let ((url (magit-read-string-ns prompt initial-input)))
+  (let ((url (magit-read-string prompt initial-input)))
     (if (string-prefix-p "~" url)
         (expand-file-name url)
       url)))


kyleam (commented):
Thank you for the PR.

> magit-remote-add: Allow whitespace in remote URL
> 
> It’s possible to add remote URLs containing whitespace when calling git
> directly on the command line (e.g. `git remote set-url something
> "host:/path/to/folder name/"`), so I believe magit-remote-add should
> also allow whitespace in URLs.

Fair enough.  And there's also simply local path names.

However, there's some history here.  These sites used to allow whitespace, but that changed with 43aa38b9 (magit-read-string-ns: new function, 2015-08-07), which was prompted by gh-2145.

So, there's really not a single behavior that fits what everyone wants.  We could add an option to control it, but I'm not really sure spaces in remote paths is a common enough case to justify another option.  That commit was almost seven years ago, but I don't recall anyone bringing this up until now.

undergroundquizscene:
> However, there's some history here. These sites used to allow whitespace, but that changed with [43aa38b](https://github.com/magit/magit/commit/43aa38b93d00161bd5f94ee792543611830aa3f7) (magit-read-string-ns: new function, 2015-08-07), which was prompted by [gh-2145](https://github.com/magit/magit/issues/2145).

Ah, good point! (And I see that some of the other places are places I would like to be able to insert spaces as well, e.g. when setting the URL of a remote.)

> So, there's really not a single behavior that fits what everyone wants. We could add an option to control it, but I'm not really sure spaces in remote paths is a common enough case to justify another option. That commit was almost seven years ago, but I don't recall anyone bringing this up until now.

Yeah, that’s fair – I relatively frequently want to be able to do this, but I wouldn’t be surprised if most people never do. A compromise that might work for what I need and to prevent the error in https://github.com/magit/magit/issues/2145 would be to error on leading/trailing whitespace, but I doubt the change is worth the effort.

I can see about maintaining a fork for myself with this change, and I’ll close this issue – thanks!

PR #4738, opened 2022-08-18
Allow magit-clone-url-format to be more dynamic

nbarrientos:
This patch permits `magit-clone-url-format` to be represented as an alist mapping hostnames to URL formats. This way Magit can be configured to translate repository names into URLs when using servers that employ different URL schemes. The current value type (string) is kept for backwards compatibility.

Example usage with sr.ht:

``` 1c-enterprise
  ELISP> (add-to-list 'magit-clone-name-alist '("\\`\\(?:sourcehut:\\|sh:\\)\\([^:]+\\)\\'" "git.sr.ht" "sourcehut.user"))
  (("\\`\\(?:sourcehut:\\|sh:\\)\\([^:]+\\)\\'" "git.sr.ht" "sourcehut.user")
   ("\\`\\(?:github:\\|gh:\\)?\\([^:]+\\)\\'" "github.com" "github.user")
   ("\\`\\(?:gitlab:\\|gl:\\)\\([^:]+\\)\\'" "gitlab.com" "gitlab.user"))

  ELISP> (setq magit-clone-url-format '(("git.sr.ht" . "git@%h:~%n") (nil . "git@%h:%n.git")))
  (("git.sr.ht" . "git@%h:~%n")
   (nil . "git@%h:%n.git"))

  ELISP> (magit-clone--name-to-url "gh:nbarrientos/dotfiles")
  "git@github.com:nbarrientos/dotfiles.git"

  ELISP> (magit-clone--name-to-url "sh:nbarrientos/dotfiles")
  "git@git.sr.ht:~nbarrientos/dotfiles"

  ELISP> (magit-clone--name-to-url "gl:nbarrientos/dotfiles")
  "git@gitlab.com:nbarrientos/dotfiles.git"
```

Maybe it makes sense to change the default value of `magit-clone-url-format` and set it to the alist and perhaps add sourcehut to the default `magit-clone-name-alist` but up to you. Dunno where you want to draw the line. Happy to keep the sr.ht specific stuff in my own configuration.

Thanks for considering this patch.

### DIFF
--- a/lisp/magit-clone.el
+++ b/lisp/magit-clone.el
@@ -98,7 +98,7 @@ as the username itself."
 the name of the owner.  Also see `magit-clone-name-alist'."
   :package-version '(magit . "3.0.0")
   :group 'magit-commands
-  :type 'regexp)
+  :type 'string)
 
 ;;; Commands
 
--- a/lisp/magit-clone.el
+++ b/lisp/magit-clone.el
@@ -93,12 +93,17 @@ as the username itself."
                        (string :tag "user name or git variable"))))
 
 (defcustom magit-clone-url-format "git@%h:%n.git"
-  "Format used when turning repository names into urls.
-%h is the hostname and %n is the repository name, including
-the name of the owner.  Also see `magit-clone-name-alist'."
+  "Format(s) used when turning repository names into urls.
+%h is the hostname and %n is the repository name, including the
+name of the owner.  The value can be a string (representing a
+single static format) or an alist with elements (HOSTNAME
+. FORMAT) mapping hostnames to formats.  When an alist is used,
+the nil key represents the default.  Also see
+`magit-clone-name-alist'."
   :package-version '(magit . "3.0.0")
   :group 'magit-commands
-  :type 'string)
+  :type '(choice (string)
+                 (alist :key-type string :value-type string)))
 
 ;;; Commands
 
@@ -311,16 +316,24 @@ Then show the status buffer for the new repository."
                   'magit-clone-name-alist)))
 
 (defun magit-clone--format-url (host user repo)
-  (format-spec
-   magit-clone-url-format
-   `((?h . ,host)
-     (?n . ,(if (string-search "/" repo)
-                repo
-              (if (string-search "." user)
-                  (if-let ((user (magit-get user)))
-                      (concat user "/" repo)
-                    (user-error "Set %S or specify owner explicitly" user))
-                (concat user "/" repo)))))))
+  (if-let ((url-format
+            (cond ((listp magit-clone-url-format)
+                   (cdr (or (assoc host magit-clone-url-format)
+                            (assoc nil magit-clone-url-format))))
+                  ((stringp magit-clone-url-format)
+                   magit-clone-url-format))))
+      (format-spec
+       url-format
+       `((?h . ,host)
+         (?n . ,(if (string-search "/" repo)
+                    repo
+                  (if (string-search "." user)
+                      (if-let ((user (magit-get user)))
+                          (concat user "/" repo)
+                        (user-error "Set %S or specify owner explicitly" user))
+                    (concat user "/" repo))))))
+    (user-error
+     "Bogus `magit-clone-url-format' (bad type or missing default)")))
 
 ;;; _
 (provide 'magit-clone)
--- a/docs/RelNotes/3.4.0.org
+++ b/docs/RelNotes/3.4.0.org
@@ -57,6 +57,9 @@ dfe3d03a14 git-commit-save-message: Report whether message was saved
   default) to support partial cloning, a feature that is available as
   of Git v2.17.  #4102
 
+- ~magit-clone-url-format~ can now be configured to an alist to
+  support servers that employ different URL schemes.  #4738
+
 - Added new command ~magit-stash-push~, which allows specifying a list of
   files to be stashed.  #4583
 
--- a/docs/magit.org
+++ b/docs/magit.org
@@ -4333,8 +4333,26 @@ The following suffixes are disabled by default. See
 - User Option: magit-clone-url-format ::
 
   The format specified by this option is used when turning repository
-  names into urls. ~%h~ is the hostname and ~%n~ is the repository name,
-  including the name of the owner.
+  names into urls.  ~%h~ is the hostname and ~%n~ is the repository
+  name, including the name of the owner.  The value can be a string
+  (representing a single static format) or an alist with elements
+  ~(HOSTNAME . FORMAT)~ mapping hostnames to formats.  When an alist
+  is used, the ~nil~ key represents the default format.
+
+  Example of a single format string:
+
+  #+BEGIN_SRC emacs-lisp
+    (setq magit-clone-url-format
+          "git@%h:%n.git")
+  #+END_SRC
+
+  Example of by-hostname format strings:
+
+  #+BEGIN_SRC emacs-lisp
+    (setq magit-clone-url-format
+          '(("git.example.com" . "git@%h:~%n")
+            (nil . "git@%h:%n.git")))
+  #+END_SRC
 
 ** Staging and Unstaging
 
--- a/docs/magit.texi
+++ b/docs/magit.texi
@@ -5389,8 +5389,26 @@ username itself.
 
 @defopt magit-clone-url-format
 The format specified by this option is used when turning repository
-names into urls. @code{%h} is the hostname and @code{%n} is the repository name,
-including the name of the owner.
+names into urls.  @code{%h} is the hostname and @code{%n} is the repository
+name, including the name of the owner.  The value can be a string
+(representing a single static format) or an alist with elements
+@code{(HOSTNAME . FORMAT)} mapping hostnames to formats.  When an alist
+is used, the @code{nil} key represents the default format.
+
+Example of a single format string:
+
+@lisp
+(setq magit-clone-url-format
+      "git@@%h:%n.git")
+@end lisp
+
+Example of by-hostname format strings:
+
+@lisp
+(setq magit-clone-url-format
+      '(("git.example.com" . "git@@%h:~%n")
+        (nil . "git@@%h:%n.git")))
+@end lisp
 @end defopt
 
 @node Staging and Unstaging
--- a/test/magit-tests.el
+++ b/test/magit-tests.el
@@ -303,6 +303,48 @@ Enter passphrase for key '/home/user/.ssh/id_rsa': "
         (should (equal (pop sent-strings) "mypasswd\n")))
       (should (null sent-strings)))))
 
+;;; Clone
+
+(ert-deftest magit-clone:--name-to-url-format-single-string ()
+  (let ((magit-clone-url-format "bird@%h:%n.git")
+        (magit-clone-name-alist
+         '(("\\`\\(?:github:\\|gh:\\)?\\([^:]+\\)\\'" "github.com" "u")
+           ("\\`\\(?:gitlab:\\|gl:\\)\\([^:]+\\)\\'" "gitlab.com" "u"))))
+    (should (string-equal (magit-clone--name-to-url "gh:a/b")
+                          "bird@github.com:a/b.git"))
+    (should (string-equal (magit-clone--name-to-url "gl:a/b")
+                          "bird@gitlab.com:a/b.git"))
+    (should (string-equal (magit-clone--name-to-url "github:c/d")
+                          "bird@github.com:c/d.git"))
+    (should (string-equal (magit-clone--name-to-url "gitlab:c/d")
+                          "bird@gitlab.com:c/d.git"))))
+
+(ert-deftest magit-clone:--name-to-url-format-bad-type-throws-error ()
+  (let ((magit-clone-url-format 3))
+    (should-error (magit-clone--name-to-url "gh:a/b")
+                  :type 'user-error)))
+
+(ert-deftest magit-clone:--name-to-url-format-alist-different-urls-per-hostname ()
+  (let ((magit-clone-name-alist
+         '(("\\`\\(?:example:\\|ex:\\)\\([^:]+\\)\\'" "git.example.com" "foouser")
+           ("\\`\\(?:gh:\\)?\\([^:]+\\)\\'" "github.com" "u")))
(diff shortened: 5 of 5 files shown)


lisp/magit-clone.el
```diff
@@ -93,12 +93,16 @@ as the username itself."
...
+  "Format(s) used when turning repository names into urls.
+%h is the hostname and %n is the repository name, including the
+name of the owner. The value can be a string (representing a
+single static format) or an alist with elements (HOSTNAME FORMAT)
+mapping hostnames to formats. When an alist is used, the nil key
+represents the default.  Also see `magit-clone-name-alist'."
   :package-version '(magit . "3.0.0")
   :group 'magit-commands
-  :type 'regexp)
+  :type '(choice (regexp)
```
kyleam, line 104: Hmm, isn't the original type wrong?  It looks like it should be `string`.  Would you mind adding an initial commit to this PR that fixes that?

lisp/magit-clone.el
```diff
@@ -311,16 +315,22 @@ Then show the status buffer for the new repository."
...
-                  (if-let ((user (magit-get user)))
-                      (concat user "/" repo)
-                    (user-error "Set %S or specify owner explicitly" user))
-                (concat user "/" repo)))))))
+  (let ((url-format
+        (cond ((listp magit-clone-url-format)
+               (cdr (or (assoc host magit-clone-url-format)
+                        (assoc nil magit-clone-url-format))))
+              ((stringp magit-clone-url-format)
+               magit-clone-url-format))))
```
kyleam, line 323: Given this is a user option, it probably makes sense to give a user error if this expression ends up nil (either because it it falls all the way through the `cond` due to an invalid type or because the user didn't add a nil fallback to their alist).

kyleam (commented):
Thanks for the patch.  Sounds like a good direction to go.

> Maybe it makes sense to change the default value of magit-clone-url-format and set it to the alist and perhaps add sourcehut to the default magit-clone-name-alist but up to you

We already have  "gh/github" and "gl/gitlab" in `magit-clone-name-alist`, so I'd be in favor of giving sourcehut the same treatment by making both the `magit-clone-url-format` and `magit-clone-name-alist` changes you suggest.

Other comments:

 * The manual might need an update (didn't check closely).

 * The example usage and text below that are helpful/appreciated as part of the PR description, but those bits can be pruned from the commit message.

 * Jonas is on break, so fair warning that we may have to revisit decisions when he gets back.

lisp/magit-clone.el
nbarrientos: I think so. I didn't change it to reduce the noise produced by the patch. It's added now as a separate commit as you suggested.

lisp/magit-clone.el
nbarrientos: Yup, that's a great idea indeed. It's added.

nbarrientos:
> Thanks for the patch. Sounds like a good direction to go.

Thanks for the review.

> We already have "gh/github" and "gl/gitlab" in magit-clone-name-alist, so I'd be in favor of giving sourcehut the same treatment by making both the magit-clone-url-format and magit-clone-name-alist changes you suggest.

It'd be cool indeed if Magit was able to translate repository names in Sourcehut to URLs out of the box. I'm happy to wait for Jonas' input on this one, for the moment I'll just keep it as local configuration.

I've also updated the manual and made the commit message slimmer.

lisp/magit-clone.el
```diff
@@ -311,16 +315,24 @@ Then show the status buffer for the new repository."
...
-                    (user-error "Set %S or specify owner explicitly" user))
-                (concat user "/" repo)))))))
+  (let ((url-format
+         (cond ((listp magit-clone-url-format)
+                (cdr (or (assoc host magit-clone-url-format)
+                         (assoc nil magit-clone-url-format))))
+               ((stringp magit-clone-url-format)
+                magit-clone-url-format))))
+    (unless url-format
+      (user-error "Bogus `magit-clone-url-format' (bad type of missing default)"))
```
kyleam, line 325: A purely cosmetic note on something I think it's okay if you leave as is: my impression (without any grepping to back it up) is that the `(let ((x ...)) (unless x ...)  ...)` form is pretty rare in this code base, with the either of the two options below being more common.

<details>
<summary>if-let</summary>

```elisp
(if-let ((url-format
          (cond ((listp magit-clone-url-format)
                 (cdr (or (assoc host magit-clone-url-format)
                          (assoc nil magit-clone-url-format))))
                ((stringp magit-clone-url-format)
                 magit-clone-url-format))))
    (format-spec
     url-format
     `((?h . ,host)
       (?n . ,(if (string-search "/" repo)
                  repo
                (if (string-search "." user)
                    (if-let ((user (magit-get user)))
                        (concat user "/" repo)
                      (user-error "Set %S or specify owner explicitly" user))
                  (concat user "/" repo))))))
  (user-error
   "Bogus `magit-clone-url-format' (bad type or missing default)"))
```

</details>

<details>
<summary>inline `or`</summary>

```elisp
(format-spec
 (or (cond ((listp magit-clone-url-format)
            (cdr (or (assoc host magit-clone-url-format)
                     (assoc nil magit-clone-url-format))))
           ((stringp magit-clone-url-format)
            magit-clone-url-format))
     (user-error
      "Bogus `magit-clone-url-format' (bad type or missing default)"))
 `((?h . ,host)
   (?n . ,(if (string-search "/" repo)
              repo
            (if (string-search "." user)
                (if-let ((user (magit-get user)))
                    (concat user "/" repo)
                  (user-error "Set %S or specify owner explicitly" user))
              (concat user "/" repo))))))
```

</details>

kyleam:
> It'd be cool indeed if Magit was able to translate repository names in Sourcehut to URLs out of the box. I'm happy to wait for Jonas' input on this one, for the moment I'll just keep it as local configuration.

Fair enough.

lisp/magit-clone.el
nbarrientos: Yes, I totally agree. I've transformed it into an `if-let` as suggested. I've also added some tests to make sure that the code does what it should be doing.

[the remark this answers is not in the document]
nbarrientos: Sure. It should look nicer now.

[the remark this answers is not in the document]
nbarrientos: Ops, indeed. Changed. The tests also validate that the sentence is correct.

lisp/magit-clone.el
```diff
@@ -93,12 +93,16 @@ as the username itself."
                        (string :tag "user name or git variable"))))
 
 (defcustom magit-clone-url-format "git@%h:%n.git"
-  "Format used when turning repository names into urls.
-%h is the hostname and %n is the repository name, including
-the name of the owner.  Also see `magit-clone-name-alist'."
+  "Format(s) used when turning repository names into urls.
+%h is the hostname and %n is the repository name, including the
+name of the owner. The value can be a string (representing a
```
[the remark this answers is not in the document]
nbarrientos: Sorry, it should look better now. Hope I'm not missing any.

kyleam (approved):
> I've also added some tests to make sure that the code does what it should be doing.

Thanks for taking the time to do that.

---

This is in a good state in my view.

I've pushed an update that does the following:

 * rebases onto current master (the unrelated CI failure should now be resolved)

 * amends the second commit with regenerated texi

 * adds an entry to docs/RelNotes/3.4.0.org

 * makes minor tweaks to the new test

   The most notable thing is that I removed the error message text.  I think checking for a user-error is sufficient.  Checking the exact message means you need to worry about things like value of `text-quoting-style`.  For example, if a user running the test interactively and had a `text-quoting-style` of `grave`, some of the tests would fail.

   For similar reasons, I tweaked the tests to avoid assuming the default value of `magit-clone-name-alist`.

Please let me know if the update looks okay to you.

<details>
<summary>range-diff</summary>

```
1:  3f3bd09a = 1:  5d35a86a Set type of magit-clone-url-format to string
2:  b16be703 ! 2:  ffbbfa1e Allow magit-clone-url-format to be more dynamic
    @@ Commit message
         that employ different URL schemes. The current value type (string) is
         kept for backwards compatibility.

    + ## docs/RelNotes/3.4.0.org ##
    +@@ docs/RelNotes/3.4.0.org: ** Changes since v3.3.0
    +   default) to support partial cloning, a feature that is available as
    +   of Git v2.17.  #4102
    +
    ++- ~magit-clone-url-format~ can now be configured to an alist to
    ++  support servers that employ different URL schemes.  #4738
    ++
    + - Added new command ~magit-stash-push~, which allows specifying a list of
    +   files to be stashed.  #4583
    +
    +
      ## docs/magit.org ##
     @@ docs/magit.org: ** Cloning Repository
      - User Option: magit-clone-url-format ::
    @@ docs/magit.org: ** Cloning Repository
      ** Staging and Unstaging


    + ## docs/magit.texi ##
    +@@ docs/magit.texi: Cloning Repository
    +
    + @defopt magit-clone-url-format
    + The format specified by this option is used when turning repository
    +-names into urls. @code{%h} is the hostname and @code{%n} is the repository name,
    +-including the name of the owner.
    ++names into urls.  @code{%h} is the hostname and @code{%n} is the repository
    ++name, including the name of the owner.  The value can be a string
    ++(representing a single static format) or an alist with elements
    ++@code{(HOSTNAME . FORMAT)} mapping hostnames to formats.  When an alist
    ++is used, the @code{nil} key represents the default format.
    ++
    ++Example of a single format string:
    ++
    ++@lisp
    ++(setq magit-clone-url-format
    ++      "git@@%h:%n.git")
    ++@end lisp
    ++
    ++Example of by-hostname format strings:
    ++
    ++@lisp
    ++(setq magit-clone-url-format
    ++      '(("git.example.com" . "git@@%h:~%n")
    ++        (nil . "git@@%h:%n.git")))
    ++@end lisp
    + @end defopt
    +
    + @node Staging and Unstaging
    +
      ## lisp/magit-clone.el ##
     @@ lisp/magit-clone.el: (defcustom magit-clone-name-alist
                             (string :tag "user name or git variable"))))
    @@ test/magit-tests.el: (ert-deftest magit-process:password-prompt-observed ()
     +;;; Clone
     +
     +(ert-deftest magit-clone:--name-to-url-format-single-string ()
    -+  (let ((magit-clone-url-format "bird@%h:%n.git"))
    ++  (let ((magit-clone-url-format "bird@%h:%n.git")
    ++        (magit-clone-name-alist
    ++         '(("\\`\\(?:github:\\|gh:\\)?\\([^:]+\\)\\'" "github.com" "u")
    ++           ("\\`\\(?:gitlab:\\|gl:\\)\\([^:]+\\)\\'" "gitlab.com" "u"))))
     +    (should (string-equal (magit-clone--name-to-url "gh:a/b")
     +                          "bird@github.com:a/b.git"))
     +    (should (string-equal (magit-clone--name-to-url "gl:a/b")
    @@ test/magit-tests.el: (ert-deftest magit-process:password-prompt-observed ()
     +                          "bird@gitlab.com:c/d.git"))))
     +
     +(ert-deftest magit-clone:--name-to-url-format-bad-type-throws-error ()
    -+  (let* ((magit-clone-url-format 3)
    -+         (error-msg (cadr (should-error
    -+                           (magit-clone--name-to-url "gh:a/b")
    -+                           :type 'user-error))))
    -+    (should (string-equal
    -+             "Bogus ‘magit-clone-url-format’ (bad type or missing default)"
    -+             error-msg))))
    ++  (let ((magit-clone-url-format 3))
    ++    (should-error (magit-clone--name-to-url "gh:a/b")
    ++                  :type 'user-error)))
     +
     +(ert-deftest magit-clone:--name-to-url-format-alist-different-urls-per-hostname ()
     +  (let ((magit-clone-name-alist
    -+         (append
    -+          magit-clone-name-alist
    -+          '(("\\`\\(?:example:\\|ex:\\)\\([^:]+\\)\\'" "git.example.com" "foouser"))))
    ++         '(("\\`\\(?:example:\\|ex:\\)\\([^:]+\\)\\'" "git.example.com" "foouser")
    ++           ("\\`\\(?:gh:\\)?\\([^:]+\\)\\'" "github.com" "u")))
     +        (magit-clone-url-format
    -+         '(("git.example.com" . "cow@%h:~%n") (nil . "git@%h:%n.git"))))
    ++         '(("git.example.com" . "cow@%h:~%n")
    ++           (nil . "git@%h:%n.git"))))
     +    (should (string-equal (magit-clone--name-to-url "gh:a/b")
     +                          "git@github.com:a/b.git"))
    -+    (should (string-equal (magit-clone--name-to-url "gl:a/b")
    -+                          "git@gitlab.com:a/b.git"))
     +    (should (string-equal (magit-clone--name-to-url "ex:a/b")
     +                          "cow@git.example.com:~a/b"))
     +    (should (string-equal (magit-clone--name-to-url "example:x/y")
    @@ test/magit-tests.el: (ert-deftest magit-process:password-prompt-observed ()
     +                          "cow@git.example.com:~foouser/c"))))
     +
     +(ert-deftest magit-clone:--name-to-url-format-alist-no-fallback-throws-error ()
    -+  (let* ((magit-clone-url-format '(("fail.example.com" . "git@%h:~%n")))
    -+         (error-msg (cadr (should-error
    -+                           (magit-clone--name-to-url "gh:a/b")
    -+                           :type 'user-error))))
    -+    (should (string-equal
    -+             "Bogus ‘magit-clone-url-format’ (bad type or missing default)"
    -+             error-msg))))
    ++  (let ((magit-clone-url-format '(("fail.example.com" . "git@%h:~%n"))))
    ++    (should-error (magit-clone--name-to-url "gh:a/b")
    ++                  :type 'user-error)))
     +
      ;;; Status

```

</details>

nbarrientos:
Thanks! It looks perfect, all your changes make sense to me. I can prepare a separate merge request proposing to change the defaults for `magit-clone-url-format` and `magit-clone-name-alist` if you'd rather to merge this now. Happy to also wait and try to do it all here. Let me know. 

It's been a pleasure to work on this with you. Thanks for your time.

kyleam:
> I can prepare a separate merge request proposing to change the defaults for `magit-clone-url-format` and `magit-clone-name-alist` if you'd rather to merge this now. Happy to also wait and try to do it all here. Let me know.

This stands on its own well enough, so let's merge it and leave the default change to a separate PR.  Thanks.

PR #4741, opened 2022-08-21
Enable cloning repositories in Sourcehut by name

nbarrientos:
The patch adds a new item to the default value of `magit-clone-name-alist` so Magit knows about Sourcehut. The default value of `magit-clone-url-format` has to be also changed to accommodate the URL scheme used by Sourcehut which differs from the ones provided by Github and Gitlab.

See #4738 for more details/discussion.

Before the patch:

``` emacs-lisp
ELISP> (magit-clone--name-to-url "sh:nbarrientos/dotfiles")
*** Eval error ***  Not an url and no matching entry in ‘magit-clone-name-alist’
```

After the patch:

``` emacs-lisp
ELISP> (magit-clone--name-to-url "sh:nbarrientos/dotfiles")
"git@git.sr.ht:~nbarrientos/dotfiles"
ELISP> (magit-clone--name-to-url "sourcehut:pkal/compat")
"git@git.sr.ht:~pkal/compat"
```

In other words, invocations like:

```
M-x magit-clone u sh:nbarrientos/dotfiles
```

should now work out of the box without any custom user configuration.

### DIFF
--- a/docs/magit.org
+++ b/docs/magit.org
@@ -4337,7 +4337,7 @@ The following suffixes are disabled by default. See
   name, including the name of the owner.  The value can be a string
   (representing a single static format) or an alist with elements
   ~(HOSTNAME . FORMAT)~ mapping hostnames to formats.  When an alist
-  is used, the ~nil~ key represents the default format.
+  is used, the ~t~ key represents the default format.
 
   Example of a single format string:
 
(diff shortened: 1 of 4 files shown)


tarsius:
I've also changed the default key to `t`.

nbarrientos:
Thanks for merging :)

PR #4751, opened 2022-09-09
Improve file renames identification

nbarrientos:
Currently Magit can only identify renames in diffstats like:

```
file1 => file2
dir1/dir2/{file1 => file2}
dir1/{dir2/file1 => dir3/file2}
```

This patch allows discovering directory renames like:

```
dir1/{dir2 => dir3}/file1
{dir1 => dir2}/file1
```

by handling any construct `{A => B}` as a replacement rule and simply applying it.

This change improves the user experience when using `magit-jump-to-diffstat-or-diff`.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -2168,9 +2168,8 @@ section or a child thereof."
             (push (magit-decode-git-path
                    (let ((f (match-string 1)))
                      (cond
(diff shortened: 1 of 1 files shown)


kyleam (approved):
Thanks, looks good to me.

I tweaked the subject a bit.  Main change was to add "magit-diff-wash-diffstat: " to the front to follow the style that's typically used in this repo, particularly when only one function is touched.

PR #4752, opened 2022-09-13
git-rebase-mode-show-keybindings: font-lock single-quoted keys

seanfarley:
Previously, the rebase todo list would have the top half of the key bindings font-locked but the bottom half would be single-quoted, e.g. \'c'.

This pull request fontifies all keybindings for uniformity (and visual prettiness).

-------

### DIFF
--- a/lisp/git-rebase.el
+++ b/lisp/git-rebase.el
@@ -806,14 +806,8 @@ By default, this is the same except for the \"pick\" command."
                 (replace-match " " t t nil 2)
                 (replace-match
                  (string-pad
-                  (mapconcat (lambda (key)
-                               (save-match-data
-                                 (substitute-command-keys
-                                  (format "\\`%s'" (key-description key)))))
-                             (cl-remove-if (lambda (key) (eq (elt key 0) 'menu-bar))
-                                           (reverse (where-is-internal
-                                                     cmd git-rebase-mode-map)))
-                             ", ")
+                  (save-match-data
+                    (substitute-command-keys (format "\\[%s]" cmd)))
                   8)
                  t t nil 3)))))))))
 


seanfarley:
Attaching some before and after screenshots for easier comparison.

Before:
Screen Shot 2022-09-10 at 6 48 08 PM

After:
Screen Shot 2022-09-13 at 11 18 38 AM

lisp/git-rebase.el
```diff
@@ -808,8 +808,10 @@ By default, this is the same except for the \"pick\" command."
                  (string-pad
                   (mapconcat (lambda (key)
                                (save-match-data
-                                 (substitute-command-keys
-                                  (format "\\`%s'" (key-description key)))))
+                                  (format (propertize "%s"
```
kyleam, line 811: * You're dropping the `substitute-command-keys` call, so the indentation is now off (i.e. functions like `indent-sexp` will change the indentation).

 * I don't see the point of the `format`.  It looks like you could propertize `(key-description)` directly.

 * Is the `save-match-data` call still needed given you're dropping the `substitute-command-keys` call?

kyleam (commented):
Thanks for spotting this issue and for the PR.

> git-rebase-mode-show-keybindings: font-lock single-quoted keys
>
> Previously, the rebase todo list would have the top half of the key
> bindings font-locked but the bottom half would be single-quoted, e.g.
> \'c'.

I'd appreciate if the commit message gave a brief analysis of what the actual issue is and what the proposed fix is.

It looks like the mis-rendering of some keys (e.g., `\‘c’` rather than `c`) is a relatively recent regression from 5ddf2518 (git-rebase-mode-show-keybindings: Propertize all keys, 2022-05-13).  The keys that are mis-rendered are those that are mapped from the Git command names to `git-rebase-*` commands (i.e. those that aren't inserted based on the `git-rebase-command-descriptions` list).

For these commands/keys, 5ddf2518 started sending the output of `key-description` to `substitute-command-keys`, but that doesn't correctly display keys:

```elisp
(substitute-command-keys
 (format "\\`%s'" (key-description "^N"))) ;; => "\\‘C-n’"
```

Given that, I'd find something like "git-rebase-mode-show-keybindings: Fix key display for mapped commands" to be a clearer commit message subject, I think.  "Mapped commands" is probably a bit unclear, but I'm trying to highlight that these are the set of commands that are being automatically mapped from the Git command name to the `git-rebase-*` command.

---

The two high-level solutions I can think of are

 * just propertize the description directly

 * send the command, not the key description, through `substitute-command-keys`, like the `git-rebase-command-descriptions` handling does

You go with the first approach, and I think that's appealing because it aligns most closely with the state before 5ddf2518.

However, it's worth noting that the source of the different font locking that 5ddf2518 was trying to address is a change in Emacs 28.1.  Before 4a112fd7a6f (Add new face 'help-key-binding' for keybindings in help, 2021-03-08), `substitute-command-keys` didn't apply a face to these keys.

So, for someone running Emacs 27, the first approach would lead to the mapped commands being highlighted while the commands from `git-rebase-command-descriptions` are not.  The second approach would have the benefit of highlighting (or not) consistently, and the only downside that I see is that only one key binding would be mentioned (which seems fine to me).

<details>
<summary>patch for second approach</summary>

```diff
diff --git a/lisp/git-rebase.el b/lisp/git-rebase.el
index 61cdf370..7ac3ab6a 100644
--- a/lisp/git-rebase.el
+++ b/lisp/git-rebase.el
@@ -806,14 +806,8 @@ (defun git-rebase-mode-show-keybindings ()
                 (replace-match " " t t nil 2)
                 (replace-match
                  (string-pad
-                  (mapconcat (lambda (key)
-                               (save-match-data
-                                 (substitute-command-keys
-                                  (format "\\`%s'" (key-description key)))))
-                             (cl-remove-if (lambda (key) (eq (elt key 0) 'menu-bar))
-                                           (reverse (where-is-internal
-                                                     cmd git-rebase-mode-map)))
-                             ", ")
+                  (save-match-data
+                    (substitute-command-keys (format "\\[%s]" cmd)))
                   8)
                  t t nil 3)))))))))

```

</details>


Perhaps the Emacs 27 highlighting mismatch is not worth worrying about (it's cosmetic and will go away with time), but it's at least worth mentioning in the commit message.

seanfarley:
> So, for someone running Emacs 27, the first approach would lead to the mapped commands being highlighted while the commands from git-rebase-command-descriptions are not. The second approach would have the benefit of highlighting (or not) consistently, and the only downside that I see is that only one key binding would be mentioned (which seems fine to me).

That approach sounds really good to me (especially the smaller code footprint). I'll fix up the commit and address the great feedback and code review you gave, thanks!

lisp/git-rebase.el
seanfarley: You're right about all of the above. I've gone with your second approach, so I hope that resolves all these comments.

seanfarley:
I've reworked the commit and its message but do let me know if you'd like any changes to the wording (or code)!

kyleam:
Thanks for the update, LGTM

kyleam (approved):

PR #4756, opened 2022-09-15
Add smerge-keep-all support to magit

FirstLoveLife:


lisp/magit-apply.el
```diff
@@ -776,6 +776,11 @@ a separate commit.  A typical workflow would be:
   (interactive)
   (magit-call-smerge #'smerge-keep-lower))
 
+(defun magit-smerge-keep-all ()
+  "Keep the both versions of the conflict at point."
```
tarsius, line 780: Drop the "the".

tarsius:
> Add smerge-keep-all support to magit

"... to magit" is a bit redundant here. :smile_cat: 

Please use the format we usually use for such commit messages: "smerge-keep-all: New command".

I was aware of the other smerge commands when I decided against adding this one, but that's mostly because I've never used it myself. What happens if the common ancestor of the two sides is shown (in between them)? Is that discarded by this command. If it does handle this case reasonably, then we should probably add it to `magit-call-smerge` as well.

tarsius:
Do you sometimes keep both sides without subsequently having to make manual edits? If not, then I don't quite see the point of adding this command, as it seems to me it would then be better to visit the file first and then invoke the raw smerge command.

FirstLoveLife:
> What happens if the common ancestor of the two sides is shown (in between them)?

You mean three way diff? I just test with
```
[merge]
    conflictstyle = diff3
```

and get a three way diff. `smerge-keep-all` simply preserves all of upper, base, and lower.

FirstLoveLife:
> Do you sometimes keep both sides without subsequently having to make manual edits? If not, then I don't quite see the point of adding this command, as it seems to me it would then be better to visit the file first and to invoke the raw smerge command from there.

Yes. When I merge upstream kernel makefile updates, I often want to preserve both theirs and mine changes. I don't hope to go to files and do smerge-keep-all, `magit-smerge-keep-all` is much simple.

lisp/magit-apply.el
FirstLoveLife: > Drop the "the".

ok, I will do it after return home.

PR #4761, opened 2022-09-16
git-commit-buffer-message: point-at-bol -> line-beginning-position

leungbk:
Per Emacs 29.1 obsoletion of point-at-bol in
b7e867b841f47dcff3aeaef9b5608a237386ce70.


tarsius:
Thanks!

(Now that I finally updated by Emacs build,) I ended up fixing this myself, while also fixing similar issues in this and other packages.

PR #4764, opened 2022-09-19
Fix display buffer action alist syntax

basil-conto:
Technically `(t)` and `t` are equally valid non-`nil` values for the `inhibit-same-window` display buffer action alist entry, but to minimize confusion and proliferation of future mistakes, use the simpler and more widespread boolean.


tarsius:
Thanks!

PR #4771, opened 2022-10-05
manual: fix a typo

kenranunderscore:
Fixed a typo: `rational` -> `rationale`.

I've tried updating the manual and attached its output, but that also changed the `magit.texi` in two other places. Those changes look odd to me. One adds (or removes?) a newline at EOF, the other mangles an s-expression's indentation. So of course just tell me and I'll just drop that part of the commit!

### DIFF
--- a/docs/magit.org
+++ b/docs/magit.org
@@ -8797,7 +8797,7 @@ to pronounce Magit like magic, while taking into account that C and T
 do not sound the same.
 
 The German "Magie" is not pronounced the same as the English "magic",
-so if you speak German then you can use the above rational to justify
+so if you speak German then you can use the above rationale to justify
 using the former pronunciation; ~Mag{ie => it}~.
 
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks!

(You commit also added two typos; I've extended the commit to remove those.)

kenranunderscore:
Yeah, that was what I was clumsily trying to describe above. Using the texinfo org export changed the texi file this way and I was wondering why that was happening, but didn't want to manually edit around in that file after.

Thanks for the amend! Did you also "just" export like the documentation described to get the amended version?

tarsius:
I think something changed in Org very recently that affects this. I did go back a few commits and then the change to the code-block did not happen anymore. But I had to add the trailing newline manually. I don't have the energy to investigate this now and just hope things sort themselves out by the time I have to regenerate the manual again. If not, I'll investigate then (or delay again ;-).

PR #4780, opened 2022-10-08
GitHub Workflow Hardening

st3fan:
This patch aims to harden the GitHub workflows a bit. The following things were done:

- Where possible, limit the scope of the GitHub token (`secrets.GITHUB_TOKEN`) as much as possible, ideally to just `read-all` so that code and actions executed from the workflow have no chance of doing anything bad.
- Consistently set `persist-credentials: false` for the `actions/checkout` action so that the token cannot be easily grabbed by code and actions executed from the workflow.
- Pin down third-party actions to their commit instead of a version branch or `main`/`master` so that builds are more reproducible and lessen the chance of an action doing something bad. (Except for actions written and owned by GitHub itself.)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,6 @@
 name: test
 on: [ push, pull_request ]
+permissions: read-all
 env:
   BUILD_MAGIT_LIBGIT: "false"
 jobs:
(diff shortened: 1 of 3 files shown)


tarsius:
I've just locked down actions a bit organization wise.

The only actions allowed now are those defined in this organization and:

```
actions/checkout@*,
aws-actions/configure-aws-credentials@*,
magit/actions/*,
purcell/setup-emacs@*,
```

I've also limited to certain repositories.

Actions are currently not enabled for `libegit2` and `magit-annex` even though they have ci configuration. I suspect ci is broken for them anyway.

/cc @kyleam Let me know if anything needs any additional permissions.

kyleam:
> Actions are currently not enabled for libegit2 and magit-annex even though they have ci configuration. I suspect ci is broken for them anyway.

Yeah, magit-annex's CI has been out of commission since the travis-ci.org was retired.  I should get around to migrating it to actions.

.github/workflows/test.yml
```diff
@@ -1,5 +1,6 @@
 name: test
 on: [ push, pull_request ]
+permissions: read-all
```
skangas, line 3: Could this be just: `permissions: {}` (that is, not even read permissions)?

tarsius:
I have done about the same now, but

* keep using `purcell/setup-emacs@master`
* use `permissions: {}`
* configure things in `magit/actions` where possible. I added this indirection exactly so that I could make such changes in only one place while having multiple repositories benefit.

Issue #4132, opened 2020-05-31
unexpected behavior based on the value of "display-buffer-alist"


brandelune:
The code that is proposed here:
https://github.com/nex3/perspective-el#some-musings-on-emacs-window-layouts
to stabilize emacs Windows behavior severely conflicts with magit's basic behavior.

```
(setq display-buffer-alist
'((".*" (display-buffer-reuse-window display-buffer-same-window))))
```
The issue I registered there involves 2 frames:
https://github.com/nex3/perspective-el/issues/129

But even with one frame, magit's behavior is unsatisfactory.

With only one frame, here is the behavior:

c c  on staged contents → should display the commit message interface
instead, it reopens the diff display and then, a new c is required to access the commit message interface.

With 2 frames, the same c c kills the frame in which it was called and calls a diff on the front buffer of the other frame.

tarsius:
```elisp
(setq magit-display-buffer-function magit-display-buffer-same-window-except-diff-v1)
```

might be all you need.

brandelune:
Thank you. That was it. Shouldn't it be default behavior ?

tarsius:
No but we should probably do the same as dd83d2dc2a3ab6492de471f2296078397bb0c75f for `magit-display-buffer-traditional` as well.

brandelune:
Ok, thank you very much for the explanations.

gcv:
@tarsius: I'm here because @brandelune used the `display-buffer-alist` setting above at my suggestion (https://github.com/nex3/perspective-el#some-musings-on-emacs-window-layouts), and has drawn my attention to its incompatibility with recent changes in Magit. I took the opportunity to re-read [The Zen of Buffer Display](https://www.gnu.org/software/emacs/manual/html_node/elisp/The-Zen-of-Buffer-Display.html).

Given that Magit now wants to display the diffs along with commit messages, and breaks under what I consider a necessary customization (and frankly, not at all drastic — especially considering that Magit has worked flawlessly with my settings for a decade, and I have not seen any other packages suffer), this seems like an appropriate time to use `display-buffer-overriding-action`. A little grepping suggests that `magit-diff.el` already uses it. And while the manual advises caution, given the critical nature of precise window control in this case, it seems like the right place to use it.

tarsius:
[...]

PR #4790, opened 2022-10-20
magit-display-buffer: inhibit-same-window for magit-display-buffer-noselect

roxma:
I prefer to use display-buffer-same-window by default via display-buffer-alist. It is the most predictable buffer switching behavior.

But there're some cases, e.g. magit-commit, magit-blame that needs to show the diff in side window. In this case, we pass `t` as `display-buffer` ACTION argument to inhibit displaying in the same window.

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -695,7 +695,9 @@ and `magit-post-display-buffer-hook'."
                                 magit-stash-mode
                                 magit-status-mode))))
               '(display-buffer-same-window)
-            nil))) ; display in another window
+            (if magit-display-buffer-noselect
+                t       ; ``t``to display in another window
+              nil))))   ; `nil` for display-buffer-alist
 
 (defun magit-display-buffer-same-window-except-diff-v1 (buffer)
   "Display BUFFER in the selected window except for some modes.


kyleam:
Thanks for the PR.

Considering `magit-display-buffer-noselect` in `magit-display-buffer-traditional` (and perhaps other `magit-display-buffer-*` functions) may be a good way to go, but given the history here (e.g., #4132) and chances for unintended consequences, I don't think it's a change we'll rush into at this point.

> I prefer to use display-buffer-same-window by default via display-buffer-alist.

In that case, you may prefer

```elisp
(setq magit-display-buffer-function #'magit-display-buffer-same-window-except-diff-v1)
```

(There's also the `magit-commit-diff-inhibit-same-window` option, but that's only relevant for the diff shown while committing.)

And if you want to implement the approach of this PR on your end, you can of course override `magit-display-buffer-function` to a custom function:

```elisp
(setq magit-display-buffer-function
      (lambda (buffer)
        (display-buffer buffer magit-display-buffer-noselect)))
```

roxma:
Thank you for your advice, I'll use this version for now:

```elisp
 ;; inhibit-same-window for diff wdinwo triggered by magit-commit
(setq magit-display-buffer-function
      (lambda (buffer)
        (if magit-display-buffer-noselect
            (display-buffer buffer t)    ; t for inhibit-same-window
          (magit-display-buffer-traditional buffer))))
```

PR #4810, opened 2022-11-06
git-commit: Add "Co-developed-by" header

whame:
This is a commonly used header, for example in patches for the Linux kernel (see
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst).


whame:
I'm not sure why the tests are failing?

kyleam:
Thanks for the PR.

> I'm not sure why the tests are failing?

master's builds have been failing for a bit now.  I've started to look at it a couple times and didn't make any progress, and I noticed that Jonas pushed a branch for debugging it, so I think he's looking at it too, but I don't think anyone knows what's going on at this point .

Anyway, your branch passes for me locally.  I have no objections to your changes but will wait a day or so before merging to give others a chance to chime in.

PR #4828, opened 2022-12-16
git-commit-timeout: New variable

L3n41c:
`git-commit-post-finish-hook` is not run until the new commit has been created. If Git takes more than 1 second to do so, the hook isn’t run.

Whereas a 1 second timeout seems reasonable for most use cases, there are use cases where a greater time is desirable.

For ex., when `git` is configured to sign commits with GnuPG, the user might be interactively prompted to enter the passphrase of the GnuPG private key.
If the user takes more than 1 second to type their passphrase, the hook isn’t run.

In order to address such use cases, the new variable `git-commit-timeout` allows to tune the value of this timeout.

### DIFF
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -230,10 +230,10 @@ to edit a commit message.  If a commit is created without the
 user typing a message into a buffer, then this hook is not run.
 
 This hook is not run until the new commit has been created.  If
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks!  Makes sense.

I've tweaked this a bit before merging.

Issue #4027, opened 2020-01-07
Adopt a convention for highlighting commit messages


rpatterson:
Magit uses `magit-log-propertize-keywords` to find parts of the commit message to highlight in log buffers and elsewhere.  I find this very useful making it much easier to visually scan long logs to distinguish related commits and find what I'm looking for.  Currently, `magit-log-propertize-keywords` only handles `sqaush`, `fixup` or "keywords" in brackets, `[foo] ...`.  So I ended up using the latter to make my logs more readable.

Worrying that I was abusing or misusing my commit messages to achieve this, I went looking for any documentation about how one is supposed to use `[foo]` in git commit messages in the wider VCS community and I was surprised I couldn't really find anything authoritative.  So maybe magit should adopt some external, more rigorous convention about commit messages for it's highlighting?  When I did a search, I found [`commitlint`](https://github.com/conventional-changelog/commitlint) which seems to be a pretty good index for references about git commit message conventions that magit might adopt.

tarsius:
I don't think Magit should enforce some commit message convention beyond the commonly accepted conventions it already (optionally!) enforces. (And it's actually `git-commit.el` that is doing that.)

I suppose highlighting `[...]` is a non-standard addition by Magit, but (a) I have seen it done elsewhere (b) it doesn't cause any (or very very minimal) harm for projects that don't use it.

That being said writing some elisp to support e.g. commitlint might be worth it. But that should be done in a separate package (which can use the appropriate hooks provided by `git-commit`).

rpatterson:
What I am asking for *is* about optional highlighting not about linting.  IOW, I am proposing we extend or replace the existing `[...]` highlighting support to highlight [the conventional commits specification](https://www.conventionalcommits.org/en/v1.0.0/) format for messages.  That convention is what  `commitlint` enforces, but I was only mentioning it as a way to point to the convention behind the linter.  I've been using these conventions for some time now, so I can better clarify that bit, sorry for the confusion.

[...]

PR #4840, opened 2023-01-05
Add conventional commits highlights

maikol-solis:
This commit aims to solve #4027 by just adding a new set of regex rules to highlight conventional commits keywords (see https://www.conventionalcommits.org/en/v1.0.0/). 

Please let me know if I missed something. 

Best.

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -1414,7 +1414,13 @@ Do not add this to a hook variable."
       (magit--put-face (match-beginning 0) (1- boundary)
                        'magit-keyword-squash msg))
     (when magit-log-highlight-keywords
+      ;; Case [...]
       (while (string-match "\\[[^[]*?]" msg boundary)
+        (setq boundary (match-end 0))
+        (magit--put-face (match-beginning 0) boundary
+                         'magit-keyword msg))
+      ;; Conventional commits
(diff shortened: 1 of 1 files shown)


tarsius:
> Please let me know if I missed something.

This highlights the keywords at the beginning of every line, but they are only valid in the summary line. This hard-codes some keywords, but projects are free to define their own additions.

Also, this doesn't support footers and while the keyword is the main feature of Conventional Commits that everyone who uses that always uses, once we add that feature it is only a matter of time until someone requests that we add support for footers as well. Highlighting the keyword and footers is only part of the story; subsequently someone will ask that we provide commands to insert them, similar to the commands we already provide for inserting git trailers.

Having all these features would be useful but it is not necessary to add them right away. (And likewise the work is not done once an initial version of all these features has been implemented.) This is why both @kyleam and I have said, that this should be implemented as a separate package, whose sole purpose is to support Conventional Commits in the `git-commit` buffer (and, possibly in the future, other parts of Magit).

maikol-solis:
Thanks Jonas for your comments. I agree that my solution is a little rough, and I just wanted to contribute to magit. Please feel to close the PR if you like. 

In #4027 there is a good enough solution just to highlight those commits in the log buffer. Maybe someone could use it as a starting point to create a separate package for this purpose. I can contribute testing or reviewing in any case.  

Best.

tarsius:
You could contribute to magit's *wiki* at https://github.com/magit/magit/wiki/Tips-and-Tricks. :smile: 

From the sound of it the solution in #4027 will more or less satisfy those who have already commented there, so I will close this. But I also think, that creating a package along the feature set I outlined above, would be a good exercise for you are someone else who wants to get a bit more into emacs (lisp).

Cheers.

PR #4842, opened 2023-01-06
Remove unnecessary Compat requires

minad:
Arguably a trivial change, but nevertheless my first PR to Magit ;)


tarsius:
Right, you mentioned that. I have to admit I already forgot, but would have remembered eventually. :stuck_out_tongue_closed_eyes:

PR #4849, opened 2023-01-20
Add hook `magit-after-successful-clone-hook'

nbarrientos:
I vaguely remember somebody trying to add something like this to Magit in the past but not sure it has been proposed in this context. My apologies in advance if this has been already deemed not necessary.

I need to run code after a new repository is cloned, but only after the Git process has successfully finished. What I want to do is to add the repository to the project's database. 

Before, I was doing something like this:

```lisp
(defun my/magit-maybe-remember-project (repository directory args)
    (when (file-directory-p directory)
      (project-remember-project (cons 'vc directory))))
(advice-add 'magit-clone-internal :after #'my/magit-maybe-remember-project)
```

However this is not good enough as there's a race condition between Git finishing (async process) and my code being executed. It could happen that `magit-clone-internal` exits but `directory` does not exist *yet*. The `when` guard is necessary because the repository may also fail to clone, hence the need to execute code only *after* a *successful* clone operation.

This patch adds a new hook allowing user code to be executed right after the Git `clone` process has exited 0, which guarantees that the repository has been correctly cloned and the target directory exists. The hook can be used like this:

```lisp
(defun my/magit-remember-project (repository directory)
  (project-remember-project (cons 'vc directory)))
(add-hook 'magit-after-successful-clone-hook #'my/magit-remember-project)
```

Thanks.

### DIFF
--- a/docs/magit.org
+++ b/docs/magit.org
@@ -4354,6 +4354,12 @@ The following suffixes are disabled by default. See
             (nil . "git@%h:%n.git")))
   #+END_SRC
 
+- User Option: magit-post-clone-hook ::
+
+  Hook run after the Git process has successfully finished cloning the
+  repository.  When the hook is called, ~default-directory~ is
+  let-bound to the directory where the repository has been cloned.
+
 ** Staging and Unstaging
 
 Like Git, Magit can of course stage and unstage complete files.
--- a/lisp/magit-clone.el
+++ b/lisp/magit-clone.el
@@ -112,6 +112,15 @@ default.  Also see `magit-clone-name-alist'."
                                           (const :tag "Default" t))
                         :value-type (string :tag "Format"))))
 
+(defcustom magit-post-clone-hook nil
+  "Hook run after the repository has been successfully cloned.
+
+When the hook is called, `default-directory' is let-bound to the
+directory where the repository has been cloned."
+  :package-version '(magit . "3.4.0")
+  :group 'magit-commands
+  :type 'hook)
+
 ;;; Commands
 
 ;;;###autoload (autoload 'magit-clone "magit-clone" nil t)
@@ -275,6 +284,8 @@ Then show the status buffer for the new repository."
            (let ((default-directory directory))
              (magit-call-git "sparse-checkout" "init" "--cone")
              (magit-call-git "checkout" (magit-get-current-branch))))
+         (let ((default-directory directory))
+           (run-hooks 'magit-post-clone-hook))
          (with-current-buffer (process-get process 'command-buf)
            (magit-status-setup-buffer directory)))))))
 
--- a/docs/magit.texi
+++ b/docs/magit.texi
@@ -5410,6 +5410,12 @@ Example of by-hostname format strings:
 @end lisp
 @end defopt
 
+@defopt magit-post-clone-hook
+Hook run after the Git process has successfully finished cloning the
+repository.  When the hook is called, @code{default-directory} is
+let-bound to the directory where the repository has been cloned.
+@end defopt
+
 @node Staging and Unstaging
 @section Staging and Unstaging
 


tarsius:
> I vaguely remember somebody trying to add something like this to Magit in the past but not sure it has been proposed in this context.

I usually oppose the addition of "after `git something`" hooks, because with a little more work one can usually use the respective githook. (`emacs` new `--script` argument makes that easier, otherwise more boilerplate is required, as described at https://gist.github.com/ctarbide/99b0ac9f7d6bef19cdd3e9f71b4cbcf7).

But this seems more useful/necessary because there is no corresponding githook.

docs/magit.org
```diff
@@ -4354,6 +4354,12 @@ The following suffixes are disabled by default. See
             (nil . "git@%h:%n.git")))
   #+END_SRC
 
+- User Option: magit-after-successful-clone-hook ::
```
tarsius, line 4357: You didn't actually define such a user option. I think you should.

I prefer a shorter name: `magit-post-clone-hook`.

lisp/magit-clone.el
```diff
@@ -268,6 +268,7 @@ Then show the status buffer for the new repository."
                (setf (magit-get "remote.pushDefault") remote))
              (unless magit-clone-set-remote-head
                (magit-remote-unset-head remote))))
+         (run-hook-with-args 'magit-after-successful-clone-hook repository directory)
```
tarsius, line 271: Why are you doing this here exactly and not earlier or later in the sentinel? Doing it at the very end seems reasonable to me.

tarsius (changes requested):

docs/magit.org
nbarrientos: > You didn't actually define such a user option. I think you should.

Actually I did but I left it in my staging area :wink: Not sure about the value of `:package-version`, though. Please let me know.

> I prefer a shorter name: `magit-post-clone-hook`.

Sure, no problem. 

Both changes pushed and history rewritten.

lisp/magit-clone.el
nbarrientos: You're right, it's now placed right after all the extra `git` calls and before the status buffer is setup. Would that be fine?

lisp/magit-clone.el
```diff
@@ -275,6 +284,7 @@ Then show the status buffer for the new repository."
            (let ((default-directory directory))
              (magit-call-git "sparse-checkout" "init" "--cone")
              (magit-call-git "checkout" (magit-get-current-branch))))
+         (run-hook-with-args 'magit-post-clone-hook repository directory)
```
kyleam, line 287: Given this passes arguments, ideally the option wouldn't be named with "-hook" at the end (see `(info "(elisp)Hooks")`).  Or perhaps the hook doesn't need to be called with arguments, and it'd be sufficient to run it within the let-bound `default-directory` from line 284.

nbarrientos: Hi Kyle,

Yes, it's probably a good idea to follow the conventions :grinning:. Indeed a hook that ends with `-hook` is expected to be argument-less.

I'm happy to change it to a "normal" hook, it'll cover sufficiently my user case assuming that `default-directory` is set to the directory where the repository has been cloned, as you're suggesting. This could be added as a note to the documentation of the hook so it's possible to know a bit about the context of the callback without having to look at the code.

However I think that adding it under the umbrella of the `let` clause you're suggesting wouldn't be enough as it's conditional to creating a sparse checkout. I could add another `let` block though right after the process finishes successfully, something like:

```diff
@@ -271,6 +271,8 @@ defun magit-clone-internal
            (magit-process-sentinel process event)))
        (when (and (eq (process-status process) 'exit)
                   (= (process-exit-status process) 0))
+         (let ((default-directory directory))
+           (run-hooks 'magit-post-clone-hook))
          (when checkout
            (let ((default-directory directory))
              (when set-push-default
``` 

It could also be at the very end as it is currently. 

Please let me know what you'd prefer.

kyleam: > This could be added as a note to the documentation of the hook [...]

Right, if you go with the argument-less function, it should certainly be mentioned in the docs.

> However I think that adding it under the umbrella of the `let` clause you're suggesting wouldn't be enough [...]

Yes, sorry,  I didn't look closely enough at the context.

> It could also be at the very end as it is currently.

Like Jonas above, I think at the end seems like a natural spot.

tarsius: It is generally preferred to use hooks without any arguments and if that is not possible, then convention is to name it `*-functions`. Here it seems unnecessary to use arguments; binding `default-directory` is enough and remote url can be retrieved from the git configuration if needed (as can other pieces of information one might need).

tarsius (changes requested):
https://github.com/magit/magit/pull/4849/files#r1089873451

lisp/magit-clone.el
nbarrientos: Agreed. The patch should be aligned now to what has been discussed :)

nbarrientos:
No hurry at all but just to make sure that we're not deadlocked waiting for each other this should be ready to be re-reviewed. Thanks again.

tarsius:
Sorry for the delay!

PR #4851, opened 2023-01-21
magit-patch-apply: apply patch with reverse option -R.

diasbruno:
This patch includes the reverse (`-R`) option to `magit-patch-apply`

The patch to the command is `a -> a -R -> a`

Screenshot from 2023-01-20 20-42-23

Screenshot from 2023-01-20 21-05-12

### DIFF
--- a/lisp/magit-patch.el
+++ b/lisp/magit-patch.el
@@ -242,7 +242,8 @@ which creates patches for all commits that are reachable from
   ["Arguments"
    ("-i" "Also apply to index" "--index")
    ("-c" "Only apply to index" "--cached")
-   ("-3" "Fall back on 3way merge" ("-3" "--3way"))]
+   ("-3" "Fall back on 3way merge" ("-3" "--3way"))
+   ("-R" "Apply reverse patch" "-R")]
   ["Actions"
    ("a"  "Apply patch" magit-patch-apply)]
   (interactive


tarsius:
Cross-reference #4850.

tarsius:
I'm not sure this one should be added by default, and have decided to not add it, at least for now.

In any case, you can add additional arguments and suffix commands without having to modify the source. It is a bit more complicated than `define-key`, but the idea is the same; users can customize key bindings to their liking.

I've just added this particular binding as an example at ~(info "(transient)Modifying Existing Transients")~. (Not linking to the web version, because I just noticed that publishing is currently broken.)

Note that this could be improved like so:

```elisp
(transient-append-suffix 'magit-patch-apply "-3"
  '("-R" "Apply in reverse" ("-R" "--reverse")))
```

It doesn't matter much, but this tells transient, that the ~-R~ key binding isn't our invention. Instead it derives from the short-argument used by the underlying command itself (here `git apply`).

diasbruno:
@tarsius I find it really useful to revert, like, selecting a diff file on magit, revert and reapply. It's like `patch -Rp1 < file.diff` and then `patch -p1 < file.diff`.

No problem, let me know on future if it makes sense to add it and I reopen the PR.

Really love magit. Hope to contribute more!

PR #4857, opened 2023-01-24
Fix bug in `magit-log-merged`

umanwizard:
When calculating the position of the given commit in the branch's history (for constructing the revision range), we were erroneously counting _all_ the commits that are in the branch but not ancestors of the given commit. However, revision descriptions like `branch~N` count `N` _first parent_ links; therefore, we need to calculate the position of the commit in the same way.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -2154,12 +2154,15 @@ exists in the current repository is considered its main branch."
                      (cons (magit-get "init.defaultBranch")
                            magit-main-branch-names))))))
 
-(defun magit-rev-diff-count (a b)
+(defun magit-rev-diff-count (a b &optional first-parent)
   "Return the commits in A but not B and vice versa.
-Return a list of two integers: (A>B B>A)."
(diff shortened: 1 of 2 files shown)


umanwizard:
The CI failures don't seem related.

tarsius:
> The CI failures don't seem related.

I've fixed those issue now and rebased this branch onto main to trigger a rebuild.

tarsius:
Please add an optional argument to `magit-rev-diff-count` instead of adding a new function and change the commit summary line to `magit-log-merged: ...`, while trying to replace the ellipsis with a more precise but still short summary.

This change seems intuitively right, but the current commit message has a bit of a gap. It mentions that we count the wrong commits, but not in what way the displayed log is wrong because of that. If you can briefly summarize that, that would be helpful.

umanwizard:
Made the changes you requested.

tarsius:
Thanks!

tarsius:
Merged (but I forgot to push here after signing and before merging).

PR #4860, opened 2023-01-25
Docs - performance notes on macOS

ccarlile:
I have been suffering degraded magit performance on macOS for a long time. This [superuser question](https://superuser.com/questions/1452865/git-is-working-painfully-slow) finally gave me the hint to check the performance of the two executables and turns out the native mac `git` is indeed much faster than the one I had installed using `nix`. This resulted in much faster response times from `magit`.

This PR updates the "MacOS Performance" section of the docs to note this


tarsius:
Thanks!

ccarlile:
Likewise, thanks for the awesome project!

PR #4861, opened 2023-01-25
Make gravatar regexes more flexible

sellout:
This extends the regex support in `magit-revision-show-gravatars` so that gravatars can be inserted before or between matched text. Existing regexes
should behave as before, but now a regex like `^\(\)Author:     ` will place the
image in column 0 (after the first matched subexpression).

Implements #4859.

### DIFF
--- a/docs/magit.org
+++ b/docs/magit.org
@@ -3409,7 +3409,9 @@ that they are available here too.
   author's image.  The top half of the image is inserted right
   after the matched text, the bottom half on the next line in the
   same column.  The cdr specifies where to insert the committer's
-  image, accordingly.  Either the car or the cdr may be nil."
+  image, accordingly.  Either the car or the cdr may be nil.  If the
+  regular expression contains grouped subexpressions, the image is
+  inserted after the first subexpression.
 
 - User Option: magit-revision-use-hash-sections ::
 
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -423,7 +423,9 @@ two regular expressions.  The car specifies where to insert the
 author's image.  The top half of the image is inserted right
 after the matched text, the bottom half on the next line in the
 same column.  The cdr specifies where to insert the committer's
-image, accordingly.  Either the car or the cdr may be nil."
+image, accordingly.  Either the car or the cdr may be nil.  If the
+regular expression contains grouped subexpressions, the image is
+inserted after the first subexpression."
   :package-version '(magit . "2.3.0")
   :group 'magit-revision
   :type '(choice (const :tag "Don't show gravatars" nil)
@@ -2771,8 +2773,9 @@ or a ref which is not a branch, then it inserts nothing."
   (save-excursion
     (goto-char beg)
     (when (re-search-forward regexp nil t)
+      (goto-char (or (match-end 1) (match-end 0)))
       (when-let ((window (get-buffer-window)))
-        (let* ((column   (length (match-string 0)))
+        (let* ((column   (length (or (match-string 1) (match-string 0))))
                (font-obj (query-font (font-at (point) window)))
                (size     (* 2 (+ (aref font-obj 4)
                                  (aref font-obj 5))))


tarsius:
I'm not sure how I feel about this implementation. On the one hand it's neat and tidy, but on the other I feel like it is not going to help anyone but you.

The change to the docstring does not make it clear that it allows placing the images at the beginning of the line and that the motivation is the alignment problem that arises from using a proportional font.

Once you add this information to the docstring, the tidy but hacky kludge is no longer contained; when reading it it will become clear that it is a bolted on kludge. If we cannot avoid that (and earlier extensions already fall into that category anyway), then we might as well "do it properly", even if that is more involved.

I drafted that last night; see https://github.com/magit/magit/tree/bol-gravatar.

sellout:
Yeah, that looks good to me. I landed on the group approach because I was originally thinking of using lookahead (which would have required no changes to the code, but elisp regex doesn’t support it).

I was also thinking maybe the match could be used for `\1`-style replacement, but didn’t get around to trying to implement that (or finding a satisfactory way to express it). E.g., something (obviously too clunky) like
```elisp
(setq magit-revision-show-gravatars
  '(("^\\(Author: \\)\\(     \\)" . ("\1\i" . "        \i"))
    . ("^\\(Commit: \\)\\(     \\)" . ("\1\i" . "        \i"))))
```
where `("\1\i" . "        \i")` is a pair of regex replacement strings (top half and bottom half) with `\i` as a pseudo backslash construct representing the image half. So that would change
```
origin/gravatar-position 5bc9d4d9eb727051153679be9dc06de6a963efce
Author:       Greg Pfeil <greg@technomadic.org>
AuthorDate:   Tue Jan 24 23:42:12 2023 -0700
Commit:       Greg Pfeil <greg@technomadic.org>
CommitDate:   Wed Jan 25 00:02:59 2023 -0700
```
to
```
origin/gravatar-position 5bc9d4d9eb727051153679be9dc06de6a963efce
Author: ⎡img⎤ Greg Pfeil <greg@technomadic.org>
        ⎣img⎦ Tue Jan 24 23:42:12 2023 -0700
Commit: ⎡img⎤ Greg Pfeil <greg@technomadic.org>
        ⎣img⎦ Wed Jan 25 00:02:59 2023 -0700
```

sellout:
One other issue I noticed with putting the image at the beginning of the line is that folding the initial section in the revision leaves the top half of the Gravatar image in place:
Screenshot 2023-01-25 at 09 11 24

I don’t ever fold that intentionally, but in case it’s obvious to you what is going on there, I thought I’d mention it.

sellout:
Or maybe this could all just be collapsed into `magit-revision-header-format` with a pseudo-format placeholder.

tarsius:
> One other issue I noticed with putting the image at the beginning of the line is that folding the initial section in the revision leaves the top half of the Gravatar image in place:

Then we have greater care of where markers go. This inserts at boarder between the header and the body. Make sure the marker stays before inserted image.

> Or maybe this could all just be collapsed into `magit-revision-header-format` with a pseudo-format placeholder.

I don't understand.

sellout:
Re: `magit-revision-header-format` …

Right now I have something like
```elisp
(setq magit-revision-headers-format
   "Author: %aN <%aE>
        %ad
Committer: %cN <%cE>
           %cd
")
(setq magit-revision-show-gravatars '("^Author: " . "^Committer: "))
```
which results in something like
```
Author: ⎡img⎤ Greg Pfeil <greg@technomadic.org>
        ⎣img⎦ Tue Jan 10 15:32:19 2023 -0700
Committer: ⎡img⎤ Greg Pfeil <greg@technomadic.org>
           ⎣img⎦ Tue Jan 10 15:32:19 2023 -0700
```
I was thinking maybe `magit-revision-show-gravatars` could be replaced by placeholders like `%ag` and `%cg`:
```elisp
(setq magit-revision-headers-format
   "Author: %ag %aN <%aE>
        %ad
Committer: %cg %cN <%cE>
           %cd
")
```
 Magit could replace them with something that Git won’t try to process (e.g., `<![MAGIT[author gravatar]]>`, then pass it with `--pretty`, and when Magit receives the response, replace `<![MAGIT[author gravatar]]>` with the top half of the image, and find the matching column in the next line to place the bottom half.

Or even make top (`%aG`) & bottom (`%ag`) half explicit:
```elisp
(setq magit-revision-headers-format
   "Author: %aG %aN <%aE>
        %ag %ad
Committer: %cG %cN <%cE>
           %cg %cd
")
```
which would make alignment more explicit, and _might_ also help with proportonal font alignment.

tarsius:
> which would make alignment more explicit, and might also help with proportonal font alignment.

It would also mean that if the image hasn't been cached yet, the placeholder would be displayed while fetching it.

tarsius:
See #4876 for why I am closing this.

I have pushed bfa63365515e045520bc01f5cdef1f3b095a80d2 instead. It obviously doesn't do what you want but at least it avoids misaligning the gravatar halves.

Figuring out why the retroactively inserted image doesn't get hidden when the section is collapsed, would be worth figuring out. I have tried but failed, and do not have the patience to investigate it any further at this time.

PR #4863, opened 2023-01-26
magit-pull: support `--no-rebase` argument to `git pull`.

umanwizard:
In recent versions of git, non-rebase merges are no longer the default. Instead, when trying to pull divergent branches, if `--no-rebase` is not passed explicitly (and if the default has not been overridden in config), an error message is printed to the effect that the user must specify how to deal with the divergent branch.

### DIFF
--- a/lisp/magit-pull.el
+++ b/lisp/magit-pull.el
@@ -42,11 +42,12 @@
 (transient-define-prefix magit-pull ()
   "Pull from another repository."
   :man-page "git-pull"
-  :incompatible '(("--ff-only" "--rebase"))
+  :incompatible '(("--ff-only" "--rebase" "--no-rebase"))
   [:description
    (lambda () (if magit-pull-or-fetch "Pull arguments" "Arguments"))
    ("-f" "Fast-forward only" "--ff-only")
    ("-r" "Rebase local commits" ("-r" "--rebase"))
+   ("-m" "Merge commits" "--no-rebase")
    ("-A" "Autostash" "--autostash" :level 7)]
   [:description
    (lambda ()


kyleam (commented):
Thanks.  I'm not a pull user, and I don't think Jonas is either, so updates in this area are appreciated.

> magit-pull: support `--no-rebase` argument to `git pull`.

convention nit: drop the trailing period

Following similar commits in this repo, this could be reduced to "magit-pull: Add --no-rebase".

> In recent versions of git, non-rebase merges are no longer the
> default.

`s/non-rebase merges/merges/`?  Otherwise I'm left wondering what a "rebase merge" is.

> Instead, when trying to pull divergent branches, if
> `--no-rebase` is not passed explicitly (and if the default has not
> been overridden in config), an error message is printed to the effect
> that the user must specify how to deal with the divergent branch.

To put some versions and commits on this:

 * In v2.27.0, `git pull` started warning if the user did not specify to rebase or merge (by giving an option or setting a config variable): d18c950a69 (pull: warn if the user didn't say whether to rebase or to merge, 2020-03-10).

 * In v2.33.1, this switched from a warning to an error: 031e2f7ae1 (pull: abort by default when fast-forwarding is not possible, 2021-07-22).

> This change allows the user to explicitly specify that the divergent
> branches should be merged.

Hmm, wouldn't the general course of action be for them to configure a relevant option (e.g., through `magit-branch-configure`) rather than explicitly pass `--no-rebase` or `--rebase` each time?  And once the user configures a default (which Git is trying hard to get you to do), it seems to me that the case for adding `--no-rebase` is the same before or after Git switched from the merge default to aborting: allow the user to do a one-time override of a configured rebase default.

In any case, I think it makes sense to allow the ability to trigger a merge via an argument, but I'm not sure that adding `--no-rebase` is the best way to go about it.  The main thing I'm wondering is whether it'd be better to make the `--rebase` option read a value.  At first this could be just "false" (aka `--no-rebase`) or "true" (same as a bare `--rebase`), but attaching a value means that later when someone comes along and asks for the "merges" value, it's just a matter of updating the reader.  (Note that the `--rebase[=false|true]` spelling has been around since Git v1.8.5.)

I suppose the main argument against that is breaking the muscle memory of anyone that regularly used `magit-pull` and `-r`.  But I'm not sure anyone uses it enough to have muscle memory for it.  Wouldn't you just set the config value (or activate the transient option by default)?

tarsius:
I agree with @kyleam.

Having both `--no-rebase` and `--rebase` seems wrong. We could offer only the one necessary to override the configured default. But I tend to think we should offer all four available values.

I don't feel like working on this now, but I have added it to my private todo list.

tarsius:
Since d79bba7665c4f878e65d17957649d2b2c05f5e27 it's possible to set `--rebase` to `false` (or `true`, `merges` or `interactive`).

umanwizard:
Thank you!

PR #4864, opened 2023-01-28
magit-auto-revert-mode autoload never sees init file

dickmao:
The magit-auto-revert-mode defvar in magit-autoloads will always see a null global-auto-revert-mode.

### DIFF
--- a/lisp/magit-autorevert.el
+++ b/lisp/magit-autorevert.el
@@ -64,16 +64,14 @@ is enabled."
                 function))
 
 (defcustom magit-auto-revert-tracked-only t
-  "Whether `magit-auto-revert-mode' only reverts tracked files."
+  "Obsolesced.  `magit-auto-revert-mode' applies only to magit tracked files.
+To auto-revert all files, use `global-auto-revert-mode'."
   :package-version '(magit . "2.4.0")
   :group 'magit-auto-revert
-  :type 'boolean
-  :set (lambda (var val)
-         (set var val)
-         (when (and (bound-and-true-p magit-auto-revert-mode)
-                    (featurep 'magit-autorevert))
-           (magit-auto-revert-mode -1)
-           (magit-auto-revert-mode))))
+  :type 'boolean)
+(make-obsolete-variable 'magit-auto-revert-tracked-only
+                        'global-auto-revert-mode
+                        "Magit 3.3.1")
 
 (defcustom magit-auto-revert-immediately t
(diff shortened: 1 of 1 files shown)


tarsius:
This code has a backstory. A user got very angry with me for "doing something so utterly stupid and and dangerous as enabling some auto-revert-mode by default. That results in dataloss! You don't care about your users!" That was very unpleasant and I would like to avoid that user showing up again, due to the "regression" that the merge of this pull-request would be in their eyes. They might even be right about that being a regression.

What problem does this pull-request solve? The commit message is very sparse. It mentions that it is "superfluous to check global-auto-revert-mode" but it does not mention any negative effect of doing it "needlessly". Are there any?

I don't think this check is actually superfluous. `global-auto-revert-mode` is enabled by default, yes. But that doesn't mean that it is impossible for a user to disable it. Users should have the choice to leave `global-auto-revert-mode` enabled, to enable `magit-auto-revert-mode` instead, or to enable neither. I think your change makes it harder to do the latter. It might still be possible to do it, but the order and techniques used to achieve that might matter. Or in other words, there is a risk that certain configuration that currently prevents both modes from being enabled, would break because of this change.

dickmao:
The following somewhat pathological case exploits the timing of the init-kludge.  The user manually calls `auto-revert-mode' on a non-magit-tracked file, turns off `magit-auto-revert-mode`, then finds the file loses auto-revert-mode.

```
${EMACS:-emacs} -Q -f package-initialize \
--visit ~/.emacs \
--eval "(cl-assert (not (featurep (quote magit-autorevert))))" \
--eval "(call-interactively (function auto-revert-mode))" \
--eval "(cl-assert auto-revert-mode)" \
--eval "(custom-set-variables (quote (magit-auto-revert-mode nil)))" \
--eval "(princ (format \"manual auto-revert-mode %s\!\n\" \
          (if auto-revert-mode \"stayed on\" \"got turned off\")) \
          (function external-debugging-output))" \
--eval "(call-interactively (function save-buffers-kill-emacs))"
```
The root problem is upstream.  `global-auto-revert-mode' should have been implemented as a proper globalized minor mode, and `magit-auto-revert-mode' tries too hard playing the foster parent filling in for GNU's deadbeat dad.

My emacs fork rewrites global-auto-revert-mode to something sensible, and the above commit is what I run for my magit fork.  It should also play nice with GNU Emacs.  Anyway, I don't expect it to get merged, so I'm closing.

PR #4875, opened 2023-02-10
Bump compat dependency, seq is loaded by Compat

minad:
Hello @tarsius,

I just tried the newest Compat with Magit and I am observing seemingly spurious compilation failures (only on 25 and 27). Any ideas what could be causing this? The difference is that Compat now requires seq, since seq is preloaded on Emacs 29. These failures only occur because I removed the seq requires in Magit, but this means that the seq loading does not work as expected.

Daniel


minad:
Okay, I believe it should be fixed by https://github.com/emacs-compat/compat/commit/5bd2c405cfbf0b7adff3dbd2316baec650271eea. I suspect that the problem is the same as last time, when interpreted usage of Compat was broken. How can I run this PR against the newest Compat commit https://github.com/emacs-compat/compat/commit/5bd2c405cfbf0b7adff3dbd2316baec650271eea?

tarsius:
> It works after I forked your workflows! I must say, your rigorous CI workflows are great at shaking out Compat issues :)

Rigorous but also still very bare bones.

E.g., as you noticed, I am currently getting dependencies from the mirror, which means I am always testing against the versions from last Monday (which is when I tend to update the mirror). I'm not doing that because it is the right thing to do but because I don't want to maintain a list of the dependencies' repositories for each package. I intend to maintain a global list of package2repo mappings, but haven't gotten around to that yet.

tarsius:
The main goal for the current incarnation of the ci workflows, was to increase the odds that future improvements would only result in changes to the workflow repository itself, as opposed to having to edit every repository that uses them. We'll see if that works out. Doing everything the right way, is more of a distant goal, and making it usable in repositories that diverge from my idiosyncrasies is a non-goal.

I think now the time has come to look into getting the dependencies from their upstreams. Because the alternative is to update packages whenever something like this comes up. That does take a while. In fact I am currently waiting for it to complete. :grin:

minad:
While I think your workflows are well thought out, I am not sure if I like the shared workflows, since this makes it harder to try something quickly like adjusting the repository where something is pulled from. 

I have decided for myself to treat each of my Emacs projects as a separate entity. This also reflects the spirit of my packages which are decoupled by design. But then, I also don't maintain automatic tests for most of the packages since testing UI interaction in Emacs seems to be quite painful. How would tests for Vertico or OSM look like? Obviously one can trivially test some of the pure functions, which do one precise thing, but this is not where the complexity is located.

I may add tests in the future, but there are also other complications like having to maintain the github workflow yammels and ensuring that they are ignored everywhere (see also https://github.com/melpa/melpa/issues/8287). Currently my repositories are very bare bone and only contain what I want to distribute also such that I don't have to mess with these ignore mechanisms. So maybe solving it once and for all centrally as you did is the way to go.

PR #4876, opened 2023-02-11
Support variable-pitch faces in magit-log

sellout:
- specify `fixed-pitch` inheritance for `magit-hash` and `magit-log-graph` to make the initial part align;
- apply properties to the graph even when `--color` isn’t specified;
- apply `magit-log-graph` even when there is an ANSI color face;
- apply `magit-hash` even when there is a signature face; and
- apply `magit-hash` to the spaces used for alignment.

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -163,7 +163,8 @@ because the latter may make use of Imenu's outdated cache."
                  function))
 
 (defface magit-log-graph
-  '((((class color) (background light)) :foreground "grey30")
+  '((default :inherit fixed-pitch)
+    (((class color) (background light)) :foreground "grey30")
     (((class color) (background  dark)) :foreground "grey80"))
   "Face for the graph part of the log output."
   :group 'magit-faces)
@@ -1242,12 +1243,11 @@ Do not add this to a hook variable."
 
 (defun magit-log-wash-log (style args)
   (setq args (flatten-tree args))
-  (when (and (member "--graph" args)
-             (member "--color" args))
+  (when (member "--graph" args)
     (let ((ansi-color-apply-face-function
            (lambda (beg end face)
              (put-text-property beg end 'font-lock-face
-                                (or face 'magit-log-graph)))))
+                                (plist-put face :inherit '(magit-log-graph))))))
       (ansi-color-apply-on-region (point-min) (point-max))))
   (when (eq style 'cherry)
     (reverse-region (point-min) (point-max)))
@@ -1303,16 +1303,17 @@ Do not add this to a hook variable."
           ('module     (oset section type 'module-commit))
           ('bisect-log (setq hash (magit-rev-parse "--short" hash))))
         (setq hash (propertize hash 'font-lock-face
-                               (pcase (and gpg (aref gpg 0))
-                                 (?G 'magit-signature-good)
-                                 (?B 'magit-signature-bad)
-                                 (?U 'magit-signature-untrusted)
-                                 (?X 'magit-signature-expired)
-                                 (?Y 'magit-signature-expired-key)
-                                 (?R 'magit-signature-revoked)
-                                 (?E 'magit-signature-error)
-                                 (?N 'magit-hash)
-                                 (_  'magit-hash))))
+                               (append (pcase (and gpg (aref gpg 0))
+                                         (?G '(magit-signature-good))
+                                         (?B '(magit-signature-bad))
+                                         (?U '(magit-signature-untrusted))
+                                         (?X '(magit-signature-expired))
+                                         (?Y '(magit-signature-expired-key))
+                                         (?R '(magit-signature-revoked))
+                                         (?E '(magit-signature-error))
+                                         (?N ())
+                                         (_  ()))
+                                       '(magit-hash))))
         (when cherry
           (when (and (derived-mode-p 'magit-refs-mode)
                      magit-refs-show-commit-count)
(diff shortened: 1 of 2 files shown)


sellout:
When I use a variable-pitch `default` face, magit-log (with `--graph --color`) looks like
Screenshot 2023-02-11 at 00 14 06
Customizing the faces locally helps a bit
Screenshot 2023-02-11 at 00 15 28
This is mostly because the existing faces aren’t applied consistently enough – the blank lines don’t get the same face as the hashes they’re aligning with and the graph doesn’t use the right face when it’s colorized. With this PR, it looks like
Screenshot 2023-02-11 at 00 17 46

tarsius:
I don't think it is a good idea to use a variable-pitch font in a code editor. I consider it a mistake to do that. For source code this is just wrong in my opinion. This extends to diffs of code, and magit buffers are full of those. Obviously you disagree and that is fine.

(I would argue that for diffs or code it is even more important to use monospace than it is for source code itself; so my recommendation would be to override the the default face for `magit` and `diff-mode` buffers.)

While looking at your proposals and working on alternatives, I had to temporarily enable a variable-pitch font. My ability to make sense of the code I was looking at instantly took a nose dive and I lost the ability to spot whitespace errors and missing parens. I would also contest that it looked any better; I would say it looked slightly worse, from a esthetic standpoint. From a practical standpoint it was much worse. Again, obviously you disagree, and if I used such a setup for a few months, I would of course learn to cope.

For Org, Info and similar buffers, switching to a variable-pitch font is something I have considered before, but I have not done it because of the work that would involve. For example, I would have to ensure that code blocks *don't* use variable-pitch.

A lot of code in Magit (and many other Emacs packages) assumes the use of a fixed-pitch font. "Fixing" all relevant places would be a lot of work, which I am not willing to put in. As I see it that would only reduce the damage done by the use of a variable-pitch font. Explicitly supporting variable-pitch would also make a lot of code slightly more complicated. To an extend I am willing to do that, but the approach I tend to favor is to override variable-pitch in places where it breaks alignment and such. (I have just done that in response to your pull-requests.)

I also do not intend to merge any pull-requests that add better support for variable-pitch fonts. There is a cost to accepting any contributions, which goes way beyond the initial cost of reviewing and merging the change. When I disagree with the goal of the change in the first place, then it is a burden to accept a contribution. Often I do it anyway, e.g., because I realize that many users disagree with my personal preference.

When it comes to the use of variable-pitch fonts across all of Emacs, I think you are in the minority, but you are certainly not the only one who prefers that. I dread variable-pitch related feature and pull requests, because they tend to be only the first in a series of such requests.

Because of that I have to clearly say *no*, even if that comes across as overly negative.

sellout:
I definitely understand the concern with code complexity, etc. and especially for changes that only affect a small minority of users. I feel like Magit already does an impressive job of supporting variable pitch faces. I actually don’t use variable-width for code or diffs[^1], either.

I’m happy to remove the `fixed-pitch` inheritance from your faces (as I mentioned, doing that locally was my first attempt). But the other parts of this I feel are still justifiable regardless. Three of the changes:

* apply properties to the graph even when `--color` isn’t specified;
* apply `magit-log-graph` even when there is an ANSI color face; and
* apply `magit-hash` even when there is a signature face

are simply semantic improvements. They use the faces specified for those components more consistently. Currently the definition of `magit-log-graph` is “face for the uncolored portion of the graph part of the log output, when `--color` is specified,” as opposed to the documented, “face for the graph part of the log output.”

The fourth change (apply magit-hash to the spaces used for alignment) isn’t a semantic improvement, but in general, aligning things should use the same face[^2] as the thing it’s aligning with. Many changes to the face can locally affect the spacing, so matching the alignment face compensates for that.

For testing (at least the parts that I recommend you keep), instead of setting `default` to a variable-pitch face, you could instead just customize `magit-hash` and `magit-log-graph` with `:height 1.2` or something, which would illustrate the issues.

[^1]: at least not for the added/removed/context bits – I do use variable pitch for the file and hunk headers.

[^2]: Or at least a similar face. Something like
    ```emacs-lisp
    (defface magit-hash-log-alignment '((t (:inherit magit-hash)))
      "Face for blank hash area when “--graph” is used with ‘magit-log’.")
    ```
    can be useful. It would allow someone to make adjustments for the blank lines, like if they have a background color or box on `magit-hash`. This could be customized to remove those in the non-hash lines.

tarsius:
> * apply properties to the graph even when `--color` isn’t specified;

Does it really look better to use variable pitch here? Monospace fonts tend to be more suitable for "ascii art".

`--color` isn't just unconditionally enabled when `--graph` is enabled for performance reasons. So the "`--graph` only case should be handled separately.

> * apply `magit-log-graph` even when there is an ANSI color face; and

I am not sure using `plist-put` is the right thing to do here; I fear it destructively modifies the face definition.

> * apply `magit-hash` even when there is a signature face

I think that is not necessary, because one can just modify the signature faces instead. (But okay, one of those faces is used in one other place.)

> The fourth change (apply magit-hash to the spaces used for alignment)

I've already done that.

sellout:
I’ve made the changes to separate the semantic improvements from the variable pitch changes. I’ll open a PR, but figured we should close this conversation first.

> > * apply properties to the graph even when `--color` isn’t specified;
> 
> Does it really look better to use variable pitch here? Monospace fonts tend to be more suitable for "ascii art". But I guess we could do it anyway.

The intent here is the opposite – the buffer as a whole is `default` (which may be variable pitch), so then `magit-log-graph` can `:inherit fixed-pitch`, but that only works if it’s applied to the graph consistently. The issue also appears with the standard value of `magit-log-graph` – it doesn’t get applied when `color` isn’t used, so the effect of dimming the graph is lost.

> `--color` isn't just unconditionally enabled when `--graph` is enabled for performance reasons. So the "`--graph` only case should be handled separately.

I separated it out in my upcoming change, but I wonder where the performance impact lies – whether it’s in `ansi-color-apply-face-function` scanning for color codes or in the repeated calls to propertize one char at a time. If it’s the former, then splitting it out is right. If it’s the latter, they could stay merged, as only one call to `put-text-property` is made when `--color` isn’t used. In any case, I didn’t want to try to figure that out, so I just separated them.

> > * apply `magit-log-graph` even when there is an ANSI color face; and
> 
> I am not sure using `plist-put` is the right thing to do here; I fear it destructively modifies the face definition. I think `(if face (list face 'magit-log-graph) 'magit-log-graph)` would work, but haven't looked at the documentation again yet.

You’re right. I’m not sure why I didn’t do `(list face 'magit-log-graph)` in the first place. It’s sufficient without the `if`.

> > * apply `magit-hash` even when there is a signature face
> 
> I think that is not necessary, because one can just modify the signature faces instead. (But okay, one of those faces is used in one other place.) Again, I guess we can just do it your way anyway.

Personally I like the separation between the faces for what a thing is (`magit-hash`) and other information you want to present over it (`magit-signature-*`). I’m happy to revert it if you’d rather consider them as alternatives to `magit-hash` – but then I would maybe suggest that the signature faces `:inherit magit-hash`.

PR #4877, opened 2023-02-11
Make it possible to include graph in magit-status

sellout:
- add a `defcustom` to toggle graph display

### DIFF
--- a/lisp/magit-status.el
+++ b/lisp/magit-status.el
@@ -35,6 +35,23 @@
   :link '(info-link "(magit)Status Buffer")
   :group 'magit-modes)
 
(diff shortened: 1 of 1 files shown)


sellout:
What it looks like (combined with the changes from #4876 and some themes)
Screenshot 2023-02-11 at 01 25 52

tarsius:
That is already possible. In the status buffer type `L - g w`.

sellout:
Ah, perfect. Thanks! I had looked around (and asked on some Emacs channels) and couldn’t figure it out. I _thought_ there must already be something. Thank you.

Issue #3430, opened 2018-04-26
Split magit-insert-revision-headers in two


juan-leon:
I would like to split `magit-insert-revision-headers` in two so that I can have
more configuration flexibility for `magit-revision-sections-hook`.

I like the default value for `magit-revision-sections-hook`, and I typically use
it.  But sometimes I need to work with huge repos (hundreds of thousands of
commits, lots of branches and tags).  So reviewing commits with those repos is
very slow when `magit-insert-revision-headers` belongs to
`magit-revision-sections-hook`, and I need to remove it for those repos.

Reason it is slow it because it spawns things like `git branch --contains SHA`
and `git branch --merged SHA` and I think something else for tags.  Those
commands are very fast for small repos, and very slow for humongous ones.

When I remove `magit-insert-revision-headers`, I lose this part of the headers,
that I find very useful, and introduces no noticeable delays:

```
583a487069aab60aa2d1200afb3c14115fbead6e
Author:     XXX
AuthorDate: Tue Apr 17 21:30:43 2018 -0300
Commit:     XXX
CommitDate: Tue Apr 17 21:30:43 2018 -0300
Parent:     aad3c18 Blah blah
```

When I just want to lose the slow part:

```
Containing: BranchA BranchB BranchC ... (81 more)
Merged:     Foo Bar ... (69 more)
Precedes:   TagX (1115)
Follows:    TagY (4)
```

If we had `magit-insert-revision-headers` and
`magit-insert-revision-extra-headers` users would have more flexibility for
configuring `magit-revision-sections-hook` to their liking.

Does it make sense?

tarsius:
You can do that using `(setq magit-revision-insert-related-refs nil)`.

PR #4885, opened 2023-02-24
Add `magit-revision-insert-related-refs-exclude' user option

phil-s:
Provide finer granularity for `magit-insert-revision-headers`.

This is related to #3430.  The suggested solution for that was `(setq magit-revision-insert-related-refs nil)` which eliminates all related refs, but the request was to eliminate only some categories of related ref, which this PR provides for.

I have found myself in a similar situation -- my repository maintains a large history of development branches, and consequently the "Merged" and "Contained" lists can be *enormous*, and it's really annoying to have to scroll past them all to see the actual diff in every revision buffer.

This PR adds a user option `magit-revision-insert-related-refs-exclude` to complement the `magit-revision-insert-related-refs` option, such that items specified by the former are ignored when processing the latter.

I went with an exclusion list so that any categories of related ref added in the future would not be affected if users customize this option, but I'm also not 100% certain of the ideal UI here.  This was a simple approach which enabled me to keep some of the related refs while hiding the really verbose ones.

A while ago I looked into making these things into collapsible sections, but couldn't figure that out at the time; but if that's possible, that would seem nice.  If that *is* possible, then perhaps a customizable alist with values specifying behaviour for each category of related ref would be good (either showing in full, showing collapsed if there are "too many", or ignoring/hiding).


tarsius:
> Provide finer granularity for `magit-insert-revision-headers`.

Seems reasonable. I think I'll merge this.

> A while ago I looked into making these things into collapsible sections, but couldn't figure that out at the time; but if that's possible, that would seem nice. If that _is_ possible, then perhaps a customizable alist with values specifying behaviour for each category of related ref would be good (either showing in full, showing collapsed if there are "too many", or ignoring/hiding).

It should be possible, but I wonder how useful that would be. I believe we already abbreviate if the list in one category gets too long, or else we probably should. Expanding the rest probably isn't possible yet, so maybe we should add that.

(If we were going to do what you suggest it would make sense to do it now, instead of risking having to later change `magit-revision-insert-related-refs-exclude` to an alist and having to deal with old flat-list values. On the other hand, we could also treat this independently; always show just one line of branches per type and add a `+` button to show the rest.)

But using up too much space to show a long list, is the smaller problem. The big problem is the amount of work it takes to determine all the relevant branches, when there is an excessive number of branches, isn't it?

phil-s:
> (If we were going to do what you suggest it would make sense to do it now, instead of risking having to later change `magit-revision-insert-related-refs-exclude` to an alist and having to deal with old flat-list values.

Fair.  I've pushed a new commit switching to an alist user option, which would allow for other values besides `nil` and `t` in future.

phil-s:
I don't know (and am not being shown) what's wrong with those test runs, btw.  I'm assuming it's all going to sort itself out, but if there's an *actual* error related to this branch, I can't see it -- opening any of the specific failures just shows "Error: " followed by nothing at all.  It sounds like failures building a test environment, though, so I'm otherwise ignoring them.

tarsius:
> (and am not being shown)

- Click on "Compile / Compile / Compile using Emacs snapshot (pull_request) Failing after 1m [Details](https://github.com/magit/magit/actions/runs/4302149446/jobs/7561850198)"
- Click on "[Compile / Compile using Emacs snaps](https://github.com/magit/magit/actions/runs/4302149446/jobs/7561850198#logs)"
- Click on "Compile (byte-compile-error-on-warn)"
- See the error: "magit-diff.el:2621:19: Error: ‘backward-delete-char’ is for interactive use only; use ‘delete-char’ instead."

To fix that, rebase your branch onto "main".

phil-s:
Thanks for that.  I've figured out why I wasn't seeing it -- github is using a separate domain for the ajax requests needed to show that specific info.

I'll rebase that now.

tarsius:
Thanks!

PR #4886, opened 2023-02-26
README.md: Add section on youtube videos by [System Crafters](https://www.youtube.com/@SystemCrafters/videos)

greg7mdp:
I have no relationship whatsoever with the author of these videos. I think they are excellent and a very good introduction to `magit`, and having them listed in the README can only make `magit` even more popular than it already is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,15 @@ screenshots as presented in the preceding article.
 This article discusses Magit's properties in somewhat more abstract
 terms.
 
+#### [Some nice videos by SystemCrafters](https://www.youtube.com/@SystemCrafters)
+
+A series of three videos to enjoy in order:
+
+- [An Introduction to the Ultimate Git Interface, Magit!](https://www.youtube.com/watch?v=_zfvQkJsYwI)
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks for the suggestion!

I am aware of these videos and though I listed them [here](https://magit.vc/screencasts/). Apparently not; I will have to update that. I also intended for a while to *recommend* a few videos on that package.

I should probably link to that page from Getting Started in the README, but it doesn't seem right to single out videos from one source. Other nice videos also exist.

Deciding which videos to recommend will be a bit of work, so it will be a while until I get around to it. But again, thanks for the push in that direction.

greg7mdp:
Ha, I see you did it. Good!

tarsius:
Due to "recent events" I had to rust it a bit. :grinning:

Such as https://news.ycombinator.com/item?id=34945086 being the current top post on the orange site!

(I think there are more recent videos from the last two years, but right now it was more important to link what we have already and recommend some of the available videos.)

greg7mdp:
Yes, I saw that on HN. Hope you get more sponsors (btw I am one and not using paypal). I think adding the link to videos will help for sure. Previously it was not obvious for the new user that there are so many good videos about using magit. And the videos help a lot because the usage is not obvious.

PR #4888, opened 2023-02-28
magit-commit-diff: Use environment variables supplied by emacsclient

yoichi:
When EDITOR is set to "emacsclient -t" or "emacsclient -c", `git commit -a` or `git commit <pathspec>...` now shows correct diffs.

# Expected behavior

When EDITOR is set to "emacsclient -t" or "emacsclient -c", `git commit -a` or `git commit <pathspec>...` shows correct diffs.

# Observed behavior

It shows already staged changes only.

# How to reproduce

1. export EDITOR="emacsclient -t"
2. edit some file in working tree
3. git commit -a
4. emacsclient displays 2 windows: "COMMIT_EDITMSG" and "magit-diff: repo-name"
5. "magit-diff: repo-name" doesn't show the modification made by step 2

# Limitation

When emacsclient doesn't create a new frame (e.g. EDITOR="emacsclient"), we cannot get the expected behavior.

### DIFF
--- a/lisp/magit-commit.el
+++ b/lisp/magit-commit.el
@@ -563,7 +563,10 @@ See `magit-commit-absorb' for an alternative implementation."
       ;; replaced with the diff buffer.  See #2632.
       (unrecord-window-buffer nil diff-buffer))
     (message "Diffing changes to be committed (C-g to abort diffing)")
-    (let ((inhibit-quit nil))
+    (let ((inhibit-quit nil)
+	  (process-environment
+	   (append process-environment
+		   (frame-parameter nil 'environment))))
       (condition-case nil
           (magit-commit-diff-1)
         (quit)))))


tarsius:
Unfortunately for you I just resolved to take a break from trying to complete all incoming requests as quickly as possible, but instead of ignoring it completely I will ask some questions that come to mind:

- Isn't the value of `GIT_INDEX_FILE` here the same as the default index anyway?
- Does this work for any version of Emacs and Git?
- It seems you are committing from the command-line.  Did you verify that this does not break commits created from Magit?
- Magit is the main focus of this package.  If what you are doing is useful when committing from the command-line, then it *might* also be use when committing from Magit, in which case I would want to use this approach for that too.
- Did you look at the heuristics used to show the right diff?  Could those be improved instead?  If not, how do they interact?  Or maybe that should be thrown out altogether?
- If it is a different index, and doing this causes another diff to be displayed, then I think that is quite likely that that would break many commands, such as "jump to this line in the appropriate blob".
- We are not just interested in the end result, i.e., what is in the index, but need to know what the range of the respective diff is.

All of this and more would have to be answered/addressed before I could merge this.  Since I won't get back to this soon, you have plenty of time to do that.  On the other hand, my first impression is that this cannot be right, but since I haven't actually investigated any of the above points, that impression might very well be wrong.

yoichi:
> * It seems you are committing from the command-line.  Did you verify that this does not break commits created from Magit?

Thank you for the comment. If we invoke magit-commit on the same frame, it will behave unexpectedly.

yoichi:
> * Isn't the value of `GIT_INDEX_FILE` here the same as the default index anyway?

For example, if we do `git add some-file` and `git commit`, then `GIT_INDEX_FILE` is same as the default index.
magit-commit-diff shows the expected diff In that case.

The problem occur when we do `git commit -a` or `git commit some-file`.  `GIT_INDEX_FILE` is not same as the default index in those cases (`GIT_INDEX_FILE=.git/index.lock`).

yoichi:
> If we invoke magit-commit on the same frame, it will behave unexpectedly.
> Normally that frame is used only for editing commit messages, but it's not restricted to doing anything else.

In [0d70ff9](https://github.com/magit/magit/pull/4888/commits/0d70ff930531b78e306df04cc125e6c489d671d0) I've changed to use the environment variables of the process that opened COMMIT_EDITMSG instead of getting them from the frame. Now commits made from magit are no longer broken.

yoichi:
> * Does this work for any version of Emacs and Git?

emacsclient sends environment variables to server since emacs-23.1 (release date: 2009-07-29).
So it works on any supported version of Emacs (magit requires Emacs 25.1 as described in docs/RelNotes/2.90.0.org).

I'll investigate about git later.

yoichi:
> > * Does this work for any version of Emacs and Git?
> 
> emacsclient sends environment variables to server since emacs-23.1 (release date: 2009-07-29). So it works on any supported version of Emacs (magit requires Emacs 25.1 as described in docs/RelNotes/2.90.0.org).
> 
> I'll investigate about git later.

Git 2.0.0 does passes GIT_INDEX_FILE to the editor. So it works on any supported version of git.
https://github.com/git/git/blob/v2.0.0/builtin/commit.c#L884

yoichi:
> * Magit is the main focus of this package.  If what you are doing is useful when committing from the command-line, then it _might_ also be use when committing from Magit, in which case I would want to use this approach for that too.

Magit itself doesn’t use temporary index file, so it will behave the same when GIT_INDEX_FILE variable is not set and when GIT_INDEX_FILE is set to the default index file.

Currently emacsclient sent environment variables to server only if it create a new frame. Magit doesn’t create a frame when it invokes emacsclient , so this change does not affect normal Magit behavior for now. 
On the other hand, commits from the command line will not display correct diffs if the EDITOR setting does not create a new frame, even after this change applied.

I'm going to discuss in emacs-devel whether it's possible to send environment variables even if emacsclient don't create a new frame. Even if it did, it will not affect magit's behavior with this change. It just sets the default path as GIT_INDEX_FILE.

yoichi:
> I'm going to discuss in emacs-devel whether it's possible to send environment variables even if emacsclient don't create a new frame.

By Eli's answer, emacsclient will not send environment variable when it does not create a new frame.
https://lists.gnu.org/archive/html/emacs-devel/2023-04/msg00055.html

zenspider:
Given eli's answer, this should be closed?

tarsius:
I would be happy to close this. I still suspect it would be a lot of work to improve this (what ever that entails) to an extend, even when not making 100% sure there are no regressions and giving up on making it work correct in each and every case. Sure it would be nice if it were perfect, but feel there are many more important things to do first. So if everyone agrees, more or less, then let's just close.

PR #4890, opened 2023-03-02
Fix navigation in status when both section content and section end are nil

### LINKED ISSUE #4181: magit-section-position-in-heading-p: Wrong type argument: number-or-marker-p, nil
In Emacs 27, in a status buffer with changes visible, navigating the buffer (with normal cursor movements), I get an error. I've just updated to Emacs 27 so this might be related.

```
Debugger entered--Lisp error: (wrong-type-argument number-or-marker-p nil)
  (< 148 nil)
  (and section (>= pos (eieio-oref section 'start)) (< pos (or (eieio-oref section 'content) (eieio-oref section 'end))) t)
  (magit-section-position-in-heading-p)
  (magit-next-line 1 1)
  (funcall-interactively magit-next-line 1 1)
  (call-interactively magit-next-line nil nil)
  (command-execute magit-next-line)
```

This happens in the last part of `magit-section-position-in-heading-p` (within the `or` and the 2 `oref` calls):

```emacs
(defun magit-section-position-in-heading-p (&optional section pos)
  "Return t if POSITION is inside the heading of SECTION.
POSITION defaults to point and SECTION defaults to the
current section."
  (unless section
    (setq section (magit-current-section)))
  (unless pos
    (setq pos (point)))
  (and section
       (>= pos (oref section start))
       (<  pos (or (oref section content)
                   (oref section end)))
       t))
```

Magit v2.90.1-1030-g798aff56, Git 2.27.0, Emacs 27.1, gnu/linux

I don't have any recipe to reproduce so you might just want to close the issue :-).

zenspider:
Fixes #4181

### DIFF
--- a/lisp/magit-section.el
+++ b/lisp/magit-section.el
@@ -1973,7 +1973,8 @@ current section."
   (and section
        (>= pos (oref section start))
        (<  pos (or (oref section content)
-                   (oref section end)))
+                   (oref section end)
+                   pos))
        t))
 
 (defun magit-section-internal-region-p (&optional section)


tarsius:
I just pushed 410e4583e01b7fac31a4748e78f4c613859d3c41 instead. Initially I used `with-demoted-errors` but decided against that.

PR #4891, opened 2023-03-04
magit-autorevert.el: don't enable if global-auto-revert-mode is on

Hi-Angel:
Judging by comments inside the `(define-globalized-minor-mode magit-auto-revert-mode …`, it wasn't supposed to be enabled when global-auto-revert-mode is on. Unfortunately that doesn't quite work out, the mode is enabled disregarding the state of global-auto-revert-mode. Let's work around that by adding another check for whether the mode is enabled.

In case anybody wants a testcase for magit **prior** to this commit failing to **not** enable the mode: given this `.emacs` file:

```lisp
(use-package autorevert
  :config
  (global-auto-revert-mode 1))

(use-package magit)
```

Upon starting Emacs and opening `*Messages*` buffer you'll see a `Turning on magit-auto-revert-mode...done` line.

### DIFF
--- a/lisp/magit-autorevert.el
+++ b/lisp/magit-autorevert.el
@@ -141,7 +141,8 @@ seconds of user inactivity.  That is not desirable."
 Do not use this function elsewhere, and don't remove it from
 the `after-init-hook'.  For more information see the comments
 and code surrounding the definition of this function."
-  (if magit-auto-revert-mode
+  (if (and magit-auto-revert-mode
+           (not global-auto-revert-mode))
       (let ((start (current-time)))
         (magit-message "Turning on magit-auto-revert-mode...")
         (magit-auto-revert-mode 1)


Hi-Angel:
As a side-note, before making that commit I spent more multiple hours trying to figure out how initial values of mode variables are being chosen, with or without the `:init-value` key-arg being set.

From what I understand, the `(define-globalized-minor-mode magit-auto-revert-mode` seems to expect that the initial value of `magit-auto-revert-mode` variable *(the one with the name of the global mode)* will be set to nil if there's `global-auto-revert-mode` enabled.

Not only that doesn't like that, but after spending multiple hours I have no idea how in the world those values should work at all. I tried creating test global modes based on other minor ones, and haven't found a pattern in whether the initial value it `t` or `nil`. At this point I think it's much easier to just add another check like I do in the commit.

tarsius:
If pushed a fix. See the commit message for my understanding of the issue.

PR #4895, opened 2023-03-04
magit-save-repository-buffers: guard against nil default-directory

progfolio:
Ensure buffer's repository directory exists before binding default-directory. Otherwise, subsequent subprocess calls will fail due to being called with a nil default-directory.

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -1194,9 +1194,13 @@ argument (the prefix) non-nil means save all with no questions."
                   ;; places us in another repository, then the below
                   ;; let-binding is needed to prevent that file from
                   ;; being saved.
-                  (let ((default-directory
-                         (file-name-directory buffer-file-name)))
+                  (let* ((dir (file-name-directory buffer-file-name))
+                         (default-directory dir))
                     (and
+                     ;; - Ensure repository still on disk.
+                     ;;   Otherwise, subprocesses called from repo dir will fail.
+                     dir
+                     (file-exists-p dir)
                      ;; - Check whether refreshing is disabled.
                      (not magit-inhibit-refresh-save)
                      ;; - Check whether the visited file is either on the


progfolio:
You can reproduce the issue this aims to fix by:

1. opening a buffer in a repository
2. delete the repository directory (the buffer is live, but it's associated directory is gone now)
3. invoke magit-status in any repository

progfolio:
This actually only partially solves the problem.
Subsequent attempts to commit (in any repository) fail with:

```
  1 git … commit --
hint: Waiting for your editor to close the file...
Waiting for Emacs...
*ERROR*: Cannot open load file: No such file or directory, darkman
error: There was a problem with the editor '/usr/bin/emacsclient --socket-name=/run/user/1000/emacs/server'.
Please supply the message using either -m or -F option.
```

In this example, darkman's repository is no longer on disk.

tarsius:
The way you are using `and-let*` here, is that actually documented or might it just fall out of the implementation that you can do that? Even if its the former, I am not sure I am comfortable doing that.

> This actually only partially solves the problem.

But you think you can solve it completely?

progfolio:
Jonas Bernoulli ***@***.***> writes:

> The way you are using and-let* here, is that actually documented 
> or might it just fall out
> of the implementation that you can do that? Even if its the 
> former, I am not sure I am
> comfortable doing that.

You'll have to be specific about which part of the usage you're 
unsure of.
 
>  This actually only partially solves the problem.
>
> But you think you can solve it completely?

I'll look into it more.

progfolio:
> This actually only partially solves the problem.
Subsequent attempts to commit (in any repository) fail with:

I haven't been able to reproduce this after restarting Emacs with the patch applied.
Things may have just gotten into a funky state when I was working on the patch after I initially encountered the first issue.

tarsius:
> You'll have to be specific about which part of the usage you're unsure of.

I am not sure we should rely on `(and-let* ((buffer-file-name)) ...)` being equivalent to `(and-let* ((x buffer-file-name)) ...)`. I don't think the documentation says that this is something we should and that it just falls out of the implementation.

progfolio:
Jonas Bernoulli ***@***.***> writes:

> I am not sure we should rely on (and-let* ((buffer-file-name)) 
> ...) being equivalent to
> (and-let* ((x buffer-file-name)) ...). I don't think the 
> documentation says that this is
> something we should and that it just falls out of the 
> implementation.

and-let* is implemented in terms of if-let.
From the docstring of if-let:

> Each element of SPEC is a list (SYMBOL VALUEFORM) that binds
> SYMBOL to the value of VALUEFORM.  An element can additionally 
> be
> of the form (VALUEFORM), which is evaluated and checked for nil;
> i.e. SYMBOL can be omitted if only the test result is of
> interest.  It can also be of the form SYMBOL, then the binding 
> of
> SYMBOL is checked for nil.

Does ((buffer-file-name)) not fit the "SYMBOL can be omitted" 
pattern described?

tarsius:
Ah, indeed. Well now that I know about it, I will have to think about how I feel about it. One thing I know is that it is confusing for people that do not know about it yet and as far as I remember, I have never seen it done before. So I currently lean towards "this should be avoided", but that might very well change.

But for this change, please go back to using both `and` and `and-let*`, at the cost of deeper nesting.

progfolio:
Jonas Bernoulli ***@***.***> writes:

> Ah, indeed. Well now that I know about it, I will have to think 
> about how I feel about it.
> One thing I know is that it is confusing for people that do not 
> know about it yet and as far
> as I remember, I have never seen it done before. So I currently 
> lean towards "this should
> be avoided", but that might very well change.

I would reconsider. It's a generally useful idiom which is 
applicable in when-let and if-let as well.
 
> But for this change, please go back to using both and and 
> and-let*, at the cost of deeper
> nesting.

I've reworked  the patch to avoid both the binding idiom and 
increased nesting.

tarsius:
> I would reconsider. It's a generally useful idiom which is applicable in when-let and if-let as well.

As I said, I will let some time pass and do some thinking before I will decide on whether to use it in my code.

> magit-save-repository-buffers: guard against nil default-directory

Actually it checks if the specified directory exists. If that actually is nil, then that results in an error from `file-exists-p`. Since `buffer-file-name` being non-nil doesn't guarantee that `(file-name-directory buffer-file-name)` is non-nil, we should make sure that it is.

progfolio:
Jonas Bernoulli ***@***.***> writes:

> Since buffer-file-name being non-nil doesn't guarantee that
> (file-name-directory buffer-file-name) is non-nil, we should 
> make sure that it is.

Good point. I've revised to ensure the target directory is non-nil 
prior to checking for its existence.

tarsius:
I've amended the commit to do it a bit differently (no extra variable) and adjusted the comment and commit message.

PR #4899, opened 2023-03-08
magit-rebase-edit-command -> magit-rebase-edit-commit

arialdomartini:
I was reading about `magit-diff-edit-hunk-commit` when I saw a comment about `magit-rebase-edit-command`, mentioned there once. I've struggled finding further information, until I realized there is command with a very similar name, `magit-rebase-edit-commit`. I might be wrong, but I assume this should be the right name to mention.

I updated the 2 files where this command is mentioned, keeping `magit.texi` unchanged.


tarsius:
Thanks!

PR #4908, opened 2023-03-23
Documentation: correct magit-toggle-margin key binding

DivineDominion:
The margin toggle command was documented as `L t` but the transient menu shows `L L`: https://github.com/magit/magit/blob/main/lisp/magit-log.el#L520


tarsius:
Thanks!

PR #4912, opened 2023-04-01
magit-base: consolidate ellipsis customisation

nbarrientos:
This patch centralises the generation of ellipses by extending the functionality provided by `magit-ellipsis`, allowing end users to customise for instance the ellipsis used in `magit-log` buffers.

See [discussions/4907](https://github.com/magit/magit/discussions/4907) for more background and extra information.

### DIFF
--- a/lisp/magit-base.el
+++ b/lisp/magit-base.el
@@ -384,21 +384,30 @@ Messages which can currently be suppressed using this option are:
   :group 'magit-miscellaneous
   :type '(repeat string))
 
-(defcustom magit-ellipsis (if (char-displayable-p ?…) "…" "...")
-  "String used to abbreviate text in process buffers.
+(defcustom magit-ellipsis
+  '((margin (?… . ">"))
+    (t      (?… . "...")))
+  "Characters or strings used to abbreviate text in some buffers.
 
-Currently this is only used to elide `magit-git-global-arguments'
-in process buffers.  In the future it may be used in other places
-as well, but not the following:
+Each element has the form (WHERE (FANCY . UNIVERSAL)).
 
-- Author names in the log margin are always abbreviated using
-  \"…\" or if that is not displayable, then \">\".
+FANCY is a single character or nil whereas UNIVERSAL is a string
+of any length.  The ellipsis produced by `magit--ellipsis' will
+be FANCY if it's a non-nil character that can be displayed with
+the available fonts, otherwise UNIVERSAL will be used.  FANCY is
+meant to be a rich character like a horizontal ellipsis symbol or
+an emoji whereas UNIVERSAL something simpler available in a less
+rich environment like the CLI.  WHERE determines the use-case for
+the ellipsis definition.  Currently the only acceptable values
+for WHERE are `margin' or t (representing the default).
 
-- Whether collapsed sections are indicated using ellipsis is
-  controlled by `magit-section-visibility-indicator'."
-  :package-version '(magit . "3.0.0")
+Whether collapsed sections are indicated using ellipsis is
+controlled by `magit-section-visibility-indicator'."
+  :package-version '(magit . "4.0.0")
   :group 'magit-miscellaneous
-  :type 'string)
+  :type '(repeat (list (symbol :tag "Where")
+                       (cons (choice :tag "Fancy" character (const nil))
+                             (string :tag "Universal")))))
 
 (defcustom magit-update-other-window-delay 0.2
   "Delay before automatically updating the other window.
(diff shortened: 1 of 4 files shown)


test/magit-tests.el
```diff
@@ -446,6 +446,46 @@ Enter passphrase for key '/home/user/.ssh/id_rsa': "
     (should (equal (get-text-property 0 'font-lock-face str) '(bold highlight)))
     (should (equal (get-text-property 2 'font-lock-face str) '(bold)))))
 
+(ert-deftest magit-base:ellipsis-default-values ()
+  (cl-letf (((symbol-function 'char-displayable-p)
+             (lambda (char) t)))
```
tarsius, line 451: Thanks for adding tests!

Use `_` instead of `char` to keep the byte-compiler happy. Also put the lambda on the same line as the symbol-function.

lisp/magit-base.el
```diff
@@ -384,21 +384,32 @@ Messages which can currently be suppressed using this option are:
...
-as well, but not the following:
-
-- Author names in the log margin are always abbreviated using
-  \"…\" or if that is not displayable, then \">\".
-
-- Whether collapsed sections are indicated using ellipsis is
-  controlled by `magit-section-visibility-indicator'."
-  :package-version '(magit . "3.0.0")
+(defcustom magit-ellipsis
+  '((magit-log (?… . ">"))
```
tarsius, line 388: Lets go with `margin` instead of `magit-log` for now. (There is no need for a package prefix, and what makes it necessary to use a one-character ellipsis is the fact that we display information in the narrow `margin`. In the future we might display other things in some margin and need to truncate that information, and that might have nothing to do with `log`.

lisp/magit-base.el
```diff
@@ -384,21 +384,32 @@ Messages which can currently be suppressed using this option are:
...
+    (nil       (?… . "...")))
```
tarsius, line 389: The key for the default should be `t` instead of `nil`. Then use `(or where t)` below.

lisp/magit-base.el
```diff
@@ -384,21 +384,32 @@ Messages which can currently be suppressed using this option are:
...
+  "Characters or strings used to abbreviate text in some buffers.
+
+Each element has the form:
+
+\(WHERE (FANCY . UNIVERSAL))
```
tarsius, line 394: Since it fits neatly, I would compress these three lines into just one.

lisp/magit-base.el
```diff
@@ -1215,6 +1226,18 @@ Like `message', except that `message-log-max' is bound to nil."
          (goto-char (or ,pos 1))
          ,@body))))
 
+(defun magit--ellipsis (&optional where)
+  "Build an ellipsis always as string, depending on WHERE."
+  (if-let ((default (alist-get nil magit-ellipsis)))
+      (pcase-let ((`(,fancy . ,universal) (car (alist-get where magit-ellipsis default))))
```
tarsius, line 1232: The "default" key is the most used key, so I find it a bit odd to always look it up twice. I also think that error message below is worded as it is because of how you implemented this here. Simply saying that `magit-ellipsis` is invalid, seems more appropriate.

Speaking of invalid values, I think this should be backward compatible. There's no need to inform users that the format has been changed, and the old format should just be silently accepted.

tarsius (changes requested):

lisp/magit-base.el
```diff
@@ -384,21 +384,32 @@ Messages which can currently be suppressed using this option are:
...
+for WHERE are `magit-log' or nil (representing the default).
+
+Whether collapsed sections are indicated using ellipsis is
+controlled by `magit-section-visibility-indicator'."
+  :package-version '(magit . "4.0.0")
   :group 'magit-miscellaneous
-  :type 'string)
+  :type '(repeat (list symbol
+                       (cons (choice character (const nil))
+                             (choice string)))))
```
tarsius, line 412: Please use `:tag "where"/"fancy"/"universal"` here, or similar.

tarsius (changes requested):

nbarrientos:
Thanks for your comments! I've pushed separate commits fixing things so it's easier to re-review. Happy to change more stuff, just let me know. Otherwise, just ping me to squash :slightly_smiling_face: . Thanks.

tarsius:
Thanks!

> I haven't resolved the conversations myself and I don't know how you normally work in this regard.

I don't really use that much myself either.

tarsius (approved):

nbarrientos:
Thanks for merging! A :beer: is waiting for you in Genève :smile:

PR #4914, opened 2023-04-02
magit-branch-delete: Improve "also on remote?" prompt with >1 ref

phil-s:
Follow-up for #4905.

Implements the change described at https://github.com/magit/magit/discussions/4905#discussioncomment-5379339.

### DIFF
--- a/lisp/magit-apply.el
+++ b/lisp/magit-apply.el
@@ -583,7 +583,7 @@ of a side, then keep that side without prompting."
 (defun magit-discard-files--resolve (files)
   (if-let ((arg (and (cdr files)
                      (magit-read-char-case
-                         (format "For these %i files\n%s\ncheckout:\n"
+                         (format "For these %d files\n%s\ncheckout:\n"
                                  (length files)
                                  (mapconcat (lambda (file)
                                               (concat "  " file))
@@ -630,7 +630,7 @@ of a side, then keep that side without prompting."
              (magit-call-git "rm" "--cached" "--force" "--" file))))))
 
 (defun magit-discard-files--rename (files status)
-  (magit-confirm 'rename "Undo rename %s" "Undo %i renames" nil
+  (magit-confirm 'rename "Undo rename %s" "Undo %d renames" nil
     (mapcar (lambda (file)
               (setq file (assoc file status))
               (format "%s -> %s" (cadr file) (car file)))
--- a/lisp/magit-base.el
+++ b/lisp/magit-base.el
@@ -791,7 +791,7 @@ ACTION is a member of option `magit-slow-confirm'."
       (setq prompt (magit-confirm-make-prompt action)))
     (magit-confirm action
       (concat prompt " %s")
-      (concat prompt " %i files")
+      (concat prompt " %d files")
       nil files)))
 
 (defun magit-confirm-make-prompt (action)
--- a/lisp/magit-blame.el
+++ b/lisp/magit-blame.el
@@ -259,7 +259,7 @@ Also see option `magit-blame-styles'."
                  (file (and (not (derived-mode-p 'dired-mode))
                             (magit-file-relative-name
                              nil (not magit-buffer-file-name))))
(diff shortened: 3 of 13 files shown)


phil-s:
n.b. I've gone with `%i` in the format string for consistency with the old code and with other usage I could see in this file, but... are you aware that `%i` is completely undocumented as a `format` construct for Emacs?  I've just been checking that, and neither the docstring nor the elisp manual mentions it.  Obviously it works (and I can see that the C code treats `%i` identically to `%d`; evidentially for `printf` consistency<sup>1</sup>), but I'd suggest that every formatted `%i` in the codebase should be changed to `%d` to avoid confusion.

<sup>1</sup> `styled_format()` in `src/editfns.c` uses `conversion == 'd' || conversion == 'i'` in each case.

tarsius:
> are you aware that `%i` is completely undocumented as a `format` construct for Emacs?

Apparently not :grinning:

> I'd suggest that every formatted `%i` in the codebase should be changed to `%d` to avoid confusion.

Agreed. Could you please do that? (Just add a second commit to this pr.)

phil-s:
I'm wondering whether `magit-confirm-make-prompt` should escape `%` chars? The result always gets passed to `magit-confirm`, which uses `format`, and my impression is that the ACTION passed to `magit-confirm-make-prompt` is not intended to contain format constructs, in which case escaping any which were present would seem correct.

phil-s:
My other question from this is what's going on with the customize `:type` for `magit-patch-save-arguments`, which includes `:format "%v%i\n"`

This is the only instance of `%i` in the codebase which doesn't wind up being processed by `format`, but the `:format` type keyword is something entirely different to the `format` function, and once again `%i` doesn't appear to be a thing:  https://www.gnu.org/software/emacs/manual/html_node/elisp/Type-Keywords.html

The nearest thing I can find is for `:entry-format` in https://www.gnu.org/software/emacs/manual/html_node/widget/editable_002dlist.html , but I don't think that's applicable.

```el
(defcustom magit-patch-save-arguments '(exclude "--stat")
  "Control arguments used by the command `magit-patch-save'.

`magit-patch-save' (which see) saves a diff for the changes
shown in the current buffer in a patch file.  It may use the
same arguments as used in the buffer or a subset thereof, or
a constant list of arguments, depending on this option and
the prefix argument."
  :package-version '(magit . "2.12.0")
  :group 'magit-diff
  :type '(choice (const :tag "use buffer arguments" buffer)
                 (cons :tag "use buffer arguments except"
                       (const :format "" exclude)
                       (repeat :format "%v%i\n"
                               (string :tag "Argument")))
                 (repeat :tag "use constant arguments"
                         (string :tag "Argument"))))
```

tarsius:
> I'm wondering whether `magit-confirm-make-prompt` should escape `%` chars?

In case the symbol-name of the ACTION contains a `%`? I don't see a reason why one would use such a symbol. If someone does that, then that results in an obvious enough error; and if they have a good enough reason to want to use a symbol containing a `%`, then we can start quoting it then.

phil-s:
> > once again `%i` doesn't appear to be a thing: https://www.gnu.org/software/emacs/manual/html_node/elisp/Type-Keywords.html
> 
> That page documents it as:
> ```
> ‘%i’ Insert the [INS] button.
> ```

It does?  That specific page?  I do not see it either in the above URL or in the info manual.

I do see it for the `:entry-format` docs, but I think that's a different thing.

tarsius:
> It does?  That specific page?  I do not see it either in the above URL or in the info manual.

Ah, no it does, I meant this one:

> I do see it for the `:entry-format` docs, but I think that's a different thing.

It is basically the same thing. `:entry-format` is the format used for an entry (aka child) of this widget. `:format` is for the widget itself. The documentation is not complete.

phil-s:
Ok, an upstream documentation bug then.  I should chase that up.

That covers off both of my questions, so nothing more from me here unless you can see any issues in the changes.

PR #4919, opened 2023-04-16
Switch Urls from Http to Https

shoefone:
Changed all remaining http:// Urls in magit.org documentation to https:// .

Each Url that was changed was checked before changing to ensure an https:// Url actually exists.

The motivation for this was the http:// Url used for the package source (http://melpa.org/packages); of all the links, that is probably the most important one in terms of security.


tarsius:
Thanks!

shoefone:
Np, thank you!

PR #4925, opened 2023-04-20
Introduce customization to `magit-insert-worktrees'

vermiculus:
A common workflow when using worktrees is to have one worktree per feature-/fix-branch. In this workflow, it can be more useful to provide details on the ticket being worked rather than the location of the worktree (which might even be generated on the ticket ID).

Introduce `magit-worktree-section-details-function` to allow customizing the details inserted into worktree sections -- extracting the default behavior into `magit-insert-worktree-path`.

---

I'm not 100% sure about the changes to the documentation. I've noticed that there's no mention of `magit-insert-worktrees` in the current documentation at all -- so depending on where that should go, the new user option here might go somewhere else. Whatever you'd like to do is obviously fine by me 😃

### DIFF
--- a/lisp/magit-worktree.el
+++ b/lisp/magit-worktree.el
@@ -39,15 +39,15 @@ Used by `magit-worktree-checkout' and `magit-worktree-branch'."
   :group 'magit-commands
   :type 'function)
 
-(defcustom magit-worktree-section-details-function
-  #'magit-insert-worktree-path
-  "A function used to insert details about a given worktree.
-This function takes two arguments: the current HEAD of the
-worktree and the path to the worktree itself. It is respected by
-`magit-insert-worktrees'."
-  :package-version '(magit . "3.3.0")
-  :type 'function
-  :group 'magit-commands)
+(defcustom magit-worktree-section-summary-function
+  #'magit-worktree-format-path
+  "Function used to format information about a given worktree.
+`magit-insert-worktrees' calls this function with two arguments,
+the current HEAD of the worktree and the path to the worktree
+itself."
+  :package-version '(magit . "4.0.0")
+  :group 'magit-commands
+  :type '(choice (function-item magit-worktree-format-path) function))
 
 ;;; Commands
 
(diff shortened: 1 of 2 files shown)


lisp/magit-worktree.el
```diff
@@ -190,23 +190,19 @@ If there is only one worktree, then insert nothing."
             (magit-insert-section (worktree path)
               (insert head)
               (insert (make-string (- align (length head)) ?\s))
-              (funcall magit-worktree-section-details-function
-                       (substring-no-properties head)
-                       path)
+              (insert (funcall magit-worktree-section-summary-function
+                               (substring-no-properties head)
+                               path))
```
vermiculus: I considered such a signature, but is it not feasible that the function might want to include newlines or even sections as tab-targets?

I have no immediate need for such a thing, but I'm not sure about the trade-off here.

vermiculus:
Added a comment to your fixup, but I don't feel too strongly about it. LGTM 😃

Thanks for taking a look!

tarsius:
Displaying more information about a worktree is also something I have thought about before. I don't remember the details, but obviously the additional information would not particularly useful, and delayed doing so until someone asked me to do so. If I added more information, then I would likely not add an alternative function but extend the current function to display that information for everyone. (But I assume in your case this has something to do with the subversion migration, so the information probably wouldn't be useful for most other users.)

I like to avoid adding new options when it can be avoided, so I think would be better to simply move the code that inserts an individual worktree to a separate function, making it trivial to override it using an advice. I've pushed that to the `worktree-info` branch as well.

vermiculus:
> But I assume in your case this has something to do with the subversion migration, so the information probably wouldn't be useful for most other users

It's definitely related :-) but my use-case is generalizable – one could imagine creating a similar override for branches that match a naming scheme based on github issues, e.g. `fix/1234`, then use `ghub` or similar to pull the issue title down.

I'm tempted to close this given the direction #4926 is headed, but I'd like to see a firm direction on that PR before doing so. I expect we'll be going forward with your advice-able approach over the new `defcustom`, but just to avoid PR spam…

Issue #3067, opened 2017-05-07
Why is git-commit-fill-column considered obsolete?


angrybacon:
Here is the commit for reference: https://github.com/magit/magit/commit/a277432e806af89c3bdd6046721c4db948443120

While `git-commit-turn-on-auto-fill` still uses it.

kyleam:
`git-commit-turn-on-auto-fill` still uses it to accomadate users that
still set it.

I think #2848 contains most of the background on this.

npostavs:
Yep, short answer is, it's obsolete because it doesn't provide any functionality over plain old `fill-column`.

angrybacon:
Thanks, I'm advicing `git-commit-turn-on-auto-fill` then. Unless you want to keep the issue around, you can close it. 👌

PR #4933, opened 2023-05-01
manual: Remove references to git-commit-fill-column

Binary-Eater:
git-commit-fill-column was deprecated in Magit 2.9.0. It should be removed from the manual to avoid confusion since the variable is no longer present in the codebase.

Link: https://github.com/magit/magit/commit/a277432e806af89c3bdd6046721c4db948443120
Link: https://github.com/magit/magit/issues/3067


tarsius:
Thanks!

PR #4934, opened 2023-05-02
Stage untracked files with status.showUntrackedFiles=no

vermiculus:
For some worktrees, searching for untracked files can take several seconds. In these cases, Git recommends setting config option `status.showUntrackedFiles=no` to forgo their listing in git-status. Magit currently respects this by skipping the 'Untracked Files' section altogether.

This patch teaches the 'Untracked Files' section to stage multiple untracked files interactively -- listing them only once and on-demand.

---

Based on the conversation at #4932.

I'm not too pleased with the special-casing I did in `magit-stage`, but it seemed far safer than modifying `magit-apply--get-selection` and has a better UX than inserting a 'fake' child section in the `"no"` case of `magit-insert-untracked-files`.

---

After re-reading my patch and noticing I can use `if-let` at `magit-stage-untracked`, I remembered how `--if-let` differs from `if-let` and now I'm *really* unhappy with what I did in `magit-stage` – what I did seems exceptionally fragile. I pushed a fixup to address this issue, but it's still a bit of a kludge. Any better ideas?


lisp/magit-status.el
```diff
@@ -722,25 +722,34 @@ value of that variable can be set using \"D -- DIRECTORY RET g\"."
...
-                (insert (propertize file 'font-lock-face 'magit-filename) ?\n)))
-            (insert ?\n)))))))
+    (pcase show
+      ("no"
+       (magit-insert-section (untracked 'unlisted)
+         (magit-insert-heading "Untracked files")
+         (insert (substitute-command-keys
+                  (concat
+                   "Untracked files are currently hidden (\"status.showUntrackedFiles=no\").\n"
+                   "Press `\\<magit-mode-map>\\[magit-stage-file]' to search for and stage them.\n")))
```
vermiculus, line 732: It's worth noting here particularly that using the correct map (`magit-untracked-section-map`) and function (`magit-stage`) yielded the improper key `<1>`. This is almost certainly due to the multiple bindings of `magit-stage` in that keymap, but I'm not aware of how to influence the output of `substitute-command-keys` to get the desired result.

tarsius:
I am quite busy right now, it might take me until the weekend until I get back to this.

tarsius:
Closing because of the reasons given at https://github.com/magit/magit/discussions/4932#discussioncomment-5922033, and the changes since implemented.

PR #4939, opened 2023-05-08
manual: update chapter Repository List

mbeppler:
Played with the _magit-list-repositories_ and encountered two differences between implementation and the manual.

Updated the manual regarding _magit-repository-directories-depth_ and _magit-repolist-column-flags_ to the best of my knowledge.


tarsius:
Thanks!

PR #4941, opened 2023-05-10
manual: compat is also a required library

aepyornis:
While installing magit manually I noticed that the manual left out that the library compat is also required. I also corrected the with-editor path.


tarsius:
Thanks!

PR #4944, opened 2023-05-16
Add user option 'git-commit-usage-hint'

protesilaos:
The intent is to make it easier for users to visualise the common key bindings of git-commit.el.  New users can benefit from the option to place the hint in the header-line, as they tend to miss what is shown in the echo area.

The idea for this change was discussed here:
<https://github.com/magit/magit/discussions/4942>.

### DIFF
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -309,6 +309,16 @@ no effect."
   :safe 'booleanp
   :type 'boolean)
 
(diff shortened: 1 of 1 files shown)


protesilaos:
Jonas is doing it in a different way: https://github.com/magit/magit/discussions/4942#discussioncomment-5920563

We can safely close this.

PR #4954, opened 2023-06-04
magit-clone-internal: correctly get origin value from args

aepyornis:
Fixes a bug in `magit-clone-internal` where the function would mistakenly ignore the `--origin` value from args


tarsius:
Thanks!

PR #4963, opened 2023-07-13
Correct "appropriat" and "unspecifed" misspellings

rdparker:
The words "appropriate" and "unspecified" were misspelled in the docstrings for `magit-repolist-columns` and `magit-submodule-list-columns`, as well as their entries in the manual.

I've also aliased my ancient email account in `.mailmap`, although I wasn't sure what the protocol was for updating that.  It only appeared in one pre-1.2.0 commit.


tarsius:
Thanks!

PR #4965, opened 2023-07-18
magit-unstage-file: Add missing colon to prompt

bcc32:
Prompts have `: ` at the end by convention; when missing, there is no space between the prompt and the minibuffer input, which is ugly.


tarsius:
Thanks!

PR #4980, opened 2023-08-10
Address XR/Relint regexp/syntax linter warnings

basil-conto:
Cc: @kyleam

1. `M-x emacs-version RET`

   ```
   GNU Emacs 30.0.50 (build 1, x86_64-pc-linux-gnu, X toolkit, cairo version 1.16.0, Xaw3d scroll bars) of 2023-08-10
   ```

2. Install GNU ELPA package [`relint`](https://elpa.gnu.org/packages/relint.html)
3. `M-x relint-directory RET path/to/magit RET`

   ```
   Relint results for /home/blc/.emacs.d/src/magit/
   lisp/magit-extras.el:733:39: In call to skip-syntax-backward: Invalid char `s' in syntax string (pos 1)
     ">s-"
      .^
   lisp/magit-extras.el:783:41: In call to replace-regexp-in-string: Duplicated `\' inside character alternative (pos 5)
     "^[ \\+\\-]"
      ......^
   lisp/magit-log.el:1269:23: In magit-log-reflog-re: Optional expression matching an empty string (pos 127)
     "^\\(?1:[^\000\n]+\\)\000\\(?5:[^\000\n]*\\)\000\\(?:\\(?:[^@\n]+@{\\(?6:[^}\n]+\\)}\000\\(?10:merge \\|autosave \\|restart \\|rewritten \\|[^:\n]+: \\)?\\(?2:.*\\)?\\)\\|\000\\)$"
      ...................................................................................................................................................................^
   lisp/magit-log.el:1463:34: In call to string-match: Suspect `[' in char alternative (pos 4)
     "\\[[^[]*?]"
      .....^
   lisp/magit-notes.el:66:36: In call to directory-files: Use \` instead of ^ in file-matching regexp (pos 0)
     "^[^.]"
      ^
   lisp/magit-process.el:155:14: Ineffective string escape `\['
   lisp/magit-process.el:161:13: Ineffective string escape `\]'
   lisp/magit-remote.el:146:58: Ineffective string escape `\.'
   lisp/magit-remote.el:146:65: Ineffective string escape `\.'
   lisp/magit-sequence.el:1037:23: In call to directory-files: Use \` instead of ^ in file-matching regexp (pos 0)
     "^[0-9]\\{4\\}$"
      ^
   lisp/magit-sequence.el:1037:36: In call to directory-files: Use \' instead of $ in file-matching regexp (pos 11)
     "^[0-9]\\{4\\}$"
      .............^
   lisp/magit-tag.el:130:59: In magit-release-tag-regexp: Optional expression matching an empty string (pos 51)
     "\\`\\(?1:\\(?:v\\(?:ersion\\)?\\|r\\(?:elease\\)?\\)?[-_]?\\)?\\(?2:[0-9]+\\(?:\\.[0-9]+\\)*\\(?:-[a-zA-Z0-9-]+\\(?:\\.[a-zA-Z0-9-]+\\)*\\)?\\)\\'"
      .............................................................^
   lisp/magit.el:559:23: Ineffective string escape `\('

   Finished -- 13 errors.
   ```

This PR tries to address all warnings (details in commit message), except for the `` Suspect `[' in char alternative `` which looks like a false positive: mattiase/xr#7

WDYT?
Thanks.

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -1266,7 +1266,7 @@ Do not add this to a hook variable."
           "\\(?:\\(?:[^@\n]+@{\\(?6:[^}\n]+\\)}\0" ; date
                                                  ;;; refsub
           "\\(?10:merge \\|autosave \\|restart \\|rewritten \\|[^:\n]+: \\)?"
-          "\\(?2:.*\\)?\\)\\|\0\\)$"))             ; msg
+          "\\(?2:.*\\)\\)\\|\0\\)$"))              ; msg
 
 (defconst magit-reflog-subject-re
   (concat "\\(?1:[^ ]+\\) ?"                       ; command
@@ -1388,8 +1388,7 @@ Do not add this to a hook variable."
             (insert (magit-reflog-format-subject
                      (substring refsub 0
                                 (if (string-search ":" refsub) -2 -1))))))
-        (when msg
-          (insert (funcall magit-log-format-message-function hash msg)))
+        (insert (funcall magit-log-format-message-function hash msg))
         (when (and refs magit-log-show-refname-after-summary)
           (insert ?\s)
           (insert (magit-format-ref-labels refs)))
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -1460,7 +1460,7 @@ Do not add this to a hook variable."
       (magit--put-face (match-beginning 0) (1- boundary)
                        'magit-keyword-squash msg))
     (when magit-log-highlight-keywords
-      (while (string-match "\\[[^[]*?]" msg boundary)
+      (while (string-match "\\[[^][]*]" msg boundary)
         (setq boundary (match-end 0))
         (magit--put-face (match-beginning 0) boundary
                          'magit-keyword msg))))
--- a/lisp/magit-remote.el
+++ b/lisp/magit-remote.el
@@ -143,7 +143,7 @@ has to be used to view and change remote related variables."
     (when (equal (magit-get "remote.pushDefault") remote)
       (magit-set new-name "remote.pushDefault"))
     (dolist (var (magit-git-lines "config" "--name-only"
-                                  "--get-regexp" "^branch\.[^.]*\.pushRemote"
+                                  "--get-regexp" "^branch\\.[^.]*\\.pushRemote"
                                   (format "^%s$" remote)))
       (magit-call-git "config" (and (not new-name) "--unset") var new-name))))
 
--- a/lisp/magit-tag.el
+++ b/lisp/magit-tag.el
@@ -127,7 +127,7 @@ defaulting to the tag at point.
 See also `magit-release-tag-regexp'.")
 
 (defvar magit-release-tag-regexp "\\`\
-\\(?1:\\(?:v\\(?:ersion\\)?\\|r\\(?:elease\\)?\\)?[-_]?\\)?\
+\\(?1:\\(?:v\\(?:ersion\\)?\\|r\\(?:elease\\)?\\)[-_]?\\)?\
 \\(?2:[0-9]+\\(?:\\.[0-9]+\\)*\
 \\(?:-[a-zA-Z0-9-]+\\(?:\\.[a-zA-Z0-9-]+\\)*\\)?\\)\\'"
   "Regexp used by `magit-tag-release' to parse release tags.
--- a/lisp/magit-extras.el
+++ b/lisp/magit-extras.el
@@ -730,7 +730,7 @@ the minibuffer too."
                     (string-replace "%N" idx eob-format)))
             (save-excursion
               (goto-char (point-max))
-              (skip-syntax-backward ">s-")
+              (skip-syntax-backward ">-")
               (beginning-of-line)
               (if (and comment-start (looking-at comment-start))
                   (while (looking-at comment-start)
--- a/lisp/magit-extras.el
+++ b/lisp/magit-extras.el
@@ -780,7 +780,7 @@ argument."
        (replace-regexp-in-string
         (format "^\\%c.*\n?" (if (< (prefix-numeric-value arg) 0) ?+ ?-))
         "")
-       (replace-regexp-in-string "^[ \\+\\-]" "")))
+       (replace-regexp-in-string "^[ +-]" "")))
     (deactivate-mark))
    ((use-region-p)
     (call-interactively #'copy-region-as-kill))
--- a/lisp/magit-notes.el
+++ b/lisp/magit-notes.el
@@ -63,7 +63,7 @@
 (defun magit-notes-merging-p ()
   (let ((dir (expand-file-name "NOTES_MERGE_WORKTREE" (magit-gitdir))))
     (and (file-directory-p dir)
-         (directory-files dir nil "^[^.]"))))
+         (directory-files dir nil "\\`[^.]"))))
 
 (transient-define-infix magit-core.notesRef ()
   :class 'magit--git-variable
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -152,13 +152,13 @@ itself from the hook, to avoid further futile attempts."
                  (const :tag "Don't start a cache daemon" nil)))
 
 (defcustom magit-process-yes-or-no-prompt-regexp
-  (concat " [\[(]"
+  (concat " [([]"
           "\\([Yy]\\(?:es\\)?\\)"
           "[/|]"
           "\\([Nn]o?\\)"
           ;; OpenSSH v8 prints this.  See #3969.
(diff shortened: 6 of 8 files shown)


lisp/magit-log.el
```diff
@@ -1388,8 +1388,7 @@ Do not add this to a hook variable."
             (insert (magit-reflog-format-subject
                      (substring refsub 0
                                 (if (string-search ":" refsub) -2 -1))))))
-        (when msg
-          (insert (funcall magit-log-format-message-function hash msg)))
+        (insert (funcall magit-log-format-message-function hash msg))
```
kyleam, line 1391: Right, based on the regexps, `msg` is always an empty string, and we can be even more confident in that because the upstream code unconditionally calls `(substring-no-properties msg)`, which would signal an error if `msg` were `nil`.  Okay, though I have a slight preference for this to come in as a separate commit.

lisp/magit-remote.el
```diff
@@ -143,7 +143,7 @@ has to be used to view and change remote related variables."
     (when (equal (magit-get "remote.pushDefault") remote)
       (magit-set new-name "remote.pushDefault"))
     (dolist (var (magit-git-lines "config" "--name-only"
-                                  "--get-regexp" "^branch\.[^.]*\.pushRemote"
+                                  "--get-regexp" "^branch\\.[^.]*\\.pushRemote"
```
kyleam, line 146: Good catch.  The single slash form leads to those dots having unintended wildcard behavior.

lisp/magit-tag.el
```diff
@@ -127,7 +127,7 @@ defaulting to the tag at point.
 See also `magit-release-tag-regexp'.")
 
 (defvar magit-release-tag-regexp "\\`\
-\\(?1:\\(?:v\\(?:ersion\\)?\\|r\\(?:elease\\)?\\)?[-_]?\\)?\
+\\(?1:\\(?:v\\(?:ersion\\)?\\|r\\(?:elease\\)?\\)[-_]?\\)?\
```
kyleam, line 130: This leads to a behavior change:

```elisp
(let* ((pre "\\`\
\\(?1:\\(?:v\\(?:ersion\\)?\\|r\\(?:elease\\)?\\)?[-_]?\\)?\
\\(?2:[0-9]+\\(?:\\.[0-9]+\\)*\
\\(?:-[a-zA-Z0-9-]+\\(?:\\.[a-zA-Z0-9-]+\\)*\\)?\\)\\'")
       (post "\\`\
\\(?1:\\(?:v\\(?:ersion\\)?\\|r\\(?:elease\\)?\\)[-_]?\\)?\
\\(?2:[0-9]+\\(?:\\.[0-9]+\\)*\
\\(?:-[a-zA-Z0-9-]+\\(?:\\.[a-zA-Z0-9-]+\\)*\\)?\\)\\'")
       (s1 "1.0")
       (s2 "_1.0"))
  (list
   (and (string-match pre s1)
        (match-string 1 s1))
   (and (string-match post s1)
        (match-string 1 s1))
   (and (string-match pre s2)
        (match-string 1 s2))
   (and (string-match post s2)
        (match-string 1 s2)))) ;; => ("" nil "_" nil)
```

The pre-s2 case may be an undefined edge case, but the s1 `"" -> nil` seems more likely to cause breakage.  Based on your commit message, I think the change in behavior is unintended.  If not, have you tested/analyzed the handling of the proposed change?

kyleam (changes requested):
Thanks.  All but one of these look good to me.

---

commit message comments:
> - Remove invalid syntax class 's' from skip-syntax-bacward call.

s/skip-syntax-bacward/skip-syntax-backward/

> Remove redundant shy groups.

Hmm, is shy groups the right term there?

The example referenced is `\\(?2:.*\\)?` vs `\\(?2:.*\\)`.  That's removing the "zero or one" postfix operator, which makes sense, but I don't see where shy groups come in.

basil-conto:
Thanks for the great feedback.

> > * Remove invalid syntax class 's' from skip-syntax-bacward call.
> 
> s/skip-syntax-bacward/skip-syntax-backward/

Fixed.

> > Remove redundant shy groups.
> 
> Hmm, is shy groups the right term there?

No, that was a brain fart.  Now fixed.

lisp/magit-log.el
basil-conto: > Okay, though I have a slight preference for this to come in as a separate commit.

Done.

lisp/magit-tag.el
basil-conto: > The pre-s2 case may be an undefined edge case, but the s1 `"" -> nil` seems more likely to cause breakage.

Do you have any ideas for what might break, e.g. 3rd-party packages, or how?

> Based on your commit message, I think the change in behavior is unintended.

Shall I call it out explicitly?

> If not, have you tested/analyzed the handling of the proposed change?

AFAICT all in-tree uses of `magit-release-tag-regexp` correctly guard against a non-match, and `""` is handled as equivalent to `nil`.  Which was to be expected, I think: `pre` defined group 1 as optional, i.e. potentially `nil`, even if in practice `match-string` returned `""` for a non-match.  So in my eyes the new `nil` value is what the callers could/should have expected all along.

basil-conto: Or am I missing something?

basil-conto:
> This PR tries to address all warnings (details in commit message), except for the `` Suspect `[' in char alternative `` which looks like a false positive: [mattiase/xr#7](https://github.com/mattiase/xr/issues/7)

Mattias suggested there what I think is a good idea regardless of the linter warning: to change
```
;; from:
"\\[[^[]*?]"
;; to:
"\\[[^][]*]"
```
in `magit-log-propertize-keywords`.  It should be faster, reads a bit more clearly to my eyes, and also avoids the linter warning.  I added the non-greedy `*?` operator in ac1ee3df54769aff61c7845df3d3dd7d131e3fe8, but I'm not sure why the original regexp in 25aa78f6189b755f9263f972c82a0d9b0c00d02f used `[^[]` rather than `[^][]`, which Mattias suggests.  @tarsius?

lisp/magit-tag.el
kyleam: > Do you have any ideas for what might break, e.g. 3rd-party packages, or how?

My concern was with in-tree uses.  I hadn't looked closely and was worried something was assuming an empty string and wouldn't handle nil...

> AFAICT all in-tree uses of `magit-release-tag-regexp` correctly guard against a non-match, and `""` is handled as equivalent to `nil`.

...which isn't the case.

> Which was to be expected, I think: `pre` defines group 1 as optional, i.e. potentially `nil`, even if in practice `match-string` returns `""` for a non-match. So in my eyes the new `nil` value is what the callers could/should have expected all along.

Good point.

> Shall I call it out explicitly?

No, I think it's fine as is.  I was just being slow on the uptake.  Thanks.

kyleam (approved):
Looks good to me.  I'll leave the merge to @tarsius, in case he has any feedback on these changes or [your comment](https://github.com/magit/magit/pull/4980#issuecomment-1674351883) about the remaining warning.

tarsius:
I think this should be split into three commits (at least):

1. syntax
2. regexp, totally safe
3. regexp, not 100% the same, but that's okay

That would make reverting easier if it turns out that (3) breaks something.

basil-conto:
> I think this should be split into three commits (at least):
> 
>     1. syntax
> 
>     2. regexp, totally safe
> 
>     3. regexp, not 100% the same, but that's okay

I've now tried doing so:
1. Syntax
2. Safe regexp changes
3. Resulting logic simplification
4. Unsafe regexp changes

WDYT?  Thanks.

tarsius:
IMO the change to `magit-log-reflog-re` (now in 2) and the resulting simplification in `magit-log-wash-rev` (current 3) belong in the same commit (new 3). It would also have been okay to make all that part of 4, but since you already do parts of it separately in 3, it makes sense now to just move all of it to 3. Also, while this is a "safe" change, it differs from the other safe changes, in that it enables further cleanup outside of regular expressions; another reason to keep it separate (but contained in a single commit). And while, I have now also concluded that this is safe, this was originally one of the changes I wanted to review in isolation, i.e., I was not immediately sure it was safe/correct.

(I realize this is entering nit-picking territory now, but since we got this far already... :grin: )

basil-conto:
> IMO the change to `magit-log-reflog-re` (now in 2) and the resulting simplification in `magit-log-wash-rev` (current 3) belong in the same commit (new 3).

Done (albeit as commit 4 rather than 3, feel free to reorder if you have a different ordering preference).

> (I realize this is entering nit-picking territory now, but since we got this far already... 😁 )

No worries, I'm familiar with the territory ;).

tarsius:
Thanks!

PR #4988, opened 2023-08-16
git-commit-setup: Fix auto-mode-alist for remote files

### LINKED ISSUE #4987: git-commit-major-mode does not work when using git over ssh with tramp
# Summary

The `git-commit-major-mode` variable, which is defined [here](https://github.com/magit/magit/blob/48818355728c48d986d74dde8b1e9fba25f0fd53/lisp/git-commit.el#L182), does not work when using git over ssh with tramp.

# Detail

The variable is used [here](https://github.com/magit/magit/blob/48818355728c48d986d74dde8b1e9fba25f0fd53/lisp/git-commit.el#L537) where the `COMMIT_EDITMSG` file's path is added to the `auto-mode-alist` which controls which files get which major modes.  For a regular file name, it should work fine.

However, the `auto-mode-alist` is not just a simply match against filenames.  Rather, it first parses the filename and strips out any weirdness, such as versions (if there are multiple of the same filename?) or tramp stuff.

So if the variable `buffer-file-name` is, for example `"/ssh:foo@1.2.3.4:/home/user/project/.git/COMMIT_EDITMSG`, then the filename that is actually checked against the `auto-mode-alist` is just `/home/user/project/.git/COMMIT_EDITMSG`, without the ssh stuff that tramp uses.

magit doesn't know about this and it is giving the entire name.  That is causing the match against auto-mode-alist to fail, which means that the `git-commit-major-mode` is not enacted.

# The fix

The code that is actually doing the match and applying the adjustment is here:

https://github.com/emacs-mirror/emacs/blob/647bcec4f537d49b7a1e6d200ec787fec11ed81a/lisp/files.el#L3320-L3324

We can do the same thing in magit, that way it will match.  The downside here is if there happens to be an identically named file (including path) on a different ssh or locally, but that shouldn't be a problem because we are only modifying the `auto-mode-alist` temporarily for this just one file.  I can't imagine a situation where this this would happen and someone would get the mode applied to the wrong file.

I'll get to work on a patch and have it ready soon.

# Version
```
Magit 20230815.1255 [>= 3.3.0.50-git], Transient 0.4.1, Git 2.25.1, Emacs 29.0.50, gnu/linux
```

eyal0:
The filename added to `auto-mode-alist` must be adjusted to account for tramp and versions because that is how the `auto-mode-alist` matching is done.
    
This code is based on [these](https://github.com/emacs-mirror/emacs/blob/647bcec4f537d49b7a1e6d200ec787fec11ed81a/lisp/files.el#L3315-L3324) lines.

This fixes #4987

I tested this change on my desktop which has a tramp connection to files that I am working on.  Before the change, the commit message buffer opened as fundamental mode, but only only remote ssh files with tramp.  After the change, both the local files and the remote ones are honoring the `git-commit-major-mode`.

### DIFF
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -534,10 +534,17 @@ Used as the local value of `header-line-format', in buffer using
       (hack-dir-local-variables)
       (hack-local-variables-apply)))
   (when git-commit-major-mode
-    (let ((auto-mode-alist (list (cons (concat "\\`"
-                                               (regexp-quote buffer-file-name)
-                                               "\\'")
-                                       git-commit-major-mode)))
+    (let ((auto-mode-alist
+           ;; `set-auto-mode--apply-alist' removes the remote part from
+           ;; the file-name before looking it up in `auto-mode-alist'.
+           ;; For our temporary entry to be found, we have to modify the
+           ;; file-name the same way.
+           (list (cons (concat "\\`"
+                               (regexp-quote
+                                (or (file-remote-p buffer-file-name 'localname)
+                                    buffer-file-name))
+                               "\\'")
+                       git-commit-major-mode)))
           ;; The major-mode hook might want to consult these minor
           ;; modes, while the minor-mode hooks might want to consider
           ;; the major mode.


lisp/git-commit.el
```diff
@@ -534,8 +534,18 @@ Used as the local value of `header-line-format', in buffer using
...
+    (let* ((name buffer-file-name)
+           ;; The auto-mode-alist procesing will remove any tramp-mode
+           ;; and version artifacts from the filename before
+           ;; processing it so the same is done here.
+           (remote-id (file-remote-p buffer-file-name))
+           (name (file-name-sans-versions name))
+           (name (if (and (stringp remote-id)
+                          (string-match (regexp-quote remote-id) name))
+                     (substring name (match-end 0))
+                   name))
```
kyleam, line 546: Could this be simplified by using `(file-remote-p ... 'localname)`?

```diff
diff --git a/lisp/git-commit.el b/lisp/git-commit.el
index b11141a3..aac66e19 100644
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -534,10 +534,13 @@ (defun git-commit-setup ()
       (hack-dir-local-variables)
       (hack-local-variables-apply)))
   (when git-commit-major-mode
-    (let ((auto-mode-alist (list (cons (concat "\\`"
-                                               (regexp-quote buffer-file-name)
-                                               "\\'")
-                                       git-commit-major-mode)))
+    (let ((auto-mode-alist
+           (list (cons (concat "\\`"
+                               (regexp-quote
+                                (or (file-remote-p buffer-file-name 'localname)
+                                    buffer-file-name))
+                               "\\'")
+                       git-commit-major-mode)))
           ;; The major-mode hook might want to consult these minor
           ;; modes, while the minor-mode hooks might want to consider
           ;; the major mode.
```

kyleam (changes requested):
Thanks, I haven't tested on my end yet, but your problem description/fix sound right.

project convention nit-pick: commit message subject should be something along the lines of "git-commit-setup: Fix auto-mode-alist for remote files".  (Note the area/function prefix, and the lack of a trailing period.)

lisp/git-commit.el
eyal0: It seems like that ought to work!  I had to squint at the docs of `file-remote-p` for a while!

You'll have to call file-name-sans-versions twice, though, which is not pretty.  Imagine in the future it changes: you'll have to type it twice!

```elisp
    (let ((auto-mode-alist
           ;; The auto-mode-alist procesing will remove any tramp-mode
           ;; and version artifacts from the filename before
           ;; processing it so the same is done here.
           (list (cons (concat "\\`"
                               (regexp-quote
                                (let ((buffer-file-name-sans-versions
                                       (file-name-sans-versions buffer-file-name)))
                                  (or (file-remote-p (file-name-sans-versions buffer-file-name) 'localname)
                                      (file-name-sans-versions buffer-file-name))))
                               "\\'")
                       git-commit-major-mode)))
```

I'll make this change.

eyal0: oops, I mean:

```elisp
    (let ((auto-mode-alist
           ;; The auto-mode-alist procesing will remove any tramp-mode
           ;; and version artifacts from the filename before
           ;; processing it so the same is done here.
           (list (cons (concat "\\`"
                               (regexp-quote
                                (let ((buffer-file-name-sans-versions
                                       (file-name-sans-versions buffer-file-name)))
                                  (or (file-remote-p buffer-file-name-sans-versions 'localname)
                                      buffer-file-name-sans-versions)))
                               "\\'")
                       git-commit-major-mode)))
```

eyal0: Let me know if you prefer this.

eyal0:
> Thanks, I haven't tested on my end yet, but your problem description and fix sound right.
> 
> project convention nit-pick: commit message subject should be something along the lines of "git-commit-setup: Fix auto-mode-alist for remote files". (Note the area/function prefix, and the lack of a trailing period.)

No problem!

eyal0:
I don't get it, it says resolved but it's not resolved?  :shrug:

tarsius:
The commend doesn't quite work for me. This seem better, but maybe could also use some tweaking:

```
          ;; `set-auto-mode--apply-alist' removes the remote part from
          ;; the file-name before looking it up in `auto-mode-alist'.
          ;; For our temporary entry to be found, we have to modify the
          ;; file-name the same way.
```

lisp/git-commit.el
```diff
@@ -534,10 +534,18 @@ Used as the local value of `header-line-format', in buffer using
...
+    (let ((auto-mode-alist
+           ;; The auto-mode-alist procesing will remove any tramp-mode
+           ;; and version artifacts from the filename before
+           ;; processing it so the same is done here.
+           (list (cons (concat "\\`"
+                               (regexp-quote
+                                (let ((buffer-file-name-sans-versions
+                                       (file-name-sans-versions buffer-file-name)))
+                                  (or (file-remote-p buffer-file-name-sans-versions 'localname)
+                                      buffer-file-name-sans-versions)))
```
kyleam, line 546: [ in addition to @tarsius's [suggestion](https://github.com/magit/magit/pull/4988#issuecomment-1682219355) to reword the comment ]

Rearranging could avoid the let-binding while keeping a single `file-name-sans-versions` call:

```elisp
(file-name-sans-versions
 (or (file-remote-p buffer-file-name 'localname)
     buffer-file-name))
```

But I don't see why we need `file-name-sans-versions` here.  For this case, won't the buffer always be visiting one of the known sets of "commit" files matched by `git-commit-filename-regexp`?

As far as I can tell, `git-commit-setup` won't even be called for a backup file because it wouldn't match `git-commit-filename-regexp`:

https://github.com/magit/magit/blob/48818355728c48d986d74dde8b1e9fba25f0fd53/lisp/git-commit.el#L455-L458

https://github.com/magit/magit/blob/48818355728c48d986d74dde8b1e9fba25f0fd53/lisp/git-commit.el#L439-L441

eyal0: > Rearranging could avoid the let-binding while keeping a single file-name-sans-versions call:

But then we'd be calling them in a different order.  I don't know if that makes a difference but I would think it best to retain the order.

Really the best thing would be if the emacs source code pushed all this stuff into a function and we could just call that function.  And then we'd stay in sync.  But it's probably too much to expect emacs to do that!  I suppose that we could ask...

We're beginning to get into [bikeshedding](https://en.wiktionary.org/wiki/bikeshedding) territory here.

eyal0:
> The commend doesn't quite work for me. This seem better, but maybe could also use some tweaking:
> 
> ```
>           ;; `set-auto-mode--apply-alist' removes the remote part from
>           ;; the file-name before looking it up in `auto-mode-alist'.
>           ;; For our temporary entry to be found, we have to modify the
>           ;; file-name the same way.
> ```
> 
> Edit: ... and suffix

Done.  d8a13fac68e094aa6d30532f1910ef5945e01781

lisp/git-commit.el
kyleam: As I said above, as far as I can tell, we would never be operating while visiting a backup file.  If that's the case, keeping the `file-name-sans-versions` just invites confusion in my eyes.  My recommendation is to drop it entirely.

lisp/git-commit.el
```diff
@@ -534,10 +534,19 @@ Used as the local value of `header-line-format', in buffer using
...
+    (let ((auto-mode-alist
+           ;; `set-auto-mode--apply-alist' removes the remote part
+           ;; from the file-name before looking it up in
+           ;; `auto-mode-alist'.  For our temporary entry to be found,
+           ;; we have to modify the file-name the same way.
+           (list (cons (concat "\\`"
+                               (regexp-quote
+                                (let ((buffer-file-name-sans-versions
+                                       (file-name-sans-versions buffer-file-name)))
+                                  (or (file-remote-p buffer-file-name-sans-versions 'localname)
```
kyleam, line 546: Regardless of whether you go with my [suggestion](https://github.com/magit/magit/pull/4988/files#r1298987443) for how to rework this, please reflow these lines to not exceed 80 characters to match style used throughout this code base.

kyleam:
One additional meta comment...  your commit is under a `users.noreply.github.com` address.  I believe this quote from Jonas (just from the [first case](https://github.com/magit/magit/pull/3834#issuecomment-484859706) I happened to find with a search) still represents the current view:

> We would like to properly attribute changes to people. It may be possible to configure Github to do that, but the way you have configured it the commit is attribute to your Github alias and some dummy email. We would like to use your real name and email; but won't force you to do so.

---

I've now had time to test this, and it worked as advertised on my end.

Thanks again for fixing this.

lisp/git-commit.el
eyal0: Okay!

eyal0:
All done.  Anything else?

tarsius:
> Anything else?

As I usually do for such minor tweaks, I have improved the filling of the comment and tweaked the commit message a bit.

Thanks for hanging in there. :stuck_out_tongue_winking_eye:

PR #4989, opened 2023-08-16
magit-revision-diff-merges-option

dickmao:
The pedants will say first-parent shouldn't be the default but they're wrong.  While there's no convincing a pedant, we should at least make first-parent possible.


tarsius:
`--diff-merges` is now available from the diff transient. You can set and permanently save it (using `C-x C-s`) from there.

PR #5019, opened 2023-10-06
magit-log-remove-graph-args: ask user to provide all variants

marc-hb:
It's not practical for the average user to inspect the magit-log.el source code and find whether which it uses `-u` or `--patch` internally. It's much faster and simpler for the user to just add all variants.

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -70,7 +70,9 @@
   :type 'hook)
 
 (defcustom magit-log-remove-graph-args '("--follow" "--grep" "-G" "-S" "-L")
-  "The log arguments that cause the `--graph' argument to be dropped."
+  "The log arguments that cause the `--graph' argument to be dropped. For
+a predictable and consistent behavior, add _all_ variants on separate lines,
+example: --patch, -p and -u."
   :package-version '(magit . "2.3.0")
   :group 'magit-log
   :type '(repeat (string :tag "Argument"))


tarsius:
The `magit-log` transient prefix command lists the arguments that the user can optionally enable. `--patch` is one of them, the aliases `-p` and `-u` are not. It doesn't make sense to suggest that the user add `-p` and `-u` to `magit-log-remove-graph-args` in order to something *when those arguments are passed to `git log`* because that never happens.

marc-hb:
> It doesn't make sense to suggest that the user add -p and -u to magit-log-remove-graph-args in order to something when those arguments are passed to git log because that never happens.

So _you_ know that but how does the user know it? Now and in the future - customizations carry across versions.

This is the user interface:

```
Formatting
-g Show graph (--graph)
-p Show diffs (--patch)
...
```

This does not tell the user what implementation detail magit uses internally.

I understand the text between (parentheses) to show what is the equivalent for a plain git log command but I before looking at the source code I had no idea what magit actually does internally. If I had to guess I would have bet on some finely controlled `--format=...` option instead or something like it.

If the text between parentheses is actually the internal implementation detail (always?) then something in the user interface should clearly state that.

marc-hb:
> The magit-log transient prefix command lists the arguments that the user can optionally enable

Anther UI issue is the large distance between the customization screen and the main user interface. When I read this:

> "The log arguments that cause the `--graph' argument to be dropped."

It was not clear to me at all where these log arguments could come from. Maybe there's some advanced input field that lets the user enter arguments directly? Because why not.

Now I know for sure but only since you just told me above.

tarsius:
I understand your argument and agree that you have a point.

But I still disagree that this extra information should be added.

One reason for that is that I wouldn't want to suggest the argument under the names `--patch`, `-p` and `-u` should be added to this option at all. Users are welcome to do that, after all it is an option, not an internal variable, but that doesn't mean that I want to suggest that this is an argument that should be added there.

This variable was first added in 297e53ea761d8743580bdfe657c92b02f9eaa091. As you can see, this functionality was added to disable `--graph` when an argument is used which is *incompatible* with that. Not arguments which one might or might not want to use together with that. Of course by making this an option.

You want to disable `--patch/-u/-p` when `--graph` is used. Other users might want to do that for other arguments. By singling out `--patch/-u/-p` I would suggest that it makes more sense to add this argument, which I do not agree with. So should we instead list all commands that one might possibly add, listing each one under its various names?

> So *you* know that but how does the user know it?

To use this interface, users have to understand in the first place that they can toggle the listed arguments by pressing the respective key sequence. When they do that, they see that the color of the argument (displayed at the end of the line in between parenthesis) changes colors. They have to further understand that the color change indicates that the state of the argument changed, i.e., whether the argument is going to be used or not. From there it is not a big jump to assume that the argument that is actually being used, is the argument that changes color when being toggled.

marc-hb:
> But I still disagree that this extra information should be added.

I should have mentioned that you convinced me of that from the beginning, I should have said sorry.  Instead I immediately jumped to state the problem better and touch on alternatives. Basically turning this PR into a github issue.

tarsius:
I've added more information to the docstring.

marc-hb:
> when an argument is used which is incompatible with that. Not just any arguments which one might or might not want to use together with --graph

Mmmm... then why is it customizable? It's either compatible or not, isn't it?

(the personal reasons I got into this are performance and readability, not compatibility)

> By singling out --patch/-u/-p I would suggest that it makes more sense to add this argument

I don't think it would have because it was prefixed with "Example:". Anyway this point is moot, forget this suggestion and current PR state entirely.

> To use this interface, users have to understand in the first place that they can toggle the listed arguments by pressing the respective key sequence.

Yes but as I tried to say earlier there's no clear indication that the options listed in the "transient" user interface are the ONLY possible way to alter the behavior. There could be some other, advanced interface or customization that provides more possibilities.


> From there it is not a big jump to assume that the argument that is actually being used, is the argument that changes color when being toggled.

I disagree with this and I will boldly argue that I have a better "user" perspective than you on this - because I barely ever looked at the implementation and code. I've used magit for years (and git for even longer) and I always expected the text between parentheses to be EQUIVALENT to the corresponding git log parameter. So it is a "big jump" to assume this is exactly what is used behind the scenes, that may or may not the case. In general: interface != implementation.

I'll think more and try to submit a (completely) different patch. The current help text is just one line long so I hope that leaves room for adding one sentence or two.

EDIT: just saw you beat me to it.

marc-hb:
> I recommend that you don't do that.  Nowadays I would define
this as a constant, but I am preserving it as an option, in case
someone actually customized it."

OK, now that option is clearly in the "internal implementation" area. This answers most of my concerns, thanks!

marc-hb:
> (the personal reasons I got into this are performance and readability, not compatibility)

Full disclosure: I'm using magit over tramp. Crazy, I know...

tarsius:
Aaah, don't remind me! There are some Tramp issues I need to investigate, and since I don't really us Tramp except to investigate issues other people are having when using Tramp with Magit, that's always rather painful.

marc-hb:
> that's always rather painful.

You mean having to setup two systems? Not painful using tramp itself? The latency is painful but the rest is fine in my experience (otherwise I wouldn't keep using it). Just some rare crashes now and then.

Despite the many round trips, the latency should be more than bearable on a LAN. But wait, there's even better: for interactive testing you can just use `/scp:localhost:usual/files`! I just tried and it worked just fine. No second system to setup and absolute minimum latency. Did you never try this? Just make sure ssh ControlMaster is on; should already be the default in Tramp.

> Just some rare crashes now and then.

Very off topic sorry but do you have recommendation(s) for an Emacs debugger? Something for a seasoned Emacs user and experienced developer but elisp rookie. `(message ...)` gets old very quickly.

marc-hb:
Back on topic: what I really miss (or couldn't find?) is a simple way to change the default log options. I ended up changing them directly in the `magit-log.el` source code.

There are many projects who use a linear git history, especially on Github thanks to its "Rebase and merge" button. Tramp or not, for this sort of project `--graph` is purely slowing down the interface for zero value.

Also, I don't see why --patch is not the default. For --one-line, magit-status is for me good enough the vast majority of the time.

Everyone has different `git log` preferences; you get the idea. this is a customize feature recommendation in disguise in case that wasn't clear enough :-)

kyleam:
> Back on topic: what I really miss (or couldn't find?) is a simple way to change the default log options.  I ended up changing them directly in the `magit-log.el` source code.

You can set the values in the transient and then [save them](https://magit.vc/manual/transient/Saving-Values.html).

marc-hb:
Thanks @kyleam , I had no idea you could do that. Much appreciated. I tried `Ctrl-h m` a couple times but transient does not work like that. I should have searched the manual!

PS: if you ever need a tramp guinea pig or have any question about it please tag me.

PR #5027, opened 2023-10-11
git-commit: Ignore leading comment when inserting headers

magnars:
It is suggested several places on the internet to use a git commit template, for instance
[here](https://gist.github.com/lisawolderiksen/a7b99d94c92c6671181611be1641c733). These often include a header comment:

    # Title: Summary, imperative, start upper case, don't end with a period
    # No more than 50 chars. #### 50 chars is here:  #

This change ensures that headers are inserted after this leading comment, instead of above it.

### DIFF
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -943,17 +943,28 @@ have the form \"NAME <EMAIL>\"."
 (defun git-commit-insert-header (header name email)
   (setq header (format "%s: %s <%s>" header name email))
   (save-excursion
-    (goto-char (point-max))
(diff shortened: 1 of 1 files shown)


magnars:
I was a bit quick on the trigger. This code will only ignore a single leading comment line. I'll brew up something a bit more robust.

magnars:
This updated version will skip past any leading comments, make a note of where they are, and then make sure to insert the header before any comments except the leading ones.

tarsius:
Thanks!

magnars:
Thank you! 🙏😊

PR #5036, opened 2023-10-27
magit-diff-wash-diffstat: Allow empty renames

nbarrientos:
This patch complements 564cff8a40c2d7c8d4e679f1c7c2974c97f5f149 by also allowing following file renames formatted as follows:

* `foo/{ => bar/baz}/qux/quux.c`
* `foo/{bar/baz => }/qux/quux.c`


tarsius:
Thanks a lot!

PR #5065, opened 2023-12-13
magit-push: add new --push-option transient

dpolad:
Adding -o (--push-option) transient to magit-push.
I have added an Infix Commands section in magit-push.el taking inspiration from other files.

### DIFF
--- a/lisp/magit-push.el
+++ b/lisp/magit-push.el
@@ -39,6 +39,7 @@
    ("-F" "Force"            ("-f" "--force"))
    ("-h" "Disable hooks"    "--no-verify")
    ("-n" "Dry run"          ("-n" "--dry-run"))
+   (magit-push:--push-option)
    (5 "-u" "Set upstream"   "--set-upstream")
    (7 "-t" "Follow tags"    "--follow-tags")]
   [:if magit-get-current-branch
@@ -371,6 +372,14 @@ You can add this command as a suffix using something like:
 
 (defun magit-push-to-remote--desc ()
   (format "using %s" (magit--propertize-face "git push <remote>" 'bold)))
+
+;;;; Infix Commands
+
+(transient-define-argument magit-push:--push-option ()
+  :description "Push Options"
+  :class 'transient-option
+  :key "-o"
+  :argument "--push-option=")
 
 ;;; _
 (provide 'magit-push)


tarsius:
Thanks, but please add this to your personal configuration for now.

We should probably add official support for this argument eventually, which should go beyond a minimal viable version right away. It should be possible to set the variable multiple times and often used values should be offered for completion. Maybe certain common settings should be available using dedicated infix arguments. I assume different forges use different options, so we should also filter the offered choices. We should probably begin by compiling a list of links to the relevant documentation of different forges.

dpolad:
Hi,
I've done most of what's been suggested in the last comment. (multi-value, git version check and collection of common options from different forges documentation)

Do you have an example in the code on how to filter the forge currently being pushed to? I haven't found any

tarsius:
> Do you have an example in the code on how to filter the forge currently being pushed to?

I don't understand what that means.

dpolad:
> I assume different forges use different options, so we should also filter the offered choices
I am referring to your comment above, my interpretation is that, given this list:

**GitLab**
ci.skip
ci.variable="<name>=<value>"
integrations.skip_ci
merge_request.create
merge_request.target=<branch_name>
merge_request.target_project=<project>
merge_request.merge_when_pipeline_succeeds
merge_request.remove_source_branch
merge_request.title="<title>"
merge_request.description="<description>"
merge_request.draft
merge_request.milestone="<milestone>"
merge_request.label="<label>"
merge_request.unlabel="<label>"
merge_request.assign="<user>"
merge_request.unassign="<user>"

**Gerrit**
topic=driver/i42
hashtag=stable-fix
t=stable-fix    
trace=issue/123
trace
deadline=10m
push-justification=id/2345
skip-validation

**SourceHut**
debug
skip-ci
submit=
description=
visibility= (public, unlisted, or private).

You would like to provide as completion option only those that are matching something like (guessing) magit-get-current-remote ?
e.g. gitlab options to gitlab forge, etc..

tarsius:
You probably need code like this somewhere:


```elisp
(cl-typecase (forge-get-repository t)
  (forge-gitlab-repository my-forge-gitlab-push-options)
  ...)
```

dpolad:
Hi, sorry for bothering again.
Do you think it makes more sense to have the change confined into magit or should I add the various push-options lists and getter method directly in forge (IMHO where they logically belong)?

dpolad:
Here's the current commit:

https://github.com/magit/magit/compare/main...dpolad:magit:push-option

tarsius:
I would say it belongs in Magit because it doesn't use the forges' APIs, but Forge is a better fix because then we can use `forge-get-repository`. (There's probably some "type-of-remote" functionality in Magit somewhere, but I don't remember where and also suspect that it is a minimal-viable implementation, so using Forge's implementation is preferable.)

That being said, I don't really want to add it to either package at this point. Looking at the available option, I think that at some point in the future I am going to add a forge-agnostic "skip-ci" option. Adding Gitlab's `merge_request.*` options like this, seem pointless to me; it would be a painful experience to create a merge-request like this.

Generally speaking I am back to thinking it is best if users add (in their personal configuration) dedicated (transient infix) arguments for the two or three push-options that they actually use.

----

Review:
- Never quote lambda.
- `:if` can be set inside `transient-define-argument`.
- Use `and` not `when` for forms whose values are used (i.e., are not called for side-effects only).
- Reduce indentation, staying below 80 characters.
- The lists of available push-options should probably be stored in variables (or one alist stored in one variable).

dpolad:
Alright, many thanks for your help and the review.
I was naively thinking this would have been easier, but the forge interaction makes it definitely more problematic.

PR #5076, opened 2024-01-09
magit-status-setup-buffer: Ensure correct position when narrowed

leungbk:
-------

When using `M-x magit-status-here` in a narrowed buffer, something like this is necessary to visit the corresponding position of the file.

~I thought about using `pcase-let*` instead to avoid having to `widen` twice, but I also didn't want to clutter the diff and `git blame` sessions.~

I changed my mind, `save-restriction` and `widen` might not necessarily be fast.


tarsius:
> I changed my mind, `save-restriction` and `widen` might not necessarily be fast.

I benchmarked it. It makes no difference.

Issue #3384, opened 2018-02-20
Add hooks to decorate log entries


snogge:
I'd like to be able to add various indicators to loglines. 
I'm thinking of test report icons from travis and review-votes and things like that.
The decoration could be characters or images added to the line, or style overlays that set a background color or add a margin or fringe.

tarsius:
I have added a new variable `magit-log-format-message-function` that you can use for this purpose. It is called with two arguments, the hash and the message. You probably want to call the default value `magit-log-propertize-keywords` in function.

I haven't documented this variable and added no entry to the release notes because this will likely change within a few months. I might, for example, replace this with a hook as you suggested, but I just don't know yet until I start refactoring this area of Magit.

snogge:
Thank you!
I will try it out and report back.

snogge:
So I made a test to display Gerrit votes for each commit in the log, and that worked.
Unfortunately it is so slow that I actually see the washing happening in the buffer. 
For this operation I first had to find the Gerrit Change-Id from the commit message.  That required a call to git through `magit-rev-insert-format`. Then I had to make a REST API call to the Gerrit server. 

I think some kind of outer scope and state object is needed to make this efficient, even if there must be something wrong with my setup to make it this slow.

It would also be useful to only trigger the format function for certain fixes.  For pull request and CI info commits that are already merged are not that interesting.

tarsius:
I think you should be able to implement those things without any further changes to Magit. Using some log related hook and a buffer-local variable should do the trick.

If you disagree, then make the necessary changes, get performance right, and then let me have a look.

PR #5084, opened 2024-02-03
Add support for custom keywords similar to `squash!` / `fixup!`

happy-barney:
to support project/user specific markings.

For example:
- `wip! this commit isn't finished yet`
- `do-not-merge! this commit (whatever reasons are)`
  - eg printf-debugging commit
- `test! this commit isn't finished yet, it requires test`

### DIFF
--- a/lisp/magit-log.el
+++ b/lisp/magit-log.el
@@ -1476,6 +1476,10 @@ Do not add this to a hook variable."
       (setq boundary (match-end 0))
       (magit--put-face (match-beginning 0) (1- boundary)
                        'magit-keyword-squash msg))
+    (when (string-match "^\\(?:[a-zA-Z_][-a-zA-Z_]+[a-zA-Z_]\\)! " msg boundary)
+      (setq boundary (match-end 0))
+      (magit--put-face (match-beginning 0) (1- boundary)
+                       'magit-keyword-custom msg))
     (when magit-log-highlight-keywords
       (while (string-match "\\[[^][]*]" msg boundary)
         (setq boundary (match-end 0))
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -197,6 +197,11 @@ and/or `magit-branch-remote-head'."
   "Face for squash! and fixup! keywords in commit messages."
   :group 'magit-faces)
 
+(defface magit-keyword-custom
+  '((t :inherit font-lock-warning-face :foreground "orange red"))
+  "Face for custom xxx! keywords in commit messages."
+  :group 'magit-faces)
+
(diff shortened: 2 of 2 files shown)


kyleam:
Thanks for PR.  I see why this is useful for your given workflow/conventions, but in my view this falls under the same category of custom highlighting discussed in gh-4027.  For widely used conventions, a dedicated package could make sense, and for more tailored things you can also do something like this in your config:

```elisp
(advice-add
 'magit-log-propertize-keywords :filter-return
 (lambda (msg)
   (when (and (not (get-text-property 0 'font-lock-face msg))
              (string-match "\\`[a-zA-Z_][-a-zA-Z_]+[a-zA-Z_]! " msg))
     (let ((beg (match-beginning 0))
           (end (match-end 0)))
       (put-text-property beg end 'face 'font-lock-warning-face
                          msg)
       (put-text-property beg end 'font-lock-face 'font-lock-warning-face
                          msg)))
   msg)
 '((name . "my/magit-log-propertize-custom-keywords")))
```

happy-barney:
Thanks for comment, you are right, there can by more rules applicable.

On other hand I think this problem is quite common so basic support should be in `magit` as well (even if not enabled by default).

What about:
- create `alist` (regex  => face) evaluated instead of series of `when` (evaluated until first match) ?
- add support for hooks into `magit-log-propertize-keywords` (evaluated until first match) ?

kyleam:
> On other hand I think this problem is quite common so basic support should be in `magit` as well (even if not enabled by default).

I don't know how common the desire is to customize this, especially outside of some common styles/conventions that could be handled by dedicated packages (such as gh-4027).

> create `alist` (regex  => face) evaluated instead of series of `when` (evaluated until first match) ?

In my view this isn't worth adding and maintaining a customization knob for.  If a user wants something tailored, I think the advice approach is fine for this case, and there's also `magit-log-format-message-function` (see gh-3384).

> add support for hooks into `magit-log-propertize-keywords` (evaluated until first match) ?

In response to gh-3384, `magit-log-format-message-function` was introduced.  Unless there's a strong reason that won't work, I'd prefer to use existing functionality instead of adding more pieces.

In the case of a dedicated package implementing highlighting a custom style/convention, `magit-log-format-message-function` would likely be the starting point, but it may be that we need to fill in some gaps or make other adjustments on Magit's side.

kyleam:
Closing for the reasons given above, but please feel free to continue the conversation here or start a discussion about alternative approaches.

happy-barney:
Got your arguments, now I'm thinking how to do it, thanks for ideas :-)

PR #5087, opened 2024-02-07
magit-process-interrupt: support sending SIGINT to running subprocess

petergaultney:
This addresses https://github.com/magit/magit/discussions/5086, by providing a nicer workflow for canceling long-running actions (often pre-commit hooks, but could be anything) while giving those processes a chance to clean up after themselves gracefully.

I don't know where/whether we'd want to add this to the docs somewhere. I'm happy to try to help with that, and also happy for a maintainer to jump in and do it if they prefer.

### DIFF
--- a/docs/magit.org
+++ b/docs/magit.org
@@ -1825,7 +1825,13 @@ sections are available.  There is one additional command.
 
 - Key: k (magit-process-kill) ::
 
-  This command kills the process represented by the section at point.
+  This command kills the process represented by the section at point
+  by sending the signal ~SIGKILL~.
+
+- Key: M-x magit-process-interrupt ::
+
+  This command kills the process represented by the section at point
+  by sending the signal ~SIGINT~.
 
 - Variable: magit-git-debug ::
 
--- a/docs/magit.texi
+++ b/docs/magit.texi
@@ -2364,7 +2364,13 @@ sections are available.  There is one additional command.
 @item @kbd{k} (@code{magit-process-kill})
 @kindex k
 @findex magit-process-kill
-This command kills the process represented by the section at point.
+This command kills the process represented by the section at point
+by sending the signal @code{SIGKILL}.
+
+@item @kbd{M-x magit-process-interrupt}
+@findex magit-process-interrupt
+This command kills the process represented by the section at point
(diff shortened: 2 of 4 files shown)


kyleam:
Thanks for the PR.  This seems like a worthwhile addition.

I wonder a bit about whether it'd be good for this to be the default because ideally the process would have a chance to clean up before exiting.  One approach could be that the command bound to `k` could do `SIGINT` by default, and `SIGKILL` with a prefix arg.  I suspect there'd be minimal fallout from that change.

I think this PR's current approach of adding a separate (and unbound) function for `SIGINT` is fine (and wouldn't make the above change harder to do at a later time), but I'm curious if others have thoughts on this.

> I don't know where/whether we'd want to add this to the docs somewhere. I'm happy to try to help with that, and also happy for a maintainer to jump in and do it if they prefer.

I'll amend the commit to add `magit-process-interrupt` to magit.org.

tarsius:
For people like me who only very rarely have to kill `git`, neither using a prefix argument wouldn't be very discoverable, nor knowing that there are two commands and having to remember their key bindings.

We could change `magit-process-kill` to first send `SIGINT` and remember that using `process-put`, and then `SIGKILL` on the second try. The user wouldn't even have to know about this, a thought process like "hm, somehow it didn't work [maybe I pressed the wrong key?], let's try again" would be enough.

kyleam:
> [...] wouldn't be very discoverable [...]

Fair enough.

> We could change `magit-process-kill` to first send `SIGINT` [...]

I like that idea.

kyleam:
@petergaultney Given the discussion above, I'll close this, mentioning the [suggested approach](https://github.com/magit/magit/pull/5087#issuecomment-1947852173) on your feature request at gh-5086.

petergaultney:
> @petergaultney Given the discussion above, I'll close this, mentioning the [suggested approach](https://github.com/magit/magit/pull/5087#issuecomment-1947852173) on your feature request at [gh-5086](https://github.com/magit/magit/discussions/5086).

This sounds reasonable to me. I'm not sure I'm the best person to make the PR to do this, unfortunately, but hopefully someone else knows enough to get it done!

PR #5092, opened 2024-02-15
magit-refs: Fix error when HEAD is not symbolic

WxNzEMof:
If the HEAD (tracking) ref of a remote is not symbolic, it will not hit the `(if head-branch` check.

We should probably not be chopping off prefixes blindly anyway.

---

Thanks!

### DIFF
--- a/lisp/magit-refs.el
+++ b/lisp/magit-refs.el
@@ -589,7 +589,9 @@ line is inserted at all."
               (when (magit-refs--insert-refname-p branch)
                 (magit-insert-section (branch branch t)
                   (let ((headp (equal branch head))
-                        (abbrev (if magit-refs-show-remote-prefix
+                        (abbrev (if (or magit-refs-show-remote-prefix
+                                        (not (string-prefix-p (concat remote "/")
+                                                              branch)))
                                     branch
                                   (substring branch (1+ (length remote))))))
                     (magit-insert-heading


kyleam:
Thanks for the patch.

> If the HEAD (tracking) ref of a remote is not symbolic

Could you capture the situation in this script (or something like it)?

```sh
#!/bin/sh

set -eu

tdir=$(mktemp -d "${TMPDIR:-/tmp}"/magit-5092-XXXXXXX)
git init --bare "$tdir/a.git"
git clone --no-local "$tdir/a.git" "$tdir/a"
cd "$tdir/a"
# Setup needed...
```

Given this code also recently had an update to deal with a Git change: what's your Git version?

---

A general comment: the commit was made on GitHub and with a `*users.noreply.github.com` address.  See this [comment](https://github.com/magit/magit/pull/3834#issuecomment-484859706) by Jonas for more details, but 1) the commit message is missing a trailing newline (as is often the case with commits created on GitHub) and 2) to properly attribute you "[w]e would like to use your real name and email; but won't force you to do so".

WxNzEMof:
> Could you capture the situation in this script (or something like it)?

Unfortunately I have not been able to find the exact circumstances that can lead to the creation of such a repository.

> Given this code also recently had an update to deal with a Git change: what's your Git version?

2.43.0

> the commit message is missing a trailing newline (as is often the case with commits created on GitHub)

Recreated in magit.

> to properly attribute you "[w]e would like to use your real name and email; but won't force you to do so".

No attribution needed for contributions from this GitHub account, thank you.

tarsius:
> If the HEAD (tracking) ref of a remote is not symbolic,

I think `HEAD` is supposed to always be a symref. Git porcelain commands seem to prevent you from creating a regular ref with that name, but plumbing commands don't remove the footgun.

```
#!/bin/sh

set -eu

tdir=$(mktemp -d "${TMPDIR:-/tmp}"/magit-5092-XXXXXXX)
echo $tdir
git init "$tdir/a"
cd "$tdir/a"
git commit --allow-empty -m init
git branch HEAD main | true
git show-ref
echo "---"
git update-ref refs/heads/HEAD main
git show-ref

git clone --no-local "$tdir/a" "$tdir/b"
cd "$tdir/b"
git "for-each-ref" "--format=\
%(symref:short)%09%(refname:short)%09%(refname)"
```

We might consider *not* displaying that ref a feature. It's a mistake on the remote end, which you shouldn't have to worry about locally.

WxNzEMof:
To clarify, the problem is that rendering the entire buffer errors out (so refs and remotes after this odd one are not displayed at all). magit is trying to blindly remove a prefix without first checking that a string starts with the prefix, which I think is not right on its own, regardless of any circumstances that can lead to such a situation.

WxNzEMof:
> Could you capture the situation in this script (or something like it)?

I think I found how I ended up in this situation. This reproduces it:

```
rm -rf repo1 repo2 repo3 repo4

git init repo1
git -C repo1 commit --allow-empty -m 'Initial commit'

git clone repo1 repo2

git init repo3
git -C repo3 checkout -b dummy
git -C repo3 fetch ../repo2 '+refs/remotes/origin/*:refs/heads/*'

git clone repo3 repo4

git -C repo4 for-each-ref --format=$'%(symref:short)\t%(refname:short)\t%(refname)\t%(subject)' refs/remotes/origin
```

kyleam:
Thanks for the reproducer.

> git -C repo3 checkout -b dummy
> git -C repo3 fetch ../repo2 '+refs/remotes/origin/*:refs/heads/*'

That's puts the repo in quite the pathological state.  It clobbers the checked out branch in addition to adding an ambiguous `HEAD` branch.

In that state, a follow-up commit would fail:

```
repo3$ git commit --allow-empty -mfoo
fatal: cannot lock ref 'HEAD': unable to resolve reference 'refs/heads/dummy'
```

> magit is trying to blindly remove a prefix without first checking that a string starts with the prefix, which I think is not right on its own

Magit asks Git for output all over the place and has expectations for how it will come back. Some level of safeguards are fine, but I think it's an overkill for Magit to try to validate all the output it gets from Git.  In this case, Git's getting confused by a corrupt state (combination of the remote's `HEAD` pointing to a non-existent branch and it having a `refs/heads/HEAD`) and displaying a `refs/remotes/$remote/HEAD` value as `$remote`, which it should only do for the remote's HEAD symref.

That's not to say I don't think we should prevent this error now that we know a state that can trigger it.  However, I think 1) the commit message should give a clearer description of what sort of state this is addressing and 2) there should be a code comment.

Instead of your guard, another option would be a guard like this, but I don't have a strong preference (and think both are unclear enough to warrant a code comment):

```
(if head-branch
    (progn (cl-assert
            (equal ref (concat "refs/remotes/" remote "/HEAD")))
           (setq head head-branch))
  ...)
```

to something like

```
(cond (head-branch
       (progn (cl-assert
               (equal ref (concat "refs/remotes/" remote "/HEAD")))
              (setq head head-branch)))
      ;; COMMENT
      ((not (string-suffix-p "/HEAD" ref))
       ...))
```

kyleam:
I said:

> But I don't have a strong preference (and think both are unclear enough to warrant a code comment).

On second thought, a guard that removes the entry entirely (like the one in my comment) is preferable to this PR's current guard because it avoids displaying this misconfigured ref (along the lines of what Jonas mentioned above in terms of not displaying the mistake on the remote end).

kyleam:
I said:

> a guard that removes the entry entirely (like the one in my comment) is preferable

I pushed a commit with that approach to [km/skip-bad-remote-head](https://github.com/magit/magit/compare/km/skip-bad-remote-head).  I'll plan to merge that in a couple days if there are no objections.

By the way, here are the two simplest cases I've come up with to create a remote that triggers this error, but they (and the original reproducer here) boil down to the same end result of a remote with 1) an ambiguous `refs/heads/HEAD` ref **and** 2) `HEAD` pointing to a non-existent ref.

<details>
<summary>cases</summary>

```sh
#!/bin/sh

set -eu

tdir=$(mktemp -d "${TMPDIR:-/tmp}"/magit-5092-XXXXXXX)

cd "$tdir"

git init "a"
(
    cd a
    git commit --allow-empty -m init
    branch=$(git symbolic-ref --short HEAD)
    git update-ref refs/heads/HEAD "$branch"
    git update-ref -d refs/heads/"$branch"
)
git clone --no-local a b
```

```sh
#!/bin/sh

set -eu

tdir=$(mktemp -d "${TMPDIR:-/tmp}"/magit-5092-XXXXXXX)

cd "$tdir"

git init "a"
(
    cd a
    git commit --allow-empty -m init
    branch=$(git symbolic-ref --short HEAD)
    git update-ref refs/heads/HEAD "$branch"
    git checkout --orphan other
)
git clone --no-local a b
```

</details>

kyleam:
> I pushed a commit with that approach to [km/skip-bad-remote-head](https://github.com/magit/magit/compare/km/skip-bad-remote-head). I'll plan to merge that in a couple days if there are no objections.

Pushed as fc485bae.

PR #5097, opened 2024-02-19
git-commit: Add handling for the cut_line constant

adamroyjones:
I'm sorry if this is in any way botched. I'm very far from being a wizard of elisp.

---

With commit.verbose set to true, a special line will be added to the commit message. All content below this will be ignored by git. This line is defined [as cut_line in wt-status.c](https://github.com/git/git/blob/3e0d3cd5c7def4808247caf168e17f2bbf47892b/wt-status.c#L38-L39). To be meta: git prepares the following buffer for this branch's commit:

```
git-commit: Add handling for the cut_line constant

With commit.verbose set to true, a special line will be added to the
commit message. All content below this will be ignored by git.

This line is defined as cut_line in wt-status.c:

https://github.com/git/git/blob/3e0d3cd5c7def4808247caf168e17f2bbf47892b/wt-status.c#L38-L39

I noticed that, for exceptionally large commits, magit was triggering
a commensurately large flyspell operation. It appears to be a small
oversight in the logic of git-commit-turn-on-flyspell.

This commit attempts to add handling for the cut_line constant to fix
this.

# Please enter the commit message for your changes. Lines starting
# with '#' will be ignored, and an empty message aborts the commit.
#
# On branch snippety-snip
# Changes to be committed:
#	modified:   lisp/git-commit.el
#
# ------------------------ >8 ------------------------
# Do not modify or remove the line above.
# Everything below it will be ignored.
diff --git a/lisp/git-commit.el b/lisp/git-commit.el
index 440bc03b..8c8bf41c 100644
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -673,6 +673,12 @@ turning on `orglink-mode'."
     (setq-local orglink-match-anywhere t)
     (orglink-mode 1)))
 
+(defvar git-commit-cut-line "------------------------ >8 ------------------------"
+  "A constant string (cut_line) defined in git/wt-status.c.
+
+This string demarcates text that will be ignored by git.  This string
+will appear when commit.verbose is set to true.")
+
 (defun git-commit-turn-on-flyspell ()
   "Unconditionally turn on Flyspell mode.
 Also prevent comments from being checked and
@@ -682,9 +688,15 @@ finally check current non-comment text."
   (setq flyspell-generic-check-word-predicate
         #'git-commit-flyspell-verify)
   (let ((end)
+        (cut-line-regex (format "^%s %s$" comment-start git-commit-cut-line))
         (comment-start-regex (format "^\\(%s\\|$\\)" comment-start)))
     (save-excursion
-      (goto-char (point-max))
+      (goto-char (point-min))
+      (while (not (looking-at cut-line-regex))
+        (forward-line))
+      (setq end (point)))
+    (save-excursion
+      (goto-char end)
       (while (and (not (bobp)) (looking-at comment-start-regex))
         (forward-line -1))
       (unless (looking-at comment-start-regex)
```

I noticed that, for exceptionally large commits, magit was triggering a commensurately large flyspell operation. It appears to be a small oversight in the logic of git-commit-turn-on-flyspell.

This branch attempts to add handling for the cut_line constant to fix this.

### DIFF
--- a/lisp/git-commit.el
+++ b/lisp/git-commit.el
@@ -675,16 +675,21 @@ turning on `orglink-mode'."
 
 (defun git-commit-turn-on-flyspell ()
   "Unconditionally turn on Flyspell mode.
-Also prevent comments from being checked and
-finally check current non-comment text."
+Also check text that is already in the buffer, while avoiding to check
+most text that Git will strip from the final message, such as the last
+comment and anything below the cut line (\"--- >8 ---\")."
   (require 'flyspell)
   (turn-on-flyspell)
   (setq flyspell-generic-check-word-predicate
         #'git-commit-flyspell-verify)
-  (let ((end)
+  (let ((end nil)
+        ;; The "cut line" is defined in "git/wt-status.c".  It appears
+        ;; in the commit message when `commit.verbose' is set to true.
+        (cut-line-regex (format "^%s -\\{8,\\} >8 -\\{8,\\}$" comment-start))
         (comment-start-regex (format "^\\(%s\\|$\\)" comment-start)))
     (save-excursion
-      (goto-char (point-max))
+      (goto-char (or (re-search-forward cut-line-regex nil t)
+                     (point-max)))
       (while (and (not (bobp)) (looking-at comment-start-regex))
         (forward-line -1))
       (unless (looking-at comment-start-regex)


tarsius:
Thanks!

Your description only says that this "fixes it" [the performance issue], but you should also mention that what this actually does is instruct flyspell to ignore all text after the cut-line.

For comments it makes sense to loop over lines in the buffers (there may be non-comments in between the comments). For finding the one and only cut-line, a single `re-search-backward` is more appropriate.

adamroyjones:
Yes, true on all points—I've updated things now. I've also tried to make the commit message's formatting hew more closely to that of the history (including LaTeX-like single-quotes, double spaces between sentences, etc.).

 Thanks for the guidance! Please let me know if there are other improvements.

lisp/git-commit.el
```diff
@@ -673,18 +673,27 @@ turning on `orglink-mode'."
     (setq-local orglink-match-anywhere t)
     (orglink-mode 1)))
 
+(defvar git-commit-cut-line "------------------------ >8 ------------------------"
+  "A constant string (cut_line) defined in git/wt-status.c.
+
+This string demarcates text that will be ignored by git.  This string
+will appear when commit.verbose is set to true.")
```
tarsius, line 680: The "demarcate" sentence is the important part, which should be placed on the first line. The "cut_line" part is an implementation detail, which should come last, or even be left out completely or be put in a comment. Also, since it is a constant, it should be defined using `defconst`.

However, since the exact number of dashes *may* change[^1], we should play it safe and don't hardcode the same number, and instead just assume a reasonable minimal number. If we do that, the string gets shorter and we don't have to define it separately anymore. (Of course one may argue that it's still a constant that should be defined explicitly, but IMO that would be overkill here.) `(format "^%s -\\{8,\\} >8 -\\{8,\\}$" comment-start)` seems reasonable to me.

[^1]: Though probably won't, provided the git maintainers realize that some tools might hardcode the number and decide to play it safe.

lisp/git-commit.el
```diff
@@ -673,18 +673,27 @@ turning on `orglink-mode'."
...
+Also prevent anything after git's cut line or commented-out from being
+checked and finally check current non-comment text."
   (require 'flyspell)
   (turn-on-flyspell)
   (setq flyspell-generic-check-word-predicate
         #'git-commit-flyspell-verify)
   (let ((end)
-        (comment-start-regex (format "^\\(%s\\|$\\)" comment-start)))
+         (cut-line-regex (format "^%s %s$" comment-start git-commit-cut-line))
+         (comment-start-regex (format "^\\(%s\\|$\\)" comment-start)))
```
tarsius, line 692: This breaks indentation.

lisp/git-commit.el
```diff
@@ -673,18 +673,27 @@ turning on `orglink-mode'."
...
     (save-excursion
-      (goto-char (point-max))
+      (setq end (re-search-forward cut-line-regex nil t))
+      (if (not end) (setq end (point-max)))
```
tarsius, line 695: This can be simplified as `(setq end (or (research-forward cut-line-regex nil t) (point-max)))`.

tarsius (changes requested):

adamroyjones:
Thanks for being patient with me: it must be frustrating to shepherd things from an elisp novice. Hopefully any future pull requests of mine will be a bit better.

I've incorporated the suggestions and folded everything into the commit. I've slightly reworded the documentation string for `git-commit-turn-on-flyspell`. There's a bit of inconsistency in spacing (e.g. `git-commit-check-style-conventions` has a double line feed after the opening sentence), but I've not tried to harmonise things.

tarsius:
> Thanks for being patient with me: it must be frustrating to shepherd things from an elisp novice.

It can be, but this isn't one of those instances. I've pushed a fixup commit, does that look good to you too?

adamroyjones:
Yep, it looks good to me! Thanks again.

PR #5098, opened 2024-02-20
Check current buffer in magit-section-post-command-hook

LemonBreezes:
This commit addresses an issue where if you run a command like `persp-switch` which uses `set-window-buffer`, the current buffer is not updated which can cause magit-post-command-hook to emit an error.

### DIFF
--- a/lisp/magit-section.el
+++ b/lisp/magit-section.el
@@ -1635,10 +1635,14 @@ evaluated its BODY.  Admittedly that's a bit of a hack."
   (setq magit-section-pre-command-section (magit-current-section)))
 
 (defun magit-section-post-command-hook ()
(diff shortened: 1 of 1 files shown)


tarsius:
I think this might only affect the menu related steps. Please give my fixup commit a try.

tarsius:
I would like to investigate this some more[^1], but since I don't have the time for that now and this commit does the job, I am merging it. Thanks, by the way. I think I ran into this myself before, but couldn't [/didn't try to] figure out what was going on.

[^1]: We might be able to further reduce the cope of the condition and/or we should move that to another function altogether. Maybe something else, well see -- at a later time. :grinning:

PR #5102, opened 2024-02-28
magit-tag-release: also read message for --sign/--local-user

corecode:
git tag requires a message for all "heavy" tags that have an associated tag object.  --annotate, --sign, and --local-user= all create tag objects.  Make magit-tag-release read a message in all of these cases.

### DIFF
--- a/lisp/magit-tag.el
+++ b/lisp/magit-tag.el
@@ -192,7 +192,9 @@ like \"/path/to/foo-bar\"."
                     (match-string 2 tag)))
           (args (magit-tag-arguments)))
        (list tag
-             (and (member "--annotate" args)
+             (and (or (member "--annotate" args)
+                      (member "--sign" args)
+                      (seq-some (lambda (x) (string-prefix-p "--local-user" x)) args))
                   (read-string
                    (format "Message for %S: " tag)
                    (cond ((and pver (string-match (regexp-quote pver) pmsg))


tarsius:
> git tag requires a message for all "heavy" tags that have an associated tag object. --annotate, --sign, and --local-user= all create tag objects.

If `git tag` is called with one of these arguments but without `-m`, then it invokes `$EDITOR` for the user to write the tag message. As for commit messages, this is convenient when we want to provide more text than fits on a single line.

- If you invoke `magit-tag-release` with one of the sign arguments (but without `--annotate`), then a buffer pops up.

- If you invoke `magit-tag-release` with `--annotate` (regardless of whether a signing argument is also used), then the user is instead required to provide the message in the minibuffer.

This difference in behavior is intentional; the user can decide *how* they want to provide the message. However, I also believe I only considered the case where the user signs tags ("because tags should always be signed"). I.e., it is currently not possible to *not* sign the tag and *also* use the minibuffer to write the message.

One solution that comes to mind, is to change `magit-tag-release` so that it does *not* use the minibuffer if `--annotate` is enabled. (Instead of, as you propose, to extend that special handling of that argument to the signing arguments as you propose.)

The special handling would not be *removed*, instead the `-m` argument should be added to the menu, and if *that* is enabled, then `magit-tag-release` would read a message in the minibuffer (instead of using a proper buffer, or not adding an annotation at all, depending on the other arguments). 

- Users could then enable `-m`, if they want to use the minibuffer.
- If they want to use a proper buffer, they could instead enable `--annotate` (or a signing argument).
- If they don't want an annotation, then they should keep all arguments disabled.

We would of course also have to consider how that affects `magit-tag-create`.

corecode:
`magit-tag-release` currently uses `magit-run-git-async` and doesn't set EDITOR, that's why I am submitting this PR.

I figured I would retain the spirit of `magit-tag-release` and read from the minibuf.  It felt like `magit-tag-create` is the function that will do the whole process with editor popup and no prepared tag/message, while `magit-tag-release` optionally prompts for a message.  Both --sign and --local-user imply --annotate, so it feels like `magit-tag-release` should act accordingly.

tarsius:
> `magit-tag-release` currently uses `magit-run-git-async` and doesn't set EDITOR,

I've fixed that.

corecode:
Do you intend `magit-tag-release` to act like `magit-tag-create` if `--sign` is used (but not `--annotate`)?  I did expect it to act as if `--annotate` was selected (because `--sign` does imply it).

tarsius:
I intend to make these commands consistent with each other, and with `git tag` (if appropriate). The topic isn't in my working memory right now, so I don't know what that entails. I'll work on it once I don't have much more important matters to attend to.

tarsius:
I've made chances to increase consistency and separation of concerns. See 2da34f1317.

PR #5112, opened 2024-03-18
* lisp/magit-section.el: Allow several functions as arguments

DamienCassou:
-------

The function `magit-disable-section-inserter` is convenient in a
.dir-locals.el file to disable a section per repository. When more
than one section should be disabled, this function can be called
several times. This commit makes disabling several sections easier by
allowing more than one section as argument.

### DIFF
--- a/lisp/magit-section.el
+++ b/lisp/magit-section.el
@@ -2179,10 +2179,15 @@ again use `remove-hook'."
 
 (defun magit-disable-section-inserter (fn)
   "Disable the section inserter FN in the current repository.
+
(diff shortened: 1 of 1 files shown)


tarsius:
I'm not convinced this is easier (copying one line and adjusting the new copy, isn't much harder than moving to the correct place in the only line).

`magit-disable-section-inserter` can be though of as a variation of `remove-hook`, which also has to be called once per hook function.

tarsius:
I might come back to this at some point, but for now I am closing.

PR #5123, opened 2024-04-09
Specify the shade of "green"

mavit:
This works around Emacs bug 66538 (https://debbugs.gnu.org/cgi/bugreport.cgi?bug=66538).

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -278,7 +278,7 @@ Also see https://github.com/magit/magit/issues/4720
 and https://debbugs.gnu.org/cgi/bugreport.cgi?bug=62093.")
 
 (defface magit-process-ok
-  '((t :inherit magit-section-heading :foreground "green"))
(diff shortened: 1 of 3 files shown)


tarsius:
Before doing anything I would want to see how this is handled upstream.

Some thoughts, assuming that things remain as they are:

1. It's not like I picked this specific shade of green, because it is better than others, but because *some* green makes sense to use. Depending on the theme, different shades would be more appropriate than others. The idea is that theme authors would pick the appropriate shade.
2. IMO even with a dark background (I tried #000000 and #333333), both shades look good/readable enough.

In other words, I tend to think, that things should remain as they are. "green" communicates better, that some shade of green should be used, than "#00ff00" does.

tarsius:
> Before doing anything I would want to see how this is handled upstream.

Well, I don't think there's really anything they can do about it -- retroactively.

> In other words, I tend to think, that things should remain as they are.

I've decided to go with that.

PR #5130, opened 2024-04-28
Cleanup expansion of magit-insert-section

tarsius:
Since forever now, `magit-insert-section` expanded to something huge. I'm finally fixing that with this pull-request.  The expansion now offloads most of the work to support functions. This is a change that is *looong* overdue. One reason for delaying it until now, was that it is likely to cause issues for many users.

Due to limitations of the package managers we use, such drastic changes are unfortunately problematic.  At least `package.el` does not use a clean environment to build packages and does not rebuild packages when their dependencies change.

- When updating packages that are already loaded in the current Emacs instance, then the old outdated code may leak into the byte-compilation of the new version. It may also leak into third-party packages. This is particularly likely when the update involves changes to macros or eieio classes; both of which is the case here.

- When updating packages, which provide macros used by other packages, and those updates involve changes to those macros, this does not cause the dependent packages to be recompiled. The macro expansions that were created using the old version of the dependency may then try to use the dependency in a way that is no longer supported by the new version.

- When this happens we end up with franken versions, which exhibit "impossible" behavior. That's because the byte-code contains combinations of code that never existed in that form under version control, and bisecting thus won't work. The exact faulty behavior depends on from what version the user updated to what version (times the number of involved packages). When the maintainer is not aware of these issues, this can lead to very frustrating debugging sessions.

These issues apply in this case. This will likely result in an error like

```
Invalid slot name: "#<magit-section ...>", :value
```

but it is possible you will see different errors.

If you encounter problems due to this update as a user, then you have to reinstall the involved packages, while ensuring you do so in a clean environment:

1. Uninstall `magit`, `magit-section` and all packages that depend on them.
2. Exit Emacs.
3. Restart Emacs.
4. Reinstall the packages.

If you maintain a package that depends on `magit` or `magit-section`, then I recommend that you bump the dependency for Melpa.  Use the snapshot "version" `"20240428"` (and then remember to change that before creating a release). [Speaking of releases, I plan to create new releases for Magit, and everything else, in May.  So you won't have to remember this for too long. ;P ] This *reduces* the likelihood of your users encountering issues, but unfortunately isn't enough to ensure it won't happen, in which case you can point them here (but please also take part in helping users).

Cheers,
Jonas


flomnes:
I applied the following instructions
>     Uninstall magit, magit-section and all packages that depend on them.
>     Exit Emacs.
>     Restart Emacs.
>     Reinstall the packages.
It worked, thanks !

PR #5137, opened 2024-05-08
One minor fix for misplaced space and quote

wtianyi:
I believe the quote should be after space and directly follow the `seq` symbol.


tarsius:
Yes, of course, and thanks for letting me know.

Funnily enough, someone else spotted and fixed (#5135) another instance of this typo. This time around I have checked whether there are other instances and fixed them all (855d924969eb6231d62ab446f727c9f7ad5c4913).

PR #5141, opened 2024-05-13
git-commit.el: Fix typo

ideapark:
Signed-off-by: Park Zhou <ideapark@petalmail.com>


tarsius:
Thanks again!

So it appears you are reading the source and are running into typos along the way. That isn't unexpected and I appreciative the fixes. I did not merge your previous pr as quickly as the first, because I expected that you would keep providing more fixes as you go, hoping you would just push additional commits to the open pull-request. Please do that going forward. Fixing these mistakes is nice but it is not important, that each fix hits the main branch as quickly as possible, and it is less work for both of us, if we don't have to deal with a pull-request per typo.

tarsius:
Some notes about commit messages:
- We don't use "signed-off-by" or similar trailers. I usually leave them in place if a contributor uses them, but when I also make some change to the patch or commit message, that makes me a bit uneasy, because depending on one's definition of what is being signed off, that makes that statement untrue.
- We only use "some-file: Do things" for changes that concern that file as a whole, like "foo.el: New library".
- I am a bit inconsistent when it comes to commit message about typo fixes, making harder for contributors to pick up on a style. But here's some examples I would use.
  - some-function: Fix docstring typos
  - manual: Fix typos
  - docs: Fix typos [When typos are fixed in both docstrings and the manual]
  - Fix docstring typos [Multiple unrelated typos in multiple more or less related docstrings]

ideapark:
Thanks for your explaination, I will follow your instructions.

PR #5151, opened 2024-06-05
Docs clarity copyedits

tsobczynski:
I put on a "copy editor" hat and made some revisions to the prose of the manual. There are two commits here, first to consistently capitalize "Git" as a proper noun where previously missed, and second to improve the clarity of the description of the variable "magit-git-debug".

My edits were to the Org source file (magit.org), and I used Make to update the derived TeX (magit.texi).


tarsius:
I don't think its an improvement.

PR #5157, opened 2024-06-20
Report detailed hook runtime

rcj:
When `magit-refresh-verbose' is set, execution times of different magit sections and hooks is written to the *Messages* buffer. For hooks only the total runtime of all functions in the hooks is reported.

In order to identify costly hooks, reporting the runtime for each function is helpful. By wrapping each function of a hook, we can report individual run times.

I'm working with some very large repositories and magit sometimes takes several seconds to display the status buffer. `magit-refresh-verbose' already helps with identifying individual sections which are expensive and might be candidates for removal from the status buffer for these repositories. For the different magit hooks only the total runtime of the hook is shown, which makes it harder to identify which function of a hook is actually contributing how much to the total runtime and requires optimization or removal.

### DIFF
--- a/lisp/magit-mode.el
+++ b/lisp/magit-mode.el
@@ -1455,10 +1455,21 @@ The additional output can be found in the *Messages* buffer."
 (defun magit-run-hook-with-benchmark (hook)
   (when hook
     (if magit-refresh-verbose
-        (let ((start (current-time)))
-          (message "Running %s..." hook)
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks!

I've done it with less code in 72e6631ca24d078c97603a26c1e07dda8fd5e108.

rcj:
Even better, thanks!

PR #5162, opened 2024-06-28
Add --force to magit-pull and magit-fetch

prati0100:
git pull and git fetch allow the --force option to force update local branches. From man git-fetch(1):

    -f, --force
        When git fetch is used with <src>:<dst> refspec it may refuse to
        update the local branch as discussed in the <refspec> part
        below. This option overrides that check.

Add that option to magit-pull and magit-fetch. Use -F as the suffix to be consistent with magit-push.

### DIFF
--- a/lisp/magit-fetch.el
+++ b/lisp/magit-fetch.el
@@ -37,7 +37,8 @@
   ["Arguments"
    ("-p" "Prune deleted branches" ("-p" "--prune"))
    ("-t" "Fetch all tags" ("-t" "--tags"))
-   (7 "-u" "Fetch full history" "--unshallow")]
+   (7 "-u" "Fetch full history" "--unshallow")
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks!

My initial impulse was to suggest that you add that locally instead, but then I remembered that I recently failed to update tags locally after an upstream had rewritten history, which was caused by (from git-fetch(1)):

           Until Git version 2.20, and unlike when pushing with git-push(1),
           any updates to refs/tags/* would be accepted without + in the
           refspec (or --force). When fetching, we promiscuously considered
           all tag updates from a remote to be forced fetches. Since Git
           version 2.20, fetching to update refs/tags/* works the same way as
           when pushing. I.e. any updates will be rejected without + in the
           refspec (or --force).

, a change in behavior I wasn't aware of.

We should be able to fetch updated tags without having to drop to the command line.

PR #5169, opened 2024-07-10
autoload magit--handle-bookmark and require magit-bookmark

jdtsmith:
Loading magit bookmarks from a new emacs session uses
`magit--handle-bookmark`.  This provides an autoload to ensure this
works even prior to loading magit.  `magit-bookmark` is also required,
as it implements the necessary generic methods.

See Discussion #5168.


tarsius:
Thanks!

In case you are wondering why I changed it to `(quote magit-bookmark)`: that hides it from the tools I use to maintain the Emacsmirror, so `epkg-describe` package won't claim there is a circular dependency.

PR #5170, opened 2024-07-11
magit-dir-checkout: new function to reset dir

danielfleischer:
- Reset a directory to a given commit.
- Reading the commit and directory from users using existing functions. 
- Adding new binding `X d`.
- Mention in the Org documentation.


tarsius:
Thanks for the suggestion.

Seems like overkill to add a whole new command for this. My initial impulse was to add a prefix argument, to also offer directories. In the end I decided `magit-file-checkout` should just always to so.

danielfleischer:
Thanks, that's great. I would add something to the docs, for discoverability.

PR #5171, opened 2024-07-11
magit--straight-chase-links: replace with magit--chase-links

progfolio:
Refactor function to work with straight.el and Elpaca.

magit-version: use magit--chase-links


tarsius:
Thanks!

I made some cosmetic changes, mostly because of line length. (I would have pushed here before merging, but it appears you haven't given me permission to do so.)

PR #4428, opened 2021-07-05
magit-push-current-to-{upstream,pushremote}: Escape remote and branch names

Nuky:
The branches and remote names are passed unescaped to `magit-confirm` from
`magit-push-current-to-{upstream,pushremote}`.  `magit-confirm` uses its
prompt as a format, which may cause an "Invalid format operation error" when
a branch or remote name contains a "%" character.

--

Magit 3.1.1, Git 2.32.0, Emacs 27.2, darwin

To reproduce on magit-push-current-to-upstream:

1. create branch: `b s` `test%branch`
2. try to push branch, creating upstream one: `P -n u` `origin/test%branch`

Similarly for magit-push-current-to-pushremote:

1. create remote: `M a` `test%remote` `https://localhost/` `n`
2. try to push branch, setting its pushremote: `P -n p` `test%remote`

 
Backtrace (`magit-push-current-to-upstream`):

```
Debugger entered--Lisp error: (error "Invalid format operation %b")
  format("Really use \"origin/test%branch\" as upstream and push \"test%branch\" there? " 0)
  (setq prompt-n (format (concat (or prompt-n prompt) "? ") (length items)))
  (progn (setq prompt-n (format (concat (or prompt-n prompt) "? ") (length items))) (setq prompt (format (concat (or prompt (magit-confirm-make-prompt action)) "? ") (car items))) (or (cond ((and (not (eq action t)) (or (eq magit-no-confirm t) (memq action magit-no-confirm) (cl-member-if #'... magit--no-confirm-alist))) (or (not sitems) items)) ((not sitems) (magit-y-or-n-p prompt action)) ((= (length items) 1) (and (magit-y-or-n-p prompt action) items)) ((> (length items) 1) (and (magit-y-or-n-p (concat (mapconcat ... items "\n") "\n\n" prompt-n) action) items))) (if noabort nil (user-error "Abort"))))
[...]

PR #5178, opened 2024-07-21
Percent escape magit-confirm prompts

Nuky:
`magit-branch-delete` fails with `Invalid format operation error` when the branch name contains a `%`. I had encountered this same issue when pushing branches back in #4428. The issue is that the `%` ends up in the prompt passed to `magit-confirm` which it uses as a format.

This PR fixes `magit-branch-delete` specifically, and also some (all?) other problematic calls to `magit-confirm`.

Repro steps for `magit-branch-delete`:
1. Create a branch called `foo%target=dev`
2. Push the branch to `origin`
3. `b k` (`magit-branch-delete`) `origin/foo%target=dev`

(Alternatively you can use a remote with a `%` in the name to trigger the same problem)

### DIFF
--- a/lisp/magit-apply.el
+++ b/lisp/magit-apply.el
@@ -548,9 +548,11 @@ of a side, then keep that side without prompting."
       (funcall apply section "--reverse" "--index"))))
 
 (defun magit-discard-hunks (sections)
-  (magit-confirm 'discard (format "Discard %s hunks from %s"
+  (magit-confirm 'discard (format "Discard %d hunks from %s"
                                   (length sections)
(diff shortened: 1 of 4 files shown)


tarsius:
Thanks!

I didn't want to complicate so many callers, so I've moved the quoting inside `magit-confirm` itself.  See ad68015aa1f7cab19eae0d86983193fa2b7038ca.

Nuky:
That's much better than my amateur attempts at writing elisp!

Thanks for all your work on magit all these years.

tarsius:
You're welcome!

PR #5216, opened 2024-09-01
Remove magit-section from Package-Requires

kohnish:
It's already provided and it is causing unnecessary installation from ELPA

### DIFF
--- a/lisp/magit.el
+++ b/lisp/magit.el
@@ -22,7 +22,6 @@
 ;;     (emacs "26.1")
 ;;     (compat "30.0.0.0")
 ;;     (dash "2.19.1")
-;;     (magit-section "4.1.0")
 ;;     (seq "2.24")
 ;;     (transient "0.7.5")
 ;;     (with-editor "3.4.2"))


tarsius:
`magit-section.el` is intentionally distributed as a separate package `magit-section`. This file is **not** included in the `magit` package, which therefore must depend on the `magit-section` package.

kohnish:
For package-vc-install users, it is a part of the package. It would be great if you make it not part of the repo as well.

tarsius:
> It would be great if you make it not part of the repo as well.

`magit-section.el` never existed outside the `magit/magit` repository. It has been like this since before `package-vc-install` was created.

> For package-vc-install users, it is a part of the package.

`package-vc-install` indeed operates under the assumption that the elisp libraries in a repository are always distributed as a single package. That assumption is not correct, see https://emacsmirror.net/stats/compare.html#org59a72e1 for a list of repositories/packages for which this assumption is not correct (that link will break, if it already has, search for "Shared repositories" on that page).

That doesn't mean that I think that what `package-vc-install` does is wrong; just that there is a mismatch. There is nothing I can do about this. It is unfortunate that this means that `magit-section` gets needlessly installed for `package-vc-install` users, but at least that doesn't *break* anything for the (I believe) not all that many users of that. On the other hand, removing the `magit-section` dependency as you propose, would *break* Magit for the several magnitude more users of plain old `package-install`.

tarsius:
> So "why is magit-section.el part of magit/magit?" you might ask.

Because `magit-section.el` was written to be used by Magit. It would be highly inconvenient to maintain them in two separate repositories.

> So "why is magit-section.el distributed as a separate package?" you might ask.

Because other package authors urged me to do that, so that they could add similar interfaces in their own (not git-related) packages without having to depend on Magit.

kohnish:
Yeah, I understand, thanks for the explanation.  
One could make a wish. Me not knowing how stable the API is and why those dependent packages need to keep depending on the latest version of magit-section, now my wish is that magit just duplicates it in its own repo or so.  

I'm sure every user of magit-section uses magit as well, magit is that great.  
It would be quite benefitial for users who don't want to experience breakages due to some inconsistency in the repo or simply ELPA going down, but github is up.  
For now I could kinda work around by setting `(setopt package-archives nil)`.

Issue #5212, opened 2024-08-30
`magit-commit` says "nothing staged" when git submodule changes are the only staged changes


dhull:
When attempting to create a commit where the only changes are updates to git submodule, `magit-commit` gives the error "user-error: Nothing staged (or unstaged)". This has been broken for a couple of weeks now; before that I was able to create a commit where the only change was an update to a submodule.

Magit version:
Magit 20240822.1737 [>= 4.0.0], Transient 0.7.4, Git 2.46.0, Emacs 29.1, darwin

(I tried running `emacs -Q --debug-init --eval '(setq debug-on-error t)' -L ../dash-20240510.1327 -L ../git-commit-20240811.1419 -L ../transient-20240821.158 -L ../with-editor-20240822.2051 -L . -l magit` but it failed with the error `Lisp error: (file-missing "Cannot open load file" "No such file or directory" "compat")`. I did run the "magit-emacs-Q-command" and it failed with same "user-error: Nothing staged (or unstaged)" error.)

dhull:
I just upgraded to magit-20240830.1753 and it has the same error.

tarsius:
I cannot reproduce that and the involved functions were not recently modified.

The check is performed in `magit-commit-assert`; you might want to insert debug statements there. `(magit-anything-staged-p)` is supposed to return t preventing the path to the mentioned error being taken. Check what it returns for you. Without any arguments, it runs `git diff --quiet --cached --`, which is expected to exit with 1 in this case. Try that in a shell.

dhull:
Thanks so much for your help. It seems like there is a bug in the git diff command involving the `--submodule=log` option (which I have configured in my `~/.gitconfig` file.

I have git 2.46.0 installed via homebrew on my Mac.
```
% brew info git
==> git: stable 2.46.0 (bottled), HEAD
Distributed revision control system
https://git-scm.com
Installed
/usr/local/Cellar/git/2.46.0 (1,678 files, 51.3MB) *
  Poured from bottle using the formulae.brew.sh API on 2024-07-30 at 07:57:59
...
```

I ran the `git diff` command you suggested above, replacing `--quiet` with `--exit-code` so I can see git's output.

First is from the system git, which produces the correct exit codes.
```
% /usr/bin/git --version
git version 2.39.3 (Apple Git-146)

% /usr/bin/git diff --exit-code --cached --submodule=short --; echo $?
diff --git a/fbt-modules/fbt-type b/fbt-modules/fbt-type
[...]

PR #5221, opened 2024-09-06
Fix staged rename detection...

jolmg:
Staging only a file rename and trying to commit caused Magit to report "Nothing staged." This fixes that by turning off rename detection in git-diff.

### DIFF
--- a/lisp/magit-git.el
+++ b/lisp/magit-git.el
@@ -1249,7 +1249,7 @@ Sorted from longest to shortest CYGWIN name."
   "Return t if there are any staged changes.
 If optional FILES is non-nil, then only changes to those files
 are considered."
-  (magit-git-failure "diff" "--quiet" "--cached"
+  (magit-git-failure "diff" "--quiet" "--cached" "--no-renames"
                      (and ignore-submodules "--ignore-submodules")
                      "--" files))
 


tarsius:
I looked at this diff and my first thought was "wouldn't that achieve the exact opposite"? The second "... if anything".

I then tried to commit a rename with the latest Magit version and it worked. Are you sure you are using the latest version?

If you can reproduce the issue with the latest version, then please provide detailed steps to reproduce. Most importantly, how do you "Stag[e] only a file rename"?

jolmg:
> Are you sure you are using the latest version?

v4.1.0, which is just a couple of commits behind main.

> If you can reproduce the issue with the latest version, then please provide detailed steps to reproduce. Most importantly, how do you "Stag[e] only a file rename"?

```sh
$ git --version
git version 2.46.0
$ mkcd /tmp/git-diff-test 
$ git init 
Initialized empty Git repository in /tmp/git-diff-test/.git/
$ touch foo 
$ git add foo 
$ git commit -m "foo"
[main (root-commit) f60d7bc] foo
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 foo
$ mv foo bar 
$ git status -s
 D foo
?? bar
$ git add foo bar 
$ git status -s
R  foo -> bar
$ git diff --quiet --cached; echo $?             
0
$ git diff --quiet --cached --no-renames; echo $? 
1
```

> I looked at this diff and my first thought was "wouldn't that achieve the exact opposite"?

If git detects the rename, git-diff returns no difference because foo and bar have the same contents. That results in magit not detecting the change and therefore not detecting something was staged.

If git doesn't detect the rename, git-diff returns a difference in the form of a new file and a deleted file. That results in magit detecting the change and therefore detecting something was staged.

jolmg:
Here's a quick little video. I unstage both (fumbling a bit), then stage both the new untracked `bar` and the deleted `foo`, and try to commit. I'm given the message "Nothing staged (or unstaged)". Then I add the argument `--no-renames`, eval, retry committing, and then I'm given the window for the commit message.

https://github.com/user-attachments/assets/560053a5-b666-448c-aee0-74daf61765c8

tarsius:
I'm still on an older release and with that both diff variants have the same exit status:

```
$ git version
git version 2.45.2
...
$ git diff --quiet --cached; echo $?
1
$ git diff --quiet --cached --no-renames; echo $?
1
```

I had a quick look at the release notes and didn't spot anything related to this. This seems to be a regression (possibly introduced with the `--ext-diff` changes).

Could you please inquire about this on the git mailing-list?

jolmg:
https://lore.kernel.org/git/0241cbcc-11b3-483a-8a63-d523291b92ff@app.fastmail.com/T/#u

tarsius:
Thanks!

I see the issue has been acknowledged.

The "submodule diff regression" refers to #5212.

tarsius:
I'm not merging this because it is a transient bug, which only affects an edge-case, and don't want to risk any issues caused by the kludge.

PR #5226, opened 2024-09-16
Avoid activating Eglot for buffers visiting blobs

pengjiz:
Hi!

Eglot may be activated for buffers visiting blobs, and most LSP servers bail out in that case. So in this commit we remove the activation function from the relevant hook.

Note that when started, Eglot checks and attempts to manage existing buffers as well. However, at that point, buffers visiting blobs have 'buffer-file-name' being nil, and Eglot will ignore them. So we only need to modify the hook instead of advising the function.

This is to achieve the same effect as the advice for 'lsp', introduced in f8353f57.

Thank you!


tarsius:
Thanks!

PR #5228, opened 2024-09-18
Proof of concept for limited magit refresh

CeleritasCelery:
I have been looking into magit performance on large repos, particularly over tramp. In my particular case running `magit-status-refresh-buffer` takes about 6 seconds. That means there is at least a 6 second delay to any command in the status buffer, which is very painful. Here is the verbose output:

```
Refreshing magit...
Running magit-pre-refresh-hook...
  magit-maybe-save-repository-buffers                0.074059
Running magit-pre-refresh-hook...done (0.074s)
Refreshing buffer ‘magit: repo_cluster’...
  magit-insert-error-header                          0.000001 
  magit-insert-diff-filter-header                    0.329797 !!
  magit-insert-head-branch-header                    0.132646 !!
  magit-insert-upstream-branch-header                0.000216 
  magit-insert-push-branch-header                    0.224527 !!
  magit-insert-tags-header                           0.966151 !!
  magit-insert-status-headers                        1.725762 !!
  magit-insert-merge-log                             0.241909 !!
  magit-insert-rebase-sequence                       0.240397 !!
  magit-insert-am-sequence                           0.064967 !!
  magit-insert-sequencer-sequence                    0.205455 !!
  magit-insert-bisect-output                         0.073950 !!
  magit-insert-bisect-rest                           0.000002 
  magit-insert-bisect-log                            0.000001 
  magit-insert-untracked-files                       1.274190 !!
  magit-insert-unstaged-changes                      0.484438 !!
  magit-insert-staged-changes                        0.210606 !!
  magit-insert-stashes                               0.340388 !!
  magit-insert-unpushed-to-pushremote                0.173376 !!
  magit-insert-unpushed-to-upstream-or-recent        0.359119 !!
  magit-insert-unpulled-from-pushremote              0.106821 !!
  magit-insert-unpulled-from-upstream                0.000005 
Refreshing buffer ‘magit: repo_cluster’...done (5.962s)
Running magit-post-refresh-hook...
  magit-auto-revert-buffers                          0.000003
Running magit-post-refresh-hook...done (0.000s)
Refreshing magit...done (6.040s, cached 54/90 (60%))
```

I initially thought that it was some issue with magit that caused these commands take so long, but it turns out this is just a big repo and that these operations take a long time. 

Most operations do not require a refresh of all the information in the status buffer. This proof of concept code enables a "limited refresh" where we only refresh the parts of the buffer that would actually change for a given command. Everything else is read from a cache. For example here is new verbose output with limited refresh when we stage a file:

```
Refreshing magit...
Running magit-pre-refresh-hook...
  magit-maybe-save-repository-buffers                0.000001
Running magit-pre-refresh-hook...done (0.000s)
Refreshing buffer ‘magit: repo_cluster’...
  magit-insert-error-header                          0.000001 
  magit-insert-diff-filter-header                    0.000083 
  magit-insert-head-branch-header                    0.000070 
  magit-insert-upstream-branch-header                0.000253 
  magit-insert-push-branch-header                    0.000066 
  magit-insert-tags-header                           0.000048 
  magit-insert-status-headers                        0.007927 
  magit-insert-merge-log                             0.000009 
  magit-insert-rebase-sequence                       0.000009 
  magit-insert-am-sequence                           0.000002 
  magit-insert-sequencer-sequence                    0.000005 
  magit-insert-bisect-output                         0.000009 
  magit-insert-bisect-rest                           0.000001 
  magit-insert-bisect-log                            0.000001 
  magit-insert-untracked-files                       0.000264 
  magit-insert-unstaged-changes                      0.850395 !!
  magit-insert-staged-changes                        0.214693 !!
  magit-insert-stashes                               0.000285 
  magit-insert-unpushed-to-pushremote                0.000369 
  magit-insert-unpushed-to-upstream-or-recent        0.000613 
  magit-insert-unpulled-from-pushremote              0.000047 
  magit-insert-unpulled-from-upstream                0.000002 
Refreshing buffer ‘magit: repo_cluster’...done (1.080s)
Running magit-post-refresh-hook...
  magit-auto-revert-buffers                          0.000006
Running magit-post-refresh-hook...done (0.000s)
Refreshing magit...done (1.083s, cached 1/4 (25%))
```

So it took the total time from 6 seconds to 1. Likewise, discarding a file now takes 0.25 seconds instead of 6. This makes working with magit so much more efficient and pleasant. 

To facilitate this, I created a new variable `magit-status--limited-refresh` that can be dynamically bound to the set of hooks that we want to run during a "limited refresh". So far I have only implemented this for the commands in `magit-apply` (stage, unstage, discard, etc).

This code is very much a proof of concept and is not ready to be merged. It is just hacked on and not nearly as elegant or well organized as the normal magit code. But I wanted to get feedback and see if this is something you would want to support in magit. It could make a huge difference for users of large repos or repos over tramp.

### DIFF
--- a/lisp/magit-apply.el
+++ b/lisp/magit-apply.el
@@ -149,6 +149,13 @@ so causes the change to be applied to the index as well."
                      (concat (magit-diff-file-header section)
                              (magit-apply--section-content section))))
 
+
+(defun magit-apply--get-refresh ()
+  (pcase (magit-diff-type)
+    ('untracked '(magit-insert-untracked-files magit-insert-staged-changes))
+    ('unstaged '(magit-insert-unstaged-changes magit-insert-staged-changes))
+    ('staged '(magit-insert-untracked-files magit-insert-unstaged-changes magit-insert-staged-changes))))
+
 (defun magit-apply--adjust-hunk-new-starts (hunks)
   "Adjust new line numbers in headers of HUNKS for partial application.
 HUNKS should be a list of ordered, contiguous hunks to be applied
@@ -215,7 +222,10 @@ adjusted as \"@@ -10,6 +10,7 @@\" and \"@@ -18,6 +19,7 @@\"."
               (magit-diff-hunk-region-patch hunk args))))))
 
 (defun magit-apply-patch (section:s args patch)
-  (let* ((files (if (atom section:s)
+  (let* ((magit-status--limited-refresh
+             (or magit-status--limited-refresh
+                 (magit-apply--get-refresh)))
+         (files (if (atom section:s)
                     (list (oref section:s value))
                   (--map (oref it value) section:s)))
          (command (symbol-name this-command))
@@ -354,13 +364,15 @@ ignored) files."
 
 (defun magit-stage-1 (arg &optional files)
   (magit-wip-commit-before-change files " before stage")
-  (magit-run-git "add" arg (if files (cons "--" files) "."))
+  (let ((magit-status--limited-refresh (magit-apply--get-refresh)))
+    (magit-run-git "add" arg (if files (cons "--" files) ".")))
   (when magit-auto-revert-mode
     (mapc #'magit-turn-on-auto-revert-mode-if-desired files))
   (magit-wip-commit-after-apply files " after stage"))
 
 (defun magit-stage-untracked (&optional intent)
-  (let* ((section (magit-current-section))
+  (let* ((magit-status--limited-refresh '(magit-insert-untracked-files magit-insert-staged-changes))
+         (section (magit-current-section))
          (files (pcase (magit-diff-scope)
                   ('file  (list (oref section value)))
                   ('files (magit-region-values nil t))
@@ -387,7 +399,7 @@ ignored) files."
                    (fboundp 'borg--sort-submodule-sections))
           (let* ((topdir (magit-toplevel))
                  (url (let ((default-directory
-                             (file-name-as-directory (expand-file-name repo))))
+                              (file-name-as-directory (expand-file-name repo))))
                         (or (magit-get "remote" (magit-get-some-remote) "url")
                             (concat (file-name-as-directory ".") repo))))
                  (package
@@ -469,9 +481,10 @@ ignored) files."
 
 (defun magit-unstage-1 (files)
   (magit-wip-commit-before-change files " before unstage")
-  (if (magit-no-commit-p)
-      (magit-run-git "rm" "--cached" "--" files)
-    (magit-run-git "reset" "HEAD" "--" files))
+  (let ((magit-status--limited-refresh (magit-apply--get-refresh)))
+    (if (magit-no-commit-p)
+        (magit-run-git "rm" "--cached" "--" files)
+      (magit-run-git "reset" "HEAD" "--" files)))
   (magit-wip-commit-after-apply files " after unstage"))
 
 (defun magit-unstage-intent (files)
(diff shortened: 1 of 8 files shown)


tarsius:
Sorry for not responding earlier.  I noticed that CI failed due to warnings from the compiler and was hoping you would address that.  And then I got busy with other things.

(While fixing the warnings, please also undo the stray whitespace changes.)

I do not intend to add something like this, but that should not discourage you from using it and maybe tweak it over time.

In a way I plan to do something similar, but where you add a cache to make it faster to recreate all sections, I plan to make it possible to only update those sections that actually need to be updated.  So in a way I plan to do the opposite of what you are doing here.  Merging this cache-based approach would make it harder to eventually put my plan into action.

I know I have been planning to do this for a while and still haven't done it, but I am quite confident, that 2025 is the year I finally get to it. Meanwhile you, and anyone stumbling into this, can use your approach.  (I thus recommend that you occasionally rebase it onto "main".)

CeleritasCelery:
> In a way I plan to do something similar, but where you add a cache to make it faster to recreate all sections, I plan to make it possible to only update those sections that actually need to be updated. So in a way I plan to do the opposite of what you are doing here.

I initially tried to take that approach. Essentially using the buffer itself as the cache. However I ran into some difficulty understanding the text properties and section object that magit uses. So I switched to this approach. 

A few observations from using this for a bit:
- which sections needed to be refreshed is a little more complex then I originally thought. For example if you are in the middle of a Cherry Pick and stage some files that can update the cherry pick status as well. 
- changing configuration options like `pull.rebase` will trigger a refresh, even though most of the time it is not needed. However all of those config options use the same function, so there is not an easy way to differentiate which ones need to refresh which sections. 
- the refresh cache (magit’s original version not mine) is only used during refresh, not during other commands. The biggest place this is an issue is with Async commands that update the mode line. The mode-line setup and tear down functions will call `magit-toplevel` about 6 times. This is about 12 extra calls to git which over tramp adds an entire extra second of latency. 
- sometimes when doing a git pull and there are conflicts there can tens of thousands of lines add to the status buffer, which makes it lock up while trying to parse. It would be great to have an option to limit that to say 5000 lines of 5 seconds or something. Because if you `C-g` out of it you don’t get the rest of the buffer and are left in a half finished state. 

I actually think I am going to test another approach that might work better for my workflow. Create a wrapper command just for staging/unstaging that will limit the sections to only untracked, unstaged, and staged, and disable automatic `magit-refresh` via an advice. That will let you do all the operations you want and then manually trigger refresh only once when you are done. This will get rid of the waiting for refresh (even a limited refresh) between each operation. Essentially let you batch commands together.

> I know I have been planning to do this for a while and still haven't done it, but I am quite confident, that 2025 is the year I finally get to it.

That is so exciting to hear! I can’t wait for this improvement!

tarsius:
> I initially tried to take that approach. Essentially using the buffer
> itself as the cache. However I ran into some difficulty understanding
> the text properties and section object that magit uses.

Yes, that has to be carefully redesigned. And doing that has far reaching consequences, which is why I (1) avoided doing it for such a long time, and (2) now that I am more motivated to do it, I still intend to take it on very slowly.

> - the refresh cache (magit’s original version not mine) is only used
>   during refresh, not during other commands. The biggest place this is
>   an issue is with Async commands that update the mode line. The
>   mode-line setup and tear down functions will call `magit-toplevel`
>   about 6 times. This is about 12 extra calls to git which over tramp
>   adds an entire extra second of latency.

That's something we might want to address separately (and earlier).

> - sometimes when doing a git pull and there are conflicts there can
>   tens of thousands of lines add to the status buffer, which makes it
>   lock up while trying to parse. It would be great to have an option
>   to limit that to say 5000 lines of 5 seconds or something. Because
>   if you `C-g` out of it you don’t get the rest of the buffer and are
>   left in a half finished state.

When I run into such a large conflict (which doesn't happen often) I usually just with the "C-g" approach, but yes, that's not particularly satisfying.

> I actually think I am going to test another approach that might work
> better for my workflow.

Just in case you have not seen it; some older thoughts on the matter can be found in https://github.com/magit/magit/issues/2985.

----

I am closing this pull-request because I do not intend to merge it.  That doesn't mean that we cannot continue talking here.

PR #5229, opened 2024-09-20
magit-revert-rev-file-buffer: Adapt to upstream changes in Emacs 30

pengjiz:
Hello!

In Emacs commit [9506b9392e], `define-globalized-minor-mode` was refactored, and in Emacs 30 only the function `MODE-enable-in-buffer` is added to `after-change-major-mode-hook`. So in this PR we remove both the new and the old functions from the hook to avoid activating `diff-hl-mode` for blob buffers.

Thank you!

[9506b9392e]: https://git.savannah.gnu.org/cgit/emacs.git/commit/lisp/emacs-lisp/easy-mmode.el?id=9506b9392e8c3548b769d9951a1c9d4abad9ee21


tarsius:
Thanks!

PR #5231, opened 2024-09-27
Fix minor typos

jacobilsoe:
Fixed minor typos


tarsius:
Thanks! I've applied the typo fix.

The second changes something that isn't a type, which I want to keep.  You might not like how it sounds, but it it quite common to find such "(which see)" after variables/functions in docstrings and Emacs manuals.

PR #5240, opened 2024-10-10
magit-start-process: Handle tramp-pipe-stty-settings being unbound

CeleritasCelery:
Currently, if you set `magit-tramp-pipe-stty-settings` to `'pty` it will bind `tramp-pipe-stty-settings` to its own symbol value. But if tramp has not been loaded yet this will result in referencing a void variable. We fix this by checking if the variable is bound before referencing it. 

## To reproduce the issue

`(setq magit-tramp-pipe-stty-settings 'pty)`

call `magit-status`.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,10 @@ Bug fixes:
   that margin, then the old width was not restored when the mode was
   disabled.  #5236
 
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks!

PR #5241, opened 2024-10-15
Fix documentation typo

fiveNinePlusR:
`with-eval-after-load 'magit-mode` should be `with-eval-after-load 'magit` in order to work

### DIFF
--- a/docs/magit.org
+++ b/docs/magit.org
@@ -639,7 +639,7 @@ whenever a buffer is saved to a file inside the respective repository
 by adding a hook, like so:
 
 #+begin_src emacs-lisp
-  (with-eval-after-load 'magit-mode
+  (with-eval-after-load 'magit
     (add-hook 'after-save-hook 'magit-after-save-refresh-status t))
 #+end_src
 


tarsius:
I don't think that is accurate. `magit-after-save-refresh-status` is defined in "magit-mode.el", so delaying until `magit-mode` is loaded, is appropriate. Delaying until after `magit` is loaded would also work, because if that is loaded, then `magit-mode` is loaded as well. But it would be a less accurate way of specifying what has to be loaded before it is safe to add to the hook.

Also, what do you mean by "in order to work"? What doesn't work otherwise? Do you get an error (please post a backtrace, if so)?

fiveNinePlusR:
I think you're right... didn't realize there was a magit-mode as well as just plain magit for that. I was basing my with-eval-after-load call off the docs after not using emacs for a bit and thought it was a typo because most with-eval-after-load calls omit the -mode part. It was not working for my use case(overriding a variable later in the load). feel free to close if you think this is not applicable or doesn't need clarification.

tarsius:
The code block you based your customization on was specifically about adding that one hook and not about customizing Magit in general, for which `(with-eval-after-load 'magit ...)` would be the safer option.

I'll keep things as-is for now.

PR #5244, opened 2024-10-17
magit-copy-section-value: fix default case

vermiculus:
If the current section type is not one of

    '(branch commit module-commit tag)

the intention here is to just copy the raw section value, but this was not being done. Instead, `magit-section-case` was unfortunately interpreting the `(kill-new (message ...))` expression as a list of section types. Whoops!

Use the default case to correct this error.


tarsius:
Thanks!

PR #5261, opened 2024-12-02
Feat/add fixup variants

x4lldux:
Adds new command for `--fixup=amend:`. During PR reviews it's often needed to fix not only code but also commit messages, fixup amend allows that.

post scriptum:
I also have a command for `--fixup=reword:` but I don't know yet how to best approach diffing. I'm partial to showing diff of a commit we are rewording, but that requires global var to save that commit used by that single command.

### DIFF
--- a/lisp/magit-commit.el
+++ b/lisp/magit-commit.el
@@ -142,6 +142,7 @@ Also see https://github.com/magit/magit/issues/4132."
     ("f" "Fixup"          magit-commit-fixup)
     ("s" "Squash"         magit-commit-squash)
     ("A" "Augment"        magit-commit-augment)
+    ("m" "Fixup amend"    magit-commit-fixup-amend)
     (6 "x" "Absorb changes" magit-commit-autofixup)
     (6 "X" "Absorb modules" magit-commit-absorb-modules)]
    [""
@@ -249,6 +250,17 @@ depending on the value of option `magit-commit-squash-confirm'."
   (interactive (list (magit-commit-at-point)
                      (magit-commit-arguments)))
   (magit-commit-squash-internal "--fixup" commit args))
+
+;;;###autoload
+(defun magit-commit-fixup-amend (&optional commit args)
+  "Create a fixup amend commit.
+
+With a prefix argument the target COMMIT has to be confirmed.
+Otherwise the commit at point may be used without confirmation
+depending on the value of option `magit-commit-squash-confirm'."
+  (interactive (list (magit-commit-at-point)
+                     (magit-commit-arguments)))
+  (magit-commit-squash-internal "--fixup" commit args nil t nil "amend:"))
 
 ;;;###autoload
 (defun magit-commit-squash (&optional commit args)
--- a/lisp/magit-commit.el
+++ b/lisp/magit-commit.el
@@ -290,7 +290,7 @@ depending on the value of option `magit-commit-squash-confirm'."
   (magit-commit-squash-internal "--squash" commit args t))
 
 (defun magit-commit-squash-internal
-    (option commit &optional args rebase edit confirmed)
+    (option commit &optional args rebase edit confirmed commit-prefix)
   (when-let ((args (magit-commit-assert args (not edit))))
     (when (and commit rebase (not (magit-rev-ancestor-p commit "HEAD")))
       (magit-read-char-case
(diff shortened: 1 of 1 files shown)


tarsius:
I wasn't aware these existed.

They were added in v2.32.0.

I'll have to think about how I want to support these. Because December is a really busy month, I do not expect to get around to that before next year.

kyleam:
@tarsius Some time ago I thought a bit about adding support for `--fixup=(amend|reword):`.  Here are some scattered thoughts (modified from local notes), in case they're of any use when you revisit this.

 * I think the `--fixup=amend:` functionality adds something useful on top of what's already available with Magit's commit variants.  It provides a way to 1) compose the final commit message at the same time as making the fixup commit and 2) put off the actual rebase.

   `magit-commit-instant-squash` covers *1* but not *2*.  `magit-commit-squash` covers *2* but not *1*.  `magit-commit-augment` covers *2* and almost covers *1*, but it doesn't provide the original message as the starting point and, on rebase, you'll still be thrown into an editor and need to remove the original message.

   (That all may be obvious to others, but, based on my notes, it seems I wasn't sure upfront whether these new fixup variants provided anything useful on top of Magit's existing commands.)

   I'd use something that covers *1* and *2* in cases where I'm making a fixup commit and want to get the message down while it's fresh in my head, but I'm not ready to take on the rebase because I anticipate needing to resolve conflicts.

   Based on the "during PR reviews" in @x4lldux's PR description, my understanding is that the motivation is a setting where reviewers expect fixup commits with final messages be pushed for review before the final rebase.  That's a scenario for wanting *1* and *2* that I hadn't considered when initially thinking about this.  (An alternative would be to do the rebase and use a range-diff to communicate the changes to the reviewers.)

 * I personally wouldn't have much use for `--fixup=reword:` because I'd just use `magit-rebase-reword-commit`.  Given `--autostash`, that can be done even with uncommitted changes, though someone may want to avoid relying on `--autostash` if they have changes they want to keep changes staged.

   But there's still the review case above where reviewers may insist on seeing fixup commits on top, leaving the rebase until later.  If that workflow (wanting to review fixup commits before the final rebase) is popular, I suppose a good number of people would have a use case for `--fixup=reword:`.

 * Assuming this functionality gets exposed as dedicated commands, the best names I've come up with are `magit-commit-amend-at` (for `--fixup=amend:`) and `magit-commit-reword-at` (for `--fixup=reword:`).  I'm not sure if those convey the meaning like I hope, but they may fit in better with the current set of names than `magit-commit-fixup-amend` and `magit-commit-fixup-reword`.

 * these commits have an `amend!` subject, so it'd be good to highlight that in the same way as `fixup!` and `squash!`

   <details>
   <summary>diff</summary>
   
   ```diff
   diff --git a/lisp/git-rebase.el b/lisp/git-rebase.el
   index 6f4af506..33740145 100644
   --- a/lisp/git-rebase.el
   +++ b/lisp/git-rebase.el
   @@ -778,7 +778,7 @@ (defun git-rebase-mode-font-lock-keywords ()
        (git-rebase-match-comment-line 0 'font-lock-comment-face)
        ("\\[[^[]*\\]"
         0 'magit-keyword t)
   -    ("\\(?:fixup!\\|squash!\\)"
   +    ("\\(?:fixup!\\|squash!\\|amend!\\)"
         0 'magit-keyword-squash t)
        (,(format "^%s Rebase \\([^ ]*\\) onto \\([^ ]*\\)" comment-start)
         (1 'git-rebase-comment-hash t)
   diff --git a/lisp/magit-log.el b/lisp/magit-log.el
   index e452e2da..e76a681c 100644
   --- a/lisp/magit-log.el
   +++ b/lisp/magit-log.el
   @@ -1476,7 +1476,7 @@ (cl-defun magit-log-wash-rev (style abbrev)
    
    (defun magit-log-propertize-keywords (_rev msg)
      (let ((boundary 0))
   -    (when (string-match "^\\(?:squash\\|fixup\\)! " msg boundary)
   +    (when (string-match "^\\(?:squash\\|fixup\\|amend\\)! " msg boundary)
          (setq boundary (match-end 0))
          (magit--put-face (match-beginning 0) (1- boundary)
                           'magit-keyword-squash msg))
   ```
   
   </details>

 * For `rebase --autosquash`, these get mapped to `fixup -C {new commit}`, so there's no changes (e.g., a new directive) needed at rebase todo-list level.

tarsius (approved):

tarsius (approved):

tarsius (approved):

tarsius (approved):

tarsius (approved):
Piu, finally got it right!

tarsius (changes requested):
Change stuff.

tarsius (approved):
Wonderful!

tarsius:
(Please ignore the excessive "reviews" above, I was testing some Forge functionality.)

@x4lldux, thanks for the suggestion.

Thanks @kyleam, your analysis was useful.

I have added two new commands. Additionally I have restructured the menu and added more polish. There's also a lot more documentation now.

PR #5272, opened 2024-12-21
* lisp/magit-commit.el (magit-commit-absorb): Remove -v option

DamienCassou:
The option "-v" is forced in the call to `magit-run-git-async`. Adding
it through the transient crashes the command because git-absorb
doesn't accept duplicate options.

How to reproduce the issue:
1. Type `C-u M-x magit-commit-absorb`
2. Type `-v`
2. Type `x` to start git-absorb

Actual: You get an error because `-v` is specified twice

-------


tarsius:
Thanks for the report. I've fixed it differently.

PR #5277, opened 2024-12-23
magit-list-special-refnames: Fix usage of seq-keep

nikital:
seq-keep applies the function to the returned sequence, resulting a list of '(t t t ...) being returned.  Adjust the lambda to return the ref when the ref exists.

Before this fix, when using Push -> Another branch you would get "t" suggested instead of "HEAD".

Fixes: 621b67fc (magit-list-special-refnames: Cosmetics, 2024-11-22)


tarsius:
Thanks!

I've decided to go with `cl-remove-if-not` instead.

PR #5282, opened 2025-01-08
Fix some broken diffs when external diff is configured

nikital:
Before this PR, some diff Git commands will be called without "--no-ext-diff" for stash and commit, which results broken diffs when one has a custom external diff program.


tarsius:
I'll review this tomorrow.

Until then I am abusing this pull-request to test creating review requests using Forge. Please ignore those requests.

tarsius:
Thanks!

I'll update the changelog soon, together with all the other changes so far.

PR #5291, opened 2025-01-19
Git commit accept ssh key passphrase

### LINKED ISSUE #5288: signing commit with ssh-key doen't ask for passphrase
# what i expected:
When having the .gitconfig setup as
```
[gpg]
	format = ssh
[commit]
	gpgsign = true
```

and creating a commit, I expect to be asked my ssh-key passphrase, just like when I do `git commit -m ""` from the terminal and a signed commit being created.

# what I observed
The commit creating process stops after:
```
Type C-c C-c to finish, C-c C-k to cancel, and M-p and M-n to recover older messages
Message saved
Saving file ~/test/.git/COMMIT_EDITMSG...
Wrote ~/test/.git/COMMIT_EDITMSG
```
When exiting emacs the following process is still active: `git             10257   run     magit-process: .dotfiles  /dev/pts/3   Main         git --no-pager --literal-pathspecs -c core.preloadindex=true -c log.showSignature=false -c color.ui=false -c color.diff=false commit --`

When I fetch from a repo in magit, I am asked my ssh-key-passphrase. After entering it once, all subsequent signed commits are performed in full.

Performing `git commit -m ""` from the terminal, I am asked my passphrase even without prior pulling/pushing etc.

# version
Magit 20250116.1243 [>= 20250116.1243], Transient 20250108.1351, Git 2.48.0, Emacs 29.4, gnu/linux

bird-dancer:
Add regexps to match the ssh-key prompts when requesting a passphrase
for ssh-key, this likely relates to this prompt:
https://github.com/openssh/openssh-portable/blob/826483d51a9fee60703298bbf839d9ce37943474/sshconnect2.c#L1533

fixes #5288

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -157,7 +157,7 @@ itself from the hook, to avoid further futile attempts."
   :type 'regexp)
 
 (defcustom magit-process-password-prompt-regexps
-  '("^\\(Enter \\)?[Pp]assphrase\\( for \\(RSA \\)?key '.*'\\)?: ?$"
+  '("^\\(Enter \\)?[Pp]assphrase\\( for \\(RSA \\)?\\(key \\)?['\"].*['\"]\\)?: ?$"
     ;; Match-group 99 is used to identify the "user@host" part.
     "^\\(Enter \\|([^) ]+) \\)?\
 [Pp]assword\\( for '?\\(https?://\\)?\\(?99:[^']+\\)'?\\)?: ?$"


tarsius:
Thanks, but I am already working on it. (I came up with a more complete solution, but want to add some tests before I push it.)

basil-conto:
> (I came up with a more complete solution, but want to add some tests before I push it.)

Commit a43c1d8d7a99dd2b9f865285bed442dddb0e5ce8 says:

> - Since we don't try to determine a key's identifier anyway, when
>   asked for a passphrase, essentially only look for a reasonably
>   anchored string "passphrase".

FWIW, in repos with multiple remotes each with a different SSH key, I no longer know which key's passphrase to enter.

All prompts are of the same form `Enter passphrase for key '~/.ssh/foo': ` so they all match the first entry in `magit-process-password-prompt-regexps` and get turned into `Enter passphrase: `.

If this was an intented consequence of the recent changes, I can of course just customise the user option to keep the key identifier.

Thanks!

tarsius:
Thanks, fixed.

PR #5294, opened 2025-01-28
Fix magit-read-char-case macro usages

gemmaro:
This update removes the redundant "or" from the prompt messages.

The `magit-read-char-case` macro has been updated to automatically format prompt messages in natural English and correctly add "or" before the last entry.  However, this change led to some prompt messages containing duplicated "or" words.  For instance, the prompt for the `magit-clone` command previously displayed:

> Clone from [u]rl or name, [p]ath, [l]ocal url, or **or** [b]undle

This update addresses that issue by removing the redundant "or."


tarsius:
Thanks!

PR #5295, opened 2025-01-28
Make magit-stash-index respect user configured function

iarenaza:
When magit-stash-read-message-function was introduced (in commit 72fbb3a579b), magit-stash-read-args was updated to use it, but magit-stash-index was not.

Which means that if the user customizes magit-stash-read-message-function, the customized value will not be used in magit-stash-index.


tarsius:
Thanks!

PR #5297, opened 2025-01-28
Add --signoff argument to rebase

Thaodan:
For example to sign-off a patch series after reviewing them.

### DIFF
--- a/lisp/magit-sequence.el
+++ b/lisp/magit-sequence.el
@@ -544,7 +544,8 @@ This discards all changes made since the sequence started."
    ("-i" "Interactive"              ("-i" "--interactive"))
    ("-h" "Disable hooks"            "--no-verify")
    (7 magit-rebase:--exec)
-   (5 magit:--gpg-sign)]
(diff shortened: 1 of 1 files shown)


Thaodan:
Thanks. I was wondering I should have moved the define argument for the sign-off option.

Now that sign-off is in general on level 6 when before it was always visible, should there be a changelog entry for this?

tarsius:
> Now that sign-off is in general on level 6 when before it was always visible, should there be a changelog entry for this?

Yes, when I go through the commits since the last changelog update, then I'll add an entry about this.

Thaodan:
OK great.

Keep up the good work.

tarsius:
> Keep up the good work.

I did, so now you can set the level for this argument across menus in one go:

```elisp
(transient-set-default-level 'magit:--signoff 1)
```

See https://magit.vc/manual/transient/Enabling-and-Disabling-Suffixes.html.

Note that you have to update both `transient` and `magit` and that `magit` must be compiled against the new `transient`.

PR #5304, opened 2025-02-02
fix `magit-log-default-arguments` being ignored

Alan-Chen99:


tarsius:
Thanks!

This had to be fixed in three more places.

Issue #5317, opened 2025-02-17
Commit message font-lock for trailers does not work


cbrune:
This bug is reproducible from the top of main using the `make emacs-Q` approach.

# what behavior you expected
When editing commit messages, the trailer tags would be highlighted properly.

# what behavior you observed

If the commit message contains a standard trailer like `Signed-off-by: Foo Bar <foo@bar.net>` you will see an error message like:

```
Error during redisplay: (jit-lock-function 51) signaled (error "No match 3 in highlight (3 'git-commit-trailer-token)")
```

If the commit message contains a free form trailer like `This-is-fake: Foo Bar <foo@bar.net>` you will see an error message like:

```
Error during redisplay: (jit-lock-function 51) signaled (error "No match 1 in highlight (1 'git-commit-trailer-token)")
```

Also it breaks further font-lock highlighting in the message, e.g. the comment block at the end of the message is not highlighten properly

# and how we can reproduce the issue.

1. `make emacs-Q`
2. edit a commit message and add a trailer
3. check the output in the `*Messages*` buffer

tarsius:
Fixed by #5318.

PR #5318, opened 2025-02-17
git-commit: make trailer regex matches lax

cbrune:
The git-commit--trailer-regexp regex is written so that either groups 1 and 2 will match or groups 3 and 4 will match, but it is impossible for all 4 groups to match.

This patch updates the font-lock regex checker to be lax about whether or not the group matched. Without this patch we have errors like:

  Error during redisplay: (jit-lock-function 77) signaled (error "No match 3 in highlight (3 'git-commit-trailer-token)")

See #5317


tarsius:
Thanks!

PR #5333, opened 2025-03-05
magit-insert-head-branch-header: Fix use in let assignment block

fledermaus:
`branch` is potentially assigned and used in the let assignment sexp, should be `let*` for this to work.

### DIFF
--- a/lisp/magit-status.el
+++ b/lisp/magit-status.el
@@ -568,8 +568,8 @@ the status buffer causes this section to disappear again."
   "Insert a header line about the current branch.
 If `HEAD' is detached, then insert information about that commit
 instead.  The optional BRANCH argument is for internal use only."
-  (let ((branch (or branch (magit-get-current-branch)))
-        (output (magit-rev-format "%h %s" (or branch "HEAD"))))
+  (let* ((branch (or branch (magit-get-current-branch)))
+         (output (magit-rev-format "%h %s" (or branch "HEAD"))))
     (string-match "^\\([^ ]+\\) \\(.*\\)" output)
     (magit-bind-match-strings (commit summary) output
       (when (equal summary "")


tarsius:
How did you find this? `HEAD` and "the current branch (if any)" point at the same commit, so the result is the same, regardless of what we pass to `git rev-parse`. If BRANCH is nil, it is set because that makes a difference later in `let`'s BODY, not ARGLIST. Am I missing something?

fledermaus:
It was actually a repo that had become corrupted. I agree it's not possible to trip up on this on a well formed repo.

tarsius:
What happens in corrupted repository if this is not applied (i.e., what kind of "corruption" are we talking about)? I assume the only effect is that information for the "wrong" commit is shown (as opposed to some *error* occuring).

fledermaus:
I was trying to figure out why I was getting no magit status buffer output, and traced it as far as the value of branch being nil, even though 'git branch' was correctly returning what _should_ have been the current branch.

So what was happening was instead of trying to show the status of the branch and showing me an error (I hope) the branch value stayed as nil, and since the HEAD object was corrupt I just got an empty magit buffer.

fledermaus:
I don't have an easy repro though.

tarsius:
We surely use `HEAD` elsewhere, so even if we avoid falling back to that here, other parts of Magit will still be broken if that somehow gets corrupted. I am inclined to close this without merging.

I'm also considering

```diff
@@ -568,13 +568,12 @@ (defun magit-insert-head-branch-header (&optional branch)
   "Insert a header line about the current branch.
 If `HEAD' is detached, then insert information about that commit
 instead.  The optional BRANCH argument is for internal use only."
-  (let ((branch (or branch (magit-get-current-branch)))
-        (output (magit-rev-format "%h %s" (or branch "HEAD"))))
+  (let ((output (magit-rev-format "%h %s" (or branch "HEAD"))))
     (string-match "^\\([^ ]+\\) \\(.*\\)" output)
     (magit-bind-match-strings (commit summary) output
       (when (equal summary "")
         (setq summary "(no commit message)"))
-      (if branch
+      (if-let ((branch (or branch (magit-get-current-branch))))
           (magit-insert-section (branch branch)
             (insert (format "%-10s" "Head: "))
             (when magit-status-show-hashes-in-headers
```

to avoid the potential impression that there's a bug here.

PR #5340, opened 2025-03-18
magit-stash-save: Use diff with "--default-prefix"

Schnouki:
Without this, if Git is configured with `diff.noprefix = true`, then saving a stash with e.g. `magit-stash-index` fails with "git diff header lacks filename information when removing 1 leading pathname component" after saving the stash, leaving the stashed data in the index.

### DIFF
--- a/lisp/magit-stash.el
+++ b/lisp/magit-stash.el
@@ -430,7 +430,8 @@ Then apply STASH, dropping it if it applies cleanly."
                            (magit-stash-create message index worktree untracked))
         (if (eq keep 'worktree)
             (with-temp-buffer
-              (magit-git-insert "diff" "--cached" "--no-ext-diff")
+              (magit-git-insert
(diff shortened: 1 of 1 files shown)


tarsius:
This has come up before. The advice we have settled on is to undo such settings in `magit-git-global-arguments`.

Schnouki:
Yeah, that should work too. If Magit assumes that the default prefix is used for diff, do you think it would make sense to add `-c diff.noprefix=false` to `magit-git-global-arguments`?

tarsius:
That probably wasn't done already because I did not want to also proactively undo `diff.srcPrefix`, `diff.dstPrefix` and `diff.mnemonicPrefix`. I did some rudimentary testing and it appears doing it for these three isn't actually necessary.

Schnouki:
Thanks a lot for this change! I really appreciate it :)

PR #5348, opened 2025-03-26
Use ansi-color in magit-process-filter

bobrowadam:
This change use ansi-colors for the magit-process buffer output while the process is sending it's output.
I know about `magit-process-finish-apply-ansi-colors` variable but it only apply the coloring at`magit-process-finish`, maybe I missed some complexity that forcing applying the coloring only then.

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -843,6 +843,7 @@ Magit status buffer."
         (setq string (substring string (1+ ret-pos)))
         (delete-region (line-beginning-position) (point)))
       (setq string (magit-process-remove-bogus-errors string))
+      (setq string (ansi-color-apply string))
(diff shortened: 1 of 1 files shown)


kyleam:
> maybe I missed some complexity that forcing applying the coloring only then

I suspect the main concern is performance.   See gh-3159.

bobrowadam:
@kyleam 
Thanks for the pointing that out.
How about applying coloring only when a custom flag is turned on? something like `magit-process-filter-apply-ansi-colors`?

tarsius:
> I suspect the main concern is performance.

That and if the output is chunked, some text won't be propertized as intended.

But this keeps coming up, so I have added an option to enable it but use its docstring to warn against actually doing it.

PR #5354, opened 2025-04-13
magit-section-show-level: Support using region of sections

x4lldux:
Often I find I'd like to see changes in some of the staged files. Changing levels one by one is time consuming and changing level for entire staged section is too verbose. This QoL improvement makes this task way faster.


tarsius:
Thanks!

PR #5357, opened 2025-04-19
manual: Remove inaccurate sentence about magit-diff-visit-file

RJSent:
From my reading, the existing documentation for `magit-diff-visit-file` accidentally copies a line from `magit-diff-visit-worktree-file`, specifically "This commands [sic] visits the worktree version of the appropriate file". 

This PR deletes that sentence and also combines the existing "This command..." sentence with the following paragraph to be consistent with the other commands/options in the same section.


tarsius:
Thanks!

I've also corrected the docstring.

PR #5360, opened 2025-04-19
Speedup refresh by prefetching git calls in parallel

Azkae:
Hello, 

This PR adds an option (`magit-prefetch`) to prefech git calls in parallel.

Before refreshing the buffer, we fill `magit--refresh-cache` with commands that will be used during refresh. This is done in parallel. 
`magit-refresh` will then use results from `magit--refresh-cache` and make very few git calls.

There is two steps because some commands depend on branch/remote names so we fetch them in parallel first.

On my machine this speeds up `magit-refresh` by approximately 100ms. 
I have been using this for a few weeks without issue.

Sadly this doesn't work on tramp right now (where I believe it could have the most impact because of higher roundtrips). I'm not a tramp user so I don't know if this could be adapted to work on tramp.

What do you think?


tarsius:
Some quick early feedback.

```elisp
(defun magit-maybe-prime-refresh-cache ()
  "...
Add to `magit-pre-refresh-hook'."
  (when (and (or magit-refresh-status-buffer
                 (derived-mode-p 'magit-status-mode))
             ;; This will certainly lead to a feature request.
             ;; On the other hand, dropping this will likely
             ;; cause the "illegal re-entry" issue.
             (not (file-remote-p default-directory)))
    (cl-flet
        ((prime (cmds)
           (dolist (cmd cmds)
             (incf running)
             (make-process
              ...
              :sentinel (... (decf running))))))
      (let* (...
             (running 0)
             (elapsed
              (benchmark-elapse
                (prime ...)
                (prime ...))))
        (with-timeout 1
          (while (> running 0)
            (sit-for 0.01)))
        (when magit-refresh-verbose
          (message "Refresh cached primed in %.3fs" elapsed))))))
```

tarsius:
> What do you think?

I have mixed feelings about it. While I have considered doing something like this before, I have also decided against doing it.

The work doesn't end after landing such a feature. E.g., I might end up having to add Tramp support in the future (and like you I don't use that myself, and it is always a huge pain when a user of my packages runs into issues with it).

There's even the risk that this has a negative impact for some users. For example you didn't consider the case of users who have disabled `magit-refresh-status-buffer`. I caught that, but there might be other complications we both don't yet realize exist.

I don't know yet whether I want to add this too Magit. In any case it should mature a bit first. In case I end up not merging, it should be possible to maintain this as a separate package, if you are willing to do that.

Azkae:
Thank you for your feedback! I have simplified the code a bit more. 
I understand if you don't want to support this in Magit directly, reliable tramp support seems a bit hard.

> There's even the risk that this has a negative impact for some users.

I originally added an option to have this feature disabled by default. Do you think I should add it back?

> it should be possible to maintain this as a separate package, if you are willing to do that.

Maybe I will look into supporting it in a separate package if we end up not merging this.
The only upside to supporting it in Magit directly is that it would be easier to update the commands list when new commands are added or modified.

tarsius:
> I originally added an option to have this feature disabled by default. Do you think I should add it back?

No, adding or removing the function from the hook is also a decent way of enabling/disabling the feature. The function should *not* be added to the hook *by default*.

lisp/magit-git.el
```diff
@@ -2854,6 +2859,103 @@ out.  Only existing branches can be selected."
...
+        (message "Refresh cached primed in %.3fs" elapsed)))))
+
+(defun magit--prime-caches-with-commands (commands)
+  "Prime the refresh cache with the provided COMMANDS."
+  (let ((buffers '())
+        (repo-path (magit-toplevel))
+        (running 0))
+    (cl-loop
+     for args in commands
+     for index from 0
```
tarsius, line 2921: That's not being used anymore. That also means that you could use `mapcar` instead of `cl-loop`, which I prefer. You'll only find one other instance of `cl-loop` in Magit (which is only there because I haven't gotten around to replacing it yet).

lisp/magit-git.el
```diff
@@ -2854,6 +2859,103 @@ out.  Only existing branches can be selected."
...
+        :buffer buffer
+        :noquery t
+        :connection-type 'pipe
+        :command (cons magit-git-executable (magit-process-git-arguments args))
+        :sentinel
+        (lambda (proc _event)
+          (when (eq (process-status proc) 'exit)
+            (let ((buf (process-buffer proc)))
+              (when (and (buffer-live-p buf) magit--refresh-cache)
+                (let ((value
```
tarsius, line 2938: I prefer to not add a variable which is then only used once.

lisp/magit-git.el
```diff
@@ -2854,6 +2859,103 @@ out.  Only existing branches can be selected."
...
+        :name (buffer-name buffer)
+        :buffer buffer
+        :noquery t
+        :connection-type 'pipe
+        :command (cons magit-git-executable (magit-process-git-arguments args))
+        :sentinel
+        (lambda (proc _event)
+          (when (eq (process-status proc) 'exit)
+            (let ((buf (process-buffer proc)))
+              (when (and (buffer-live-p buf) magit--refresh-cache)
```
tarsius, line 2937: Checking whether `magit--refresh-cache` is non-nil should probably happen earlier, at least in the outer `when`. Does this have to be checked at all?

lisp/magit-git.el
```diff
@@ -2854,6 +2859,103 @@ out.  Only existing branches can be selected."
...
+                                (goto-char (point-min))
+                                (buffer-substring-no-properties
+                                 (point) (line-end-position)))))))
+                  (push (cons key value)
+                        (cdr magit--refresh-cache))))
+              (cl-decf running)))))
+       (cl-incf running)
+       (push buffer buffers)))
+
+    (with-timeout (1)
```
tarsius, line 2951: If this does not timeout then all buffers are already killed, so drop the following `dolist` and change this to `(with-timeout (1 (mapc #'kill-buffer buffers)) ...)`.

lisp/magit-git.el
Azkae: You're right, I already check that the cache exists in `magit-maybe-prime-refresh-cache`. 
I think we still need to verify that the cache exists because `magit-refresh-all` runs magit-pre-refresh-hook without setting a cache.

lisp/magit-git.el
Azkae: I may be missing something but I don't think we are killing buffers when there is no timeout.

Azkae:
I realized that there is something I need to fix regarding errors:
For some calls that are expected to fail (calls that use magit-git-string-p) it's fine to push nil in the cache in case of an error.

But I shouldn't put nil in the cache for calls made by `magit-git-string`. If the git call fails, we should make it twice (one in the prime caching, and one in normal call path) and allow for the real call to handle the error.

I will most likely add a flag in the commands, something like:

```
("rev-parse" "--verify" ,(concat "refs/tags/" branch) :cache-errors)
```

lisp/magit-git.el
tarsius: Yes exactly, if there is no timeout, then there are no buffers left to be cleaned up. That is why we should move the cleanup to the "this is only done if a timeout actually occurs" place (it helps others (including future us) realize that that step is only needed in case of a timeout).

Azkae:
Sorry for the radio silence on this PR.
I went ahead and created a separate package for this: [magit-prime](https://github.com/Azkae/magit-prime)

tarsius:
Great thanks!

You might want to add that to Melpa and/or mention it at https://github.com/magit/magit/wiki/Tips-and-Tricks#performance-hints.

PR #5395, opened 2025-06-13
magit-branch: Add --force argument to the transient

DamienCassou:
The "magit-branch" transient (containing "checkout" commands) didn't have support for the `--force` parameter that both `git branch` and `git checkout` accept. This PR adds this new parameter to the transient but keep it invisible by default (level = 7).

### DIFF
--- a/lisp/magit-branch.el
+++ b/lisp/magit-branch.el
@@ -224,7 +224,9 @@ has to be used to view and change branch related variables."
     :inapt-if-not magit-get-some-remote)]
   ["Arguments"
    (7 "-r" "Recurse submodules when checking out an existing branch"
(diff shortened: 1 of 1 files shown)


tarsius:
I don't think this should be added. It is too dangerous and it is trivial to work around it not being offered: First throw away local change manually and explicitly using Magit's "discard" commands.

I did not respond earlier because you had marked this as a draft and the argument's description is excessively long. I assumed you're intention was to bring that to a reasonable length, before requesting a review. I don't think it is possible to make this shorter, while also conveying how dangerous it is, but I figured I would wait to see what you would come up with.

Please instead add this argument in your personal configuration.

PR #5414, opened 2025-07-21
magit-status-show-untracked-files: Fix documentation

DamienCassou:
I believe the documentation was wrong so I fixed it.


DamienCassou:
I didn't update magit.texi because running `make texi` as suggested in the PR template produces a huge diff with unrelated changes.

tarsius:
Thanks!

PR #5423, opened 2025-08-17
Add option to disable inserting cherry commits in magit-refs

aarikere:
This PR adds an option to disable inserting cherry commits in magit-refs. This was done in order to get around the issue described here.

https://stackoverflow.com/questions/79691880/is-there-a-way-to-list-branches-in-magit-without-comparing-them-with-any-other-b?noredirect=1#comment140568761_79691880

In short, in a very large mono-repo that I work with `magit-show-refs` didn't complete even after 10 minutes. With these changes it completes in around 10s to list 18k odd branches and tags.


tarsius:
The cherries are only calculated and inserted once they become visible, i.e., when their parent sections are expanded to show the cherries.

Do those sections start out expanded for you? If so, try to reproduce that with `emacs -Q` (see https://magit.vc/manual/magit/Debugging-Tools.html). Have you customized `magit-section-set-visibility-hook` and/or `magit-section-initial-visibility-alist`?

aarikere:
You're correct! I had `(magit-section . show)` in my initialization to have the "Recent commits" section visible in the `magit-status`buffer. But I see that all the entries in the `magit-refs` buffer are of the same type. Removing `(magit-section . show)` solves the problem with `magit-refs`, but then is there a way to keep "Recent commits" section visible while keeping the sections in `magit-refs` collapsed?

tarsius:
Since 936a211ada2045d24193656e94914be0e9cd1cbd `magit-describe-section` describes the "Recent commit" section so:

```
#<magit-unpushed-section "@{upstream}.." [unpushed status] 291-[306-]957>
  is inserted by `magit-insert-unpushed-to-upstream-or-recent'
  from `magit-status-sections-hook'.
```

Before that we got.

```
#<magit-section "@{upstream}.." [recent status] 291-[306-]957>
  is inserted by `magit-insert-unpushed-to-upstream-or-recent'
  from `magit-status-sections-hook'.
```

(This is something I probably should revisit, but there were reasons why I made this less correct, see the commit message.)

So it appears you configured this seven years ago or more.

----

You used `magit-section` as the key (which is the *class* of the section). The docstring of `magit-section-initial-visibility-alist` instructs you to use the type or lineage, here `unpushed` or `[unpushed status]`.

The class can also be used, which you maybe learned from also reading the docstring of `magit-match-section`. But that's not necessary or recommended here.

You should definitely not use the class `magit-section`. That's the root class of all section classes, so it matches every section.

So use `unpushed` as the key, and to future-proof it, maybe add a second entry using `recent`.

aarikere:
Yes, that worked! 

Sorry for the trouble and thanks for the help!

PR #5424, opened 2025-08-18
magit-section: Display visibility indicators in left margin

emreyolcu:
Magit currently allows displaying visibility indicators either in the fringe or at the end of section headings, configured via `magit-section-visibility-indicators`. For those using high-DPI displays, fringe indicators get scaled up naively and appear blocky. Ideally, Emacs would support displaying SVGs as fringe indicators, and there was some talk on emacs-devel about this, but my understanding is that it is seen as low priority and there is not much progress towards it. 

To work around this issue, I prefer to disable fringes and use margins for indicators instead. This works quite well with a font with broad Unicode coverage (like Iosevka). Some packages (like Flycheck) have built-in support for margin indicators (precisely for the reason I describe above). Magit currently does not support margin indicators, but I have been using my fork with margin indicators hardcoded into `magit-section` for about a year now. I have not run into any issues so far; would you be open to adding support for this? 

I have only created this PR to start a discussion; of course either there would need to be a new variable to control this or we could allow an additional form for the value of `magit-section-visibility-indicators`. There are also other decisions like whether the change would need to reside in `magit-section.el` or `magit-margin.el`.

Here's how it looks with my current setup. What do you think?

Screen Shot 2025-08-18 at 13 18 57

### DIFF
--- a/lisp/magit-section.el
+++ b/lisp/magit-section.el
@@ -1961,19 +1961,17 @@ When `magit-section-preserve-visibility' is nil, return nil."
            (eoh (magit--eol-position beg)))
       (cond
        ((symbolp (car-safe magit-section-visibility-indicator))
-        (let ((ov (magit--overlay-at beg 'magit-vis-indicator 'fringe)))
+        (let ((ov (magit--overlay-at beg 'magit-vis-indicator 'margin)))
           (unless ov
             (setq ov (make-overlay beg eoh nil t))
             (overlay-put ov 'evaporate t)
-            (overlay-put ov 'magit-vis-indicator 'fringe))
+            (overlay-put ov 'magit-vis-indicator 'margin))
           (overlay-put
            ov 'before-string
-           (propertize "fringe" 'display
-                       (list 'left-fringe
-                             (if (oref section hidden)
-                                 (car magit-section-visibility-indicator)
-                               (cdr magit-section-visibility-indicator))
-                             'fringe)))))
+           (propertize "margin" 'display
+                       `((margin left-margin)
+                         ,(propertize (if (oref section hidden) "🞂" "🞃")
+                                      'face 'default))))))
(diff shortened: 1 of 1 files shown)


tarsius:
I've worked on this a bit and have gotten pretty far.  I've pushed the result to the `margin` branch. (The first commit is unrelated, except that it also deals with a margin. I wrote that in response to the pull-request before this one and had to put it somewhere.)

Working on this was procrastination. I have to finish some better changes now, and won't work on this until I am done with that.

Closing this pull-request, since my changes are further along the way.

tarsius:
Oh, but please give my branch a try and report any issue (here).

emreyolcu:
> Oh, but please give my branch a try and report any issue (here).

Thanks, will do!

emreyolcu:
I've tried your branch and haven't encountered any issues.

Just an FYI, margins have the property that, since they are considered to be part of the window geometry, their pixel width is not affected by `text-scale-adjust`. However, `text-scale-adjust` *does* affect the font height in the margins. This causes the text in the margins to get clipped when increasing font size via `text-scale-adjust`. I find this behavior unintuitive, but [Eli disagrees](https://yhetil.org/emacs/86r0dv8kes.fsf@gnu.org/). If you want to avoid this kind of clipping, then it might be a good idea to implement the workaround suggested by Eli in the thread I linked to, by making the margin indicators have some face with an absolute `:height` attribute.

tarsius:
Merged it (after some more tweaking and cleanup in the surrounding).

I haven't added a text scaling workaround, but users can add it themselves by changing the face to something like:

```elisp
`((t :inherit default :height ,(face-attribute 'default :height)))
```

emreyolcu:
Looks great, thanks so much!

tarsius:
You're welcome!

PR #5429, opened 2025-08-23
Update melpa testing instructions in bug_report.md

foudfou:
The bug reporting instructions include a section to test magit in isolation when downloaded from MELPA.

As a bug reporter, I want it to be easy and straightforward to make pre-submission assessment. While reporting a bug, attempting to run the instructions in my environment failed. This PR includes the updated command line call.


tarsius:
Thanks! There are more places were Cond-Let still had to be mentioned, which I have done in 2ab34a757503436904b80d0a24feeb45ab1e01c5 (without merging this pr). 

I added the dependency on Cond-Let last night, in the wee hours, and still have to update other packages (which depend on Magit) accordingly.

PR #5440, opened 2025-09-14
Function to show the difference between HEAD and n commits.

lanceberge:
Function to show the difference between HEAD and n commits.

This will prompt the user for input n then show the diff of
range HEAD~n.

Since this is a common operation, this reduces the friction of running `magit-diff-range` then typing HEAD~n

### DIFF
--- a/docs/magit.org
+++ b/docs/magit.org
@@ -3151,6 +3151,11 @@ Also see [[man:git-diff]]
   revisions, choose a revision to view changes along, starting at the
   common ancestor of both revisions (i.e., use a "..."  range).
 
+- Key: d h (magit-diff-head) ::
+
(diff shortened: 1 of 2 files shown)


tarsius:
> Since this is a common operation,

I cannot recall doing that. :grin:

Are you aware that you can mark a few commits in a log using the region and then type `d d` to compare the end-points?

lanceberge:
I do `git diff HEAD~1`  or `HEAD~2` all the time - different workflows I guess. Didn't know about `d d` but that's takes a few more keys than just `C-u 2 M-x magit-show-head`

tarsius:
Thanks for the suggestion, but I won't add this command. Please instead define it in your init file.

PR #5453, opened 2025-10-01
magit-diff-mode-map: bind `S-SPC` to `scroll-down`

younger-1:
In `magit-revision-mode` buffer, `SPC` is bound to `scroll-up`, but `S-SPC` is bound to `magit-diff-show-or-scroll-down`.


tarsius:
Thanks!

PR #5469, opened 2025-11-03
fix(blame): correct custom type of `magit-blame-styles`

LuciusChen:
The `magit-blame-styles` variable was incorrectly declared with `:type 'string`, even though its value is an alist of style definitions. This caused warnings when customizing the variable or setting it with `setopt`.

This commit changes the custom type to a proper nested alist type to match the actual data structure and suppress spurious warnings.


tarsius:
Thanks!

I've merged this with style fixes.

PR #5487, opened 2025-11-28
magit-margin: Require 'magit-section

sergiodj:
This is needed because otherwise we will see an error saying that
`magit-section-visibility-indicator` is undefined.

### DIFF
--- a/lisp/magit-margin.el
+++ b/lisp/magit-margin.el
@@ -32,6 +32,7 @@
 (require 'magit-base)
 (require 'magit-transient)
 (require 'magit-mode)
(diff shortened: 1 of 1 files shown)


tarsius:
`magit-margin` already require `magit-base`, which requires `magit-section`.

I am guessing you updated Magit and didn't restart Emacs afterwards.

sergiodj:
@tarsius Huh, I had updated magit but also did restart Emacs.  I was seeing this error before, but now it's not there anymore.  I'm assuming this was caused by some weird problem with native compilation?  Either way, sorry about the noise.

PR #5488, opened 2025-11-30
magit-log: Fix check whether magit-wip-mode is enabled

mjoerg:
`magit-wip-mode` is a minor mode, for which transient's `if-mode` predicate cannot be used.


tarsius:
Thanks!

PR #5492, opened 2025-12-12
magit-insert-worktrees: mention in documentation

huonw:
This section inserter is handy, and has existed since 2.7.0 / 82bfdd46295c5d6fa5d1a26a5058994e5b4f7c56... but appears to not yet be documented.

This documentation is modeled in the style of `magit-sparse-checkout-insert-header`, where it is mentioned in the docs for the relevant commands rather than the docs of `magit-status`, since most people who wish see a list of worktrees are likely using `magit-worktree` / reading its documentation.

---

Thanks for Magit!

### DIFF
--- a/docs/magit.org
+++ b/docs/magit.org
@@ -7519,6 +7519,11 @@ Also see [[man:git-worktree]]
   If the worktree at point is the one whose status is already being
   displayed in the current buffer, then show it in Dired instead.
 
+If you want the status buffer to list worktrees, add the function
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks!

huonw:
Woohoo, thank you for merging so quickly!

tarsius:
I was looking at Github just when you opened this. And now again. But not in between. :grinning:

PR #5504, opened 2026-01-15
magit-worktree-delete: Use git to remove worktree when not trashing

leezu:
The previous implementation used Emacs's delete-directory to remove the worktree directory, which is very slow for worktrees containing many files (e.g., a Python .venv with thousands of files).

When magit-delete-by-moving-to-trash is nil, use git-worktree-remove instead, which is much faster.  The --force flag is used because git-worktree-remove otherwise refuses to remove worktrees containing untracked files.  This is backward compatible since the old behavior also removed untracked files without warning.

### DIFF
--- a/lisp/magit-worktree.el
+++ b/lisp/magit-worktree.el
@@ -204,8 +204,10 @@ The primary worktree cannot be deleted."
       (magit-confirm-files (if magit-delete-by-moving-to-trash 'trash 'delete)
                            (list worktree))
       (when (file-exists-p worktree)
-        (let ((delete-by-moving-to-trash magit-delete-by-moving-to-trash))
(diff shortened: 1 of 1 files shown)


tarsius:
I didn't believe it was slow at first, because I have always assumed that when the documentation talked about "deleting recursively", that just meant deleting the repository even though it contains files. But they mean that literally, deleting everything one by one, instead of just the directory and be done with it. That seems a like a questionable choice.

So yeah, I'll likely merge this. (After some sleep.)

tarsius:
> the old behavior also removed untracked files without warning.

That seems too dangerous, so I've changed that now.

leezu:
Thank you, @tarsius.

PR #5513, opened 2026-02-04
Fix syntax highlighting for staged changes

ArthurHeymans:
## Summary

This PR fixes two bugs in the syntax highlighting feature for staged changes:

- **Fix typo in `magit-find-file-noselect`**: `length>` requires a number as its second argument, not `t`. Changed `(length> (magit--file-index-stages file) t)` to `(length> (magit--file-index-stages file) 1)`.

- **Handle `{index}` revision in `magit-diff--get-hunk-syntax`**: When viewing staged changes, the revision is `"{index}"` which cannot be dereferenced as a commit. For syntax highlighting purposes, we treat it as `"{worktree}"` since we only need the syntax rules from the file's major mode, not the actual index content.

## Testing

Tested with a Rust project that had staged changes - syntax highlighting now works correctly for both staged and unstaged diffs.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -3684,6 +3684,10 @@ actually a `diff' but a `diffstat' section."
 
 (defun magit-diff--get-hunk-syntax (hunk side rev file)
   (let ((args (magit-diff--get-hunk-text hunk (eq side 'old))))
+    ;; For syntax highlighting, treat "{index}" as "{worktree}" since
+    ;; we only need syntax rules, not actual index content.
+    (when (equal rev "{index}")
+      (setq rev "{worktree}"))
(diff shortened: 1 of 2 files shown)


tarsius:
Thanks, but I already fixed the typo on `main` and I haven't rebased the `syntax` feature branch yet, which means that I have not finished this intermezzo yet.  There are questions to be answered, before I can do that, such as how to deal with {index} "commit" oid, and I don't think {index} ~= {worktree} as you have done here is the answer, I will arrive at.

I recommend you use the reflog to go back to an earlier version of the `syntax` feature branch. Just because I rebase, that does not currently mean that I necessarily think it works after the transplant.  I very frequently rebase because if there are conflicts, it easier to resolve them if I still have the reasoning behind the conflicting change in my short-time memory.

PR #5514, opened 2026-02-04
magit-repos: use uniquify-separator if set

purcell:
When the user has set `uniquify-separator`, use it in `magit-list-repos-uniquify` for consistency. Fixes a small nit where magit's uniquification is jarringly different from what I usually see. This change stops short of explicitly loading `uniquify`, since that seemed unnecessary.


tarsius:
Thanks!

PR #5522, opened 2026-02-16
magit-branch, magit-diff: Use alist custom :type

basil-conto:
In the custom `:type` of
- `magit-branch-adjust-remote-upstream-alist` and
- `magit-diff-highlight-indentation`,

this replaces
- `(repeat cons)` -> `alist`,
- `string` -> `regexp`,
- `integer` -> `natnum`.

The UI difference is barely noticeable, but the latter types are more explicit and may better help validate input.

### DIFF
--- a/lisp/magit-diff.el
+++ b/lisp/magit-diff.el
@@ -305,10 +305,10 @@ If the used INDENT is `tabs', highlight indentation with tabs.
 If INDENT is an integer, highlight indentation with at least
 that many spaces.  Otherwise, highlight neither."
   :group 'magit-diff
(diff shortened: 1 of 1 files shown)


tarsius:
Thanks!

basil-conto:
Thanks!  BTW, could you please add a `.mailmap` entry like the following for me?
```
<basil@contovou.net> <contovob@tcd.ie>
```
If you prefer starting with a name, it would be
```
Basil L. Contovounesios <basil@contovou.net> <contovob@tcd.ie>
```

tarsius:
Sure, done.

PR #5541, opened 2026-03-23
doc: update stats link in readme

confusedalex:
updated the old link to the new stats page


tarsius:
Thanks!

There was a second instance and the anchor also isn't necessary anymore, so I've done this in my own commit.

PR #5542, opened 2026-03-24
Fix rewording empty commits during rebase

Silex:
Hello,

This commit is AI generated. It supposedly fixes my problem: when I try to reword the root commit of a repository it fails. E.g doing `rw` on the root commit. Everything "works" but then the commit is still not reworded.

I'll let you judge if fix is appropriate.

---

This fixes a bug in Magit's single-commit reword flow for empty commits, including the root commit case the user reported. Invoking `rw` currently rewrites the rebase todo entry to use Git's `reword` action, but when the targeted commit is empty Git later tries to amend that commit without `--allow-empty`. The result is that the rebase stops with Git claiming that amending the commit would make it empty, even though the commit was already empty and the user only wanted to change its message.

The root cause is that `magit-rebase--perl-editor` always rewrote the selected todo entry to `reword`, regardless of whether the target commit was empty. That is fine for normal commits, but it routes empty commits through a broken Git path. This patch teaches Magit to detect empty target commits and, for the reword action only, keep the original `pick` entry and append an `exec git commit --amend --allow-empty --only --edit` step instead. That preserves the existing interactive-rebase workflow for non-empty commits while using a safe amend path for empty commits. The change covers empty root commits and empty non-root commits with the same logic.

I added regression tests that execute Git using the exact `GIT_SEQUENCE_EDITOR` string Magit generates. One test covers rewording an empty root commit with `git rebase -i --root`; the other covers rewording an empty non-root commit during an interactive rebase range. Both tests verify that the rebase completes successfully and that the commit history remains intact.

Validation for this patch consisted of running the two new ERT tests in batch mode and byte-compiling the touched files.

### DIFF
--- a/lisp/magit-sequence.el
+++ b/lisp/magit-sequence.el
@@ -805,16 +805,29 @@ argument, prompt for the first commit to potentially squash into."
       nil nil t)))
 
 (defun magit-rebase--perl-editor (action since)
-  (let ((commit (magit-rev-abbrev (magit-rebase--target-commit since))))
-    (format "%s -i -p -e '++$x if not $x and s/^pick %s/%s %s/'"
+  (let* ((commit (magit-rebase--target-commit since))
+         (abbrev (magit-rev-abbrev commit))
+         (replacement
+          (cl-case action
+            (edit   (format "edit %s" abbrev))
+            (remove (format "noop\n# pick %s" abbrev))
+            (reword (if (magit-rebase--empty-commit-p commit)
+                        (concat "$&\\n"
+                                "exec git commit --amend "
+                                "--allow-empty --only --edit")
+                      (format "reword %s" abbrev)))
+            (t      (error "Unknown action: %s" action)))))
+    (format "%s -i -p -e '++$x if not $x and s/^pick %s\\S*.*$/%s/'"
             magit-perl-executable
-            commit
-            (cl-case action
-              (edit   "edit")
-              (remove "noop\n# pick")
-              (reword "reword")
-              (t      (error "Unknown action: %s" action)))
-            commit)))
+            abbrev
+            replacement)))
+
+(defun magit-rebase--empty-commit-p (commit)
+  (let ((tree (magit-rev-parse (concat commit "^{tree}"))))
+    (equal tree
+           (if-let ((parent (car (magit-commit-parents commit))))
+               (magit-rev-parse (concat parent "^{tree}"))
+             (magit-git-string "mktree")))))
 
 ;;;###autoload
 (defun magit-rebase-continue (&optional noedit)
--- a/test/magit-tests.el
+++ b/test/magit-tests.el
@@ -25,6 +25,7 @@
 (require 'tramp-sh)
 
 (require 'magit)
+(require 'magit-sequence)
 
 (defun magit-test-init-repo (dir &rest args)
   (let ((magit-git-global-arguments
@@ -69,6 +70,12 @@
         (unless raw
           (setq result (string-trim result)))
         result))))
+
+(defun magit-test-rebase-with-editor (editor &rest args)
+  (let ((process-environment (copy-sequence process-environment)))
+    (push "GIT_EDITOR=true" process-environment)
+    (push (concat "GIT_SEQUENCE_EDITOR=" editor) process-environment)
+    (apply #'magit-git-success "rebase" "-i" args)))
 
 ;;; Git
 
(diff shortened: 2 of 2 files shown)


Silex:
Q: Why is the fix about empty commit, the root commit is not empty usually?

A: Because the failure mode is not “root commit” by itself, it is “root commit with the same tree as its parent context during rebase”, which for the actual Git machinery becomes the same problematic class as rewording an empty commit.

More concretely:

`rw` on an older commit uses interactive rebase and rewrites the todo entry to `reword`.
When Git stops at a `reword` step, it effectively amends that commit message.
For commits whose tree does not change relative to the replay state, Git can hit its “amend would make it empty” path unless `--allow-empty` is used.

The root commit you reported is one way to hit that, but the underlying bug is broader than “rootness”. If I special-cased only the root commit, Magit would still fail on non-root empty commits during the same reword flow. That is why I fixed the actual failing condition instead of the symptom.

So the short version is: root commit was the reproducer, empty-commit rewording was the root cause.

tarsius:
I can reword any one empty commit using `r w`, including the root commit.

If you can actually reproduce the alleged issue, then provide human written and verified steps to reproduce the issue.

It's possible that the issue is due to the use of an ancient Git version. Also see `--no-keep-empty` and `--keep-empty` in the `git-rebase` manpage.

Silex:
Manual reproduction:

```sh
tmpdir=$(mktemp -d)
cd "$tmpdir"
git init
git config user.name 'A U Thor'
git config user.email 'a.u.thor@example.com'
git commit --allow-empty -m 'empty-root'
```

In Magit, on that repository:

1. Move point to the root commit.
2. Invoke `r w` (`magit-rebase-reword-commit`).
3. Even with `--keep-empty`, the rebase stops with:

```text
  1 git … rebase -i --root --autostash --autosquash --keep-empty
hint: Waiting for your editor to close the file...
interactive rebase in progress; onto de9fcf1
Last command done (1 command done):
   reword 185b690 empty-root OMG # empty
No commands remaining.
You are currently editing a commit while rebasing branch 'main' on 'de9fcf1'.


Initial commit

No changes
You asked to amend the most recent commit, but doing so would make
it empty. You can repeat your command with --allow-empty, or you can
remove the commit entirely with "git reset HEAD^".
You can amend the commit now, with

  git commit --amend 

Once you are satisfied with your changes, run

  git rebase --continue
```

The relevant part is that Magit uses `reword <sha>` for an already-empty commit, and Git then tries `commit --amend` without `--allow-empty`.

This PR adds a testcase for that behavior in `test/magit-tests.el`:

- `magit-rebase-reword-commit:empty-root`
- `magit-rebase-reword-commit:empty-past`

Verification:

- `HEAD~1` = `df9e539f`: both tests fail
- `HEAD` = `30ece690`: both tests pass

So the testcase reproduces the issue on `HEAD~1`, and this commit fixes it.

tarsius:
> 1. Move point to the root commit.
> 2. Invoke `r w` (`magit-rebase-reword-commit`).
> 3. Even with `--keep-empty`, the rebase stops with:

I did that once more and once more it works. A commit message editing buffer popped up. I added some text and finished with `C-c C-c`. Back in the status buffer the new message is now displayed for the one and only commit.

Some relevant documentation from `git-rebase(1)`:

> BEHAVIORAL DIFFERENCES
>
>        git rebase has two primary backends: apply and merge. (The apply
>        backend used to be known as the am backend, but the name led to
>        confusion as it looks like a verb instead of a noun. Also, the merge
>        backend used to be known as the interactive backend, but it is now used
>        for non-interactive cases as well. Both were renamed based on
>        lower-level functionality that underpinned each.) There are some subtle
>        differences in how these two backends behave:
> 
>    Empty commits
>
>        The apply backend unfortunately drops intentionally empty commits, i.e.
>        commits that started empty, though these are rare in practice. It also
>        drops commits that become empty and has no option for controlling this
>        behavior.
> 
>        The merge backend keeps intentionally empty commits by default (though
>        with -i they are marked as empty in the todo list editor, or they can
>        be dropped automatically with --no-keep-empty).
> 
>        Similar to the apply backend, by default the merge backend drops
>        commits that become empty unless -i/--interactive is specified (in
>        which case it stops and asks the user what to do). The merge backend
>        also has an --empty=(drop|keep|stop) option for changing the behavior
>        of handling commits that become empty.

What does `git config rebase.backend` say.?
What Git version do you use?

Silex:
> git config rebase.backend

This returns nothing.

Git version is `2.43.0`, it's old but not *that* old. It's the default version from ubuntu 24.04, and nothing in the changelog suggest something was fixed around this, but maybe I just didn't find it.

Did you really test with an empty root commit or just a regular root commit?

Not sure what to tell you, here is my experience (I typed `M-x magit-version` still visible in 1st screenshot).

## Before pressing `r`

image

## After pressing `r`, before pressing `w`

image

## After pressing `w`

image

And the error is the one above:

```
  1 git … rebase -i --root --autostash --autosquash --keep-empty
hint: Waiting for your editor to close the file...
interactive rebase in progress; onto de9fcf1
Last command done (1 command done):
   reword 185b690 empty-root OMG # empty
No commands remaining.
You are currently editing a commit while rebasing branch 'main' on 'de9fcf1'.


Initial commit

No changes
You asked to amend the most recent commit, but doing so would make
it empty. You can repeat your command with --allow-empty, or you can
remove the commit entirely with "git reset HEAD^".
You can amend the commit now, with

  git commit --amend 

Once you are satisfied with your changes, run

  git rebase --continue
  ```
  
  What I don't quite understand is that the tests I added in this PR should fail by you too if you run them on `main`, does it?
  
  For information, `c w` works fine, just `r w` does not (or `r i` followed by `reword` in the interactive rebase).

tarsius:
> What I don't quite understand is that the tests I added in this PR should fail by you too if you run them on `main`, does it?

No, they succeed for me.

What Git version do you use? (Use Magit to determine the version, using `! ! git version RET`.)

Silex:
2.43.0, the version from ubuntu 24.04

tarsius:
Sounds recent enough. Just in case I verified that there were no relevant changes in the quoted documentation since that release.

My guess is that you somehow end up using the `apply` rebase backend, instead of the `merge` backend. Try forcing the latter by setting `rebase.backend`.

Silex:
Still same issue after setting `rebase.backend`.

If I eval-buffer my modifications in `magit-sequence.el` then it works.

It's not like I often edit the root commit anyway, and I know how to work around it with `cw` so I'll just give up 🤷‍♂️ 

Thanks for the time, have a nice day.

tarsius:
It's weird, indeed. At this point I don't see any reason why you would see different behavior.

Cheers!

PR #5560, opened 2026-04-15
Fix `magit-read-module-path` to check for `predicate`

pedrormjunior:
Fix problem introduced in d527f833.

Discovered when trying to delete a submodule.


tarsius:
Thanks!

PR #5565, opened 2026-04-27
Add magit-branch-new-branch-initial-name value

matthewbauer:
This is useful in setups where many different users share the same upstream. The prefix makes it easy to know what is your work vs. someone else's.

For instance, I might set this in .dir-locals.el for a repo using this setup:

```
((nil . ((magit-branch-new-branch-initial-name . "matthew/"))))
```

### DIFF
--- a/lisp/magit-branch.el
+++ b/lisp/magit-branch.el
@@ -202,6 +202,13 @@ has to be used to view and change branch related variables."
   :group 'magit-commands
   :type '(repeat string))
 
+(defcustom magit-branch-new-branch-initial-name ""
+  "Initial name for the new branch."
+  :package-version '(magit . "4.6.0")
+  :group 'magit-commands
+  :type 'string
+  :safe 'stringp)
+
 ;;; Commands
 
 ;;;###autoload(autoload 'magit-branch "magit" nil t)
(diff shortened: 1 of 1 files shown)


tarsius:
Please do not change the indentation of `cond`. (When using Emacs 31 it is different from the default in my libraries due to `(lisp-indent-local-overrides . ((cond . 0) (interactive . 0))` in `.dir-locals.el`. Also use `magit-branch-new-branch-initial-name: New option` as the summary, for consistency with other commit messages.

tarsius:
I've instead added `magit-branch-name-suggestions` and switched to using `magit-completing-read`.

matthewbauer:
so one limitation here is it's not setting "INITIAL-INPUT", so there's no prefilling going on, at least with my vertico setup. Potentially, if only one element exists, it could be the INITIAL-INPUT? 

For now I have this advice:

```
(advice-add 'magit-branch--read-name :override
    (defun my-magit-branch--read-name (prompt &optional start-point)
      (let ((taken (magit-list-local-branch-names))
            (choices magit-branch-name-suggestions))
        (when (and start-point
                   (magit-remote-branch-p start-point)
                   (string-match "/" start-point))
          (cl-pushnew (substring start-point (match-end 0))
                      choices :test #'equal))
        (magit-completing-read
         prompt (seq-difference choices taken) nil
         (lambda (choice)
           (cond
             ((member choice taken)
              (run-with-timer
               0 nil (##minibuffer-message "conflicts with existing branch"))
              nil)
             ((not (magit-git-success "check-ref-format" "--branch" choice))
              (run-with-timer
               0 nil (##minibuffer-message "not a valid branch name"))
              nil)
             (t)))
             (when (= (length choices) 1) (car choices))))))
```

PR #5580, opened 2026-05-18
magit--find-buffer condition changes during iteration

XeCycle:
Hi, I updated from MELPA today and found ediff-dwim on a file line inside a magit-revision buffer broke; it displays the parent commit of that file twice, compare it against itself and says no change.  Followed it in edebug and found this function likely broken.


tarsius:
Thanks a lot!

Issue #5220, opened 2024-09-06
magit hangs when staging a hunk with `tramp-direct-async-process` enabled


CeleritasCelery:
`magit-version` : Magit v4.1.0-2-g4d054196, Transient 0.7.4, Git 2.37.1, Emacs 29.4, darwin
`tramp-version` : 2.7.1.1

This is similar to #3624, but is **not** the same issue. That issue was related to an older version of tramp and this is with `2.7.1`. However it only manifests when you have [tramp-direct-async-process](https://www.gnu.org/software/tramp/#Improving-performance-of-asynchronous-remote-processes-1) enabled.

starting with `magit-emacs-Q-command` and tramp `2.7.1`, add the following config:

```lisp
(connection-local-set-profile-variables
 'remote-direct-async-process
 '((tramp-direct-async-process . t)))

;; Set "remotehost" to match your host.
(connection-local-set-profiles
 '(:application tramp :machine "remotehost")
 'remote-direct-async-process)
```
 Then run `magit-status` on a remote repo and try to stage a hunk. Magit will hang, and I got the following backtrace using `debug-on-quit`:

```lisp
Debugger entered--Lisp error: (quit)
  magit-run-git-with-input("apply" ("--cached") "-p0" nil "--ignore-space-change" "-")
  magit-apply-patch(#<magit-file-section "README.md" [file unstaged status] 695-[716-]1906> ("--cached") "diff --git README.md README.md\nindex 10fde69a01..0...")
  magit-apply-hunk(#<magit-hunk-section (nil (20 7) (20 7)) [hunk file unstaged status] 1178-[1277-]1672> "--cached")
  magit-stage(nil)
  funcall-interactively(magit-stage nil)
  command-execute(magit-stage)
```
The backtrace doesn't seem particularly useful since all it tells us is that magit is hanging while running the command.

tarsius:
> However it only manifests when you have [tramp-direct-async-process](https://www.gnu.org/software/tramp/#Improving-performance-of-asynchronous-remote-processes-1) enabled.

That feature improves performance, but as described, it breaks quite a few things. It looks like "staging individual hunks using magit", has to be added to that list. To generalize that a bit, the feature that seems to break is "sending text to the remote process on its stdin". You could ask Tramp's maintainer, if this indeed is something that is inherently impossible with this optimization.

CeleritasCelery:
[...]

Issue #4720, opened 2022-07-05
Emacs hangs on staging chunks over Tramp 


impaktor:
I've noticed this a long time (+1 year, ever since I started relying on tramp), on several computers, and emacs versions.

In M-x magit-status, I select the chunks I want to stage, of a file, minibuffer will say "Git finished" after a few seconds, but then I must C-g C-g C-g C-g to get emacs responsive again, and re-enter magit-status to stage next chunk.

- I rarely stage whole file, but when I do this, it works (usually/always(?), I believe.
- I just tried staging a subset of a chunk (C-spc), worked the first time, but not the second, in which case it freezes on "git finished" message. 
- My version:
`Magit 20220625.1219 [>= 3.3.0-git], Git 2.30.2, Emacs 27.1, gnu/linux`

I just tried reproducing this on a "clean" emacs (emacs -Q with magit), and I could stage 2 chunks, but when I try to stage a single line in a third chunk emacs freezes for minutes (I gave up and did C-g after a while). Doesn't look like emacs is using 100% CPU, so there's something else causing it to lock up.

kyleam:
Hmm, it certainly isn't zippy on my end, but I'm not able to trigger a hang when staging region hunks (tried five stage operations in a row).

impaktor:
Is there some code I could run in the background to monitor things to help debug?
I guess not, since Emacs doesn't multithread, and debug-on-error doesn't help, since it doesn't throw an error.

tarsius:
Try to use the profiler or debug statements to see where the hang happens.

impaktor:
Sorry for silence, I have not forgotten, more a question of me not knowing exactly what to do to further diagnose this. I've been running the emacs profiler for an hour or two today. Result below. (I did a lot of magit today, picking chunks and files for stageing which is very slow. Also I note just tab compleation in terminal (outside of emacs) on e.g. git branch name can take seconds, so maybe it's my git that's slow and not magit? (Often when stageing chunks I get "git finished" in mini-buffer, yet it keeps hanging for a long time and when I C-g to abort I see in mini-buffer: `error in process sentinel: Quit: ""`).

```
- command-execute                                              571538  86%
 - call-interactively                                          571483  86%
  - funcall-interactively                                      555432  83%
   + magit-discard                                             299647  45%
   + counsel-find-file                                         138882  20%
   + ivy-switch-buffer                                          31150   4%
   + magit-status                                               19231   2%
   + save-buffer                                                19200   2%
[...]

PR #5591, opened 2026-06-12
Make magit-tramp-pipe-stty-settings customizable

NicholasBHubbard:
`magit-tramp-pipe-stty-settings` is documented as a user-facing variable but it is defined as a `defvar`. This PR makes it a `defcustom` so users can discover and configure it through Customize.

### DIFF
--- a/lisp/magit-process.el
+++ b/lisp/magit-process.el
@@ -302,7 +302,7 @@ string in the heading of its section."
   :type '(choice (const :tag "None" nil) string))
 
 (defvar tramp-pipe-stty-settings)
-(defvar magit-tramp-pipe-stty-settings ""
+(defcustom magit-tramp-pipe-stty-settings ""
   "Override `tramp-pipe-stty-settings' in `magit-start-process'.
 
 The default for that Tramp variable is \"-icanon min 1 time 0\",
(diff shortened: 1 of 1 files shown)


tarsius:
This variable is intentionally not defined as an option to discourage users from changing the value. The default value is believed to be the correct value for (almost?) everyone and changing it means possibly opting in to this breaking.

It is possible that in some edge cases another value is better, that's why I made it a variable, instead of hard-coding it. It appears you have run into such a case. Could you please describe why the default does not work for you, what value you use instead, and why that helps.

Re #4720.

NicholasBHubbard:
> Could you please describe why the default does not work for you, what value you use instead, and why that helps

I am using `tramp-direct-async-process` so I want this value set to `pty` to prevent hangs as is discussed in issue #5220.

tarsius:
That is also just a variable and users are discouraged from setting it globally.

```lisp
(defvar tramp-direct-async-process nil
  "Whether direct asynchronous processes should be used.
It is not recommended to change this variable globally.  Instead, it
should be set connection-local.")
```

There currently isn't any support in Magit for dynamically determining whether to use a pty based on that variable.

I still think this variable should not be advertised.

NicholasBHubbard:
Ok, thanks for the consideration and for working on magit. Closing this PR.

tarsius:
I've added a bookmark to remind me to look into maybe setting the variable dynamically.

PR #5603, opened 2026-07-05
Ediff not cleaning up properly when quiting out of conflict resolution

steve-downey:
At least in my configuration, using `e` to use ediff on a merge conflict would leave the control panel open after saying Y to closing everything. Then discovered there are additional runtime connections between operations around window configuration restoration, which is the story these 4 commits tell. 

LLM was used to track down the issue, which I though had to be in my code, and review fix, not for this code.

### DIFF
--- a/lisp/magit-ediff.el
+++ b/lisp/magit-ediff.el
@@ -229,9 +229,7 @@ and alternative commands."
            (when (buffer-live-p ediff-buffer-C) (kill-buffer ediff-buffer-C))
            (when (buffer-live-p ediff-ancestor-buffer)
              (kill-buffer ediff-ancestor-buffer))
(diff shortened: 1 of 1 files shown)


tarsius:
If pushed my own commits to fix these regressions. 7caa310a4e6c7fabc503e3246bc003df3b994836~..5e84551b187e2c99da5b30a68e44a3edf11673aa

Your commits and their messages were created using a LLM, weren't they? The last sentence in in your report above might say so, or the opposite; I am having difficulties making sense of the last third of that sentence.

These commits throw out the baby with the bathwater. The LLM seems to have correctly identified the regressions but not the reasons behind the intentional changes in the commits that introduced those regressions.

Please note that I do not accept LLM output and that the pull-request template says so.

PR #5612, opened 2026-07-30
Update worktree deletion to handle a locked worktree

3rdLaw:
Magit currently will happily let you delete a locked worktree, as it trashes the directory by default*. 
However, if the worktree is locked, the subsequent `prune` call will not get rid of the worktree's metadata.
This change proposes having magit unlock the worktree before deleting, when applicable.

The attached `magit-git.el` lines update argument order when searching for lock/prunable (`string-prefix-p` is prefix first, then match).

*Per `defcustom magit-delete-by-moving-to-trash t`. However, even if that is default is `nil`, the subsequent `(magit-call-git "worktree" "remove" "--force" worktree)` will fail to remove it, since the command needs double force to remove a locked worktree (git's manual states "To remove a locked worktree, specify --force twice.").


tarsius:
I hate that I have to ask, but you didn't use a LLM for this, did you?

3rdLaw:
AI did do the investigation and wrote a much-larger fix, but then I trimmed it down and cleaned it up and tested it; so I'd say the provenance of the code is 50/50.

My apologies if this is against policy! The impression I got from reading the contributing docs was that you wanted humans writing the words and vouching for the contents, which is the case here, but feel free to close if LLM code is not accepted. Cheers!

tarsius:
LLM code is not accepted.

PR #5616, opened 2026-08-05
Fix typo in stash apply merge message

wojciech-graj:
This maps to the --3way flag, so three-way is correct instead of tree-way.


kyleam:
Thanks

PR #5623, opened 2026-08-13
Use git common dir when adding private gitignore rules

Lenbok:
Fixes: #5614

Applies the fix suggested by @kyleam in the linked issue/discussion, with some documentation updates. When adding a private git ignore rule from a worktree, this ensures the rule is stored in the exclude file that git actually uses when determining file excludes. Without this fix, the ignore rule gets added to a worktree-specific exclude file that actually does not get used (possibly only for some versions of git and/or worktrees created with some versions of git).


tarsius:
See https://github.com/magit/magit/discussions/5614.

PR #5626, opened 2026-08-20
magit-xref-restore: Remove redundant second refresh

nagyation:
magit-setup-buffer-internal already refreshes the Magit buffer, which makes the trailing refresh redundant and needlessly doubles the work. Removing it makes magit-go-forward/backward almost twice as fast, which is especially noticeable in large repositories.


tarsius:
Thanks a lot!

PR #5628, opened 2026-08-21
fix a typo

Ossifragus:
This makes magit run correctly on my Emacs 32.2.50.


tarsius:
Thanks!

"Merged", because I couldn't push to your "master" after changing the commit message a bit.

PR #5635, opened 2026-09-17
magit-stage-untracked: Handle borg-user-emacs-directory being unbound

renatofdds:
After  https://github.com/magit/magit/commit/9ae2d599943bef206389a1b00500b55481a2b036,  when user does not have borg loaded, `magit-stage-untracked` errors with:

`magit-stage: Symbol’s value as variable is void: borg-user-emacs-directory`


tarsius:
Thanks!

renatofdds:
Thank you!

